PR #13, opened 2015-08-11
Fix potential null pointer dereference in netfilter

pmillerchip:
In netfilter.c, the memset call on line 70 is before the check for null pointer, so it could cause a null pointer dereference crash. The statement has been moved to after the null pointer check.


netblue30:
Merged, thanks!

PR #16, opened 2015-08-12
Block access to history files

dewbasaur:
In CVE-2015-4495 — the Firefox PDF.js exploit — an examined payload searched for various history files in order to find potentially sensitive information.


netblue30:
Merged, thanks!

PR #17, opened 2015-08-13
A bit more for CVE-2015-4495

dewbasaur:
The payload I saw for it also searched for these specific directories as well. Two notes:
* Is the ```+``` in ```${HOME}.config/psi+``` going to be an issue?
* I couldn't actually find what ```.tconn``` is for.


dewbasaur:
Oh, an exclusion for ```.purple``` will have to be added to ```pidgin.profile```, I forgot.

PR #26, opened 2015-08-15
Call realpath to resolve symlinks correctly in disable_file

pmillerchip:
Firejail wasn't handling relative symlinks correctly (like when a symlink is `../file`) and also wasn't handling multiple symlinks correctly (it could only handle one link, not a link to a link). This commit fixes that by calling the `realpath` function to do all the hard work for us.


netblue30:
realpath - I learn something new every day! Thanks!

PR #30, opened 2015-08-17
Update disable-mgmt.inc

mjudtmann:
Jailed applications should not be able to change the configuration of firejail.


netblue30:
Merged, thanks!

PR #32, opened 2015-08-18
Support ~ in blacklist and profile includes

pmillerchip:
The `~` character is now supported in the `--blacklist` option and in profile includes for specifying the home directory of the current user.


netblue30:
Merged, thanks!

PR #33, opened 2015-08-18
Support filenames with spaces in the blacklist option

pmillerchip:
Filenames with spaces can now be blacklisted on the command line:

`firejail "--blacklist=My Documents"`

...or in a profile file:

`blacklist My Documents`


netblue30:
Thanks!

PR #34, opened 2015-08-20
Fix the 'make install' command for new bash completion location

pmillerchip:
Copied the new location for the bash completion files into the `Makefile.in` file, so that they get included in the `make install` command.


pmillerchip:
That was quick, thank you!

netblue30:
My bad, I've forgot to check in a file!

pmillerchip:
No problem, it happens to all of us! I thought it must have been something simple like that. I had to do the quick fix on my repo just to get it to install, so I thought I might as well put a pull request up just to be safe.

PR #35, opened 2015-08-20
Compile with -W -Wall -Werror

pmillerchip:
Firejail now compiles with the `-W -Wall -Werror` option in `CFLAGS`. This option enables all warnings and turns them into errors, i.e. the compile stops if there are any warnings. I've removed all existing warnings so the code is now clean, and this means that in future the code will remain clean (as it's now impossible to introduce new warnings without breaking the build).

### DIFF
--- a/src/firejail/Makefile.in
+++ b/src/firejail/Makefile.in
@@ -12,7 +12,7 @@ H_FILE_LIST       = $(wildcard *.[h])
 C_FILE_LIST       = $(wildcard *.c)
 OBJS = $(C_FILE_LIST:.c=.o)
 BINOBJS =  $(foreach file, $(OBJS), $file)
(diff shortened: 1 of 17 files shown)


netblue30:
That's cool, thanks!

netblue30:
I had to remove them from src/lib/Makefile.in. It is the old libnetlink.c code I took from iproute2. I'll give it a run on some other linux platforms. What distro are you using? I am on Debian Wheezy.

pmillerchip:
I'm on Ubuntu 14.04 amd64. The version I checked in compiled on that but I didn't test it on any other distros, sorry!

netblue30:
No problem, I'll adjust it here. I have several distros set in Virtual Box, it will be easy.

pmillerchip:
That's interesting, I tried it on Debian Wheezy and it seems the problem is actually in the netlink header from Linux! Here's some more info:

http://permalink.gmane.org/gmane.linux.kernel/1901067

We could change firejail to have `-W -Wall` and drop the `-Werror` just for `lib/src`, that way we see the warning but it doesn't stop the build. What do you think? I'm just trying to get firejail to compile as cleanly as possible, but I don't mind if you think it's too harsh.

Issue #40, opened 2015-08-22
Support files in current directory and paths with private.keep


pigmonkey:
I want to use firejail and `--private.keep` to view a PDF file within a sandbox. If my PDF reader is `pdfreader` and the file is in `~/`, this works great:

    $ cd ~/
    $ firejail --private.keep=dirty.pdf pdfreader dirty.pdf

However, this fails when the file is somewhere other than `~/`

    $ cd ~/tmp/random/directory
    $ firejail --private.keep=alsodirty.pdf pdfreader alsodirty.pdf
    Error: file /home/pigmonkey/alsodirty.pdf not found.

I would like `--private.keep` to accept files in the current directory, as well as paths. For instance, in addition to the previous example I'd like to be able to do stuff like:

    $  firejail --private.keep=~/tmp/random/directory/alsodirty.pdf ...
    $  firejail --private.keep=../alsodirty.pdf ...

I understand that `--private.keep` was probably originally intended just to copy over dotfiles from the real home, but I think expanding it to support file paths like this would make it much more useful.

netblue30:
private.keep is being renamed to --private-home in the next version. It makes "private" only files in home directory. I have a --private-etc doing the same for /etc directory. I'll implement a --private-tmp, it shouldn't be a big deal.

> I understand that --private.keep was probably originally intended just to copy over dotfiles from the real home

Some users came up with the idea to have only .mozilla directory visible. I guessthe main usage so far is "firejail --private-home=.mozilla firefox"

What is pdfreader? I could build a default profile for it. Also, use --net=none with pdf readers - the bad guys are interested in sending data back.

pmillerchip:
@netblue30 The `tmp` directory in his example is relative to the home directory, i.e. it's `/home/user/tmp/random...` - it tripped me up at first! Currently the `--private-home` code doesn't expand the `~` or `${HOME}` at the front, and just appends it to `homedir`. So `~/tmp` in his example expands to `/home/user/~/tmp` which doesn't work.

I've already been thinking of making a function called something like `expand_user_supplied_path` as there's several places where we expand `~` and `${HOME}` using the same code. We could call that from the `--private-home` code and then check that the result starts with whatever's in the `homedir` variable. That would also take care of calling `realpath` and expanding out any paths like `../dir` as well. Would you like me to put up a pull request for it?

netblue30:
yes, thanks!

pigmonkey:
[...]

PR #42, opened 2015-08-22
Implement the expand_home util function

pmillerchip:
Implemented a new function called `expand_home` in util.c, in order to help with issue #40 . The following now work, that didn't before:

`firejail --private-home=~/dir/file`
`firejail --private-home=${HOME}/dir/file`

As a nice side-effect, the following now also works (previously `--private` only supported `~` for homedir):

`firejail --private=${HOME}/dir`

Hopefully I haven't broken anything else but please give it a test!


netblue30:
All seems to be fine, thanks!

PR #43, opened 2015-08-24
Fix typo in usage.c

pmillerchip:
The `--env` commit introduced a typo, this patch reverts it back to normal


netblue30:
Thanks!

PR #48, opened 2015-08-28
Rewrite globbing code to fix various minor issues

sarneaud:
* Plug a memory leak.
* Remove the short-circuit.  (This breaks when someone uses [] or ?
  patterns without using *.  I figure it's best to use the principle of
  least surprise and just let the system glob() implementation do what
  it does.)
* Stop sorting results.

I've also replaced a lot of disable-history.inc with a glob pattern.
Now it catches files like .sqlite_history and whatever the user runs
under rlwrap.


netblue30:
Thanks!

PR #52, opened 2015-09-01
Clean up some fragile uses of strncmp.

sarneaud:
In some places the code compares the first n characters of a string and
then assumes a valid string starts from the n+2th character.  I didn't
find any places where this wasn't justifiable, but I think it's better
to stick to safer patterns, especially in SUID code.


netblue30:
Thanks!

PR #60, opened 2015-09-10
added install-strip, make install now without strip.

pstn:
The default in autotools is not to strip executables when doing make install, but only when doing make install-strip. Not doing that confuses a lot of automated build systems like source based package managers or tools that build packages automatically. It would be very nice if you could pull this.


netblue30:
Thanks, I'll merge it.

PR #66, opened 2015-09-23
Add seccomp errno filter support

wader:
No tests yet. Tried to get the existing tests running without much success :(
Would users like to mix errno and black/white -lists? i currently only need errno filter and default allow.
Tried to keep same code style, hope it is ok.

### DIFF
--- a/src/firejail/seccomp.c
+++ b/src/firejail/seccomp.c
@@ -109,6 +109,10 @@ struct seccomp_data {
 	BPF_JUMP(BPF_JMP+BPF_JEQ+BPF_K, syscall_nr, 0, 1), \
 	BPF_STMT(BPF_RET+BPF_K, SECCOMP_RET_ALLOW)
 
+#define ERRNO(syscall_nr, nr) \
+	BPF_JUMP(BPF_JMP+BPF_JEQ+BPF_K, syscall_nr, 0, 1), \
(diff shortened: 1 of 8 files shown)


src/firejail/seccomp.c
```diff
@@ -157,6 +161,11 @@ void filter_debug(void) {
 			printf("  BLACKLIST %d %s\n", *nr, syscall_find_nr(*nr));
 			i += 2;
 		}
+		else if (*ptr	== 0x15 && *(ptr +14) == 0x5 && *(ptr + 15) == 0) {
```
wader: maybe this function could be rewritten using sock_filter struct pointer instead hmm would also take care of the endian stuff

wader:
For some context, i myself use this to workaround buggy virtualbox shared filesystem that has broken sendfile support, the application will fallback to userland buffered IO if it gets ENOTSUP.

netblue30:
Looks fine, I'll merge it over the weekend. Thanks!

wader:
Ok. Please run the tests to make sure it didn't break anything. Any pointers how to run the tests? i installed expect and messed around but did not really work hmm

netblue30:
I go in test directory and run:
`````
sudo ./configure
`````

Take a look at configure script first, it creates a number of bridge devices. To get rid of them I usually reboot my computer after the test.

I start the test as "./test.sh | grep TESTING"

PR #73, opened 2015-10-07
Add a profile for Spotify

greigdp:
This PR proposes a firejail profile for Spotify.

Since Spotify, when used for streaming, only requires access to a very small group of folders (config, local and cache for itself), I created the profile based around a whitelist. 

Finally, it allows access to the per-user pulseaudio config directory, in case anyone is using the [workaround for PulseAudio 7](https://github.com/netblue30/firejail/blob/master/README.md#pulseaudio-70)


netblue30:
Merged, thanks!

PR #77, opened 2015-10-10
use configure options in Makefile

pyther:
The Makefile should use the autoconf values for `bindir`, `libdir`, `datarootdir`, and `mandir`.

It looks like using lowercase variable names for the above variables is acceptable based on the following examples:
  * https://www.gnu.org/prep/standards/html_node/DESTDIR.html
  * http://www.netperf.org/svn/netperf2/trunk/doc/examples/Makefile.in


netblue30:
fixed, thanks!

PR #78, opened 2015-10-10
standalone rpm spec

pyther:
Depends on PR #77.

Provides a more reproducible way to build rpm package.


netblue30:
Merged, thanks!

PR #88, opened 2015-10-23
Update pidgin.profile

andrew160:
Pidgin's data directory is blacklisted in disable-common.inc, so it couldn't access it.


netblue30:
Merged, thanks!

PR #89, opened 2015-10-23
add a few new items to blacklist

g4jc:
I've found a few more files that should be blacklisted as they contain potentially useful information to an attacker.


netblue30:
Merged, thanks!

PR #96, opened 2015-10-25
fix libtrace for musl libc

Duncaen:
Hi, I'm sorry for wasting your time with the previous solution,
this patch is alot smaller and doesn't involve the musl_defs.h and or a configure option.


netblue30:
Thanks!

PR #100, opened 2015-10-26
Create Steam profile

dewbasaur:
seccomp not included due to the 32/64bit issue noted seen in the bug reports.

Also included a little snippet regarding my experiment with using private directories for it.

The two noblacklist lines are just in case.

### DIFF
--- a/etc/steam.profile
+++ b/etc/steam.profile
@@ -1,0 +1,20 @@
+# Steam profile (applies to games/apps launched from Steam as well)
+noblacklist ${HOME}/.steam
+noblacklist ${HOME}/.local/share/steam
+include /etc/firejail/disable-mgmt.inc
+include /etc/firejail/disable-secret.inc
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-history.inc
+caps.drop all
+netfilter
+noroot
+
+# seccomp breaks Steam runtime due to 32/64bit syscall incompatibilties
+
+
+## Author note:
+##   If you wish to use a private directory for Steam, e.g.:
+##       private /path/to/steam-home
+##   ... be aware that games will not launch from this directory without
+##   execute-permission trickery. In this case, you should store the games
+##   in a separate (whitelisted/noblacklisted) directory.


netblue30:
If I understand you correctly, you would need steam directory to be mounted without noexec flag set. Is steam directory by default in /home/user? I can make the mount for whitelist without noexec.

netblue30:
Sorry, it looks like whitelist mount already permits exec. We should probably go whitelist on it. What is the problem?

dewbasaur:
Default Steam directory on new installs is ~/.steam, used to be ~/.local/share/steam.

I was actually trying to make the private home in a /media directory. Could that be the problem?

dewbasaur:
Oh, right, I think understand now. You mean if we had:
```whitelist ${HOME}/.steam```
... instead of noblacklist, then exec will be allowed in the private home dir?

netblue30:
Yes, the only directory visible in /home/user would be .steam

dewbasaur:
Well what I was trying to was was having everything compartmentalized on the second drive.

I created ```/media/user/drive/steam-home```, moved ../Steam and ~/.steam into that directory, and ran ```firejail --private=/media/user/drive/steam-home```.

Doing that, Steam gave me an error saying that it couldn't load games off a noexec filesystem.

pmillerchip:
@dewbasaur `/media/user/` is for removable drives, are you really trying to store your Steam games on a removable drive? When I plug in an ext4 formatted external USB drive it mounts without `noexec` but I'm not sure about other formats. Check that the drive isn't formatted as FAT or NTFS as this might cause a `noexec` mount. Steam for Linux needs a Linux formatted drive.

I have my Steam games on a dedicated hard drive but it's physically inside the case, attached via SATA and formatted as ext4. I have a directory called `/bigdrive/steam` and I start Steam with `firejail --private=/bigdrive/steam`, everything works fine here.

dewbasaur:
@pmillerchip My drive is not external; it is internal, connected by SATA, and formatted ext4, just as yours. Ubuntu automatically selected that mountpoint, I think. Regardless, I've been running my games off it just fine, and running with a private home directory in my $HOME at the moment, I am presently still accessing those games just fine.

It's when I tried to have my games folder within the private home folder, which I was trying to use on the second drive as well originally, that I had problems.

pmillerchip:
Can you run the `mount` command to see if your drive mounted with the `noexec` flag or not please?

PR #102, opened 2015-10-27
Use configured libdir instead of $prefix/lib

reinerh:
When a non-default libdir is used, firejail installs
some components in this directory, but the hardcoded
paths to ftee etc. still point to $prefix/lib.
Use the configured libdir path instead.


netblue30:
Thanks!

PR #113, opened 2015-11-01
Added profile for Conkeror Browser

emacsomancer:
Add profile for Conkeror browser: http://conkeror.org/


netblue30:
merged, thanks!

PR #129, opened 2015-11-10
Allow firefox theming with non-global themes

satai:
User can have gtk themes installed locally. This directory is pretty safe to read, no private data there.


netblue30:
Merged, thanks!

PR #130, opened 2015-11-11
Fixed Skype profile: was a copy of Steam profile

al42and:
It seems that default Skype profile, introduced in 4faa338ca738a0bf40d07b6d4b9b98f62a3cb34f, is and exact duplicate of Steam profile. Thus, Skype has access to Steam directories.
Besides, file containing Skype account credentials could be stolen and used to login from another machine.

Thus in this commit, I:
1) blacklisted ~/.Skype directory in `disable-common.inc`
2) revoked Skype access to Steam dirs, and granted it access to Skype dirs in `skype.profile`


netblue30:
Merged, thanks!

PR #135, opened 2015-11-15
add some other whitelisting for theme and core firefox related functionality on Linux

ilikenwf:


netblue30:
Thanks for the patch. I'll merge it and move some whitelists in whitelist-common.inc.

ilikenwf:
Nice, thanks!

This fixes some visual issues with Firefox on Arch, under xfce4, as well as
the mime types.

PR #145, opened 2015-11-16
Fix #144

### LINKED ISSUE #144: Whitelisting a symlink that is in a subdirectory doesn't work
Example: my `~/.cache/mozilla` directory is a link to `~/Private/.cache/mozilla`.
Whitelisting `~/.cache/mozilla` correctly whitelists `~/Private/.cache/mozilla` and creates the parent folders, but the symlink itself fails because its parent folder doesn't exist.

I've already fixed this myself, so I will submit a pull request soon.

brunonova:
When whitelisting symlinks, ensure the parent directories exist.

This should fix it. At least for me it's working.
Please review the changes.


netblue30:
Thank you  for the bug and for the fix.

Issue #58, opened 2015-09-08
Firejailed Gwenview cannot open filenames with blanks


curiosity-seeker:
I've noticed a strange problem with my firejailed VLC. Whenever I try to open, e.g., a video whose filename contains a blank or is located in a directory with a blank, VLC opens and I get an error like the following (for the file "/media/Multimedia/Filme/Die Tribute von Panem - The Hunger Games.avi"):
![panem](https://cloud.githubusercontent.com/assets/14075215/9736722/3c135db4-5643-11e5-8807-fca8394a7999.png)
Note that this problem does **not** occur if I open VLC directly and navigate to that file, and it does **not** occur with an un-firejailed VLC. Rather, it only occurs if I want to open that file with firejailed VLC in a filemanager (Krusader or Dolphin - I'm running KDE Plasma 5 on Arch Linux). A similar problem happens if I want to open pictures with blanks in their filenames in firejailed Gwenview.

pmillerchip:
If you could please post the exact command line that you use to start firejail, we can help diagnose the problem.

It sounds like a situation where firejail is getting started from a shortcut like `firejail vlc $1` - in this case the `$1` gets expanded without protecting the spaces in the filename. Is it something like that?

curiosity-seeker:
Thanks for your reply. The exact command line in the KDE menu entry is: 

firejail --profile=/home/xxx/.config/firejail/vlc.profile /usr/bin/vlc --started-from-file %U

curiosity-seeker:
FWIW, I've just tried to start that video without the

 --started-from-file %U

switches but with the the same result.

netblue30:
Add a --shell=none, like this:

`````
firejail --shell=none --profile=/home/xxx/.config/firejail/vlc.profile /usr/bin/vlc --started-from-file %U
`````

curiosity-seeker:
Wow! That solved it, indeed! Thank you very much! Can you explain why?

However, that similar problem with Gwenview is not yet really solved. The command line for it is:

firejail --profile=/home/xxx/.config/firejail/gwenview.profile gwenview %U

When opening pictures with blank(s) Gwenview opens with just a line that say something like

Basisordner > tmp > org.kde.gwenview-hNwKkl

After adding --shell=none Gwenview displays the folder symbol where the picture is located and I can navigate to it. An improvement, indeed, but the picture itself isn't immediately opened as it should. Do you have an idea how to circumvent this problem?

netblue30:
[...]

PR #148, opened 2015-11-18
Add '"' chars around every argument passed to bash

brunonova:
This seems to fix the problem with the spaces when a program is started through a shell (Bug #58).
Starting a program without using a shell is still working.

I haven't tested this thoroughly though, so I'm not sure if this causes any other problems.

---

However, spaces are not the only characters that can cause problems when starting a program through a shell.
If, for example, you open a file that contains `"` somewhere, it will probably fail.

So, to fully fix this, `"` characters will have to be escaped. Other characters, like `\` and `!` may also have to be escaped.
*Or* set "shell=none" as default, which would be even easier :smile: (but I suppose this is not the default for a reason).

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -1188,11 +1188,12 @@ int main(int argc, char **argv) {
 	}
 	else {
 		// calculate the length of the command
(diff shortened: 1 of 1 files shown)


netblue30:
We'll leave " characters for some other time, thanks for the fix!

Issue #149, opened 2015-11-19
Restrictive umask breaks subdirectory whitelisting


avoidr:
Allow me to reopen #121 (I don't know if you get notifications for closed issues, I'm new) as I have more info now.

Basically the problem is that when you whitelist a subdirectory, then the topmost directory is owned by root. When whitelisting the topmost directory only, it is owned by the user.
Now with a more restrictive umask, the directory becomes inaccessible, since owned by root and only root has the rights.

```
$ umask 0022
$ firejail --whitelist=.config/
[anon@jail ~]$ ls -ld .config/
drwx------ 22 anon anon 4096 Nov 18 22:20 .config/

$ firejail --whitelist=.config/deadbeef/
[anon@jail ~]$ ls -ld .config/
drwxr-xr-x 3 65534 anon 60 Nov 19 18:48 .config/


$ umask 0077
$ firejail --whitelist=.config/
[anon@jail ~]$ ls -ld .config/
drwx------ 22 anon anon 4096 Nov 18 22:20 .config/

$ firejail --whitelist=.config/deadbeef/
[anon@jail ~]$ ls -ld .config/
drwx------ 3 65534 anon 60 Nov 19 18:48 .config/
```

netblue30:
Yes, this is a bug. Thanks!

netblue30:
I think I've fixed it. Give it a try, and if necessary I'll reopen it. Thanks.

avoidr:
You did fix the bug, but now you ignore umask by setting perms to 0755 regardless. It was enough to just change the owner from root to user, which I think would be better that way. If you could only change the ownership and leave permissions alone, that would be great!

And thanks for the fix nonetheless. You're fast and I like what you do.

avoidr:
This should do it.
```
diff --git a/src/firejail/fs_whitelist.c b/src/firejail/fs_whitelist.c
index fccb827..740bdab 100644
--- a/src/firejail/fs_whitelist.c
+++ b/src/firejail/fs_whitelist.c
@@ -27,12 +27,9 @@
 #include <fcntl.h>
 #include <errno.h>
 
-static int mkpath(const char* path) {
+static int mkpath(const char* path, mode_t mode) {
        assert(path && *path);
        
-       // create directories with a 0755 mode
-       mode_t mode = 0755;
-       
        // create directories with uid/gid as root or as current user if inside home directory
        uid_t uid = getuid();
        gid_t gid = getgid();
[...]

PR #150, opened 2015-11-19
keep original file permissions

avoidr:
Don't force 0755 permissions for whitelisted files. (#149)

### DIFF
--- a/src/firejail/fs_whitelist.c
+++ b/src/firejail/fs_whitelist.c
@@ -27,11 +27,8 @@
 #include <fcntl.h>
 #include <errno.h>
 
-static int mkpath(const char* path) {
+static int mkpath(const char* path, mode_t mode) {
 	assert(path && *path);
-	
-	// create directories with a 0755 mode
-	mode_t mode = 0755;
 	
 	// create directories with uid/gid as root or as current user if inside home directory
 	uid_t uid = getuid();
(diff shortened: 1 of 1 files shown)


netblue30:
I am sure there is more to come, for example:
`````
$ mkdir aaa
$ touch bbb
$ rm -fr aaa bbb
$ mkdir aaa
$ touch aaa/bbb
$ firejail --whitelist=~/aaa/bbb
Reading profile /etc/firejail/generic.profile
Reading profile /etc/firejail/disable-mgmt.inc
Reading profile /etc/firejail/disable-secret.inc

** Note: you can use --noprofile to disable generic.profile **

Parent pid 8187, child pid 8188
Child process initialized
$ ls -l
total 0
drw-r--r-- 2 netblue netblue 60 Nov 19 19:18 aaa
`````
aaa inherits the permissions of bbb file.
````` 
$ ls -l aaa/bbb
ls: cannot access aaa/bbb: Permission denied
$ 
`````

So I had to add a 0111 to the mode in mkpath() function:
`````
static int mkpath(const char* path, mode_t mode) {
	assert(path && *path);
	mode |= 0111;
	
	// create directories with uid/gid as root or as current user if inside home directory
...
`````
I put "mode |= 0111" fix in, thanks!

avoidr:
That's weird though, and I want to see it work properly, not halfass-hackedly. Can't leave it like that. Damn details.

You're welcome!

PR #174, opened 2015-11-29
clarify firejail-profile manpage

avoidr:
I took the liberty to reword and reorder the caps part in the manpage, to make it more clear and less self-contradictory about what it does.

Renamed /etc/password to /etc/passwd, too.


netblue30:
Merged, thanks.

PR #186, opened 2015-12-06
add parole.profile

avoidr:
Why are those entries scattered around in some profiles? Shouldn't they go to some common profile like disable-secret.inc?
```
blacklist ${HOME}/.pki/nssdb
blacklist ${HOME}/.lastpass
blacklist ${HOME}/.keepassx
```


netblue30:
Because I allow them only in browsers, mail and chat.

PR #187, opened 2015-12-06
whitelist keysnail config for firefox

rahiel:
This whitelists [keysnail](https://github.com/mooz/keysnail/wiki)'s config (extension like vimperator and pentadactyl). This is the default place for the config.


netblue30:
Merged, thanks.

PR #193, opened 2015-12-16
Make the build reproducible

reinerh:
Firejail is currently not building reproducibly [1].
The order of object files is random while linking. This is easily fixed by sorting the filelists.

A timestamp from the current date is also embedded into manpages.
By making use of the SOURCE_DATE_EPOCH variable [2] (when available), the timestamp can
be set to a deterministic one.

[1]: https://reproducible.debian.net/rb-pkg/unstable/armhf/firejail.html
[2]: https://reproducible-builds.org/specs/source-date-epoch/

### DIFF
--- a/mkman.sh
+++ b/mkman.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 sed "s/VERSION/$1/g" $2 > $3
-MONTH=`date +%b`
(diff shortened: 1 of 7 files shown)


netblue30:
Merged, thanks!

PR #198, opened 2015-12-19
use UTMP_FILE to check for its existence

Duncaen:
With musl libc `UTMP_FILE` is defined as `/dev/null/utmp`, on my system `/var/run/utmp` existed too but its just an empty file.
`fs_var_utmp` checks `/var/run/utmp` and then later opens `UTMP_FILE` resulting in following error:
```
Error fopen:fs_var_utmp(324): Not a directory
```


netblue30:
Merged, thanks!

PR #210, opened 2016-01-04
Disallow access to kdbx files

jrabe:
As of version 2.0, KeePassX now uses the `.kdbx` file extension instead of the already blacklisted `.kdb`.
See https://www.keepassx.org/news/2015/12/533 for more information.


netblue30:
Merged, thanks!

PR #213, opened 2016-01-08
dynamic allocation of noblacklist buffer

SeriousBug:
Instead of a statically allocated buffer for noblacklist entries, a buffer of same size is allocated dynamically instead. The size of the buffer is doubled every time it runs out of space.


netblue30:
Merged, thanks!

PR #217, opened 2016-01-11
Adds support for tmpfs-based profiles (profile-sync-daemon)

CrisBRM:
Useful for the people using profile-sync-daemon with chromium, which moves the folder to /tmp and symlinks it to ~/.config/chromium

Can anything be done to keep chromium from using its atrocious blue theme though?

### DIFF
--- a/etc/chromium.profile
+++ b/etc/chromium.profile
@@ -11,5 +11,6 @@ include /etc/firejail/disable-common.inc
 netfilter
 whitelist ${DOWNLOADS}
 whitelist ~/.config/chromium
+whitelist /tmp/home-chromium
 whitelist ~/.cache/chromium
 include /etc/firejail/whitelist-common.inc


netblue30:
Unfortunately we cannot whitelist /tmp directory. There are all kind of files there, mainly sockets for various system programs:
`````
$ ls -al /tmp/
total 36
drwxrwxrwt  7 root    root    4096 Jan 12 10:52 .
drwxr-xr-x 25 root    root    4096 Feb  8  2015 ..
drwxrwxrwt  2 root    root    4096 Jan 12 07:56 .ICE-unix
srwxr-xr-x  1 netblue netblue    0 Jan 12 10:52 .lxterminal-socket:0-netblue
srwxr-xr-x  1 netblue netblue    0 Jan 12 10:52 .menu-cached-:0-netblue
srwxr-xr-x  1 netblue netblue    0 Jan 12 10:52 .pcmanfm-socket--0-netblue
drwx------  2 netblue netblue 4096 Jan 12 10:52 pulse-WFDzkQE7jxwP
-rw-r--r--  1 root    root     804 Jan 12 07:56 rc.local.log
drwx------  2 netblue netblue 4096 Jan 12 10:52 ssh-PLi6xiNtNdW1
-r--r--r--  1 root    root      11 Jan 12 10:52 .X0-lock
drwxrwxrwt  2 root    root    4096 Jan 12 10:52 .X11-unix
````
This is also the place where X11 keeps the socket. Whitlisting will brake the desktop for a large number of users.

CrisBRM:
Whitelisting /tmp/home-chromium would retroactively whitelist /tmp?

netblue30:
Yes, if you whitelist anything under /tmp, all directory gets whitelisted.

johnp:
I'm using [profile-sync-daemon](https://wiki.archlinux.org/index.php/profile-sync-daemon) as user service which links `~/.mozilla/firefox/{profile}` to `/run/user/$UID/$USER-firefox-{profile}` and [anything-sync-daemon](https://wiki.archlinux.org/index.php/Anything-sync-daemon) as system service to link `~/.cache/mozilla/firefox` to `/run/user/$UID/asd-$USER/home/$USER/.cache/mozilla/firefox`. Both set the correct owner/group, but firejail only works if anything-sync-daemon is not running. Otherwise it errors out with `Error: invalid whitelist path /home/$USER/.cache/mozilla/firefox/`.
Things work if I set up asd to sync `~/.cache/mozilla/firefox/{profile}` instead. Otherwise the default profile causes an error at [fs_whitelist.c#L393-L398](https://github.com/netblue30/firejail/blob/63e0859305118a1c59d4f47eea4fb9352081b496/src/firejail/fs_whitelist.c#L393-L398).

Don't know if it makes sense to change firejails behaviour in this case, but just wanted to at least mention this workaround here.

netblue30:
OK, so you are using symlinks. The problem with whitelists is both the symlink and the real directory should be under /home/user directory - without this rule we run in security problems! Instead of /run/user/$UID, can you keep them in your home directory, something like /home/user/sync-daemon?

johnp:
I'd like to have avoided an extra fstab entry for a tmpfs in $HOME... and just to make this clear: At least everyone on Arch who is using profile-sync-daemon and firejail with default configuration has the (firefox) profile symlinked outside of $HOME without firejail complaining.
Is there a (security) difference between linking a whitelisted directory to somewhere outside of $HOME vs linking something _inside_ a whitelisted directory outside of home?

PR #236, opened 2016-01-19
Adding Seamonkey profiles

mahdi1234:
Adding both normal and -bin since gentoo has both packages and binary has suffix in executable.


netblue30:
Merged, thanks!

PR #243, opened 2016-01-20
Fixing lintian warnings

al42and:
Lintian (debian package static analyzer) has been reporting two issues:
 * malformed changelof: added required info to RELNOTES based on commit date
 * setuid binary: added it to lintian-overrides


netblue30:
Merged, thanks!

PR #244, opened 2016-01-20
changed typo in man file (namely --debug-blackilsts)

hheinz:
This fixes a typo in the man file. The option read --debug-blackilsts instead of --debug-blacklists.


netblue30:
Merged, thanks!

PR #250, opened 2016-01-21
Added Telegram profile

BaderSZ:
First commit for telegram.profile, with a basic (tested) configuration.


netblue30:
Merged, thanks!

PR #251, opened 2016-01-22
Blacklisting ~/.local/share/kwalletd

sinkuu:
(at least on Arch Linux) Kwallet saves encrypted passwords at `~/.local/share/kwalletd`.


netblue30:
Merged, thanks!

PR #255, opened 2016-01-23
Fix symlink invocation for programs placing symlinks in $PATH

sinkuu:
Some programs place symlinks in $PATH.


netblue30:
Merged, thanks!

Issue #256, opened 2016-01-23
Defunct orphans never reaped


creideiki:
I'm trying to run my e-mail klient (KMail) in Firejail 0.9.36. It works, but trying to use GnuPG leaves a lot of defunct processes around, so after a while my (global) process tree looks like this:

```
13943 creidei+  \_ /bin/bash
14230 root          \_ firejail kmail
14231 creidei+          \_ firejail kmail
14233 creidei+              \_ kmail
14276 creidei+              \_ [gpgconf] <defunct>
14278 creidei+              \_ [gpgconf] <defunct>
14280 creidei+              \_ [gpg2] <defunct>
14282 creidei+              \_ [gpgsm] <defunct>
14284 creidei+              \_ [gpgconf] <defunct>
14287 creidei+              \_ [gpg2] <defunct>
14290 creidei+              \_ [gpg2] <defunct>
14295 creidei+              \_ [gpg2] <defunct>
...
```

Apparently, KMail expects Init to wait() for those processes. However, PID 1 in the jail is the Firejail child, which does get the SIGCHLD signals, but ignores them since they aren't for the jailed application itself:

```
# strace -p 16071
Process 16071 attached
wait4(3, 0x7cf193bf34, 0, NULL)         = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=29, si_uid=1000, si_status=0, si_utime=7, si_stime=6} ---
wait4(3, 0x7cf193bf34, 0, NULL)         = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=32, si_uid=1000, si_status=0, si_utime=3, si_stime=3} ---
wait4(3, 0x7cf193bf34, 0, NULL)         = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=36, si_uid=1000, si_status=0, si_utime=8, si_stime=5} ---
wait4(3, 0x7cf193bf34, 0, NULL)         = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=39, si_uid=1000, si_status=0, si_utime=0, si_stime=3} ---
[...]

PR #257, opened 2016-01-23
Make the sandbox process reap all children.

creideiki:
Quick patch for netblue30/firejail#256

Instead of running waitpid(app) in the sandbox, run waitpid(-1) in a loop and check the return value until the original app dies.

I'm not sure you should trust me to write C code, but this patch has been running my KMail for several minutes now without eating any kittens :)


netblue30:
Merged, thanks.

PR #262, opened 2016-01-26
add Mathematica profile

rahiel:
This adds a profile for [Wolfram Mathematica](https://en.wikipedia.org/wiki/Mathematica).


netblue30:
Merged, thanks!

PR #265, opened 2016-01-27
remove duplicate include from uGet profile

pszxzsd:
Hey, thanks for the merge. I just saw you duplicated the whitelist include that was already in the profile, this deletes the first entry.


netblue30:
Sorry, my bad!

PR #267, opened 2016-01-28
Add alternative location for muttrc

manevich:
By default mutt looks for ~/.muttrc or ~/.mutt/muttrc files.


pigmonkey:
Is there any reason to not make all of `~/.mutt` read only?

manevich:
Well, `.mutt` directory can contain some non-rc files like alias file, that less sensitive and user may want to write there.
About blacklisting rc file - this is good idea, as `.muttrc` may contain passwords.
For some reason, no MUA in `/etc/firejail/disable-secret.inc`.
Probably I need to commit first one?

netblue30:
Let's put it in, if somebody complains about it we'll fix it later. Thanks!

PR #268, opened 2016-01-29
Fix for systems that don't have CAP_SYSLOG

mcarpenter:
CAP_SYSLOG was retroactively split from CAP_SYSADMIN (Linux
kernel commit ce6ada35bdf710d16582cc4869c26722547e6f11). Existing
supported systems might not yet have this commit (eg RHEL 6.6) in
which case compilation fails.


netblue30:
Merged, thanks!

PR #269, opened 2016-01-29
Include <sys/socket.h> for sa_family_t (RHEL 6.6)

mcarpenter:
Seems this is a long-running dependency issue: http://comments.gmane.org/gmane.linux.network/203185

I have tested on Ubuntu 14.04 and RHEL 6.6 but this probably needs also testing on RH/CentOS 7 and a more recent Ubuntu/Debian before merge.


netblue30:
Merged, thanks!

Issue #296, opened 2016-02-15
Various issues with building RPM files


jgriffiths:
Currently:

- There are 2 scripts to build rpms
- mkrpm.sh is completely broken and contains committed merge conflicts
- Both scripts are overly verbose
- Both scripts duplicate data held elsewhere (version, changelog) and so will always be out of date
- Both scripts fail to cleanup if interrupted
- The spec file is missing a library

The rpm building really should be/needs to be simplified and added to the make machinery.

h1z1:
See  #238 .  I believe they are working on it.

jgriffiths:
@h1z1 That issue isn't directly related although they obviously had issues with the RPM creation in 0.9.38 as shown by their comments about e.g. needing to add the the missing library to the spec file. The patch that closed that issue didn't touch the rpm creation scripts or the spec file at all.

The 0.9.39/HEAD rpm creation script has a merge conflict checked into it and cannot be run. Even if it was fixed and could be run, the spec file is currently incorrect. If that is fixed, the procedure to create an rpm is manual, laborious and error prone. In addition every time the current version or RELNOTES is changed the rpm script will need to be manually updated. This issue and associated pull request are to fix those problems.

netblue30:
https://github.com/netblue30/firejail/issues/238 - there was a problem with 0.9.38 on CentOS 6, /run directory does not exist on Centos 6, the user would have to create it manually. The problem is fixed now. I'll go ahead and merge https://github.com/netblue30/firejail/pull/297

netblue30:
All merged, thanks!

PR #297, opened 2016-02-15
RPM build fixes

jgriffiths:
Fixes the issues noted in issue #296 and in particular brings rpm building into the make system. With these changes the sequence './configure && make rpms' produces the app/debug/source rpms in cwd. I have built and installed the resulting rpm on Fedora 23 and verified that it works running Firefox.

More generally this removes redundant information that needs to be kept up to date manually, e.g. the version is passed by the Makefile automatically and added to the spec rather than being duplicated there. A future enhancement could add always-up-to-date changelog entries by munging RELNOTES.

Finally, this version is much, much simpler than the existing scripts at only 40 LOC.


netblue30:
Merged, thanks!

PR #299, opened 2016-02-16
Add epiphany profile

jrabe:
Added a profile for [GNOME Web](https://wiki.gnome.org/Apps/Web/). I've done _some_ testing, and everything seems to work fine. Since the developers are trying to keep the browser as simple as possible, I don't expect anything to break..


netblue30:
Merged, thanks!

PR #302, opened 2016-02-17
Add Polari profile

jrabe:
Added a profile for [Polari](https://wiki.gnome.org/Apps/Polari), an IRC client for GNOME.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -76,6 +76,7 @@ realinstall:
 	install -c -m 0644 .etc/audacious.profile $(DESTDIR)/$(sysconfdir)/firejail/.
 	install -c -m 0644 .etc/clementine.profile $(DESTDIR)/$(sysconfdir)/firejail/.
 	install -c -m 0644 .etc/epiphany.profile $(DESTDIR)/$(sysconfdir)/firejail/.
+	install -c -m 0644 .etc/polari.profile $(DESTDIR)/$(sysconfdir)/firejail/.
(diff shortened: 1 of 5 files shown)


netblue30:
Merged, thanks!

jrabe:
@netblue30 For me, Firejail does not automatically load this profile even though it's in the same directory (`/etc/firejail`) as the other profiles. The file permissions are also identical. It works if I manually specify it with `profile=/etc/firejail/polari.profile`. Is this just me or is anyone else having this issue?

netblue30:
Start polari in the sandbox, and add a --debug on the command line (firejail --debug polari). Cut&paste the output here.

jrabe:
Thanks, I didn't realize `/usr/bin/polari` was a symlink to `/usr/share/polari/org.gnome.Polari`. Unfortunately, it seems like Firejail discards everything after the first dot and only uses `org` as the command name. Do you think it's possible to have Firejail check for the full file name if `org.profile` doesn't exist? (If you want, I'll look into that this weekend when I have more time and open a pull request.)

jrabe:
Looks like GNOME uses `gapplication` to launch Polari (`gapplication launch org.gnome.Polari`). This makes checking the full profile name pretty useless, unfortunately.

netblue30:
Moving the issue here: https://github.com/netblue30/firejail/issues/324

jrabe:
Thanks! :+1:

PR #314, opened 2016-02-21
Epiphany: Fix settings being saved but not loaded

jrabe:
For some reason, Epiphany was able to save the settings even without access to the `.config/dconf` directory, but it did not show them in the settings window.


netblue30:
Merged, thanks!

netblue30:
I moved
`````
mkdir ~/.config
mkdir ~/.config/dconf
whitelist ~/.config/dconf
`````
in etc/whitelist-common.inc and modified epiphany and polari profiles.

jrabe:
Great, that makes it a lot easier. Thank you! :+1:

PR #317, opened 2016-02-23
man/firejail.txt: note you don't need --ip6= with SLAAC

yumkam:
If you obtain ipv6 connectivity from router via SLAAC/RA, IPv6 works without this option (and worked even before e669dbee639c4956430bc90345e1f05687c9d50a). Moreover, manual twiddling with IPv6 addresses/routes can do more harm than good in this configuration (e.g. when prefix is not static and changes midway due to ADSL reconnect, kernel handles this transparently, while with `--ip6=` option you are stick to (obsolete and non-working) address specified on firejail start).
Disclaimer: feel free to fix/polish wording/expand/etc, /me not a native speaker.

### DIFF
--- a/src/man/firejail.txt
+++ b/src/man/firejail.txt
@@ -471,6 +471,8 @@ Example:
 .br
 $ firejail \-\-net=eth0 \-\-ip6=2001:0db8:0:f101::1/64 firefox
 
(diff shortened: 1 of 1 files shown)


netblue30:
Merged, thanks!

Issue #290, opened 2016-02-11
firejail allows regular users to assign arbitrary address (IP, MAC) on any network interface


yumkam:
...and also disrespects any policy set by DHCP/etc.
(Obviously, it is only possible because firejail is setuid root, it is not possibly to do from within unprivileged userns (`unshare -r -n`) [unless "real root" will give them "real" netdev to play, with something like `ip link set $XXX netns $YYY`]).
(Sure, anyone with physical access to network can do the same, but still it requires intent, knowledge and can be noticed; but any little-educated user can run `firejail --net=eth0` and wreak havoc on LAN).
IMO, there should be (root-only writable) policy file, that restrict set of allowed interfaces, MAC and IP pools. By default, it should allow only `--net=none`; others are nice toy to play with [and I love them], but a way too powerful for regular users.
(There can be other ideas, as allow/enforce running some dhclient, but it's too complex, let's leave it for later).
Otherwise, IMO, firejail is unsafe to install as suid-root (unless it is built with --disable-network, or restricted to "power-user"/`sudo` group, etc).

genodeftest:
Wouldn't it be easier to drop the `setuid` bit? Linux namespaces doesn't require this stuff anyway.

netblue30:
It is not a security issue. It allows the user to set IP addresses on new interfaces created in a new network namespace inside the sandbox. Interfaces in the main network namespace are not affected.

yumkam:
@netblue30:
(nitpicking) Intefraces in main namespaces are affected: they are switched to promiscuous mode (with some network cards).
(Not nitpicking) It affects LAN security and breaks (some) local network policy. It is not possible to add new devices with new random MAC/IP addresses with user rights, it is possible with firejail.
Real problem: I have another machine on LAN, it has fixed (by dhcp) address. If I ran application in firejail while that machine is off, it happily hijacks address, and results in problems when this machine is turned on.
Real problem: some ISP forbids devices with unknown MACs on their network (and, of course, attempt to hijack someone's else IP address will be frowned upon).

yumkam:
[...]

PR #319, opened 2016-02-23
Add compile-time option to restrict --net= to root only

yumkam:
./configure --enable-network=restricted allows only `--net=none` to non-root users.

Other variants delegate too much power to non-root users and dangerous (it completely bypasses system-wide firewall and routing, it allows introducing arbitrary-chosen MAC and IP interfaces on LAN [disregarding DHCP policy], etc).

Root already had power to twiddle with anything, so no sense to restrain her, and `--net=none` looks safe enough (and still useful) for ordinary users (and permitted via `unshare -rn` anyway).

See #290


netblue30:
Merged, thanks!

PR #337, opened 2016-02-28
Added profiles for vivaldi

Fred-Barclay:
Hi netblue30. I've created two profiles for the vivaldi browser: etc/vivaldi.profile and etc/vivaldi-beta.profile. Since vivaldi is still in beta (and resides in /usr/bin/vivaldi-beta) I made vivaldi-beta.profile; however, it points back to vivaldi.profile in the same way that google-chrome-stable.profile points to google-chrome.profile.

I'm pretty new at GitHub so hopefully this is the correct way to go about this. I had to Google (well, DDG actually) to find instructions. :-) Let me know if this is okay! I've got a few other profiles I was thinking of committing if all goes well. 

Thanks,
Fred Barclay

### DIFF
--- a/etc/vivaldi-beta.profile
+++ b/etc/vivaldi-beta.profile
@@ -1,0 +1,2 @@
+# Vivaldi Beta browser profile
+include /etc/firejail/vivaldi.profile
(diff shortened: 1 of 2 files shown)


netblue30:
Yes, using include is fine, we already do exactly the same thing for firefox & iceweasel. Send all the profiles you have, thanks!

Fred-Barclay:
Will do. Palemoon will be done first and then (possibly) a few others.

Working on one for Virtual Box too but that's going to take longer than the others.

PR #345, opened 2016-03-02
More rpm fixes

jgriffiths:
A couple of rpm fixes, the first is needed to build rpms from the rc as it stands, the second saves space in /tmp while building.


netblue30:
Merged, thanks!

PR #346, opened 2016-03-02
add hedgewars profile (whitelist)

vn971:


vn971:
BTW, offtopic question. It wouldn't make sense to add "seccomp.keep" lists to profiles, would it?

I could add `seccomp.keep` filters for pidgin, qbittorrent, maybe some others. Possibly commented out with a notion on applicable OS (ubu14.04 in my case).

netblue30:
Merged, thanks!

seccomp.keep will break every time you update your software. It is strictly dependent on the version of the program, including all the libraries the program is using.

PR #363, opened 2016-03-10
Forward exit code from child process

### LINKED ISSUE #358: Firejail discards exit code
Hello,

we are using firejail to securely [run a compiler and linter](http://clippy.bashy.io/docs/clippy.html#section-4) in the [clippy-service](http://clippy.bashy.io/). First off, I wanted to thank you for this great project and the work you've put in it. It works great for us and allowed us to quickly setup a sandbox to allow third-party access without security risks. Thank you very much!

---

Coming to our specific problem: firejail currently **swallows the exit-code of the program it runs internally**. [As this gives us an indication whether the compiler exited successfully](https://github.com/ligthyear/clippy-service/issues/22) or whether there was a problem, this is a big problem for us.

I've identified that it would only require to keep the variable from the waitpid (instead of NULL) and forward that on process exit, but [that is currently not done](https://github.com/netblue30/firejail/blob/fca9d8e346cc4525159b9b2257387bf679a218a5/src/firejail/main.c#L1774-L1795). Would you accept a Pull-Request implementing this behavior? And/or do you want this new behavior be hidden behind a command-line option (like `--preserve-exit-code`)? I could prepare a PR about that but first wanted to make sure you'd want this, or maybe you had other good reasons to discard the exit code...

Highly appreciate you looking at this and all the work you've put in this already.
Thanks for your time!

ligthyear:
This PR reads the status code of the child process run inside the
sandbox and forwards it to the outer process. Which in turn makes that
result its own exiting status code.

This changes the behaviour of firejail in the sense that the following would prior give you the (misleading) status `0`:
```
$> firejail 'echo 1'
Parent pid 13488, child pid 13489
The new log directory is /proc/13489/root/var/log

Child process initialized
/bin/bash: echo 1: command not found

parent is shutting down, bye...
$> echo $?
127
```

Fixes #358.

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -1778,7 +1778,8 @@ int main(int argc, char **argv) {
 	signal (SIGTERM, my_handler);
 	
 	// wait for the child to finish
-	waitpid(child, NULL, 0);
+	int status = NULL;
+	waitpid(child, &status, 0);
 
 	// free globals
 #ifdef HAVE_SECCOMP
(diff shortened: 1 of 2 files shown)


netblue30:
Merged, thanks!

PR #368, opened 2016-03-17
add restrictions to chromium browser profile

vn971:
Note: this is a question-like PR. I see chromium working good
with those lines added, but it may be I'm missing something.

Merge with care.

### DIFF
--- a/etc/chromium.profile
+++ b/etc/chromium.profile
@@ -11,6 +11,8 @@ include /etc/firejail/disable-terminals.inc
 # include /etc/firejail/disable-devel.inc
 #
 
+caps.drop all
+noroot
 netfilter
 
 whitelist ${DOWNLOADS}


netblue30:
It breaks badly on Debian Jessie:
`````
$ firejail --caps.drop=all chromium
Reading profile /etc/firejail/chromium.profile
Reading profile /etc/firejail/disable-mgmt.inc
Reading profile /etc/firejail/disable-secret.inc
Reading profile /etc/firejail/disable-terminals.inc
Parent pid 1699, child pid 1700

Child process initialized

$ firejail --noroot chromium
Reading profile /etc/firejail/chromium.profile
Reading profile /etc/firejail/disable-mgmt.inc
Reading profile /etc/firejail/disable-secret.inc
Reading profile /etc/firejail/disable-terminals.inc
Parent pid 1712, child pid 1713

Child process initialized
[2:2:0317/084129:FATAL:setuid_sandbox_host.cc(158)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /usr/lib/chromium/chrome-sandbox is owned by root and has mode 4755.
`````

vn971:
I see. Thanks for the answer!

PR #371, opened 2016-03-20
fs_etc: continue to copy files if one fails.

fbarriga:
Currently if one file fails when copying private-etc files the process stop and doesn't copy the other files. Also it doesn't log which file failed.

# Example:
```
private-etc localtime,foo,resolv.conf
```

# Before:
only copied `localtime`

## log
```
/run/firejail/mnt/cp: failed to restore the default file creation context: No such file or directory
```

# Now:
copy `localtime` and `resolv.conf`

## log:
```
Warning (fs_etc): error copying file /etc/foo, skipping...
```


netblue30:
Merged, thanks!

PR #391, opened 2016-03-28
fix seccomp filter (32bit/64bit)

vn971:
It sees that seccomp filter numbers were misplaced
between different architectures.

### DIFF
--- a/src/firejail/seccomp.c
+++ b/src/firejail/seccomp.c
@@ -469,10 +469,10 @@ int seccomp_filter_drop(int enforce_seccomp) {
 	// default seccomp
 	if (cfg.seccomp_list_drop == NULL) {
 #if defined(__x86_64__)
-		seccomp_filter_32();
+		seccomp_filter_64();
 #endif
 #if defined(__i386__)
-		seccomp_filter_64();
+		seccomp_filter_32();
 #endif
 
 #ifdef SYS_mount		


netblue30:
I have to reject it. The ideais to always build two seccomp filters, one for 32-bit and one for 64-bit, in order to be able to filter 32-bit programs on 64-bit architectures - for example Steam and Wine are always 32-bit. Thanks anyway!

vn971:
But wait, the current implementation of `firejail` forbids the system call 123 (setfsgid) just to block a different system call on a different architecture.
[1] https://github.com/torvalds/linux/blob/1e75a9f34a5ed5902707fb74b468356c55142b71/arch/x86/entry/syscalls/syscall_64.tbl#L132
[2] https://github.com/torvalds/linux/blob/1e75a9f34a5ed5902707fb74b468356c55142b71/arch/x86/entry/syscalls/syscall_32.tbl#L132
Is it really intended?

Are you sure that a 32-bit program (wine or whatever) would be able to make the system call `modify_ldt` if we forbid the system call 154 (`modify_ldt`) on our 64bit system? Note that the 64 source file I linked [1] says "The abi is "common", "64" or "x32" for this file", and the system call `modify_ldt` is marked as "common".
Anyway, shortening my question. Do you have a test to cover this use case?

netblue30:
Yes, the test is in test/seccomp-dualfilter.exp for mount syscall. During the test two messages are generated in syslog:
`````
Mar 29 08:58:42 debian kernel: [64376.848556] audit: type=1326 audit(1459256322.359:14): auid=1000 uid=1000 gid=1000 ses=1 pid=29693 comm="syscall_test" exe="/home/netblue/work/github/firejail/src/tools/syscall_test" sig=31 arch=c000003e syscall=165 compat=0 ip=0x7f6c5aae9fda code=0x0
Mar 29 08:58:43 debian kernel: [64377.935602] audit: type=1326 audit(1459256323.451:15): auid=1000 uid=1000 gid=1000 ses=1 pid=29697 comm="syscall_test32" exe="/home/netblue/work/github/firejail/src/tools/syscall_test32" sig=31 arch=40000003 syscall=21 compat=1 ip=0xf776fbb5 code=0x0
`````

PR #392, opened 2016-03-28
Fix xephyr methods referring to xpra

jrabe:
A few lines in the Xephyr methods were still referring to or using xpra methods (most importantly line 176).
Does the if-statement after line 169 ("xpra does a number of weird things") also apply to Xephyr?


netblue30:
Merged, thanks!

PR #405, opened 2016-04-02
"/etc/password" -> "/etc/passwd"

avoidr:
Rather unfortunate commit message, I think, but please don't kill me.


netblue30:
Merged, thanks!

PR #406, opened 2016-04-02
add cmus.profile

avoidr:
At last, goddamn.
Or is there something I'm still missing? RELNOTES? disable-programs.inc?

Does it have to be ${HOME}, by the way? Or can it be ~/ ? Is any one preferable?


netblue30:
Merged, thanks!

PR #409, opened 2016-04-04
Allow alternative opera config path

reinerh:
Petter Reinholdtsen reported in [1] that ~/.opera is used by older Opera versions, so this path should also be allowed.

[1] https://bugs.debian.org/819950


netblue30:
Merged, thank you for the patch.

Issue #343, opened 2016-03-01
--blacklist=~/.mozilla warns of invalid file


Fred-Barclay:
Firejail 0.9.38 on LMDE Betsy (closely related to Debian Jessie.)

I came across this while working on the profile for the Pale Moon browser. ~/.mozilla is viewable and read-write even when not whitelisted. Attempts to blacklist the directory, whether in the profile or in the startup arguments, fail with the warning that ~/.mozilla is an invalid file (which is half-way correct; it's a directory). However, blacklisting other directories, such as ~/.Downloads, succeeds.

`fred@aussie ~ $ firejail --blacklist=~/.mozilla --debug palemoon | grep mozilla`
`Reading profile /home/fred/.config/firejail/palemoon.profile`
`Reading profile /etc/firejail/disable-mgmt.inc`
`Reading profile /etc/firejail/disable-secret.inc`
`Reading profile /etc/firejail/disable-common.inc`
`Reading profile /etc/firejail/disable-devel.inc`
`Reading profile /etc/firejail/whitelist-common.inc`
`...`
`Debug 337: new_name #/home/fred/.themes#`
`Warning: /home/fred/.mozilla is an invalid file, skipping...`

When I blacklist a different directory such as Downloads:
`fred@aussie ~ $ firejail --blacklist=~/Downloads --debug palemoon | grep Downloads`
`Reading profile /home/fred/.config/firejail/palemoon.profile`
`Reading profile /etc/firejail/disable-mgmt.inc`
`Reading profile /etc/firejail/disable-secret.inc`
`Reading profile /etc/firejail/disable-common.inc`
`Reading profile /etc/firejail/disable-devel.inc`
`Reading profile /etc/firejail/whitelist-common.inc`
`...`
`Disable /home/fred/Downloads`

In case the profile is to blame:
`# Pale Moon Browser profile`
`include /etc/firejail/disable-mgmt.inc`
`include /etc/firejail/disable-secret.inc`
`include /etc/firejail/disable-common.inc`
`include /etc/firejail/disable-devel.inc`
`caps.drop all`
`seccomp`
`protocol unix,inet,inet6,netlink`
`netfilter`
`tracelog`
`noroot`
`whitelist ${DOWNLOADS}`
` `
`whitelist ~/.moonchild productions`
`whitelist ~/.cache/moonchild productions/pale moon`
` `
`# Uncommented in Firefox profile. If you run into trouble you may want to enable`
`# (some of) them.`
[...]

PR #432, opened 2016-04-13
Pale Moon profile && python blacklists

Fred-Barclay:
Hey netblue30! This should (hopefully) be working well now. I grabbed the firejail_0.9.40-rc1_1_amd64.deb from Sourceforge and replaced my usual 0.9.38 install from the debian testing repos--you've been pretty busy since my last attempt at a Pale Moon profile!

Re: https://github.com/netblue30/firejail/issues/343
`.mozilla` is still visible, but the only contents visible in it is the `extensions` subdirectory. The contents of `extensions` are not visible to Pale Moon so I feel pretty comfortable with this.
(Though I still would like to be able to blacklist `.mozilla` once and for all.) :-)

EDIT: just added python2 and python3 blacklists to etc/disable-devel. Would have made a separate pull request but apparently GitHub is smarter than I... :-)

### DIFF
--- a/etc/disable-devel.inc
+++ b/etc/disable-devel.inc
@@ -34,3 +34,18 @@ blacklist /usr/lib/php*
 # Ruby
 blacklist /usr/bin/ruby
 blacklist /usr/lib/ruby
+
+# Python 2
+blacklist /usr/bin/python2*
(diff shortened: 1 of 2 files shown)


netblue30:
Merged, thanks!

PR #434, opened 2016-04-13
Separated thunderbird/icedove profiles

Fred-Barclay:
This makes it easier for the weirdos like me who use both Thunderbird and Icedove. :) TB can't see ID's profile; ID can't see TB's config.


netblue30:
Merged, thanks.

Fred-Barclay:
No worries, mate. :)
I think there might be a bit of confusion, though. `/etc/firejail/icedove.profile` already existed, but only pointed to the thunderbird profile via `include /etc/firejail/thunderbird.profile`. The TB profile then whitelisted both the TB config files and the ID config files.

All I did was rewrite icedove.profile to contain its own profile rather than point to the thunderbird.profile, and removed the icedove config from the thunderbird profile.

PR #450, opened 2016-04-17
Blacklist vera crypt

Fred-Barclay:
Found this while working on another profile and thought it might be a good idea to blacklist VeraCrypt and its directories. Just in case... :)

I took the somewhat unusual step of blacklisting the veracrypt uninstall script too--I could think of a couple of rare instances to make it an attack vector.

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -26,6 +26,14 @@ blacklist ${HOME}/.VirtualBox
 blacklist ${HOME}/VirtualBox VMs
 blacklist ${HOME}/.config/VirtualBox
 
+# VeraCrypt
+blacklist /usr/bin/veracrypt
+blacklist /usr/bin/veracrypt-uninstall.sh
+blacklist /usr/share/veracrypt
+blacklist /usr/share/applications/veracrypt.*
+blacklist /usr/share/pixmaps/veracrypt.*
+blacklist ${HOME}/.VeraCrypt
+
 # var
 blacklist /var/spool/cron
 blacklist /var/spool/anacron
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -27,8 +27,8 @@ blacklist ${HOME}/VirtualBox VMs
 blacklist ${HOME}/.config/VirtualBox
 
 # VeraCrypt
-blacklist /usr/bin/veracrypt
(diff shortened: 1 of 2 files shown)


etc/disable-common.inc
```diff
@@ -26,6 +26,14 @@ blacklist ${HOME}/.VirtualBox
 blacklist ${HOME}/VirtualBox VMs
 blacklist ${HOME}/.config/VirtualBox
 
+# VeraCrypt
+blacklist /usr/bin/veracrypt
+blacklist /usr/bin/veracrypt-uninstall.sh
```
avoidr: You probably meant to use ${PATH} instead of an absolute path, didn't you?

Fred-Barclay: Yep-should have done that. Thanks!

Fred-Barclay: Better now? Thanks for checking!

avoidr: Yes, better. You're welcome and also thanks!

avoidr:
What rare instances could you think of, by the way?

Fred-Barclay:
To be fair it would (generally) involve changing the script. But if we can assume an attacker has rwx privs. to the uninstall script (no compiler needed, it's not a binary) , even if he doesn't to the other binaries^, then he could modify and execute the script.
`mv /my/evil/veracrypt /usr/bin/veracrypt`
clean log files
And so on

Of course that's more of a script kiddie approach. A slightly smarter guy would use *any other* binary if at all possible. Or maybe it's the other way around... lol
 And I haven't actually tried this myself yet (might be fun, but don't have time).

^Highly unlikely, but work with me here. To do this he'd likely have root privileges, and if he has root, then why not take one of the hundreds of other approaches?

OR-and this is the more likely approach- we don't insist on taking control right then and there, but wait. What about adding a line to the script that shuts down his firewall (or simply opens a port and disables logging)
`ufw disable`
`systemctl disable ufw.service #For systemd machines`
Adequate log inspection would quickly reveal that ufw wasn't running, of course.

Or if we want to have some fun (especially on a UEFI machine)
`rm -rf /`
or `mv / /dev/null`
The script runs with root privs after all, and the user isn't going to double-check it before running.

Or, or...

So not a big threat as it's all script-kiddie level, but I figured "Why not?"

netblue30:
Merged, thanks!

PR #451, opened 2016-04-17
Removed thunderbird todo

Fred-Barclay:
"11. cleanup thunderbird profile - disable-common was commented out" isn't an issue now that the thunderbird profile includes the firefox profile and therefore includes disable-common.


netblue30:
merged, thanks!

PR #457, opened 2016-04-19
Aweather && Stellarium

Fred-Barclay:
Profiles for aweather and stellarium. Aweather is a weather programme (shows radar, warnings, etc; though unfortunately only for the States) and Stellarium is a planetarium.

I also added 0ad to the list of games in src/firecfg/firecfg.config since I'd missed that earlier.


netblue30:
merged, thanks.

PR #466, opened 2016-04-21
HexChat/Atril profile fix

Fred-Barclay:
Hello netblue. I've started building firejail from this repo every day, so I found that my recent commit blacklisting python had broken hexchat (which apparently requires python libs). I've just fixed that with this commit. 

For Atril, I added the printing config directory, and changed `netfilter` to `net none.` I can't think of a good reason why a pdf viewer would need to go online. 

Cheers!

### DIFF
--- a/etc/hexchat.profile
+++ b/etc/hexchat.profile
@@ -1,5 +1,6 @@
 # HexChat instant messaging profile
 noblacklist ${HOME}/.config/hexchat
+noblacklist /usr/lib/python2*
 include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-programs.inc
 include /etc/firejail/disable-devel.inc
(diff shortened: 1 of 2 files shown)


etc/hexchat.profile
```diff
@@ -1,5 +1,6 @@
 # HexChat instant messaging profile
 noblacklist ${HOME}/.config/hexchat
+noblacklist /usr/lib/python2*
```
reinerh: What about python3, perl, and other scripting languages supported?

reinerh: I don't use hexchat, so I don't know what is really required.
I just thought other scripting languages would also be required to run plugins/addons.

Fred-Barclay: Gotcha.

Well, the others would have been blacklisted anyways, so I think it's
best to leave it like this.

Cheers!

netblue30:
Merged, thanks.

PR #474, opened 2016-04-22
Create quiterss.profile

curiosity-seeker:
I suggest to add a profile for QuiteRss. The attached profile works for me on Manjaro.

### DIFF
--- a/etc/quiterss.profile
+++ b/etc/quiterss.profile
@@ -1,0 +1,29 @@
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-programs.inc
+include /etc/firejail/disable-passwdmgr.inc
+include /etc/firejail/disable-devel.inc
+
+whitelist ${HOME}/quiterssfeeds.opml
+mkdir ~/.config
+mkdir ~/.config/QuiteRss
+whitelist ${HOME}/.config/QuiteRss/
+whitelist ${HOME}/.config/QuiteRssrc
+mkdir ~/.local
+mkdir ~/.local/share
+whitelist ${HOME}/.local/share/
+mkdir ~/.cache
+mkdir ~/.cache/QuiteRss
+whitelist ${HOME}/.cache/QuiteRss
+
+caps.drop all
+seccomp
+protocol unix,inet,inet6
+netfilter
+tracelog
+noroot
+nogroups
+shell none
+private-dev
+private-bin quiterss
+#private-etc X11,ssl
+include /etc/firejail/whitelist-common.inc


netblue30:
Merged, thanks!

vn971:
Nice profile!
Am I right that you intend to have a "principle of least privelege" profile, aka "whitelist" ?

If that is the goal, you may be interested to add `private-tmp=/tmp/.X11-unix`

Also, does this exact profile really work for you? I think I also needed `whitelist ~/.ICEauthority` on my system.

And my last consideration. How do you think, would it make sense to move your ompl file just inside `~/.config/QuiteRss`? The reason for my question is this configuration `whitelist ${HOME}/quiterssfeeds.opml` which seems to apply to your computer only.

curiosity-seeker:
@vn971 : Thanks for your suggestions!

Regarding private-tmp: According to man firejail-profile this switch doesn't support arguments. Unless @netblue30 changed this recently. Anyways, with or without this argument QuiteRss doesn't start if I add that to the profile.

Yes, the profile works for me without `whitelist ~/.ICEauthority`. As a matter of fact I don't have that file on my system. Which distro are you using?

Regarding quiterssfeeds.opml: I couldn't find a setting in QuiteRss where this file should be stored. Isn't ${HOME} the default? I don't remember that I moved it manually.

vn971:
@curiosity-seeker, Sorry, you're right, it's not `private-tmp /tmp/.X11-unix`, it's `whitelist /tmp/.X11-unix`. Like this: https://github.com/netblue30/firejail/blob/5836a8892ea1cbeb6fddab71ced4634ba1d1bbcd/etc/wesnoth.profile#L18

Regarding ompl files. Well, there shouldn't be `quiterssfeeds.opml` at all, that's the trick. An *.opml file is created when a user tries to export RSS feeds into this (commonly used) format. If you never export - it will never be created. Furthermore, if you _do_ export, quiterss will ask you to enter the file name (it will NOT be "quiterssfeeds" by default).

Regarding `whitelist ~/.ICEauthority`. I was needed for me on both Ubuntu-14.04 and ArchLinux. But if it works for you now, let's leave it as is. I have a custom profile for quiterss anyway, maybe I messed something up a bit.

vn971:
BTW, now that I looked into the profile, I also see a custom file here:
`whitelist ${HOME}/.config/QuiteRssrc`

It's not a default one, too. Check out creation/modification time of the file if in doubt.

curiosity-seeker:
@vn971 : 
>it's whitelist /tmp/.X11-unix

If I add this, QuiteRss doesn't launch.

>Regarding ompl files. Well, there shouldn't be quiterssfeeds.opml at all, that's the trick. An *.opml file is created when a user tries to export RSS feeds into this (commonly used) format. If you never export - it will never be created. Furthermore, if you do export, quiterss will ask you to enter the file name (it will NOT be "quiterssfeeds" by default).

Aha, thanks. I had done this once but didn't remember. That's why I have ~/.config/QuiteRssrc. It just contains:

`[KFileDialog Settings]
Recent Files[$e]=quiterssfeeds.opml,$HOME/quiterssfeeds.opml`

>Regarding whitelist ~/.ICEauthority. I was needed for me on both Ubuntu-14.04 and ArchLinux. 

That's strange I'm using Manjaro which is an Arch derivate but I don't have this file ...

curiosity-seeker:
Unfortunately, opening links in the external browser doesn't work anymore. Adding 

`include /etc/firejail/firefox.profile`

Issue #472, opened 2016-04-22
Can escape jail with mate-terminal


Fred-Barclay:
This is somewhat similar to #471 but works from **inside** the jail:  
I can blacklist a directory, start a terminal (I've tried mate-terminal) from inside the firejail, and then escape the jail completely with that separate terminal. I can even gain root privs via sudo, which is normally impossible inside the jail:

```
fred@aussie ~ $ firejail --blacklist=$HOME/secret
Reading profile /etc/firejail/generic.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-passwdmgr.inc

** Note: you can use --noprofile to disable generic.profile **

Parent pid 6613, child pid 6614

Child process initialized
[fred@aussie ~]$ ls secret
ls: cannot open directory secret: Permission denied
[fred@aussie ~]$ pwd
/home/fred
[fred@aussie ~]$ mate-terminal
```
As you can see, I can't get into `secret` and have done nothing unusual from terminal (so none of the `mkdir...cd` and having another, un-jailed terminal mv the directory, as in #471 ).  
Having started a terminal, I can now have the real fun:

```
[fred@aussie ~]$ pwd
/home/fred
[fred@aussie ~]$ cd secret
[fred@aussie secret]$ pwd
/home/fred/secret
[fred@aussie secret]$ ls
TOPSECRET
[fred@aussie secret]$ cat TOPSECRET
TOP SECRET!
TO...
President Petrov, Президент Российской Федерации

Operation Lovejoy has been finalised by the Company. The Spetsnaz and Delta 
Forces will apprehend General Radek tonight at 2100. President Marshall offers
his full support and best wishes.
[fred@aussie secret]$ sudo su root
[sudo] password for fred:
aussie secret # whoami
root
```
I can even create and delete files, and change their permissions
```
[fred@aussie secret]$ echo ":(){ :|:& };:" >> fork.sh
[fred@aussie secret]$ cat fork.sh
:(){ :|:& };:
[fred@aussie secret]$ ls -l fork.sh
-rw-r--r-- 1 fred fred 14 Apr 22 23:58 fork.sh
[fred@aussie secret]$ chmod 777 fork.sh
[fred@aussie secret]$ ls -l fork.sh
-rwxrwxrwx 1 fred fred 14 Apr 22 23:58 fork.sh
[fred@aussie secret]$ rm fork.sh
```
[...]

PR #476, opened 2016-04-22
blacklisted additional terminals

Fred-Barclay:
These are the additional terminals that can escape firejail as in #472 as tested by xenopeek.

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -143,3 +143,8 @@ blacklist ${PATH}/xfce4-terminal
 blacklist ${PATH}/xfce4-terminal.wrapper
 blacklist ${PATH}/mate-terminal
 blacklist ${PATH}/mate-terminal.wrapper
+blacklist ${PATH}/lilyterm
+blacklist ${PATH}/pantheon-terminal
+blacklist ${PATH}/roxterm
+blacklist ${PATH}/roxterm-config
+blacklist ${PATH}/terminix
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -148,3 +148,5 @@ blacklist ${PATH}/pantheon-terminal
 blacklist ${PATH}/roxterm
 blacklist ${PATH}/roxterm-config
 blacklist ${PATH}/terminix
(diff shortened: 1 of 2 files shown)


etc/disable-common.inc
```diff
@@ -143,3 +143,8 @@ blacklist ${PATH}/xfce4-terminal
 blacklist ${PATH}/xfce4-terminal.wrapper
 blacklist ${PATH}/mate-terminal
 blacklist ${PATH}/mate-terminal.wrapper
+blacklist ${PATH}/lilyterm
+blacklist ${PATH}/pantheon-terminal
+blacklist ${PATH}/roxterm
+blacklist ${PATH}/roxterm-config
+blacklist ${PATH}/terminix
```
reinerh: Please also add urxvtc.

Fred-Barclay: Will do. Can you tell me where it is?
`which urxvtc`

reinerh: /usr/bin/urxvtc

Fred-Barclay: One final thing (I think): `ls /usr/bin/ | grep *urxvtc*`

Some of these guys (like gnome-terminal) have additional files

reinerh: There is also /usr/bin/urxvtcd, which is a wrapper (shell script) around urxvtc, which starts a server process if none is running. I don't think this one has to be added.

Fred-Barclay: How's this?
I felt safer blacklisting urxvtcd as well. :-) If urxvtc is blacklisted, urxvtcd has no use, so why not? That was my reasoning, anyways... :)

reinerh: Thanks for adding it.
My thought was that when urxvtc is blacklisted, urxvtcd is also no danger, so it's not necessary to blacklist it.
But blacklisting also doesn't harm...

Fred-Barclay: Yeah, I'm a worrywart. :laughing:

netblue30:
Merged, thanks.

PR #478, opened 2016-04-23
Fixes for hexchat.profile

ruany:
HexChat uses python3 under Arch Linux, so this is required for it to launch.
Also, links can't be clicked on since whitelist-common.inc isn't currently included (launching default browser requires mime types, otherwise it falls back to launching "perl" and fails).


netblue30:
Merged, thanks.

PR #479, opened 2016-04-24
cyberfox profile

figue:
I was testing this fork of Firefox these days. Could be useful for some people.
I didn't modified Firefox base profile, just the properly directories.


netblue30:
Merged, thanks.

PR #486, opened 2016-04-27
Make some tests more robust

reinerh:


netblue30:
Merged, thanks!

netblue30:
I had to remove "cd /home" from test/environment/zsh.exp, csh.exp and rlimit-profile.exp. It was braking the test.

reinerh:
Hm, without chaning the directory the test doesn't work when it's run in /tmp for example, because of pwd.
Do you have a different workaround for that?

netblue30:
We can take out pwd. I'll give it a try under /tmp.

netblue30:
"make test-environment" is fixed. I'll bring in a number of similar fixes for all other tests, I'll make sure everything runs fine under /tmp.

PR #490, opened 2016-04-29
Mark skipped tests differently

reinerh:
This makes it easier to see in the result if tests were skipped.


netblue30:
Merged, thanks!

PR #499, opened 2016-05-05
Use locale-independent sorting

reinerh:
On systems without installed/configured locales, the file
list is sorted by byte comparison.
On some locales it is sorted differently (e.g. X11 at the end).


netblue30:
Merged, thanks.

PR #502, opened 2016-05-07
Xapps

Fred-Barclay:
I noticed a somewhat serious bug with atril and firejail (atril wouldn't launch, if that counts as "serious" :laughing:) so this should patch that. I'm going to try and test deeper over the next few days and see if I can get a better profile, but I wanted to go ahead and get this patched ASAP.

I also added support for three of the four new Linux Mint xapps--xplayer, xreader, and xviewer (forked from totem, atril, and eog, respectively) and cleaned up a little in etc/disabled-programs.inc.

Hope this helps!
Fred


netblue30:
Merged, thanks.

Fred-Barclay:
#505 will handle this. Thanks again. :smile:

PR #521, opened 2016-05-15
Revert "cmus.profile: use empty /etc"

avoidr:
This reverts commit 471bba3242b01d91e6b1f52b9b12d2e88b5cf533.

Fucked up testing before, sorry.
Tried `--private-etc=null` on the CLI, while I had `--private-etc=group` in the profile and thought the CLI would override the profile, but it turns out it's complementary.


netblue30:
Merged, thanks.

PR #526, opened 2016-05-18
fix manual: --whitelist dir inside --read-only dir

vn971:
The original command did not make any sense (as I think), because
`--whitelist=~/work` would prevent `--read-only=~` from
having any real effect anyway.

I assume it was a typo.

### DIFF
--- a/src/man/firejail.txt
+++ b/src/man/firejail.txt
@@ -1138,7 +1138,7 @@ A short note about mixing \-\-whitelist and \-\-read-only options. Whitelisted d
 should be made read-only independently. Making a parent directory read-only, will not
 make the whitelist read-only. Example:
 .br
-$ firejail --whitelist=~/work --read-only=~/ --read-only=~/work
+$ firejail --whitelist=~ --read-only=~ --whitelist=~/work
 
 .TP
 \fB\-\-rlimit-fsize=number


netblue30:
No, It is like this: let's say you whitelist ~/.config/transmission, and make ~ read-only:
`````
$ firejail --read-only=~ --whitelist=~/.config/transmission
`````
In this case the files in your home directory will be read-only, with the exception of ~/.config/transmission. This one is a separate mount point, and it is still read-write. You would also have to make this mount point read-only:
`````
$ firejail --read-only=~ --whitelist=~/.config/transmission --read-only=~/.config/transmission
`````

vn971:
> In this case the files in your home directory will be read-only

AFAIK, they will not be readable at all. `--whitelist=~/.config` forbids access to other parts of `~`.

netblue30:
You are right here, ~/.config/transmission is a bad example. That's why I used a generic ~/work directory in the example.

vn971:
But in the current example with `~/work`, there would only be read-only access, and only to `~/work`. No access to anything else, no real whitelisting.

netblue30:
Yes, because it is a whitelist.

vn971:
@netblue30 how is this example different from `firejail --whitelist=~/work read-only=~/work` then? Why add `--read-only=~` if it does not do anything?

PR #536, opened 2016-05-25
Enable using the NO_NEW_PRIVS prctl(2) flag

KellerFuchs:
> Any task can set `no_new_privs`.  Once the bit is set, it is inherited across `fork`,
> `clone`, and `execve` and cannot be unset.  With `no_new_privs` set, `execve`
> promises not to grant the privilege to do anything that could not have been
> done without the `execve` call.  For example, the setuid and setgid bits will no
> longer change the uid or gid; file capabilities will not add to the permitted set,
> and LSMs will not relax constraints after `execve`.
> -- [Linux kernel documentation](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt)

Judicious use of `no_new_privs` can both:
- prevent sandboxed applications from (attempting to) exploit privileged (setuid, setgid or fcaps) binaries;
- prevent malicious users from (attempting to) exploit privileged binaries *using firejail's features*, by setting `force-nonewprivs` system-wide.

This PR introduces the support for `NO_NEW_PRIVS` and makes it configurable through the `nonewpriv` directive, documents it, and introduces a `force-nonewprivs` configuration item.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,13 @@ The following features can be enabled or disabled:
 
        x11    Enable or disable X11 sandboxing support, default enabled.
 
+       force-nonewprivs
+              Force use of theh NO_NEW_PRIVS prctl(2) flag.
+              This mitigates the possibility of a user abusing firejail's
+              features to trick a privileged (suid or file capabilities)
+              process into loading code or configuration that is partially
+              under their control.  Default disabled
+
        xephyr-screen
               Screen    size    for   --x11=xephyr,   default   800x600.   Run
               /usr/bin/xrandr for a full list of resolutions available on your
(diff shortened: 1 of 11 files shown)


KellerFuchs:
It would be super-helpful for this to land in 0.9.40, because it augments the `firejail.config` mechanism that gets introduced in that release in, IMO, a very useful way for administrators of shared systems.

netblue30:
Merged! I'll push the new release over the weekend, thanks.

KellerFuchs:
@netblue30 Awesome, thanks  <3

netblue30:
I'll add also compile-time support for it, thanks!

KellerFuchs:
@netblue30 Ah, something like the compile-time `restricted-network`?
That's indeed a pretty great idea.

PR #537, opened 2016-05-25
Make `restricted-network` prevent use of netfilter

KellerFuchs:
I suggest `netfilter` to be affected by `restricted-network` because it is a traditionally-privileged interface to the kernel, and the security impact of letting untrusted users run their own netfilter scripts (even on their own network namespace) is non-obvious.

Is that OK, or should I introduce a separate config flag for this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,9 +197,9 @@ The following features can be enabled or disabled:
        restricted-network
               Enable  or disable restricted network support, default disabled.
               If enabled, networking features should also be enabled  (network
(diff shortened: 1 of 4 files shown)


KellerFuchs:
PS: Since users can only use `--net=none` when `restricted-network` is set, there is no feature-loss with this change.

netblue30:
Merged, thanks.

KellerFuchs:
Thanks a bunch.

PR #538, opened 2016-05-25
Extend profiles to use the new `nonewprivs` feature

KellerFuchs:
This adds `nonewprivs` in profiles where I thought it made sense.

**WARNING**: I didn't test most profiles, simply because I don't use the corresponding software.
I'm only submitting this PR here so that other people know where it might be interesting to add `nonewprivs`.

One specific caveat deserves mention: some `sendmail` implementations are broken by `NO_NEW_PRIVS`, because they rely on the setgid bit to be able to write new mails directly in the mail queue, so in general this cannot be set on any program that may call (directly or not) `sendmail`.

### DIFF
--- a/etc/0ad.profile
+++ b/etc/0ad.profile
@@ -12,6 +12,7 @@ protocol unix,inet,inet6,netlink
 netfilter
 tracelog
 noroot
+nonewprivs
 
 # Whitelists
 noblacklist ~/.cache/0ad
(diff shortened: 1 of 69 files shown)


netblue30:
Merged, thanks.

KellerFuchs:
@netblue30 You read the part about me not having tested it, right?

netblue30:
It works fine, thanks!

KellerFuchs:
Great, then.

PR #539, opened 2016-05-25
Various

Fred-Barclay:
This adds a profile for the Brave browser, and I added "nosound" to the atril and xreader profiles (since pdf's shouldn't really be making noise :smile_cat: )

I also blacklisted "shutdown" in seccomp for amd64 machines--apparently it's not a syscall for 32-bit kernels??? "Reboot" is already blacklisted and I figured that shutdown should be too. ;)

Cheers!
Fred


reinerh:
shutdown syscall is not what you think it is.
It is used for shutting down network connections, not for halting the system.
Please don't blacklist it, it is harmless.

Fred-Barclay:
Thanks reinerh. :)

PR #540, opened 2016-05-26
Brave

Fred-Barclay:
This adds a profile for the Brave browser, and I disabled sounds in atril and xreader since a (normal) pdf shouldn't be singing. :smile_cat: Maybe this would be a good idea for Evince as well?


netblue30:
Merged, thanks.

PR #542, opened 2016-05-27
Various

Fred-Barclay:
Gitter now has a profile.
I was looking into src/firecfg/firecfg.conf and realised that several profiles weren't mentioned in it. I've added them and alphabetised the file.

I also alphabetised the security filters in all the profiles, to hopefully make them even easier to read. There were a few redundancies (weechat had netfilter twice IIRC) so they *should* be gone as well. :wink: 
Hope this helps!

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -188,6 +188,7 @@ realinstall:
 	install -c -m 0644 .etc/konversation.profile $(DESTDIR)/$(sysconfdir)/firejail/.
 	install -c -m 0644 .etc/psi-plus.profile $(DESTDIR)/$(sysconfdir)/firejail/.
 	install -c -m 0644 .etc/brave.profile $(DESTDIR)/$(sysconfdir)/firejail/.
+	install -c -m 0644 .etc/gitter.profile $(DESTDIR)/$(sysconfdir)/firejail/.
 	sh -c "if [ ! -f $(DESTDIR)/$(sysconfdir)/firejail/login.users ]; then install -c -m 0644 etc/login.users $(DESTDIR)/$(sysconfdir)/firejail/.; fi;"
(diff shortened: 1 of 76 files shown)


KellerFuchs:
FWIW, I'm not sure that alphabetizing all profiles makes them easier to read: for instance, most profiles currently have `protocol` and `netfilter` next one to another because they are both network related.
With the alphabetization, I would have to scan through the whole file to find the 2 lines relevant to network.

Fred-Barclay:
Perhaps so. Maybe I shouldn't say "easier" as an absolute--it's easier for me at any rate. :D

netblue30:
Merged! It didn't make it however in 0.9.40,

Fred-Barclay:
Thanks netblue30. 
BTW: what's the difference between the red "closed" tag after you merged this PR, and the normal purple "merged" tag? Is there something I should do to make it more consistent?

netblue30:
It was a manual merge, the automatic one didn't work because of the release I made after you submitted the pull request. So, all is fine, just github getting confused.

Fred-Barclay:
Thanks mate. :)

avoidr:
I wanted to say the same as KellerFuchs, but I didn't, because I was too busy wetting my pillow and did not prevent a catastrophe in the profile environment and now it's all dirty and there's smog all over the place, but now I need to go to the bakery.
Terrible.

Fred-Barclay:
@avoidr Ouch! Sorry, mate. :) I do wish you had mentioned it a bit earlier... :laughing: 

Let's see what @netblue30 says. I like the newer setup though I can see the logic behind your concerns. It just got confusing when reading profiles having to skip around looking for filters instead of them being queued. At any rate, some of it (esp. the duplicates and the missing entries) needed to be fixed.

netblue30:
I don't see the problem here, is a cleanup on 80 files. From time to time we need to do it, /etc/firejail seems to go up with 30 new files per release. Thanks for the patch.

PR #558, opened 2016-06-08
added profile for franz messenger

maces:
added profile for franz messenger app based n electron


netblue30:
Merged, thank you for the patch.

PR #564, opened 2016-06-11
Set $APPIMAGE and $APPDIR environment variables

### LINKED ISSUE #560: Missing environment variables in AppImage support
@netblue30 

Just discovered that you added AppImage support to firejail. Very useful, great work! 

The software inside AppImages might may the following [environment variables to be set:](https://github.com/probonopd/AppImageKit/blob/34defc0d9fc0b79735b3cdf84cd5c4d3234f9639/runtime.c#L360-L361)
* `$APPIMAGE` should contain the full path of the AppImage that has been mounted
* `$APPDIR` should contain the full path to the mountpoint where the AppImage that has been mounted

Also, running an AppImage [normally extracts its icon](https://github.com/probonopd/AppImageKit/blob/34defc0d9fc0b79735b3cdf84cd5c4d3234f9639/runtime.c#L286-L357) so that the thumbnailer in the system can use it; although I am not sure if this exceeds the scope of firejail.

Questions

* Is it required to write a separate profile for each AppImage in order for firejail to add security over normal AppImage execution without a sandbox?
* One of the main use cases would be for sandboxed AppImages to not inadvertently access/delete content in `$HOME` while still being able to use `$HOME` if explicitely allowed by the user. Do you have an idea/suggestion on how to achieve this? E.g., have a GUI dialog box to ask the user for permission?

probonopd:
Partly fixes #560

The software inside AppImages might may the following [environment variables to be set:](https://github.com/probonopd/AppImageKit/blob/34defc0d9fc0b79735b3cdf84cd5c4d3234f9639/runtime.c#L360-L361)
* `$APPIMAGE` should contain the full path of the AppImage that has been mounted
* `$APPDIR` should contain the full path to the mountpoint where the AppImage that has been mounted

Please [squash commits](https://help.github.com/articles/about-pull-request-merge-squashing/) when merging, thanks.

### DIFF
--- a/src/firejail/appimage.c
+++ b/src/firejail/appimage.c
@@ -88,7 +88,7 @@ void appimage_set(const char *appimage_path) {
 		printf("appimage mounted on %s\n", mntdir);
 	EUID_USER();
 
-	if (mntdir && setenv("APPIMAGE", appimage_path, 1) < 0)
+	if (appimage_path && setenv("APPIMAGE", appimage_path, 1) < 0)
 		errExit("setenv");
 	
 	if (mntdir && setenv("APPDIR", mntdir, 1) < 0)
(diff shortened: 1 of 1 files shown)


netblue30:
Merged, thanks!

PR #577, opened 2016-06-17
noqueue not exposed on older kernels; pfifo_fast is default qdisc

reinerh:


KellerFuchs:
It looks to me as if the `test/utils/seccomp.exp` is an unrelated fix.
Perhaps it should be in a separate commit?

reinerh:
I think it was trivial enough to keep it in one commit.
But have split it up now.

KellerFuchs:
@reinerh It's more than it is kinda confusing when using `git-blame`, since it would say “this line comes from this seemingly-unrelated commit”.
Thanks a lot.

netblue30:
Merged, thanks.

Issue #576, opened 2016-06-17
LibreOffice profile?


rekixex:
Firejail already includes many profiles. Why there is no LibreOffice profile? Could you add it?

Fred-Barclay:
That sounds like a good idea. I'm on it! :)

netblue30:
Thanks @Fred-Barclay, let me know when you are done and I'll merge you in.

Fred-Barclay:
Done!  #580

netblue30:
I added profiles for loffice, lowriter, lodraw, etc. Try it out!

PR #580, opened 2016-06-18
added libreoffice profile

Fred-Barclay:
Here's the libreoffice profile I promised in #576. :)

I thought about doing a more restrictive profile but there's no telling where a user will want to save files. What do you think?


netblue30:
Merged, thanks.

PR #582, opened 2016-06-19
.

requiredregistration:


netblue30:
OK, now I see your point. It is a little bit different. In order to enforce the configuration enforced by the administrator, firejail will fail to start if a global config file is not found. The configure flag allows the sandbox to start even if a global config is not present. In this case it will start with hardcoded defaults.

requiredregistration:
you 'saw my point' but you did not do shit about it to correct it.

the description is not accurate. i told you and you did not do anything to write an accurate one. i even suggested a few descriptions, and you just add the profiles for every shitty application they send your way.

do you fucking care about the documentation of your project or not?

netblue30:
You are wrong, that thing does not disable system wide configuration support. It only disables enforcing it.

requiredregistration:
`if the global config file firejail.cfg is not present, continue the program using defaults`

that is what i am talking about. you did not think carefully before implementing a solution. you could have just added a run-time option for this. is there a real need to have it as a compile-time option?

the goal is to keep it simple, not to increase complexity.

requiredregistration:
you are either not very smart and/or your english is not very good.

either way, you continue to do the same shit, caring more about profiles than developing firejail carefully.

that shitty description is proof that you do many things hastily. it seems that you care more about how popular firejail is and can become instead of how good you can make it be.

PR #585, opened 2016-06-20
Proposed

Fred-Barclay:
Profile for pix (one of Linux Mint's xapps). Pix was forked from gthumb so the profiles are practically identical and it already uses private-bin.

It worked fine on my end without `whitelist /tmp/.X11-unix` but I figured you had a good reason for including it on gthumb. Do you think it's necessary here?


netblue30:
Merged, thanks.

OK, this is the rule: if it is working on your setup, send a pull request with "whitelist /tmp/.X11-unix", we'll fix it later if necessary.

netblue30:
Also, keep an eye on Mint, we need to bring in all xapps!

Fred-Barclay:
Alright, thanks. 
LMDE Betsy is my main distro so I'll keep on top of the xapps. We've got all that have been released so far (xviewer, xreader, xed, and pix) and I don't know if any more are scheduled.

PR #591, opened 2016-06-23
Fixed conffiles warning for soffice

Fred-Barclay:
When building a .deb package from source, lintian warns that "file in etc not marked as conffile **etc/firejail/soffice.profile**". This should handle that.


netblue30:
Merged, thanks!

PR #596, opened 2016-06-25
add a firejail profile for strings

technoLord:
From: https://lcamtuf.blogspot.com/2014/10/psa-dont-run-strings-on-untrusted-files.html

Many shell users, and certainly most of the people working in computer forensics or other fields of information security, have a habit of running /usr/bin/strings on binary files originating from the Internet. Their understanding is that the tool simply scans the file for runs of printable characters and dumps them to stdout - something that is very unlikely to put you at any risk.

### DIFF
--- a/strings.profile
+++ b/strings.profile
@@ -1,0 +1,12 @@
+noblacklist ~/.config
+
+include /usr/local/etc/firejail/disable-common.inc
+include /usr/local/etc/firejail/disable-programs.inc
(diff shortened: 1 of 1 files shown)


avoidr:
What about `net none`, `shell none`, and `private-{bin,etc,dev}`?

netblue30:
Merged, thanks.

PR #597, opened 2016-06-25
xz decompressor

kouul:
Decompressing .xz file within a sandboxing environment.

### DIFF
--- a/etc/xzdec.profile
+++ b/etc/xzdec.profile
@@ -1,0 +1,13 @@
+# Firejail profile for XZ decompressor
+# xzdec.profile
+
+include /etc/firejail/disable-mgmt.inc
+include /etc/firejail/disable-secret.inc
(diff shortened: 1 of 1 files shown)


reinerh:
What about network access? Sounds like this can be dropped too.

Fred-Barclay:
Looks nice! :) A couple of quick suggestions:
Did you develop this on an older version of firejail? disable-mgr and disable-secret aren't part of firejail's code anymore (they were replaced, I don't remember with what) so the profile won't work on the current firejail code. 
And like reinerh, I think that dropping network access, or at least using `netfilter`, would help too.

Nice work!

avoidr:
Also `nonewprivs` and possibly `private-bin`, `private-etc`, or `private-dev`.

netblue30:
Merged, thanks.

PR #599, opened 2016-06-25
cpio sandbox profile for decompression

jmutkawoa:
For decompression purposes, this will allows CPIO to execute in a sandbox environment.
FYI, according to CVE 2016-2037, the cpio_safer_name_suffix function in util.c in cpio 2.11 allows remote attackers to cause a denial of service (out-of-bounds write) via a crafted cpio file.
Other vulnerabilities listed are:
https://www.cvedetails.com/vulnerability-list/vendor_id-72/product_id-5080/GNU-Cpio.html

### DIFF
--- a/etc/cpio.profile
+++ b/etc/cpio.profile
@@ -1,0 +1,7 @@
+include /usr/local/etc/firejail/server.profile
+include /usr/local/etc/firejail/disable-common.inc
+include /usr/local/etc/firejail/disable-programs.inc
+include /usr/local/etc/firejail/disable-passwdmgr.inc
+caps.drop all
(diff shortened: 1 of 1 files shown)


jmutkawoa:
Added net none option to disable network access as CPIO does not require network access

avoidr:
Why did you include server.profile? I would rather include disable-devel.inc. server.profile has private-tmp, which I personally would leave out as a default for programs that may be used by humans (I often use /tmp for testing shit).
Also what about `noroot`, `nonewprivs`, etc?

netblue30:
Merged, thanks.

PR #613, opened 2016-07-04
Fix improper quoting of arguments

manevich:
Using double quote instead of single quote for `cfg.command_line` allows  arbitrary command execution inside of jail if command args (eg. filenames) contain special sequences (backtick, $(), etc).

How to reproduce:
`touch '`uname -a`'`
`touch '$(cal)'`
in same directory:
`firejail cat *uname*`
command `uname -a` will be executed in jail
`firejail cat *cal*`
command `cal` will be executed in jail.

This hopefully fixes it.


netblue30:
This was a cool one! Double quotes allow expansion of arguments, so we need to use single quotes. Thanks!

PR #614, opened 2016-07-04
Telegram

Fred-Barclay:
I didn't realise at the time I redid the telegram profile that you had just written etc/Telegram.profile. :fearful: Anyways since it's identical to etc/telegram.profile I just set it to `include /etc/firejail/telegram.profile`.

I also tightened the qtox profile a bit. Hope this helps!
Cheers!


netblue30:
Merged, thanks.

PR #616, opened 2016-07-06
disable-common: Blacklist ~/.config/keybase

haasn:
This is used by keybase.io's client to store secrets.


netblue30:
Merged, thanks!

Issue #326, opened 2016-02-26
Vim has wrong CWD when run inside Firejail with the "tracelog" option


brunonova:
Strange issue. I'm running Ubuntu 14.04.

This runs perfectly:

    cd ~
    firejail --noprofile bash
    vim .gitconfig

But if I add the `--tracelog` option to firejail, vim doesn't find the file and creates a new one instead.

Running `:!pwd` inside vim in normal mode, which should print the working directory, outputs a different path than the expected one.
And it's not always the same path. The last time I executed it, the result was `/usr/share/vim/vim74/indent`.

I noticed this in Vim, but other programs may be affected as well.

chiraag-nataraj:
That's really odd...which version of Firejail are you running? I just tried what you suggested and it worked fine for me (`:!pwd` outputted `/home/chiraag/` as expected even with the `--tracelog` option).

brunonova:
I'm running the latest version from git `master`.
What distro are you using?

I forgot to say that this issue doesn't occur in `vim.tiny`, which is the default version of `vi` in Ubuntu before installing the `vim` package.

If I execute `vim -D` inside firejail with `--tracelog`:

    line 10: runtime! debian.vim
    >pwd
    /home/$USER
    >s
    line 3: set runtimepath=~/.vim,/var/lib/vim/addons,/usr/share/vim/vimfiles,/usr/
    share/vim/vim74,/usr/share/vim/vimfiles/after,/var/lib/vim/addons/after,~/.vim/a
    fter
    >pwd
    /usr/share/vim
    >cont

The working directory changed after "line 3 ...".
If I execute this without that option or outside Firejail, the working directory stays the same.

chiraag-nataraj:
I just tried this with vim (not vim.tiny) and I didn't get the issue with firejail 0.9.38 on Debian sid/experimental and Vim 7.4.

brunonova:
Hmm... I tested 0.9.38 (backported from Debian sid/Ubuntu Xenial) and I don't get the issue as well.

So this bug was introduced somewhere after 0.9.38 was released.
It also doesn't seem to me that the bug lies in the Debian packaging, since it doesn't apply any patches.

brunonova:
It seems the bug was introduced in commit d4e6b2e. I tested that and the previous commit.
The bug still exists in the latest commit.

netblue30:
I'll look into it, thanks!

manevich:
I'm somehow overlooked this bug before.
[...]

PR #617, opened 2016-07-06
Fix chdir bug in libtracelog

manevich:
Fix [bug](https://github.com/netblue30/firejail/issues/326) in libtracelog (introduced by me while fixing another).

* Add fchdir wrapper and convert file descriptor to pathname to keep as CWD.
* Instead of doing chdir (that was really bad decision) just add CWD to relative path.


netblue30:
Merged, thanks!

PR #621, opened 2016-07-08
Proposed

Fred-Barclay:
This is a profile for Atom and should also fix a warning when building a firejail.deb from source: "conffile '' is not a plain file".

Cheers!


netblue30:
All merged, thanks.

PR #622, opened 2016-07-08
correction no. 2

Fred-Barclay:
Apparently you can't comment conffiles without problems.  This should fix my last commit that swapped one warning for another. :stuck_out_tongue_winking_eye:

EDIT: converted qtox to private-bin and tightened the profile a bit more.

EDIT 2: okay, apparently I'm editing more than I planned! So here's what's in this PR:
1. A fix for the debian conffiles that fixes my original fix (fixed by a fixated fixer). 
2. Conversion to private-bin: atril, xreader, evince, and qtox.
3. Fixed permissions warning in 0ad that brought up a debug window when launching.
4. Removed network capabilities from atril, evince, gthumb, pix, and xreader.
5. Added some security filters where appropriate.

Cheers!

### DIFF
--- a/README
+++ b/README
@@ -55,6 +55,7 @@ Fred-Barclay (https://github.com/Fred-Barclay)
 	- added audacity profile
 	- fixed Telegram and qtox profiles
 	- added Atom Beta and Atom profiles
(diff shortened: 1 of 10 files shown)


netblue30:
Merged, thanks.

PR #626, opened 2016-07-10
restrict Dropbox to its own directories

rahiel:
This restricts Dropbox's file access to its own directories. I think it's worth doing this by default because Dropbox is a proprietary app. 

The whitelisted directories are the default locations used by Dropbox.


netblue30:
Merged, thanks.

PR #628, opened 2016-07-11
Jitsi

Fred-Barclay:
Profile for jitsi. I'll try and add private-bin later but it didn't want to work well at the moment. :)


netblue30:
Merged, thanks.

PR #629, opened 2016-07-12
Additional fixes of command line quoting

manevich:
[Previous fix](https://github.com/netblue30/firejail/pull/613) of command line quoting was only partial.
Code execution still was possible using constructions like this:
`firejail 'echo $(uname -a) && echo'`
`firejail -c 'echo $(uname -a)'`

Changes to `main.c`:
I see only 2 cases for quoting:
1. Arguments passed to shell should always be quoted by single quotes.
2. Arguments passed directly to program (--shell=none) should never be quoted.

Code removed from `run_symlink.c` not just redundant, but also caused obscure problems when symlinked program used with `shell none` in profile.
For example:
```
touch '1&2'
$SUCH_PROGRAM 1\&2
```
results in  "file not found" error, while
`firejail $SUCH_PROGRAM 1\&2`
and
`firejail $SUCH_PROGRAM '1&2'`
both work fine.

`--join` command's behaviour is still not fixed, will look at it later.

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -2022,11 +2022,7 @@ int main(int argc, char **argv) {
 		char *ptr1 = cfg.command_line;
 		char *ptr2 = cfg.window_title;
 		for (i = 0; i < argcnt; i++) {
-			// detect bash commands
-			if (strstr(argv[i + prog_index], "&&") || strstr(argv[i + prog_index], "||")) {
-				sprintf(ptr1, "%s ", argv[i + prog_index]);
-			}
-			else if (arg_command){
+			if (arg_shell_none){
 				sprintf(ptr1, "%s ", argv[i + prog_index]);
 			}
 			else {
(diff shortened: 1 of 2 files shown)


netblue30:
I'll set up some testing framework for all this argument processing. It will take me some time until I'm ready to merge it.

manevich:
OK, so I will put additional patches in this pull request along with explanation and test cases.
I already noted some additional problems.
Hope I can help you this way.

netblue30:
Sure, put them here, it will get me at least one week until I start. Thanks.

manevich:
New commit fixes quoting problem that can lead to command execution in jail:
`firejail echo "'\$(uname)'"`
by enclosing single quotes found in arg by double quotes like this:
`"'"'$(uname)'"'"`
There still room for improvement - runs of single quotes are encoded inefficiently.

manevich:
Just committed [better version](https://github.com/netblue30/firejail/pull/629/commits/1f4e6d3888130eff5485cbdd6ff4e84aa6e69f96) of [previous commit](https://github.com/netblue30/firejail/pull/629/commits/7669b9410df5639365967c57326a3fc3bb192810), that was severe overengineered.
Correctly handles of single quotes in args and is much more simple and readable.

netblue30:
All merged, thanks!

PR #634, opened 2016-07-15
Pidgin private-bin conversion

Fred-Barclay:
This converts pidgin to private-bin, and tightens the profile a bit.

I included `private-dev` which will break video chatting on pidgin, but searching online indicated that this is already broken in pidgin for the most part so `private-dev` will probably have no practical effect on users. What do you think?


netblue30:
Merged. private-dev should be fine in this case, thanks.

PR #643, opened 2016-07-24
Eom

Fred-Barclay:
Profile for the Eye of Mate picture viewer.
Either `protocol unix` or `net none` works fine with eom. I found that `net none` did not affect the eom plugins **if** they were marked for use beforehand, but it did not allow changing whether they should be activated or not (the user would have to do that outside of firejail) so I went with `protocol unix`... reluctantly. :frowning: 

EDIT: found some missing/mispelt programmes in src/firecfg/firecfg.config and fixed 'em.


netblue30:
Merged, thanks.

PR #646, opened 2016-07-24
Fix spelling errors found by lintian

reinerh:


reinerh:
Also noticed that test/sysutils was missing from the rc1 tarball, so I simplified the dist target a bit.

netblue30:
All merged, thanks!

Issue #632, opened 2016-07-13
the ‘nice’ option is ignored when joining


netblue30:
From wordpress

PR #652, opened 2016-07-27
Two fixes to --join behaviour

manevich:
* Honor `--nice` option when joining. Fixes https://github.com/netblue30/firejail/issues/632.
This is simplest implementation, and niceness is set independently on each join:
```firejail --name=test --nice=10```
and then
```firejail --join=test cmd1```
will start `cmd1` with default niceness (usually 0).
```firejail --nice=5 --join=test cmd2```
will start `cmd2` with niceness 5.

* Honor `--shell` option when joining, including `--shell=none`


netblue30:
Merged, thanks.

PR #653, opened 2016-07-27
disable-passwdmgr.inc: Don't leak keepassx config

thomasjfox:
It contains the path to the last used database.


netblue30:
Merged, thanks.

PR #654, opened 2016-07-28
Add profile for uudeview

thomasjfox:
uudeview might access unsafe email content,
therefore restrict it as much as possible.

In fact it's best to call firejail with a private home dir, too.


netblue30:
Merged, thanks.

PR #658, opened 2016-07-30
Allow BitlBee to write /var/lib/bitlbee

pigmonkey:
Bitlbee stores its configuration in /var/lib/bitlbee. It must be able to write to this directory in order to save config changes (adding accounts, writing OTR keys, etc).


netblue30:
Merged, thanks.

Issue #656, opened 2016-07-29
Viability of use with skype4linux alpha?


xee5ch:
I am a big fan of this package.  I am running a current installation of Debian 8.5.0 amd64 that was updated prior to testing.  I tested with firejail 9.4.0 using `jessie-backports` to stay current and Skype for Linux Alpha 1.3.0 downloaded from their new repo.

> $ aptitude show skypeforlinux 
> Package: skypeforlinux                   
> New: yes
> State: installed
> Automatically installed: no
> Version: 1.3.0.0
> Priority: extra
> Section: non-free/net
> Maintainer: Skype Technologies S.A. <info@skype.net>
> Architecture: amd64
> Uncompressed Size: 145 M
> Depends: gconf-service, libasound2 (>= 1.0.16), libatk1.0-0 (>= 1.12.4), libc6 (>= 2.12), libcairo2 (>= 1.6.0), libcups2 (>= 1.4.0),
>          libdbus-1-3 (>= 1.2.14), libexpat1 (>= 2.0.1), libfontconfig1 (>= 2.9.0), libfreetype6 (>= 2.4.2), libgcc1 (>= 1:4.1.1),
>          libgconf-2-4 (>= 2.31.1), libgdk-pixbuf2.0-0 (>= 2.22.0), libglib2.0-0 (>= 2.31.8), libgnome-keyring0 (>= 2.22.2), libgtk2.0-0
>          (>= 2.24.0), libnspr4 (>= 2:4.9-2~) | libnspr4-0d (>= 1.8.0.10), libnss3 (>= 2:3.13.4-2~) | libnss3-1d (>= 3.12.4), libpango1.0-0
>          (>= 1.14.0), libstdc++6 (>= 4.6), libx11-6 (>= 2:1.4.99.1), libxcomposite1 (>= 1:0.3-1), libxcursor1 (> 1.1.2), libxdamage1 (>=
>          1:1.1), libxext6, libxfixes3, libxi6 (>= 2:1.2.99.4), libxrandr2 (>= 2:1.2.99.2), libxrender1, libxss1, libxtst6, gnome-keyring,
>          apt-transport-https

I have tried the following executions.

```
$firejail --profile=/etc/firejail/skype.profile --private=firejail/skype skypeforlinux
$firejail --profile=/etc/firejail/default.profile --private=firejail/skype skypeforlinux
$firejail --profile=/etc/firejail/skype.profile --private=firejail/skype skypeforlinux --debug
[...]

PR #660, opened 2016-07-30
Add new skypeforlinux profile.

xee5ch:
Per recommendation of @netblue30, allow use of the netlink protocol in order for skypeforlinux to properly function in a firejail environment, per discussion in Github issue #656.


netblue30:
merged, thanks!

PR #661, opened 2016-07-30
Fix command line quoting on joining, move quoting code to functions

manevich:
Fix command line quoting on joining.
Quoting code is moved to functions in separate file to avoid code duplication.
Exit with error if length of quoted command line is longer than `ARG_MAX`.


netblue30:
Merged, thanks.

PR #663, opened 2016-07-30
Include mkuid.sh in "make dist"

thomasjfox:
Fixes the rpm build from the tarball.

I'm fine if you cherry-pick this commit instead of a merge commit :)


netblue30:
merged, thanks.

PR #664, opened 2016-07-30
Allow recursive mkdir (Closes #305)

### LINKED ISSUE #305: make mkdir  recursive
This is a question-like issue currently, but I think a change is needed to avoid confusion:

Why isn't `mkdir` invoked by default when doing `whitelist`?
The possible pitfall I see with the current approach is that people might get confused:

* somebody takes his system, tests his new-created profile. Since directories exist on his computer, everything goes fine
* he submits the profile so it would be available for everyone
* people start to complain because this will not work on their fresh computers.

Thoughts?

reinerh:


netblue30:
Merged, thanks.

PR #665, opened 2016-07-30
Add profiles for tar (gtar), unzip and unrar

thomasjfox:
I've tested compression and uncompression of
various tar formats and also straced unzip/unrar
regarding their file access in /etc.

-> should be fine.

If you want to unpack files in /usr/bin or so,
then use the --ignore=private-bin switch.


thomasjfox:
I've squeezed in a profile for file(1), too.

netblue30:
merged, thanks.

PR #668, opened 2016-07-30
Improve libtrace / libtracelog

thomasjfox:
Small improvements for libtrace / libtracelog.

If a program would have called stat() and then stat64(),
the traced program probably would crash with a NULL pointer access on orig_stat64.


netblue30:
merged, thanks.

Issue #676, opened 2016-08-01
Closing Xephyr window when exiting?


vn971:
Should firejail close its Xephyr window when exiting? The current version of firejail does NOT close the window (for me).

Behavior in question:
* start firejail with `firejail --noprofile --x11=xephyr xterm`
* close xterm by typing Ctrl+D  (`exit`)
* check out firejail processes. There should be none corresponding to the test
* observe the Xephyr window still active on the desktop, though it does not allocate anything in `/tmp/.X11-unix` anymore.

manevich:
`-terminate` option of `Xephyr` solves this problem.
But, for some reason `-noreset` (opposite of it) was added by default.
[Here](https://github.com/manevich/firejail/commit/5ef59391e909ca4bdc1794b97bb9c829e30276e8) is quick fix.
I don't see negative consequences of this patch.
Please check it, and I will send pull request to upstream.

vn971:
BTW, since we already raised a question on Xephyr parameters.. May it make sense to add a possibility to override firejail Xephyr parameters? Like `xephyr-params-override -fullscreen -terminate` in `firejail.config`.

manevich:
We can just change order of arguments to make `xephyr-extra-params` be last arg before screen number, and they will override hardcoded. See https://github.com/manevich/firejail/commit/531c3401e4ead91118e33610c97128efdc786a53.

netblue30:
all merged in, thanks.

PR #679, opened 2016-08-02
Change hardcoded Xephyr options to close window at exit of last client.

manevich:
See https://github.com/netblue30/firejail/issues/676.
* Change  default hard-coded option from `-noreset` to `-terminate`.
* Allow override of hard-coded options.

Tested it a bit with both single apps and openbox WM, everything works fine for me.
Old behavior can be restored by adding `xephyr-extra-params -noreset` to `/etc/firejail/firejail.config`.


netblue30:
Merged, thanks.

PR #681, opened 2016-08-04
add --private-template=directory option

nextime:
--private mount a tmpfs over the user home, so, things like browser extensions and configurations are back to the defaults and you cannot pre-configure it easily.

--private=directory, on the other side, make the user home persistent.

--private-template=directory aim to /opt/firejail/templates/usernamebe an alternative, it act like --private mounting a tmpfs over the user home, and then it copy recursively all the files and dirs from a template directory to the created empty user home.

This way you can put, for example, a preconfigured .config/cromium directory let's says in /opt/firejail/templates/username, and then use 
`firejail --private-template=/opt/firejail/templates/username cromium` 

NOTE: as the copy from template is done after mounting the tmpfs to mask the user home, the template directory cannot be copied from inside the original user home.

### DIFF
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -136,6 +136,7 @@ typedef struct config_t {
 	char *bin_private_keep;	// keep list for private bin directory
 	char *cwd;		// current working directory
 	char *overlay_dir;
+   char *private_template; // template dir for tmpfs home
 
 	// networking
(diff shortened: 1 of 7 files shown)


netblue30:
Merged, thanks for the patch. I used to have this feature a long time ago and scrapped it. I will rename it private-home.

PR #684, opened 2016-08-05
Checkmate

Fred-Barclay:
This adds support for gnome-chess, and the VLC profile has been tightened a bit too.


netblue30:
Merged, thanks.

PR #695, opened 2016-08-09
Busybox workaround + expand ${PATH} macro in noblacklist entries

manevich:
* Workaround for systems where common UNIX utilities are symlinks to busybox.
Aside of embedded, this includes Alpine Linux, for example.

* Expand `${PATH}` macro in `noblacklist` entries.
My busybox workaround depends on this.


netblue30:
Merged, thanks.

PR #696, opened 2016-08-10
Fixed & tightened gnome-chess

Fred-Barclay:
This should fix some issues I introduced in the initial commit (oops!) plus add `private-etc` and `private-tmp` :smile:


netblue30:
Merged, thanks

Issue #671, opened 2016-07-31
No need to open xephyr/xpra window when under `--audit`


vn971:
A nested X window (xephyr/xpra) is opened when starting firejail with `--audit`. Maybe avoid creating the window?

netblue30:
I'll have to document it and prevent --audit when --x11 is requested.

netblue30:
@manevich added a fix, you should be able to use --audit for x11 sandboxes.

Issue #693, opened 2016-08-08
--x11=xpra prevents the jailed program to output to stdout/stderr


l29ah:
I get xpra's screams instead.

netblue30:
All fixed!

Issue #648, opened 2016-07-26
xpra persists after firejailed app is quit


xahare:
firejail version 0.9.42~rc1 ubuntu 16.04.1 amd64, same behavior in 0.9.40.1
running firejail like this,

    firejail --name=workbrowser --x11 --net=fj0 --dns=8.8.8.8,8.8.4.4 --private=~/firejail/workbrowser google-chrome

the xpra session persists after the jailed app is quit.

netblue30:
moved to: https://github.com/netblue30/firejail/issues/648

xahare:
Ubuntu 16.04.1 LTS
xpra v0.17.4

netblue30:
It is possible it was fixed by this commit: https://github.com/netblue30/firejail/pull/679

There is more testing coming.

xahare:
those changes only apply to xephyr.

netblue30:
You are right, sorry.

netblue30:
All fixed.

PR #701, opened 2016-08-12
rewrite of X11 support

manevich:
Sorry for for pull request that can't be automatically merged.
I pulled your latest commits too late. But my rewrite, as seems, also fixes https://github.com/netblue30/firejail/issues/671.

Main points:
* no shell used to start commands
* don't end starting process. Old behavior can be mimicked by `firejail --x11 ... &`
* wait for jail process to terminate, and then stop X11 processes, this seems to work way better than relying on X11 servers for this.
* no problems with stdout/stderr output (except of `xpra` being noisy)
* `firejail --x11 --audit` works, leaving no dormant X11 servers behind
* should fix https://github.com/netblue30/firejail/issues/671, https://github.com/netblue30/firejail/issues/693, https://github.com/netblue30/firejail/issues/648

P.S: I used `firejail --x11 echo` and `firejail --x11=xephyr echo` to test edge case "GUI program that failed before opening any X11 connection".

### DIFF
--- a/src/firejail/x11.c
+++ b/src/firejail/x11.c
@@ -352,7 +352,7 @@ void x11_start_xephyr(int argc, char **argv) {
 
 void x11_start_xpra(int argc, char **argv) {
 	EUID_ASSERT();
-	int i;
+	size_t i;
 	struct stat s;
 	pid_t client = 0;
 	pid_t server = 0;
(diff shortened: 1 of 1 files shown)


netblue30:
All merged, thanks.

Issue #239, opened 2016-01-20
Re-attach to an overlay after the sandbox is closed


netblue30:
From wordpress:

Hi,

Is it possible to re-attach to an overlay after the sandbox is closed? Now, a new overlay is always created, but I would like to continue where I left of.

Thanks,
eli

dshmgh:
I have experimented with the overlayfs mount and if I manually remount over a previous sandbox, the merge appears to work properly.  If an option is added --reoverlay and then that creates the /home/$USER/.firejail/persist dir and that dir is used instead of /home/$USER/.firejail/$PID for the overlay then I think this enhancement would work.  But does this reuse of a previous sandbox somewhat defeat the purpose of using firejail?

I'll get some code for this change to you at some point.

netblue30:
I'm really interested in this feature, so if you send me some code I'll merge it in. The reason I've been sleeping on it is remounting the overlay is a huge security problem. For example, in the overlay you can modify /etc/shadow, then start firejail and become root. I'll have to restrict --reoverlay to root user.

Sidnioulz:
@netblue30  would the issue exist if user namespaces were in place?

netblue30:
Overlayfs and user namespace should be two different subsystems.

netblue30:
Implemented in 0.9.42~rc2

PR #702, opened 2016-08-12
Added more overlay options

hamzadis:
This pull request adds two new overlay options:
-  `--overlay-named` - creates and uses a named overlay in the ~/.firejail directory
- `--overlay-path` - creates and uses an overlay in an arbitrary location

These changes fulfil the enhancement requests outlined in #239. 
Now, the comments in that issue did mention a security problem with this feature, so I am not sure if my changes qualify as safe. Would love to see some comments about that.

### DIFF
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -221,7 +221,8 @@ extern int arg_debug_whitelists;	// print debug messages for whitelists
 extern int arg_nonetwork;	// --net=none
 extern int arg_command;	// -c
 extern int arg_overlay;		// overlay option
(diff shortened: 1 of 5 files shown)


netblue30:
Thanks for the patch, I'll merge it in the next few days.

Don't worry about the security implications, they are easy to solve. Basically, for both options we will enable seccomp by default. This will prevent the user to become root if he has modified /etc/shadow in the overlay stored.

netblue30:
Merged, thanks!

hamzadis:
Awesome, thank you.

netblue30:
I'll make it run-time and compile-time configurable, probably some minor fixes, but it looks good. Thanks.

Issue #703, opened 2016-08-12
When using non-default shell, FireJail shouldn't throw an error if Bash is missing inside the chroot


nuxwin:
We use FireJail on top of our jail builder (MakeJail) to isolate our restricted shells. We use makejail to create the chroot and then, we set the login shell to `/usr/local/bin/firejail`. Depending on the chroot configuration, we can have `busybox` inside our chroot or bash. We tells firejail to use the correct shell using the `--shell` option in the login.users file. For instance:

```
nuxwin:--chroot=/var/chroot/InstantSSH/shared_jail --quiet --dns=8.8.8.8 --dns=8.8.4.4 --name=nuxwin --noprofile --nonewprivs --ipc-namespace --rmenv=container --seccomp --shell=/bin/ash
```

```
root@srv01:/usr/local/etc/firejail# ls -la /bin/ash 
lrwxrwxrwx 1 root root 12 août  12 22:48 /bin/ash -> /bin/busybox
```

The problem is that if we use busybox (ASH) without copying bash inside the chroot, FireJail throws an error. This shouldn't occurs.

Thank you.

manevich:
Can you provide more detail about error? Which version of firejail you are using?
`firejail --shell=/bin/sh` works for me on busybox system without `/bin/bash`.

nuxwin:
@manevich 

I used latest release:

```
root@srv01:/var/chroot/InstantSSH/shared_jail/etc# firejail --version
firejail version 0.9.42~rc1
X11 support is disabled
```
I created a chroot with makejail script and I've done some bind mounts as usually. I copied the busybox binary inside the chroot and created a symlink on it (/bin/ash to /bin/busybox ) inside the chroot. Then, I've set the `--shell` option to `/bin/ash` (not /bin/sh) but this should not change anything. In the chroot, I've not copied `/bin/bash` (expected since I want only ASH).

The problem is that when login through SSH, firejail throw and error because `/bin/bash` is missing inside the chroot. If I copy it, that works as expected and login shell is ASH as expected...

I'll reproduce the problem and show you the exact output in few minutes.

nuxwin:
@manevich 

Exact output:

```
nuxwin@dev:~$ ssh nuxwin@192.168.1.133
nuxwin@192.168.1.133's password: 

  ___           _              _     ____ ____  _   _
 |_ _|_ __  ___| |_ __ _ _ __ | |_  / ___/ ___|| | | |
[...]

PR #709, opened 2016-08-13
Gather shell selection code in one place

manevich:
Shell selection code scattered everywere in code and causes bugs like this https://github.com/netblue30/firejail/issues/703.
So:
* new `guess_shell` function
* shell selection done once (well, except join), kept in `cfg.shell` variable.
* check for `/bin/bash` in chroot re-enabled as check for `cfg.shell`

For now calls `guess_shell` independently; also I noted that join never honored `--shell` and `--shell=none`. Will try to fix this latter.

### DIFF
--- a/src/firejail/bandwidth.c
+++ b/src/firejail/bandwidth.c
@@ -459,13 +459,15 @@ void bandwidth_pid(pid_t pid, const char *command, const char *dev, int down, in
 	if (setregid(0, 0))
 		errExit("setregid");
 
+	assert(cfg.shell);
+
 	char *arg[4];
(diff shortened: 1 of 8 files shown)


manevich:
Sorry, just found bug in my code, `firejail --private-bin=something cmd` fails.
Please, refrain of merging this before I fix it.
Thanks.

manevich:
Fixed, I just uncommented wrong line...

manevich:
Now `--join` honors `--shell` option including `--shell=none`.
Everything works fine for me, no regressions that I aware of.

netblue30:
Merged, thanks.

PR #713, opened 2016-08-14
Fixed #712

### LINKED ISSUE #712: Test for shell interpreter inside chroot is wrong
The current test for the shell availability inside the chroot (fs.c) is wrong, leading to the following error:

```
Error: cannot find (null) in chroot directory
```

if the interpreter is not also present in the system fs.

Indeed, the test is made on the system fs (chroot() system call is done later on) while it should be done inside the chroot. A shell available inside the chroot can be missing in system fs (busybox ash case). There is also a pointer problem in the code (on error, (null) is showed instead of the cfg.shell value).

I'll do a PR soon.

nuxwin:
This PR fix the #712 issue.

Thank you for your interest.

### DIFF
--- a/src/firejail/fs.c
+++ b/src/firejail/fs.c
@@ -1110,13 +1110,16 @@ int fs_check_chroot_dir(const char *rootdir) {
 	}
 	free(name);
 
-	// check shell
-	if (!arg_shell_none) {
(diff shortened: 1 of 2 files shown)


manevich:
@nuxwin Thank you for reporting and fixing.

netblue30:
Merged, thanks.

PR #716, opened 2016-08-15
Create inox.profile

stelariusinfinitek:
Inox is a privacy-focused browser based on Chromium. This profile will work with Inox, the Chromium profile is not compatible.


netblue30:
Merged, thanks.

PR #717, opened 2016-08-15
x11 fixes

manevich:
* don't mess terminal when Xephyr window closed.
* quiet `xpra` when `--quiet` option used
* xpra stopped correctly,no more runaway processes, even under high load. I tested it this way: applied load on system, then
`for i n $(seq 1 20); do firejail --quiet --x11 false; done` and checked for runaway processes with `xpra list` and `ps`.


netblue30:
Merged, thanks.

PR #722, opened 2016-08-16
tightened profiles

Fred-Barclay:
G'day netblue30! This is the first of (I hope) several commits that should tighten profiles when appropriate. In particular, I'm targeting `nosound`, `nogroups`, and `private-tmp` being added.


netblue30:
Merged, thanks.

PR #723, opened 2016-08-17
Add profile support for Slack

greigdp:
Add a profile for Slack, and include Slack's config directory within the global blacklist


netblue30:
Merged, thanks.

PR #724, opened 2016-08-17
Fix Spotify - "private-bin spotify" prevents Spotify loading

greigdp:
The rule `private-bin spotify` appears to prevent Spotify from loading. This seems to be a regression on the git version, compared to the release, and removing this rule will let Spotify run again.


netblue30:
Merged, thanks.

PR #727, opened 2016-08-18
Small fixes (icecat.profile, disable-common.inc and whitelist-common.inc)

icaroperseo:
Hi!

Three pretty small fixes:

1. The `icecat.profile` was updated (some dir paths mismatch with the generic `firefox.profile`).
2. New KDE terminal emulators added (_Konsole_ and _Yakuake_) besides adding a new blacklist rule.
3. GTK rendering issues fixed (related to _oxygen-gtk_) in some distributions still use KDE4, Slackware for example.

I appreciate your effort. Have a nice day!

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -14,6 +14,7 @@ blacklist /etc/xdg/autostart
 blacklist ${HOME}/.kde4/Autostart
 blacklist ${HOME}/.kde4/share/autostart
 blacklist ${HOME}/.kde/Autostart
+blacklist ${HOME}/.kde/share/autostart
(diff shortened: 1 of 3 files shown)


netblue30:
Thanks for the patch. Why do you disable konsole and yakuake?

icaroperseo:
@netblue30 _konsole_ is the default terminal emulator of KDE desktop envirointment, like _gnome-terminal_ is for Gnome (desktop envirointment). _yakuake_ is a drop-down terminal emulator of KDE (_yakuake_ use to _konsole_ as backend).

PR #728, opened 2016-08-18
Add profile for Gajim IM client

greigdp:
Add profile for Gajim (Python GTK-based IM client), as well as adding Gajim's files to the disable-programs common list


netblue30:
Merged, thanks.

Issue #718, opened 2016-08-15
thunderbird use "wrong" pdf viewer


frank-github:
after running thunderbird started by firejail version 0.9.40

> Redirecting symlink to /usr/bin/thunderbird
> Reading profile /etc/firejail/thunderbird.profile
> Reading profile /etc/firejail/firefox.profile
> Reading profile /etc/firejail/disable-common.inc
> Reading profile /etc/firejail/disable-programs.inc
> Reading profile /etc/firejail/disable-devel.inc
> Reading profile /etc/firejail/whitelist-common.inc
> Parent pid 1900, child pid 1901
> Blacklist violations are logged to syslog

I can't view attached pdf files by double click or "open", because thunderbird is not using my preferred pdf viewer anymore. It's starting the ubuntu default pdf viewer evince with the first part of pdf file name (white space in original name like "my test.pdf"

Found in syslog
Aug 16 01:16:24 ubuntu kernel: [228493.464107] audit: type=1400 audit(1471302984.440:780): apparmor="DENIED" operation="open" profile="/usr/bin/evince" name="/etc/ld.so.preload" pid=1512 comm="evince" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
Aug 16 01:16:24 ubuntu kernel: [228493.638309] audit: type=1400 audit(1471302984.616:781): apparmor="DENIED" operation="open" profile="/usr/bin/evince" name="/etc/xfce4/defaults.list" pid=1512 comm="evince" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0

anything to do with my problem? not sure, /etc/xfce4/default.list is empty 

If I start thunderbird without firejail everything works fine.

any hints and tips are welcome how to config firejail, that thunderbird is using my preferred applications for attached files.

frank-github:
tried it with private-Option:     private /home/myuser

now my favourite pdf viewer is started, but with wrong file-name, splitted at white space

netblue30:
The two audit messages you printed above are from AppArmor, and I would say the messages are normal. I assume you are on a Ubuntu platform - they have all kind of bugs in AppArmor, but in the end, evince manages to start.

> now my favourite pdf viewer is started, but with wrong file-name, splitted at white space

I think this is already fixed in the version here on github.

Thunderbird profile is a whitelisted profile,  and allows access only to firefox and thunderbird configuration files. Evince will not find its own config files, and many things will go wrong. Do the following test - in a terminal start thunderbird as follows:
`````
[...]

PR #729, opened 2016-08-18
run_no_sandbox fix

manevich:
Fixed problem with cmdline quoting, (see https://github.com/netblue30/firejail/issues/718#issuecomment-240412040)
To avoid code duplication `start_application` function from `sandbox.c` used.
Shell choice taken from `SHELL` env variable or guessed by `guess_shell` function.
Parses and honors `--shell`, `--zsh`, and `--csh` functions.

Should fix https://github.com/netblue30/firejail/issues/718


netblue30:
Merged, thanks.

PR #734, opened 2016-08-20
join fixes

manevich:
* simplify `join` function by reusing `start_application`. Honor `--audit` option on joining as result.
* fix caps filter setting being affected by setting user namespace, same way as in https://github.com/netblue30/firejail/blob/master/src/firejail/sandbox.c#L789. Don't know if setting caps after setting user namespace is enough, or one on line 286 is required too. I left if as is currently.


netblue30:
Merged, thanks.

PR #735, opened 2016-08-20
Fix error in Slack profile

greigdp:
noblacklist must appear before the includes, to avoid the default
profiles from invoking blacklists. This commit resolves this.


netblue30:
Merged, thanks.

PR #742, opened 2016-08-24
Tighten security

manevich:
**appimage.c**:
* remove redundant `mkdir`, `chown`, `chmod` calls, last two also pose race condition risk.
* open appimage file as user, to avoid race condition, that may alllow attacker mount appimage/iso file that unreadable to user.

**firejail.h**:
* add `ASSERT_PERMS` macros

**util.c**:
* change `copy_file` function to take `uid`, `gid`, and `mode` for new file, use `fchown` and `fchmod` to prevent race condition when setting owner and permissions.

**euid_common.h**:
* set `egid` when switching `euid`, so files created as root won't need additional `chown` to change group.
* exit with error if unable to switch `euid`/`egid` to user - if we continue to run code than intended to run as user with root privileges - this is disaster.

**fs.c** and other:
* Avoid `chown` and `chroot` where possible, using functions and macros mentioned above.

### DIFF
--- a/src/firejail/appimage.c
+++ b/src/firejail/appimage.c
@@ -81,6 +81,7 @@ void appimage_set(const char *appimage_path) {
 		fprintf(stderr, "Error: cannot create temporary directory\n");
 		exit(1);
 	}
+	ASSERT_PERMS(mntdir, getuid(), getgid(), 0700);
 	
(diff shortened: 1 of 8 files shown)


netblue30:
Merged, thanks.

Issue #749, opened 2016-08-29
0.9.42~rc2: Runtime Errors


githububub:
```
firejail: fs_dev.c:78: fs_private_dev: Assertion (s.st_mode & 07777) == (0755) failed.
Error: cannot establish communication with the parent, exiting...
```

Arch. Stock kernel. Stock grsec kernel. Custom grsec kernel. Same results. Comment out private-dev in the profile and standard output kicks back a "fs_private_etc" error. Reinstalled 0.9.42~rc1. Everything works as expected. 

Build log and a few test commands: http://pastebin.com/c7YBfchM

UPDATE: Rebuilt without -fPIE as I noted -fPIC and -pie were altready being passed. Same errors.

manevich:
Thanks for reporting.
My commits, especially https://github.com/netblue30/firejail/commit/0ee599684c4ca622ca22d09eba565eb07c1a2b12 are source of this problem.
I am downloading Arch now to check it further.
I did tested it on Debian Linux with stock kernel and on Alpine Linux with grsec kernel before commiting.

manevich:
Unable to reproduce on fresh ArchLinux installation in VirtualBox, using [latest git](https://github.com/netblue30/firejail/commit/80c8015d7d3158dcd44b48f83658f2c61e14eddf) compiled/installed manually.
```
$uname -r
4.7.2-1-ARCH
```
I an not familiar with Arch, so can you provide further details/suggestion?

netblue30:
It is affected by umask. This is a small test program:
`````
#include <stdio.h>
#include <sys/stat.h>
#include <sys/types.h>

int main(void) {
	mkdir("blablabla", 0777);
	return 0;
}
`````
Compile and run I get:
`````
$ umask
0022
$ ls -l
drwxr-xr-x  2 netblue netblue   4096 Aug 29 13:18 blablabla
`````
umask setting is extracted from 0777 we intended for our directory. In "man 2 umask" they say:
`````
       The  umask  is  used  by open(2), mkdir(2), and other system calls that
       create files to modify the permissions placed on newly created files or
       directories.   Specifically,  permissions  in  the umask are turned off
       from the mode argument to open(2) and mkdir(2).
`````
So, we need to make a chmod after mkdir/open calls, something like this:
#include <stdio.h>
#include <sys/stat.h>
#include <sys/types.h>

int main(void) {
	mkdir("blablabla", 0777);
        chmod("blablabla", 0777);
	return 0;
}
`````

[...]

PR #753, opened 2016-08-29
Fix chmod/umask problem

manevich:
Put back previously removed chmod calls.
Fixes https://github.com/netblue30/firejail/issues/749


netblue30:
Merged, thanks.

PR #755, opened 2016-08-31
Profile tightening

Fred-Barclay:
These tighten many profiles a bit--mostly via `private-tmp`. I've also added a profile for dosbox, and removed private-tmp from the gthumb profile since it already had a /tmp whitelist. 
The profiles in the "tested and stable" commit have all been tested and I don't expect them to break anything.
I've also committed a revised hexchat profile in "may break on some systems" and added a private-bin filter. It works perfectly on my machine, but I suspect it may break on other distros that require python or perl for hexchat since they are not included in the private-bin filter (I don't know which distros these would be, but I remember something about this from an old bug report). If you do commit this, we'll probably need to revise the profile a couple of times based on the bug reports that might trickle in. 
Then again, it might work perfectly as-is.

Cheers!
Fred

### DIFF
--- a/etc/0ad.profile
+++ b/etc/0ad.profile
@@ -19,8 +19,8 @@ whitelist ~/.local/share/0ad
 
 caps.drop all
 netfilter
-nonewprivs
 nogroups
+nonewprivs
 noroot
 protocol unix,inet,inet6
 seccomp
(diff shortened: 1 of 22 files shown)


netblue30:
This is cool, thanks!

PR #757, opened 2016-09-02
extra dosbox files

Fred-Barclay:
I forgot the extra files for the new dosbox profile in my last PR.


netblue30:
Merged, thanks.

PR #758, opened 2016-09-02
Option to fix .desktop files for firecfg

manevich:
Some `.desktop` files use full path to executable. As result, just running `firecfg` is not enough to run those apps in firejail using menu or file manager.
With `--fix` option firecfg will check `.desktop` files in `/usr/share/applications/`, replace full path by name if it is in PATH, and write result to `$HOME/.local/share/applications/`.


netblue30:
Merged, thanks.

PR #763, opened 2016-09-02
small fixes

manevich:
* Add shell selection code to `--join-network` and `--join-filesystem`
* Update examples in firejail man page broken by changes in arg quoting


netblue30:
Merged, thanks.

Issue #768, opened 2016-09-06
xpra initialization error


chiraag-nataraj:
With xpra version 0.17.5 and firejail version 0.9.42-rc2, firejail fails to create an xpra instance. The exact output is attached [here](https://github.com/netblue30/firejail/files/457117/xpra.txt).

manevich:
In [this pull request](https://github.com/netblue30/firejail/pull/769) I fixed error that, most likely, caused this bug.
Would like to hear if this fixes it for you.

chiraag-nataraj:
Yup, that works!

PR #769, opened 2016-09-06
xpra fix

manevich:
Add missed `NULL` to `argv`.
May fix https://github.com/netblue30/firejail/issues/768


netblue30:
Merged, thanks.

PR #773, opened 2016-09-10
Add option to block X11

manevich:
Add `--x11=block` command line option, `x11 block`, `x11 xephyr`, `x11 xpra` profile options.

X11 blacklisting implemented simplest way:
* Blacklist `/tmp/.X11-unix` and `.Xauthority`
* It's up to user to decide how to deal with abstract sockets, just stop with explanatory error message if abstract socket will be accessible in jail.

### DIFF
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -315,6 +315,7 @@ extern int arg_audit;		// audit
 extern char *arg_audit_prog;	// audit
 extern int arg_apparmor;	// apparmor
 extern int arg_allow_debuggers;	// allow debuggers
+extern int arg_x11_block;	// block X11
 
 extern int login_shell;
 extern int parent_to_child_fds[2];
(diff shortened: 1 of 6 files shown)


xahare:
some distros also use abstract sockets for x11. 

firejail --blacklist=/tmp/.X11-unix --blacklist=${HOME}/.Xauthority xlogo
(xlogo displays on screen)

manevich:
@xahare 
> some distros also use abstract sockets for x11. tried this on ubuntu 16.04 running unity

AFAIK there no way firejail can block abstract sockets without breaking other things.
So it's up to user to decide how to do this.
My implementation **stops with error message** if it detects that X11 abstract sockets will be accessible in jail.
```
$ firejail --x11=block

ERROR: --x11=block specified, but abstract X11 socket still accessible.
Additional setup required. To block abstract X11 socket you need either:
 * use network namespace (--net=none, --net=...)
 * add "-nolisten local" to xserver options (eg. /etc/X11/xinit/xserverrc)
```
Suggestions appreciated.

netblue30:
All merged, thanks.

xahare:
for most linux desktops this is handled by the display manager. for example in lightdm on ubuntu its /usr/share/lightdm/lightdm.conf.d/50-xserver-command.conf

manevich:
@xahare Thanks, will update the message to mention display managers.

PR #790, opened 2016-09-17
Update okular.profile

pwnage-pineapple:
Gives Okular read-only access to kdeglobals file to fix theme issue


netblue30:
Merged, thanks.

PR #798, opened 2016-09-20
Quiet SSH config

daradib:
firejail output in ssh client breaks git+ssh for me, e.g.,

    $ git clone git@github.com:netblue30/firejail.git
    Cloning into 'firejail'...
    Reading profile /etc/firejail/ssh.profile
    fatal: protocol error: bad line length character: Pare

The "Pare" comes from "Parent pid x, child pid y".


netblue30:
Merged, thanks.

PR #805, opened 2016-09-23
join-or-start option

manevich:
Command line and profile option to run apps in one sandbox without need to handle starting and joining separately.

### DIFF
--- a/src/man/firejail-profile.txt
+++ b/src/man/firejail-profile.txt
@@ -467,6 +467,12 @@ available in the new namespace is a new loopback interface (lo).
 Use this option to deny network access to programs that don't
 really need network access.
 
+.SH Other
+.TP
+\fBjoin-or-start sandboxname
+Join the sandbox identified by name or start a new one.
+Same as "firejail --join=sandboxname" command if sandbox with specified name exists, otherwise same as "name sandboxname".
+
 .SH RELOCATING PROFILES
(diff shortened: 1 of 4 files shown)


src/man/firejail-profile.txt
```diff
@@ -467,6 +467,12 @@ available in the new namespace is a new loopback interface (lo).
 Use this option to deny network access to programs that don't
 really need network access.
 
+.SH Other
+.TP
+\fBjoin-or-start sandboxname
+Join the sandbox identified by name or start a new one.
+Same as "firejail --join=sandboxname" command if sandbox with specified name exists, otherwise same as "name sandboxname".
```
Fred-Barclay: Would `otherwise same as "name sandboxname"` be better as`otherwise same as "firejail --name=sandboxname"`?

manevich: Well, this is about profiles, so I wrote `name sandboxname` as in profile.
My English is pretty bad, so I gladly accept any suggestion about wording. If you still think it's misleading I will change this.

Fred-Barclay: Oh, okay. I wasn't thinking of that.
It is perhaps a bit confusing (since there is no `join` option for the profiles) but I would leave it like you have it.

Oops!

netblue30:
Merged, thanks.

PR #813, opened 2016-09-25
Quickfix for problem found in SELinux that affects Firejail too.

manevich:
Quickfix for problem found in SELinux (CVE-2016-7545) that affects Firejail too.
See https://security-tracker.debian.org/tracker/CVE-2016-7545 and http://seclists.org/oss-sec/2016/q3/606.
Fixed by calling `setsid()` as done in SELinux. Side effect of this is broken job control in shell:
```
$ firejail --quiet
bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
$ fg
bash: fg: no job control
$ bg
bash: bg: no job control
```
But this should be fine until somebody comes with better solution.

EDIT: grsecurity prevents exploitation of this problem.

### DIFF
--- a/src/firejail/sandbox.c
+++ b/src/firejail/sandbox.c
@@ -290,6 +290,10 @@ void start_audit(void) {
 }
 
 void start_application(void) {
+	// fix problem found in SELinux (CVE-2016-7545) that affects Firejail too
+	if (setsid() == -1)
+		errExit("setsid");
+
 	//****************************************
 	// audit
 	//****************************************


netblue30:
Thanks for the bug! I put another fix in here: https://github.com/netblue30/firejail/commit/cf1e38c210b12a504bebf4b63b2a0abfd7d023e0

Test program:
`````
#include <unistd.h>
#include <sys/ioctl.h>
#include <termios.h>

int main()
{
    char *cmd = "ls /proc\n";
    while(*cmd)
     ioctl(0, TIOCSTI, cmd++);
     
//     ioctl(0, TCFLSH, TCIFLUSH);
     return 0;
}
`````
The fix is to flush the input queue as we exit the sandbox. If it doesn't hold, we will have to go with setsid you suggested.

manevich:
Thanks, your fix is much better then mine! I have lot to learn...

netblue30:
Same here.

PR #815, opened 2016-09-26
CVE-2016-7545 fix improvement

manevich:
Improved your fix a bit: it's better to flush `/dev/tty` than `stdin`, as `stdin` descriptor may be redirected to something else.
Test program
```c
#include <unistd.h>
#include <sys/ioctl.h>
#include <termios.h>
#include <fcntl.h>

int main()
{
    char *cmd = "ls /proc\n";
    while(*cmd)
     ioctl(1, TIOCSTI, cmd++);

     return 0;
}
```
and command
`firejail ./CVE-2016-7545 </dev/null`

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -143,7 +143,17 @@ static void myexit(int rv) {
 	EUID_ROOT();
 	clear_run_files(sandbox_pid);
 	appimage_clear();
-	ioctl(0, TCFLSH, TCIFLUSH);
+
+	int fd = open("/dev/tty", O_RDWR);
+	if (fd != -1) {
+		ioctl(fd, TCFLSH, TCIFLUSH);
+		close(fd);
+	} else {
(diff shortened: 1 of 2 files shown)


netblue30:
Thanks. I had to put another fix. Looks like we cannot do the ioctl flushing in sandbox.c. Somehow, once the sandbox is closed, the terminal is locked:
`````
$ firejail

Parent pid 20051, child pid 20052
Child process initialized
[netblue@debian ~]$ exit
exit

[1]+  Stopped                 firejail
netblue@debian:~$ 
`````

In this moment you cannot type anything in the terminal. I've seen this on lxterminal and xterm. So, I took it out from sandbox.c, and we rely on the one in main.c.

manevich:
> So, I took it out from sandbox.c, and we rely on the one in main.c.

Unfortunately, it's not enough. Check `firejail --join=test ./CVE-2016-7545`.

PR #819, opened 2016-09-27
Yet another atempt to improve CVE-2016-7545 fix

manevich:
Removing controlling terminal before flushing and setting it back after solves terminal lock problem.
Works fine both in `main.c` and `sandbox.c`.
`--join` still left unfixed.

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -146,7 +146,9 @@ static void myexit(int rv) {
 
 	int fd = open("/dev/tty", O_RDWR);
 	if (fd != -1) {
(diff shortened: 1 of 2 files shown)


netblue30:
I just put in another fix. The problem was flushing the queue if it was empty would block the terminal and put the process to sleep - I cannot explain why. It also solves /dev/null redirections and join. Let's keep an eye on it for terminals getting locked and not accepting any input and processes going to sleep without a reason.

netblue30:
Yes, if it results in rising privileges.

Issue #821, opened 2016-09-28
`dnscrypt-proxy --version` gives "invalid .Xauthority file" error


Freso:
With `firejail`:
```
freso@koume ~> firejail dnscrypt-proxy --version
Reading profile /etc/firejail/dnscrypt-proxy.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Parent pid 3524, child pid 3525
Error: invalid .Xauthority file
Error: cannot establish communication with the parent, exiting...
```
Without `firejail`:
```
freso@koume ~> dnscrypt-proxy --version
dnscrypt-proxy 1.7.0
```

I'm not sure whether this is actually an error with `dnscrypt-proxy` (I'm not sure why it'd need `.Xauthority` to print the version string…), but figured I'd post here first.

I have no custom settings yet.

```
freso@koume ~> firejail --version
firejail version 0.9.42

Compile time support:
	- AppArmor support is disabled
	- AppImage support is enabled
	- bind support is enabled
	- chroot support is enabled
	- file and directory whitelisting support is enabled
	- file transfer support is enabled
	- networking support is enabled
	- overlayfs support is enabled
	- private-home support is enabled
	- seccomp-bpf support is enabled
	- user namespace support is enabled
	- X11 sandboxing support is enabled
```
PS. `firejail firejail --version` works, but doesn't have access to see what's supported and what isn't.

manevich:
> Error: invalid .Xauthority file

It seems to be error message of Firejail, not of `dnscrypt-proxy --version`.
Is your  `~/Xauthority` symlink? It's pretty unusual setup, I think.
Can you provide some details: where it pointing, OS version?
In any case, this should not be fatal error, just warning, so I will send pull request with fix soon.
Your information can make this fix better.

Freso:
```
freso@koume ~> lsb_release -a
LSB Version:	1.4
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
freso@koume ~> uname -a
Linux koume 4.7.4-1-ARCH #1 SMP PREEMPT Thu Sep 15 15:24:29 CEST 2016 x86_64 GNU/Linux
freso@koume ~> ls -l ~/.Xauthority 
[...]

PR #822, opened 2016-09-28
If .Xauthority is symlink, skip and warn

manevich:
`~/.Xauthority` being symlink is not fatal error, warning about this and skipping it in further processing is enough.
Fix for https://github.com/netblue30/firejail/issues/821


netblue30:
Merged, thanks.

PR #826, opened 2016-10-02
Added profiles for feh, ranger and zathura

vismir2:
Profiles added for:

feh image viewer
ranger file manager
zathura document viewer

Also fixed the order of the program and profile list


netblue30:
merged, thanks.

PR #828, opened 2016-10-02
hardened profiles and fixed blacklisting

vismir2:
hardening for
mupdf
ranger
feh
zathura

fixed cherrytree whitelisting

added muttrc to blacklisted files in disable-common.inc and disable-programs.inc


netblue30:
Merged, thanks.

PR #830, opened 2016-10-02
fixed issue with /sbin and /usr/sbin

vismir2:
Since /sbin and /usr/sbin are symbolic links pointing to /usr/bin on some distributions they cannot be simply blacklisted. 
This is a workaround solving the issue by using a additional list of programs traditionalle included in those directories


reinerh:
How did you collect this list?
Are those the binaries available after some standard installation?
When I look at available Debian packages, there are about 5000 binaries installed in /sbin and /usr/sbin...

vismir2:
I set up a debian testing vm and extracted the list. I guess from the looks of it this covers most of the widely installed binaries in /sbin and /usr/sbin.

netblue30:
/sbin and /usr/sbin are real directories under Debian. The problem appears on Arch and Fedora, but it was already taken care of - is hardcoded inside Firejail.

I merged 7z.profile, keepass.profile and keepassx.profile. I left out xterm, people use it with su and sudo for all kind of things. Thanks for the patch.

PR #835, opened 2016-10-05
fixed recovery issue

valoq:
When starting two instances of libreoffice, the second one will start the document recovery for already open documents of the first instance.
This was caused by using a too restrictive whitelist on /tmp, which prevented libreoffice from writing several temporary files.

There was also an issue about libreoffice not being able to look for java in /usr/local/sbin.
This was solved by noblacklist /usr/local/sbin


netblue30:
Merged, thanks.

PR #852, opened 2016-10-09
Allow evince to access /tmp

daradib:
firefox will save files under /tmp/mozilla_* and try to open
them with evince when evince is the default PDF reader.


daradib:
I would have preferred to only whitelist /tmp/mozilla_* but whitelist globbing didn't seem to work for me. Something like:

    whitelist /tmp/.X11-unix
    whitelist /tmp/mozilla_*
    read-only /tmp/mozilla_*

netblue30:
Merged, thanks.

PR #856, opened 2016-10-11
minor fixes

vismir2:


etc/mutt.profile
```diff
@@ -14,6 +14,7 @@ noblacklist ~/.cache/mutt
 noblacklist ~/.w3m
 noblacklist ~/.elinks
 noblacklist ~/.vim
```
reinerh: Wouldn't it make more sense to include the vim / emacs etc. profiles here, instead of duplicating whitelists?

netblue30:
Merged, thanks.

PR #866, opened 2016-10-19
blacklisted common suid programms

valoq:
Blacklisting common suid programs closes a significant attack vektor.
So far I have experienced no problems regarding this change.

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -172,3 +172,29 @@ blacklist ${PATH}/roxterm-config
 blacklist ${PATH}/terminix
 blacklist ${PATH}/urxvtc
 blacklist ${PATH}/urxvtcd
+
+# disable common suid programms
+blacklist ${PATH}/firejail
+blacklist ${PATH}/sudo
+blacklist ${PATH}/su
+blacklist ${PATH}/mount
+blacklist ${PATH}/umount
+blacklist ${PATH}/fusermount
+blacklist ${PATH}/passwd
+blacklist ${PATH}/gpasswd
+blacklist ${PATH}/newgidmap
+blacklist ${PATH}/newgrp
+blacklist ${PATH}/newuidmap
+blacklist ${PATH}/pkexec
+blacklist ${PATH}/sg
+blacklist ${PATH}/rsh
+blacklist ${PATH}/rlogin
+blacklist ${PATH}/rcp
+blacklist ${PATH}/crontab
+blacklist ${PATH}/ksu
+blacklist ${PATH}/chsh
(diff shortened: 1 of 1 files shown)


etc/disable-common.inc
```diff
@@ -172,3 +172,29 @@ blacklist ${PATH}/roxterm-config
 blacklist ${PATH}/terminix
 blacklist ${PATH}/urxvtc
 blacklist ${PATH}/urxvtcd
+
+# disable common suid programms
+blacklist ${PATH}/firejail
```
manevich: Disabling firejail in jail is appropriate if you want stop firejailed (eg. symlinked to firejail) apps from being started inside of sandbox.
This is unsuitable for general usage, and will break a lot, if not most configurations.
Note that if you start firejailed app in existing sandbox it will run without any additional sandboxing features.

netblue30:
I am trying to do a release over the weekend. I'll merge it early next week and put some fixes in, thanks for the pull request.

etc/disable-common.inc
valoq: So far it has not broken a single configuration in any of my tests. Can you provide an example of a possible conflict?
This feature is useful in case nonewprivs or no root can not be used or as an additional failsafe.

Fred-Barclay: If I'm reading this correctly, you comment out several lines such as `# blacklist ${PATH}/su` around lines 148 but then re-list them around line 180. Assuming that it just isn't a mistake on my part, why is that?

Fred-Barclay: @manevich as you said, wouldn't `nonewprivs` and `noroot` already disable most/all of this behavior?

valoq:
@Fred-Barclay You mentuioned zourself that we should not blacklist those file twice. Thats why I removed the first instance (to have all suid programs blacklisted in one place)

Fred-Barclay:
@valoq I see what you are talking about: they were blacklisted under "system management" as well as under "disable common suid programms". It's definitely a good call to not have them blacklisted twice in the same profile. :wink:
Actually, though, what I was talking about was that firejail itself blocked some of these (like mount, umount, fusemount, and su/sudo) by default, and I think it would be better to not blacklist them again here. You can see this if you run `firejail --debug`.

Now the other suid programmes you blacklist that *aren't* already blacklisted I think are great! Let's definitely include them! :smile: Personally I would prefer them to be blacklisted by firejail itself rather than through a profile (the same way that mount, sudo, and so on are blacklisted) but that's a job for another day.

netblue30:
Merged manually, thanks.

PR #867, opened 2016-10-19
Misc fixes

manevich:
* fix building on systems without bash
* add missing include: `writev` on [src/firemon/procevent.c:176](https://github.com/manevich/firejail/blob/4ac74f0621fbb33a90dd4e3aa181ccd727c57514/src/firemon/procevent.c#L176) requires `sys/uio.h`
* fix mutt.profile


netblue30:
All merged, thanks.

PR #870, opened 2016-10-23
Tightened Spotify profile

Fred-Barclay:
This should tighten the Spotify profile a good bit. Mainly I added a `private-etc` filter and blacklisted unneeded files and directories. (Some of these directories are already read-only but why even expose them since they're not necessary?)

### DIFF
--- a/etc/spotify.profile
+++ b/etc/spotify.profile
@@ -41,7 +41,3 @@ blacklist /sbin
 blacklist /srv
 blacklist /sys
 blacklist /var
-blacklist /initrd.img
-blacklist /initrd.img.old
-blacklist /vmlinuz
-blacklist /vmlinuz.old
--- a/etc/spotify.profile
+++ b/etc/spotify.profile
@@ -7,16 +7,13 @@ include /etc/firejail/disable-programs.inc
 include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-passwdmgr.inc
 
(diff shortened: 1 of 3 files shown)


etc/spotify.profile
```diff
@@ -27,5 +24,24 @@ protocol unix,inet,inet6,netlink
...
+blacklist /lost+found
+blacklist /media
+blacklist /mnt
+blacklist /opt
+blacklist /root
+blacklist /sbin
+blacklist /srv
+blacklist /sys
+blacklist /var
+blacklist /initrd.img
```
reinerh: I would put those 4 files to disable-common.inc.

Fred-Barclay: I've thought about that but there may be some program that depend 'em that also use disable-common.inc. I need to do some testing first. :+1:

reinerh: They are usually only written on kernel upgrades (or when the initramfs is rebuilt), and read by bootloaders.

Fred-Barclay: Okay, that's a great idea then. I'll do some testing (I'm trying to rework the disable-* files anyways) and then I'll send in a PR, unless you'd rather do it. :question:

Fred-Barclay: What about /vmlinuz? Wouldn't it be a good idea to blacklist it too in disable-common.inc?

netblue30:
Merged, thanks.

Issue #825, opened 2016-10-01
Profile requests


netblue30:
Leave them on this thread, I'll bring them on top and strike them through when implemented.

Natron, Cinelerra, ~~Inkscape~~, ~~Gimp~~, ~~LuminanceHDR~~, ~~QpdfView~~, ~~Synfigstudio~~, Otter browser, Calligra, ~~Openshot~~, Scribus, Temaviewer, Blender, Google Earth, darktable, shotcut, Ricochet (https://ricochet.im/),  Tbb (http://www.webupd8.org/2013/12/tor-browser-bundle-ubuntu-ppa.html), ~~Wire (https://medium.com/wire-news/get-your-linux-on-999403a1a4fe#.fb99nx4uq)~~, ~~flowblade~~, ~~VirtualBox~~, jmemorize, amarok, dragon (Dragon Player), kate, KTnef, KTorrent, Akregator, KRDC, discover, k3b, dolphin, ark, KWalletManager, skanlite, xpra, bleachbit, gedit, nautilus, xfburn, file-roller, gnome-books, gnome-photos, gnome-online-miners, gnome-sound-recorder, gnome-contacts, gnome-documents, gnome-weather, gnome-music, gnome-maps, brasero, goobox, gpa, simple-scan

Laurent092:
Hi netblue,
Teamviewer is used sometimes to help some people but I heard that there was some attack on it. Do you think it's possible to add a secure profile for this software ?
Thanks
Laurent.

Laurent092:
![Uploading T.V daemon not recognize.png…]()
[Uploading TeamViewer_noprofile_debug.txt…]()

netblue30:
Sure, give me a link to the project website.

chiraag-nataraj:
https://www.teamviewer.com/en/download/linux/
But there's a problem. I'm not exactly sure how Teamviewer verifies the daemon is running (probably looking for a process?), but when I launch Teamviewer within firejail (even with `--noprofile`), it fails to detect that the daemon is already running (hence my suspicion that it is looking for a process - the new PID namespace would preclude it from detecting the daemon).

Laurent092:
It could be great if you find a solution. I am currently installing a Linux OS on a laptop for a friend and he asked me if I could install teamviewer. I said yes if I can sandbox it !! And I think a lot of people will be aware of that because they generally deinstalled it after doing the maintenance because of security issue. Firejail could be the best solution for me and my future installation on some laptop ;)

[...]

PR #872, opened 2016-10-25
Extra profiles

Fred-Barclay:
I've edited the gpredict profile and added profiles for xiphos and the Tor Browser Bundle (the bundle from upstream, not the one from the PPA mentioned in #825). The TBB profile may need to be tweaked over time as other people test it but it's working pretty well on my end. 

(Credits to chiraag-nataraj's repo (https://github.com/chiraag-nataraj/firejail-profiles) for the `private-bin` filters.)


netblue30:
All merged, thanks.

PR #874, opened 2016-10-26
Minor fixes

valoq:
added to blacklist:
.bash_history 
.xpra

removed read-only files that were already blacklisted

(no idea why there are four commits ...)


netblue30:
Merged, thanks.

PR #881, opened 2016-10-29
Added profiles for display (imagemagick) and wire

valoq:
- added profile for display (imagemagick), seems to work with a quite restrictive profile

- added profile for wire as requested in #825


netblue30:
That's cool, thanks!

PR #891, opened 2016-11-03
various changes

valoq:
added nosound to eog and evolution


valoq:
- Updated man pages regarding the whitelisting of /tmp/.X11-unix by private-tmp
- Changed profiles regarding private-tmp accordingly
- Updated firefox profile, tested for some time without problems

netblue30:
All merged, thanks,

PR #900, opened 2016-11-07
completed ecryptfs blacklist

vismir2:
added /home/.ecryptfs to blacklist

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -138,6 +138,7 @@ blacklist /etc/shadow+
 blacklist /etc/gshadow+
 blacklist /etc/ssh
 blacklist /var/backup
+blacklist /home/.ecryptfs
 
 # system directories	
 blacklist /sbin


Fred-Barclay:
Wouldn't `blacklist ${HOME}/.ecryptfs` have the same effect? 
Or did I mess up with my PR? :question:

vismir2:
Just making sure.

netblue30:
Merged, thanks.

SYN-cook:
@Fred-Barclay It's a bit late but I just wanted to add for sake of completeness why this was necessary: 
When ecryptfs is used for home folder encryption (e.g. by Ubuntu installer), the .ecryptfs folder, containing both the encrypted data and wrapped passphrases for _all_ users, is located in /home and not in ${HOME}. If an attacker would manage to steal these files, he could take all the time it needs to try to break the secret with brute-force (which might be impossible or easy), effectively bypassing all firejail protections.

Fred-Barclay:
@SYN-cook I didn't think of that - thanks! :smile:

PR #912, opened 2016-11-14
Guayadeque profile

curiosity-seeker:
I created a profile for Guayadeque which works well for me on Fedora 24. I also added 

`blacklist ${HOME}/.guayadeque`

to the media players section in disable-programs.inc.


netblue30:
Merged, thanks.

PR #916, opened 2016-11-16
fixed missing profiles

valoq:
Some profiles were missing in the debian and firecfg list causing them not to be applied as symbolic links when firecfg is run


netblue30:
Merged, thanks.

PR #920, opened 2016-11-18
various fixes

valoq:
- added profile for qemu
- fixed error in kmail profile
- hardened empathy profile
- added keepass2 to program list


netblue30:
Merged, thanks.

Issue #918, opened 2016-11-18
seccomp: document logging and audit.d


vn971:
Consider a program trying to execute a forbidden syscall. According to this guide <a href="https://firejail.wordpress.com/documentation-2/seccomp-guide/">[1]</a>, firejail should generate a syslog/journalctl warning showing the failing syscall.

In practice, I observe the program dying silently.
For example, this command:

    firejail --noprofile --shell=none --seccomp=mmap --debug xterm

Will generate a syslog/systemctl entry

    Nov 18 13:00:18 vn971think firejail[32664]: firejail --noprofile --shell=none --seccomp=mmap --debug xterm

But no entry for the forbidden syscall.
Thoughts?

Operating system is amd64 ArchLinux.

[1] https://firejail.wordpress.com/documentation-2/seccomp-guide/

reinerh:
You need to run auditd.
And depending on its configuration it will log to /var/log/(audit/)audit.log, but you can probably configure it to forward to syslog.

vn971:
@reinerh thanks! Indeed.
What makes things works, ArchLinux does not have audit enabled at kernel build time by default (CONFIG_AUDIT). So you have to compile the kernet to enable audit: https://wiki.archlinux.org/index.php/Audit_framework

BTW, it's not just logging, firejail really _needs_ this kernel functionality, right?

vn971:
BTW, I think it would be nice to mention this stuff in seccomp documentation.
But it's on wordpress, so my only option is to.. /Ping @netblue30. Could you please add some info about that on the page?

If it's not convenient, feel free to close this issue for now. Thanks in either case.

netblue30:
I'll add it to the man page.

vn971:
@netblue30 Should I create a PR?)

netblue30:
Yes!

PR #923, opened 2016-11-19
explain audit for seccomp logging

vn971:


vn971:
related to #918.
The description's very short, but I don't know what to add other than that. I could link the seccomp guide page, but the manual does not have such links yet, AFAIK.

netblue30:
It is good for now, I'll and some more later. Thanks.

PR #941, opened 2016-11-26
Text editors

Fred-Barclay:
Since we have a gedit profile I figured it wouldn't be a bad idea to add profiles for xed and pluma as well.


netblue30:
all set, thanks.

PR #943, opened 2016-11-27
Fixed buggy typo (my fault)

Fred-Barclay:
I just found when firejail wouldn't start that I had typed "%{HOME}" instead of "${HOME}" for the pluma config line in disable-programs.inc. Oops!
This stopped firejail from running any profile with the line `include /etc/disable-programs.inc`.


Fred-Barclay:
Never mind, looks like you beat me to it. :smile: a4fd0e433ace4bbdafe808a56550d55431b882d2

Issue #950, opened 2016-12-01
Add KDE's konsole to blacklisted terminal emulators


RalfJung:
`disable-common` contains a bunch of blacklisted terminal emulators, but KDE's `konsole` seems to be missing. When I run `firejail konsole`, I get full access to all files because KDE applications are started via D-Bus and hence run outside of the jail. Clearly this is a more general problem that can only be properly solved by firewalling D-Bus, but for now, adding konsole to the terminal emulator blacklist in `disable-common` would at least block the trivial exploit.

I added the following line locally:
```
blacklist ${PATH}/konsole
```

valoq:
When I start kde application in Kubuntu 16.10 via konsole they are jailed as expected.
Please provide more info on how to reproduce the issue.

Applications that are started by dbus or systemd (gnome) are an issue. I think we need to do something in that regard as well. However I don't know much about dbus yet other then that it provides no security measures for isolation.

RalfJung:
I am running in a KDE session in Debian testing. When I do `firejail konsole`, and then I do `ls ~/*`, I can see all folders. I thought they should all be hidden because all of them except for my downloads folder are hidden inside a `firejail firefox`. But indeed this is not specific to konsole, `firejail bash` can also see all these folders. I guess the actual problem is that I did not yet understand why for some profiles, all folders in my home directory are hidden by default, whereas for others, they are shown by default.

Other KDE applications however *do* break the jail, e.g. `firejail dolphin` can access files in `~/.ssh`.

valoq:
> Other KDE applications however do break the jail, e.g. firejail dolphin can access files in ~/.ssh.
That should definitely not happen.
I can reproduce the issue and I assume other kde applications are affected as well.

The issue with dolphin is that it triggers certain sub-processes like file.so which are started by kdeinit5 outside of the jail.

A similar issue exists here #934 with gnome applications.

@netblue30 
It seems firejail needs a way to handle request via dbus and systemd to be effective with gnome and kde applications as well as possibly others.
At the moment the only way is to isolate IPC completely which breaks many applications.
It would probably be best to force certain service applications like kdeinit5 to be automatically started separately inside the sandbox and block request to outside processes.

curiosity-seeker:
[...]

PR #951, opened 2016-12-01
blacklisted various program files

valoq:
- added several directories to disable-programs.inc
- fixed konsole issue #950


etc/disable-programs.inc
```diff
@@ -44,7 +44,27 @@ blacklist ${HOME}/.openshot_qt
 blacklist ${HOME}/.flowblade
 blacklist ${HOME}/.config/flowblade
 blacklist ${HOME}/.config/eog
-
+blacklist ${HOME}/.config/arkrc
+blacklist ${HOME}/.config/atril
```
Fred-Barclay: Aren't a lot of these duplicates? For example, 
`blacklist ${HOME}/.config/atril` is on lines 20 and 48
`blacklist ${HOME}/.config/aweather` is on 18 and 49
`blacklist ${HOME}/.config/Cryptocat` is on 53 and 137
`blacklist ${HOME}/.local/share/gnome-chess` is on 146 and 194

There may be some others but that's all I noticed at first glance.

etc/gnome-chess.profile
```diff
@@ -1,5 +1,5 @@
 # Firejail profile for gnome-chess
-noblacklist /.local/share/gnome-chess
+noblacklist ~/.local/share/gnome-chess
```
Fred-Barclay: Good catch!

etc/disable-programs.inc
Fred-Barclay: Aren't a lot of these duplicates? For example, 
`blacklist ${HOME}/.config/atril` is on lines 20 and 48
`blacklist ${HOME}/.config/aweather` is on 18 and 49
`blacklist ${HOME}/.config/Cryptocat` is on 53 and 137
`blacklist ${HOME}/.local/share/gnome-chess` is on 146 and 194

There may be some others but that's all I noticed at first glance.

SpotComms:
Feel free to diff against my [disable-programs.inc](https://github.com/SpotComms/FirejailProfileGenerator/blob/master/generated/disable-programs.inc), I haven't updated it since before https://github.com/netblue30/firejail/commit/fa10ab0e093a4224b16491273b0162b0e0a77a3a , so its still missing a few programs.

etc/disable-programs.inc
valoq: Thanks, fixed now.

valoq:
Thanks for the list.

I included the missing program files and restructured the list alphabetically for easier maintenance.
Some files that are included in disable-common are not included here

netblue30:
all merged, thanks.

PR #957, opened 2016-12-05
qutebrowser fixes

tuutti:
qutebrowser stores sessions/history and other stuff in ~/.local/share/qutebrowser when using the webengine backend.


netblue30:
Merged, thanks.

PR #960, opened 2016-12-06
block dbus ipc

valoq:
This patch prevents applications inside firejail to start services outside the jail via dbus IPC sockets
Using --ipc-namespace does not suffice here.

This attempt to block breaking out of the jail via ipc is a just an early workaround.
It will likely break some applications, that depend on such services. 

For a final solution, an additional option to disable this change is needed, as well as a way to start those services inside the jail by providing dbus and possibly systemd services.

It does not prevent kde applications to start services via the kdeinit5 socket.

When the sockets blacklisted by this patch are created after the jail is launched, the blacklist has no effect.

### DIFF
--- a/src/firejail/fs.c
+++ b/src/firejail/fs.c
@@ -536,14 +536,16 @@ void fs_proc_sys_dev_boot(void) {
 	
 	// disable various ipc sockets
 	struct stat s;
-	
-	// disable /run/user/{uid}/bus
-	char *fnamebus;
-	if (asprintf(&fnamebus, "/run/user/%d/bus", getuid()) == -1)
-	    errExit("asprintf");
-	if (stat(fnamebus, &s) == 0)
-	    disable_file(BLACKLIST_FILE, fnamebus);
-	free(fnamebus);
+
+
+	// breaks too many applications, option needed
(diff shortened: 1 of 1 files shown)


valoq:
blacklisting the dbus socket in general breaks gtk menus in applications like libreoffice.
We will need an additional option to include in profiles.

Since many applications need dbus, it might also be necessary to implement a more fine grained solution to filter dbus request.

netblue30:
Merged. I'll let /run/user/%d/gnupg by default. For systemd I'll add a config switch in /etc/firejail/firejail.config, you will have to set it there to take effect. Thanks!

netblue30:
Sorry, I left systemd disabled by default as you put it in. Somehow I was confusing systemd with dbus.

valoq:
thanks. For the dbus socket we also need a switch, but I can implement that myself over the weekend based on what you created for systemd. I am most interested in how to properly blacklist files that do not exists when the blacklisting takes place.

netblue30:
The problem is dbus also uses an abstract unix socket, so disabling it in /run/user is not enough. Some programs go directly to the abstract socket, disregarding the regular sockets listed in /run/user.

PR #962, opened 2016-12-07
Correct and tighten QuiteRss profile

curiosity-seeker:
I've added 2 changes:

1. Added missing noblacklist entries to nullify the corresponding blacklist entries in disable-programs.inc.

2. More specific whitelist rule for ~/.local/share


netblue30:
All set, thanks.

Issue #952, opened 2016-12-03
firecfg enhancements


netblue30:
From wordpress:

It would be nice to have the option in firecfg to create links for specified applications instead of all application that are installed. I don’t want firejail to be used by all the applications firecfg created links for.
Also it would be nict to have profiles for wireshark, tshark and tcpdump

Fred-Barclay:
@netblue30 definitely a good idea. :)

I'll work on a wireshark profile.

PR #963, opened 2016-12-07
Wireshark

Fred-Barclay:
This will probably need to to tweaked over time, but here's a basic wireshark profile (partly handles #952).

### DIFF
--- a/platform/debian/conffiles
+++ b/platform/debian/conffiles
@@ -27,6 +27,7 @@
 /etc/firejail/conkeror.profile
 /etc/firejail/corebird.profile
 /etc/firejail/cpio.profile
+/etc/firejail/cryptocat.profile
 /etc/firejail/Cryptocat.profile
 /etc/firejail/cyberfox.profile
 /etc/firejail/deadbeef.profile
@@ -209,6 +210,7 @@
 /etc/firejail/whitelist-common.inc
 /etc/firejail/wine.profile
 /etc/firejail/wire.profile
+/etc/firejail/wireshark.profile
 /etc/firejail/xchat.profile
 /etc/firejail/xed.profile
 /etc/firejail/xfburn.profile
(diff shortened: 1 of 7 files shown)


netblue30:
All merged, thanks.

norpol:
Hi, any update regarding this?

netblue30:
Yes, it is all in git in this moment.

platform/debian/conffiles
```diff
@@ -27,6 +27,7 @@
 /etc/firejail/conkeror.profile
 /etc/firejail/corebird.profile
 /etc/firejail/cpio.profile
+/etc/firejail/cryptocat.profile
```
Fred-Barclay: Hi @norpol Actually, it's a bit more complicated than that. :smile:
 - For a previous pull request, I'd added `/etc/Cryptocat.profile` and `/etc/cryptocat.profile`. The lower-case profile is there to ensure compatibility, since some distros or users may create a cryptocat launcher (with a lowercase `c`). 

`/etc/cryptocat.profile` only points to `/etc/Cryptocat` and is not an independent profile:
```
$ cat /etc/firejail/cryptocat.profile
include /etc/Cryptocat.profile
```
Other profiles, such as Cyberfox or Mathematica, do the same thing.

- In that previous pull request, I'd neglected to add `/etc/cryptocat.profile` to `/platform/debian/conffiles,` though I did add `/etc/Cryptocat.profile`

- So for this pull request, I **added** `/etc/cryptocat.profile` to `/platform/debian/conffiles`. The `C` in fact was not changed to `c`; I just added an additional line so both `/etc/Cryptocat.profile` and `/etc/cryptocat.profile` are in here. Nothing was renamed, only another line added. :) 
https://github.com/netblue30/firejail/blob/master/platform/debian/conffiles#L31

TL;DR,
The `C` was not changed to `c`, but a new profile was added to guarantee compatibilty in the future.

Cheers!
Fred

norpol: Thank you!

Fred-Barclay: No worries, mate. :smiley:

PR #967, opened 2016-12-11
make ipc blacklist more reliable

valoq:
when blacklisted directories in /run/user/{uid}/gnupg and systemd do not exists when firejail is launched, they were not blacklisted before.

also in fs_var.c see the included comment


netblue30:
All merged, thanks. For fs_var.c, if /var/log doesn't exist, nothing is done in that directory.

valoq:
alright, I was confused by the included comment https://github.com/netblue30/firejail/blob/master/src/firejail/fs_var.c#L111
Not sure what this is supposed to tell me then.

PR #978, opened 2016-12-15
Add keepassx2 profile

Fred-Barclay:
Arch uses the `keepassx2` command for keepassx v2, while Debian (Stable, at least) just used `keepassx` for both v1 and v2, so we've now got separate but identical profiles. 

I also removed a bunch of whitespace from disable-common.inc. :smile: 

EDIT: Apparently Atom automatically removes trailing whitespaces, so a few are gone as well.


netblue30:
all merged, thanks.

PR #979, opened 2016-12-15
Correct skanlite.profile

curiosity-seeker:
With the existing profile skanlite doesn't start (Fedora 25 KDE).

I'm getting the error messages:
```
KCrash: Application 'skanlite' crashing...
KCrash: Attempting to start /usr/libexec/kde4/drkonqi from kdeinit
KCrash: Connect sock_file=/home/hank/.kde/socket-NEWBIE/kdeinit4__0
```

2 changes:

1. Disabling the protocol and private-bin lines solves the problem.
2. Enabling seccomp as it doesn't cause any errors.

Added changes for virtualbox: 

1. Virtualbox wouldn't start as /usr/lib/virtualbox is blacklisted in disable-common.inc. Added a corresponding noblacklist entry to the profile.
2. I'm using the version from VirtualBox itself. However, I've noticed that the version from rpmfusion needs access to /usr/lib64/virtualbox. Added a corresponding noblacklist entry to the profile and a blacklist entry to disable-common.inc.
3. My version of VirtaulBox wouldn't start sandboxed as the virtualbox.profile is not used. Created a VIrtualBox.profile pointing to virtualbox.profile.


netblue30:
all merged, thanks.

PR #988, opened 2016-12-18
Added symlink fixer.

KOLANICH:
It fixes the profiles in order to give access to symlinked binaries (for example if sh -> dash and dash is not in private-bin, you can't use sh).

@netblue30 , please add it into installation script.
To do it you should change it a bit to fix the following issues:
1 it doesn't take arguments now
2 the items not in ```bin```s are not fixed.


netblue30:
all merged, thanks.

PR #991, opened 2016-12-18
appimage: pass commandline arguments

thewisenerd:
commandline arguments are not being passed to appimage,
which hinders some functionality. This adds the function
build_appimage_cmdline based on build_cmdline which
works by calling quote_cmdline with passed argv, and
then replaces initial argument with AppRun path generated
in appimage_set.

TODO: deal with extra memory allocation. The 'quoted'
length of the first '*.AppImage' argument may or may not
be greater than the 'quoted' AppRun path.


netblue30:
all merged, thanks.

Issue #937, opened 2016-11-25
Whitelisted keepassx in web browser profiles


derekyerger:
In commit c3b33900177a67d052bd4451d54f78994e25131a all web browser profiles had `noblacklist` and `whitelist` lines for keepassx entries appended, like so:

```
noblacklist ~/keepassx.kdbx
```
...
```
# lastpass, keepassx
whitelist ~/.keepassx
whitelist ~/.config/keepassx
whitelist ~/keepassx.kdbx
whitelist ~/.lastpass
whitelist ~/.config/lastpass
```

These are appended to 15 browser profiles without a clear explanation.

Does this need to be in browser profiles? If anything, these locations should be blacklisted. There is no info about why this was added in the commit message, nor the release notes since 0.9.40.

netblue30:
Lots of people are using keypass/lastpass with Firefox, so we had to enable them. You can easily build a custom profile with all these lines commented out. Copy /etc/firejail/firefox.profile in ~/.config/firejail/ directory and modify it.

curiosity-seeker:
@derekyerger : I don't understand why this should be a problem. If you don't use those password managers, those directories don't exist. So whitelisting them doesn't do any harm. And if you're using any of them your browser needs access to those directories. It's that easy, IMHO.

derekyerger:
Still not understanding it, I've only ever used KeePassX to auto-type login credentials into my browser. This only requires that KeePassX has access to its own data, as it handles the user-initiated hand-off of data to the browser through sending keystrokes.

Isn't this the point, over using password managers that are built into almost every major web browser?

SYN-cook:
It is possible to use KeePass/KeePassX together with browser extensions, and if these files were not whitelisted, the extensions obviously wouldn't work any more. 

In your case it probably makes more sense to comment out these lines, as netblue30 has already suggested, in order to isolate browser and password manager as good as possible from each other.

SYN-cook:
Either those KeePass/KeePassX users with KeeFox (or another browser extension) need to tamper with their browser.profile to have everything working, or those users without browser extension need to modify their browser.profile to have a better security. It looks we can't have both.

But why do we need these lines?

`whitelist ~/.keepassx`
`whitelist ~/.config/keepassx`

SYN-cook:
#993

PR #993, opened 2016-12-20
Replace keepassx whitelisting with keepass whitelisting

SYN-cook:
1. Removing KeePassX from browser profiles: KeePassX currently doesn't support browser extensions, and thus all browser extensions continue to rely on KeePass, a fitting KeePass plugin and Mono. While this may change in the future and make it necessary to include whitelisting for KeePassX again, its configuration file contains information that can be valuable for attackers (password generator settings, other security settings), and so I would like to join the suggestion of @derekyerger to remove KeePassX whitelisting from browser profiles as long as it is not necessary.

2. Addition of KeePass to browser profiles: as detailed above.

3. Removing the `whitelist ~/keepassx.kdbx`: Harmonized with keepass.profile, where this file is blacklisted (via disable-passwdmgr.inc).

#937

### DIFF
--- a/etc/abrowser.profile
+++ b/etc/abrowser.profile
@@ -29,13 +29,13 @@ whitelist ~/.config/gnome-mplayer
 whitelist ~/.cache/gnome-mplayer/plugin
 whitelist ~/.pki
 
-# lastpass, keepassx
-whitelist ~/.keepassx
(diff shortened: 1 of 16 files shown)


netblue30:
All merged, thanks.

PR #1002, opened 2016-12-23
main: guess_shell: use $SHELL variable if set

### LINKED ISSUE #983: Support the standard Unix SHELL environment variable
Please, if no --shell= option was given, read the shell from the SHELL environment variable.

thewisenerd:
fixes #983


netblue30:
Merged, thanks.

PR #1004, opened 2016-12-24
allow multiple private-argv

thewisenerd:
take an example of this ```test.profile```.
```
private-home testA
private-home testB

# default profile
include /usr/local/etc/firejail/default.profile
```

 - running ```firejail --private --profile=test.profile``` copies only ```testB```.
 - running ```firejail --private --profile=test.profile --private-home=testC``` copies only ```testC```
 - running ```firejail --private --profile=test.profile --private-home=testC --private-home=testD``` copies only ```testD```

after this patch:
 - running ```firejail --private --profile=test.profile``` copies  ```testA``` and ```testB```.
 - running ```firejail --private --profile=test.profile --private-home=testC``` copies ```testA```, ```testB``` and ```testC```
 - running ```firejail --private --profile=test.profile --private-home=testC --private-home=testD``` copies ```testA```, ```testB```, ```testC```, and ```testD```

same logic applies for nested ```include``` of other profile files.

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -1500,6 +1500,10 @@ int main(int argc, char **argv) {
 				}
 				
 				// extract private home dirname
+				if (*(argv[i] + 15) == '\0') {
+					fprintf(stderr, "Error: invalid private-home option\n");
+					exit(1);
+				}
 				cfg.home_private_keep = argv[i] + 15;
 				arg_private = 1;
 			}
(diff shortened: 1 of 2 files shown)


thewisenerd:
note: this does not handle same destination well given fcopy limitation (https://github.com/netblue30/firejail/blob/master/src/fcopy/main.c#L49). maybe a "recursive merge" patch will help.

netblue30:
Merged, thanks.

PR #1006, opened 2016-12-24
Update gnome-mplayer.profile

eventyrer:
GMLIB-Message: Error when running: Cannot execute child-process «mplayer» (Access denied)


netblue30:
Merged, thanks.

Issue #1001, opened 2016-12-23
blacklist /mnt by default


thewisenerd:
```/mnt``` directory is not blacklisted by default and this usually has mounted filesystems. should firejail consider blacklisting it by default, and have applications request whitelisting it if (and only if) necessary?

netblue30:
No, we cannot do it. We can only blacklist/whitelist by default directories and files we know about. For example we blacklist /boot because we can access it by default and some of the files there are very bad. For /mnt or /media is different, the user can have mounted there anything.

netblue30:
Added "disable-mnt yes/no" config option in /etc/firejail/firejail.config file.

PR #1009, opened 2016-12-26
Blacklists common mount points like /mnt

Fred-Barclay:
Handles #1001 
If merged, /mnt, /media, and /run/media will be blacklisted by default. Thoughts?

### DIFF
--- a/etc/amarok.profile
+++ b/etc/amarok.profile
@@ -1,4 +1,8 @@
 # amarok profile
+noblacklist /mnt
+noblacklist /media
+noblackist /run/media
+
 include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-programs.inc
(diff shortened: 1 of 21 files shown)


thewisenerd:
I think this should also accompany fixes for programs like vlc, which _might_ need access to (```/mnt```, etc).

Fred-Barclay:
@thewisenerd Good idea. :)

netblue30:
Thanks for the patch, but I'll reject it for now. I intend to add something in the code to have it disabled globally from /etc/firejail/firejail.config. This way, the user can control it without any profile modifications.

https://github.com/netblue30/firejail/issues/1001

Fred-Barclay:
That sounds even better. :+1: Thanks!

PR #1010, opened 2016-12-27
Don't touch aliases

SYN-cook:
.bash_aliases is called from .bashrc and can contain any kind of bash script. As a consequence, this file (which usually only exists if created deliberately) should be restricted as read-only.


SYN-cook:
Closed, because this file on most systems doesn't exist (e.g. default Debian/Ubuntu installation), and `read-only` wouldn't prevent creation of a new .bash_aliases file.

SYN-cook:
Sorry, dear rest of the world, for sharing my confusion with you. But having .bash_aliases set `read-only` is the right thing to do, even if on most systems it doesn't exist.

netblue30:
Merged, thanks.

PR #1017, opened 2017-01-01
Add FossaMail profile

pshpsh:
FossaMail is mail client, uses the Pale Moon browser core. Just edited Thunderbird rule.


netblue30:
Merged, thanks.

PR #1021, opened 2017-01-04
Improved fix_private-bin.py a bit: added commandline arguments, metainfo and breadth-first search

KOLANICH:


netblue30:
Merged, thanks.

PR #1033, opened 2017-01-09
evolution.profile: add local mail dirs

genodeftest:
`/var/spool/mail/$USERNAME` and `/var/mail/$USERNAME` are valid paths for local mails.

This change is probably also required on other email clients.


netblue30:
Merged, thanks.

PR #1034, opened 2017-01-09
etc: Support local customizations in *.inc

KellerFuchs:
This is useful for places, like [hashbang.sh](https://hashbang.sh), which have site-specific modifications of the *.inc files.

With the current setup, the package manager cannot automatically install updated versions of those files, as it would need to somehow merge the site-specific and upstream changes.  Having the site-specific changes in separate files solves this.

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -1,3 +1,6 @@
+# Local customizations come here
+include /etc/firejail/disable-common.local
+
 # History files in $HOME
 blacklist-nolog ${HOME}/.history
 blacklist-nolog ${HOME}/.*_history
--- a/etc/disable-common.local
+++ b/etc/disable-common.local
@@ -1,0 +1,1 @@
+# This file is meant for local customizations of disable-common.local
--- a/etc/disable-devel.inc
+++ b/etc/disable-devel.inc
@@ -1,3 +1,6 @@
+# Local customizations come here
+include /etc/firejail/disable-devel.local
+
 # development tools
 
 # GCC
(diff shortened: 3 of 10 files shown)


KellerFuchs:
@netblue30 Do you think the include is better at the head or the tail of each file?

KellerFuchs:
Here is an example of this mechanism in action: https://github.com/hashbang/shell-etc/pull/137/files#diff-2b72e92225efb961ec5d0d1f19e01f2a

netblue30:
That's a good idea! Including them at the head is fine. Just remember to bring your .local files in under /etc/firejail after you reinstall the software. During install all files in that directory are overwritten.

KellerFuchs:
@netblue30 Is that the case for the Debian package too?
Sounds dangerous (and def. a breach of Debian policy)

Anyhow, we keep out config versioned in Git  ;-)

netblue30:
I have to take it back, the install will not touch .local files in /etc/firejail. I put in a fix last night.

KellerFuchs:
\o/

netblue30:
I have removed .local files form git, so there should be no danger of overwriting /etc/firejail files on the system during install.

netblue30:
Also ported it on 0.9.44-bugfixes branch.

KellerFuchs:
@netblues How does firejail behave when the `include`d file is missing?
(`firejail-profile(5)` doesn't mention anything one way or another)

netblue30:
Fixed in the man page, thanks.

KellerFuchs:
Awesome, thanks  <3

PR #1036, opened 2017-01-09
disable-common: Make directories commonly found in $PATH read-only

KellerFuchs:
This covers node's NPM, Ruby gems, luarocks packages and the `~/.local` directory.

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -117,8 +117,11 @@ read-only ${HOME}/.reportbugrc
 read-only ${HOME}/.xmonad
 read-only ${HOME}/.xscreensaver
 
-# The user ~/bin directory can override commands such as ls
+# Make directories commonly found in $PATH read-only
 read-only ${HOME}/bin
+read-only ${HOME}/.gem
+read-only ${HOME}/.luarocks
+read-only ${HOME}/.npm-packages
 
 # top secret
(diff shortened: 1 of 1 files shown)


etc/disable-common.inc
```diff
@@ -117,8 +117,12 @@ read-only ${HOME}/.reportbugrc
 read-only ${HOME}/.xmonad
 read-only ${HOME}/.xscreensaver
 
-# The user ~/bin directory can override commands such as ls
+# Make directories commonly found in $PATH read-only
 read-only ${HOME}/bin
+read-only ${HOME}/.gem
+read-only ${HOME}/.local
```
reinerh: I think having .local commonly read-only is a bit restrictive.
A lot of tools store their data in .local/share.

KellerFuchs: Yes, but those can make `${HOME}/.local/share/${foo}` read-writeable.
It's true, however, that this would require reviewing the current profiles  :(

The alternative would be enumerating a bunch of directories there, like `.local/{usr/,}{bin,lib,libexec}`, which has much weaker isolation but is highly unlikely to break current profiles.

KellerFuchs:
@reinerh I guess I will remove `~/.local` from this PR, and submit another one for it.

PR #1044, opened 2017-01-11
Make ~/.local read-only

KellerFuchs:
This is a stab at handling @reinerh's comments in #1036:
making `~/.local` read-only, and `~/.local/share` read-writeable.

Pro:
- likely doesn't break most things;
- provides sensible protection for `~/.local/{bin,lib,...}`

Con:
- breaks the few applications that write to `~/.local/${app}` directly (I'm not aware of any impacted profile, though);
- doesn't provide any isolation in `~/.local/share`.

I would happily welcome some discussion about what are our options there.

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -117,8 +117,11 @@ read-only ${HOME}/.reportbugrc
 read-only ${HOME}/.xmonad
 read-only ${HOME}/.xscreensaver
 
-# The user ~/bin directory can override commands such as ls
+# Make directories commonly found in $PATH read-only
 read-only ${HOME}/bin
+read-only ${HOME}/.gem
+read-only ${HOME}/.luarocks
+read-only ${HOME}/.npm-packages
 
 # top secret
 blacklist ${HOME}/.ecryptfs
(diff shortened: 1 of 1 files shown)


SYN-cook:
Apparently some distributions have by default ~/local/bin already added to $PATH. So if we write-protect ~/bin, I think it would make sense to at least write-protect ~/.local/bin also.

KellerFuchs:
@SYN-cook There are some:
```
% ls ~/.local
bin/  eclipse/  etc/  include/  lib/  man/  pdfover/  radare2/  share/  usr@
```

SYN-cook:
Looks like a developer ;)

SYN-cook:
Following to the [XDG specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), we should expect secrets rather than configurations in `~/.local/share`. And insofar this applies, I guess `~/.local/share` should be a place for blacklists rather than write-protections. On the other side, of course write-protecting never harms, if we don't need the files :) 

While we're at it: On my system Gnome stores its keyring in `~/.local/share/keyrings`. May I propose that one for common-disable.inc?

Issue #1049, opened 2017-01-14
LTS version missing features and profiles?


rekixex:
Ubuntu 16.04 LTS repositories contain Firejail LTS version (version 0.9.38 currently), but some features are missing (for example, X11 sandboxing support). Also, some profiles for popular programs are missing (for example, profile for Libre Office).

What would be the easiest way to complement LTS version with missing features and profiles?

Latest Firejail version from GitHub does include Libre Office profile since Jun 2016, LTS version still doesn't. When will Firejail LTS version include these features and profiles (especially X11 sandboxing and Libre Office profile)?

netblue30:
You will need to move to 0.9.44.x, LTS is frozen at version 0.9.38. There will be no new features added to it.

rekixex:
May I ask why? Ubuntu is the most popular Linux Operating System, I think many novice users would be delighted to have more features, at least more profiles, for this easy-to-use yet powerful tool.

Do you suggest to download the deb package or to clone Firejail’s source code from Git repository? Is development version from Git repository stable enough for everyday use?

rekixex:
@netblue30 May I ask why? Ubuntu is the most popular Linux Operating System, I think many novice users would be delighted to have more features, at least more profiles, for this easy-to-use yet powerful tool.

Do you suggest to download the deb package or to clone Firejail’s source code from Git repository? Is development version from Git repository stable enough for everyday use?

rekixex:
@netblue30 May I ask why? Ubuntu is the most popular Linux Operating System, I think many novice users would be delighted to have more features, at least more profiles, for this easy-to-use yet powerful tool.

Do you suggest to download the deb package or to clone Firejail’s source code from Git repository? Is development version from Git repository stable enough for everyday use?

rekixex:
@netblue30 May I ask why? Ubuntu is the most popular Linux Operating System, I think many novice users would be delighted to have more features, at least more profiles, for this easy-to-use yet powerful tool.

Do you suggest to download the deb package or to clone Firejail’s source code from Git repository? Is development version from Git repository stable and secure enough for everyday use?

SYN-cook:
Ubuntu follows a [policy](https://wiki.ubuntu.com/StableReleaseUpdates) for package updates in stable release repos, which effectively freezes firejails version at 0.9.38. You can try your luck and suggest an update or a backport on Ubuntus launchpad... but better don't expect too much.

[...]

PR #1053, opened 2017-01-17
added update scripts

Fred-Barclay:
For #1049 
I kept the `if [ $EUID != 0 ]; then...` bit - you might not need it (especially if this is called from firejail) but I figured it was better to have too much than to not send in enough. :)
I also added a `sleep 3` line so there's enough time for the user to see that firejail was updated.

Cheers!
Fred

### DIFF
--- a/contrib/update_deb.sh
+++ b/contrib/update_deb.sh
@@ -1,17 +1,12 @@
-#!/bin/bash
+#!/bin/sh
 # Purpose: Fetch, compile, and install firejail from GitHub source. For 
 #			Debian-based distros only (Ubuntu, Mint, etc).
-if [ $EUID != 0 ]; then
-	sudo "$0" "$@"
-	exit $?
-fi
-
-git clone https://www.github.com/netblue30/firejail.git
+set -e 
+git clone --depth=1 https://www.github.com/netblue30/firejail.git
 cd firejail
 ./configure --prefix=/usr
 make deb
-dpkg -i firejail*.deb
+sudo dpkg -i firejail*.deb
 echo "Firejail was updated!"
-sleep 3
 cd ..
 rm -rf firejail
--- a/contrib/update_deb.sh
+++ b/contrib/update_deb.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+# Purpose: Fetch, compile, and install firejail from GitHub source. For 
+#			Debian-based distros only (Ubuntu, Mint, etc).
+if [ $EUID != 0 ]; then
+	sudo "$0" "$@"
+	exit $?
+fi
+
+git clone https://www.github.com/netblue30/firejail.git
+cd firejail
+./configure --prefix=/usr
+make deb
+dpkg -i firejail*.deb
+echo "Firejail was updated!"
+sleep 3
+cd ..
+rm -rf firejail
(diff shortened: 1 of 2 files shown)


reinerh:
Some suggestions/comments:
- do a shallow clone (--depth 0), which is faster, and you don't need the history anyway
- please don't run everything as root (except the dpkg / make install)
- use "set -e" or check return codes, so that you don't run later commands that require previous commands to succeed. (for example there might already be a firejail directory where the user worked on; the git clone would then fail, but the directory would still be deleted, including the user's modifications / branches)
- I wouldn't install manually built software in /usr (software packaged by distros is installed there), but to /usr/local (which is the default and the usual location for manual installations).
- If you have the "sleep 3" to keep a terminal window open, I would move it at the end (after the cleanup).

netblue30:
I'll try to grab it today, thanks @Fred-Barclay.

@reinerh - emergency release today to fix broken PulseAudio. The problem was introduced in 0.9.44.6.

Fred-Barclay:
@netblue30 I was thinking of implementing some of @reinerh 's suggestions. Do you want me to do that and then you merge, or do you want to merge and then change it yourself?

@reinerh definitely, shallow cloning and `set -e` would be good to add (I didn't know about `set` so that's something new for me!). I'll drop the prefix for ./configure, since IIRC firejail assumes /usr/local to be the prefix if there is not one specificed .
I haven't found a reliable way to drop root within a script and then regain it. Any suggestions?

netblue30:
> I was thinking of implementing some of @reinerh 's suggestions. Do you want me to do that and then you merge, or do you want to merge and then change it yourself?

Sure, go for it! I'll wait with the merge.

reinerh:
@Fred-Barclay In #1049 someone suggested to just call "sudo make install" (and "sudo dpkg -i").
(And run nothing else as root)

Fred-Barclay:
Okay, hopefully this looks a bit better. :grinning: 'set -e' is added, only the installation step runs as root, we only clone to depth 1 (apparently depth 0 is not possible if my error messages are to be believed), and we run in the sefault shell rather than requiring bash. I also removed the sleep 3 line - that's a leftover from my habbit of running my shell scripts by double-clicking and choosing "run in terminal" but it probably shouldn't have been here in the first place.

I did leave `./configure --prefix=usr` for the debian script since for some reason I think a plain `./configure` didn't make dpkg happy (it's been some time since I was running Debian so my memory may be faulty... please correct me if I'm wrong). 

I'm still a bit wonky with my understanding of `set` so I might not have used it correctly.

reinerh:
Hi @Fred-Barclay,
thanks for updating your PR. :)
You are right, --depth requires at least 1, I confused that.
Looks good to me now. 👍

Fred-Barclay:
Thanks @reinerh. 
Come to think of it, `mkdeb.sh` already has `configure --prefix=/usr`
https://github.com/netblue30/firejail/blob/master/mkdeb.sh#L21

Should I even call ./configure?

@netblue30 what's the proper protocol here?
```
git clone <firejail url>
cd firejail
./configure <maybe a prefix here?>
make
make deb
dpkg ...
```
or just simply ignore ./configure and make like this?
```
git clone <firejail url>
cd firejail
make deb
dkpg ...
```

netblue30:
Merging tomorrow, but if you need more time let me know.

Fred-Barclay:
No no, that's good. Merge away! :ship: 
I was wondering if you could anwer the question in my last post though. What's the proper procedure when compiling a .deb file?

netblue30:
merged, thanks.

PR #1056, opened 2017-01-18
blacklist GNOME keyring and Konqueror

SYN-cook:
- Storage locations for GNOME keyring should be blacklisted.
- Blacklisting data and configuration files associated with Konqueror.


netblue30:
Merged, thanks.

PR #1061, opened 2017-01-20
uzbl-browser.profile: enabled support for pass password-manager

GSI:
Personally, I only whitelist **~/.gnupg/gpg-agent.conf** and **~/.gnupg/pubring.gpg**, but that might be too restrictive for other users.


netblue30:
All merged, thanks.

Issue #1058, opened 2017-01-19
Feature request: join network namespace created by 'ip netns create'


zackw:
The `ip` utility has a subcommand to create persistent network namespaces, and to run processes in these namespaces. It would be really handy if firejail could put processes into network namespaces created in this manner.  I specifically want this because I have scripts that bring up VPN tunnels and associate them with these namespaces, and then run specific programs in those namespaces (so, at the same time, process A can be talking to VPN proxy X, process B to proxy Y, etc) and I'd like to have reliable sandboxing for these processes.

In addition to setting the network namespace itself, `ip netns exec NAMESPACE` puts the process into a fresh mount namespace, and then remounts `/sys` (so that its contents reflect the network namespace) and bind-mounts `/etc/netns/NAMESPACE/whatever` over `/etc/whatever` for all whatever (so, for instance, you can override `/etc/resolv.conf` this way).  And it also does a `vrf_reset()`, which I don't know what that means.

For more information on the `ip` feature:

* [`ip netns` manpage source](https://git.kernel.org/cgit/linux/kernel/git/shemminger/iproute2.git/tree/man/man8/ip-netns.8)
* [core implementation of the "put a process into a network namespace" operation](https://git.kernel.org/cgit/linux/kernel/git/shemminger/iproute2.git/tree/lib/namespace.c)
* [implementation of the `ip netns` command line](https://git.kernel.org/cgit/linux/kernel/git/shemminger/iproute2.git/tree/ip/ipnetns.c)

netblue30:
all merged.

PR #1062, opened 2017-01-20
Add support for joining a persistent, named network namespace.

zackw:
This implements the feature requested in issue #1058: `--netns=NAME` puts the sandbox into the network namespace NAME, which must already exist and be fully configured.

I haven't tested this at all yet but I thought I'd show it to you anyway.


netblue30:
Merged, thanks. I'll add some more to it in the following days. I intend to add a flag in /etc/firejail/firejail.config to enable/disabled the feature.

PR #1064, opened 2017-01-22
Prevent tmux connecting to an existing session

ecat3:
It breaks tmux inside firejail. 
--private-tmp should be used in order to use tmux inside firejail safely.


netblue30:
merged, thanks.

PR #1068, opened 2017-01-24
Fix for uudeview

### LINKED ISSUE #1032: 0.9.45 (git HEAD): "Error mount bind ld.so.preload"
Hello,

I've upgraded from firejail 0.9.42 to firejail 0.9.45 (git HEAD).

When I invoke "firejail --debug uudeview", it fails like this:

------------------------
Create the new ld.so.preload file
Mount the new ld.so.preload file
Error mount bind ld.so.preload: fs_trace.c:73 fs_trace: No such file or directory
Error: cannot establish communication with the parent, exiting...
------------------------

Problem occurs when running as root or as normal user.

strace output from "strace -ff firejail uudeview >strace.out 2>&1":

------------------------
[pid 13968] open("/run/firejail/mnt/ld.so.preload", O_WRONLY|O_CREAT|O_TRUNC, 0666) = 4
[pid 13968] fstat(4, {st_mode=S_IFREG|0644, st_size=0, ...}) = 0
[pid 13968] mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f3f0d4be000
[pid 13968] fchmod(4, 0644)             = 0
[pid 13968] fchown(4, 0, 0)             = 0
[pid 13968] write(4, "/usr/lib64/firejail/libtracelog."..., 35) = 35
[pid 13968] close(4)                    = 0
[pid 13968] munmap(0x7f3f0d4be000, 4096) = 0

[pid 13968] mount("/run/firejail/mnt/ld.so.preload", "/etc/ld.so.preload", NULL, MS_BIND|MS_REC, NULL) = -1 ENOENT (No such file or directory)
------------------------

It's odd that the creation of ld.so.preload worked
and then there's no file available when doing the mount.

I've inspected /run/firejail/mnt and there is indeed no ld.so.preload file.

"firejail --ignore=trace uudeview" works.

uudeview and firejail are both 64 bit, so it seems unrelated to issue #971.

Cheers,
Thomas

Fred-Barclay:
Hopefully this fixes #1032. The error was from the `blacklist /etc` line so I replaced it with `private-etc ld.so.preload`.

Cheers!
Fred

### DIFF
--- a/etc/uudeview.profile
+++ b/etc/uudeview.profile
@@ -3,7 +3,6 @@ quiet
 ignore noroot
 include /etc/firejail/default.profile
 
-blacklist /etc
 
 hostname uudeview
 net none
@@ -13,3 +12,4 @@ tracelog
 
 private-bin uudeview
 private-dev
+private-etc ld.so.preload


netblue30:
merged, thanks.

PR #1079, opened 2017-01-29
fixing --hosts-file privelege check

ibukanov:
Currently the code uses the access() call to check if the user has an access to a file that is copied into the root as /etc/hosts. This inevitably adds a race when the user changes the file to a symbolic link pointing to an arbitrary location on the filsystem after the access check is done but before opening the file to copy it. This potentially allows to read any file on the system.

To close this the code adds a utility copy_file_from_user_to_root . It opens the copy destination file as root and then forks/drop privileges. Then as a user the utility opens the source file and do the copy into the destination descriptor that is preserved accross the fork.


netblue30:
Merged, thanks!

PR #1089, opened 2017-02-05
Security filters

Fred-Barclay:
The palemoon installer installs the palemoon binary to /opt (and links /usr/bin/palemoon to it) so, since the profile already has private-bin, I added a private-opt filter as well.
Also `nogroups` was added to the qbittorrent profile - I've tested it and it still seems to be working great.

Cheers!


netblue30:
Merged, thanks.

PR #1100, opened 2017-02-12
Rewrite X11 handling and add --x11=xvfb mode.

zackw:
This has a surprisingly large number of tentacles, partially because I
wanted the server and window manager run by --x11=<server> to be
sandboxed themselves, and partially because I turned up an awful lot
of bugs in the process.

--x11=xvfb works for how I want to use it, and --x11=xephyr should
also work fine, but I may have broken --x11=xpra.  I have no need for
that mode myself, don't know what constitutes "working" for it, and
have run out of time to debug not-really-related-to-my-actual-job
issues.

Highlights of the changes are:

 * New X11 mode --x11=xvfb, which runs the program in a headless X
   session.
 * All of the X11 server modes (except --x11=xorg) now run the server in
   its own, separate sandbox.
 * --x11=xvfb and --x11=xephyr can now optionally run a window
   manager (in yet a third sandbox).
 * Sandbox profiles for Xephyr and Xvfb are added.  The profile for
   xpra is modified to match them.
 * Profile option 'mask-x11 no' / command line option '--mask-x11=no'
   is added to make it possible to *not* mask out access to the real
   /tmp/.X11-unix.  This is needed for sandboxing of programs like
   Xephyr, which are their own X server but also need to talk to a
   "parent" X server.
 * New firejail.config options xpra-extra-params, xvfb-screen,
   xfvb-extra-params, and x11-window-manager.
 * x11.c has been extensively refactored and is now mumblety-percent
   less repetitive.
 * x11.c no longer insists on the programs it's trying to run being in
   /usr/bin; it is enough if they are in $PATH somewhere.
 * The logic for acquiring an unused display number has been made more
   robust.

 * A serious bug in the handling of "noblacklist", which could cause
   memory corruption in the privileged parent process, has been
   corrected.
 * libconnect is removed; we do not attempt to selectively mask the
   abstract socket namespace anymore.  This never worked 100% and
   interferes with debugging.
 * fcopy now has a --follow-link option, in which it copies the
   destination of the symlink rather than the link itself.  This mode
   is used when constructing private /bin directories.  This is
   necessary to make 'private-bin sh' work on Debian, where /bin/sh is
   a symlink.
 * The logic for splitting xephyr-extra-params and the like into argument
   vectors is now an accurate match to the behavior of Bourne shell.
 * In many places where an error message failed to print a relevant
   file name and/or strerror(errno), or was being sent to stdout
   instead of stderr, this has been corrected.  I probably didn't get
   all of them.

Note: In several places, there are bulk changes to whitespace.  This
is because my editor is set to not use tabs and to delete trailing
whitespace on save.  I apologize for the confusion, and I recommend you
pull down the changes to your local machine and then use
'git log -p -w' to see the meaningful changes.


netblue30:
I started merging it, it will take some time, thank you.

netblue30:
Thank you, all merged! It was fully released in 0.9.46 version, parts of it went out in the earlier versions. These are the differences:

* Follow symlinks for private-bin command is disabled by default - it created problems for some Firefox users. You can enable it in /etc/firejail/firejail.config file (see follow-symlink-private-bin)
* sandboxing Xpra, Xephyr and Xvfb is done outside firejail code, by setting symbolic links to firejail executable in /usr/local/bin. The information how to set it up is in the profile file for each of the servers.

Thanks again!

PR #1103, opened 2017-02-15
Update unbound profile to block 3D acceleration.

Ferroin:
There is no legitimate reason for a caching DNS resolver to need 3D acceleration.  Unbound adheres to this already, so any attempts to access GPU hardware from it are by definition either bugs or the result of an exploit, so let's just block access to the GPU.

Tested on my local systems with about 2 dozen different permutations of unbound configuration, double checked by looking through the source code for Unbound.


netblue30:
Merged, thanks.

PR #1106, opened 2017-02-17
Tighten keepassx

Fred-Barclay:
G'day @netblue30 . This tightens the keepassx profiles a little bit.

I was under the impression `net none` meant the netlink socket was closed, but `firejail --audit` says:
```MAYBE: I can connect to netlink socket. Network utilities such as iproute2 will work fine in the sandbox. You can use "--protocol" to disable the socket.```
So I've kept `protocol unix` in addition to adding `net none`. Is that correct?

Cheers!


netblue30:
all merged, thanks.

PR #1108, opened 2017-02-19
Thunar

Fred-Barclay:
Here's a profile for thunar. It's definitely a lot nicer of a file manager to sandbox than, say, nautilus. :smile: 

The copyright notice in README was also updated to 2017.


netblue30:
Merged, thanks.

PR #1118, opened 2017-02-28
fixing and tidying up Keepass(x) profiles

SYN-cook:
*.kdb and *.kdbx files were blacklisted in ${HOME} via diable-common.inc, but this was not disabled for Keepass and KeepassX.

Further, ${HOME}/keepassx.kdbx doesn't need to be blacklisted separately, as it is already contained in above mentioned blacklist.

### DIFF
--- a/etc/keepass.profile
+++ b/etc/keepass.profile
@@ -1,6 +1,7 @@
 # keepass password manager profile
-noblacklist ${HOME}/.config/keepass
 noblacklist ${HOME}/.keepass
+noblacklist ${HOME}/.config/keepass
+noblacklist ${HOME}/.config/KeePass
 noblacklist ${HOME}/*.kdbx
 noblacklist ${HOME}/*.kdb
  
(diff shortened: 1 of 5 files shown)


etc/keepass.profile
```diff
@@ -1,6 +1,9 @@
 # keepass password manager profile
-noblacklist ${HOME}/.config/keepass
 noblacklist ${HOME}/.keepass
+noblacklist ${HOME}/.config/keepass
+noblacklist ${HOME}/.config/KeePass
+noblacklist ${HOME}/*.kdbx
+noblacklist ${HOME}/*.kdb
```
Fred-Barclay: I don't think we should make .kdbx files available to keepass, or .kdb files available to keepassx.

netblue30:
merged, thanks.

etc/keepass.profile
SYN-cook: You are right, Keepass 1 can't open .kdbx files. From a security standpoint the way I suggested it is clearly not ideal. 
But my idea was that maybe not all distributors name their executables the way firejail names their keepass profiles, so I took here rather a functional standpoint.

Issue #1120, opened 2017-03-01
Keepass section in browser profiles


SYN-cook:
I was playing around with KeePass and browser extensions (KeeFox, PassIFox and ChromeIPass) today, and I found that the following section was obsolete in the Firefox and Chrome profiles:

`# lastpass, keepass`
`# for keepass we additionally need to whitelist our .kdbx password database`
`whitelist ~/.keepass`
`whitelist ~/.config/keepass`
`whitelist ~/.config/KeePass`

Everything worked nicely without... Since Keepass configuration files are blacklisted via disable-passwdmgr.inc, and blacklisting was never disabled, it seems the code in this section was anyways always broken (already before #993). So do I miss something here, or is this section safe to remove? 

We should probably remove these lines if they are not absolutely necessary (I could prepare a pull request. And I would also introduce `noblacklist ~/.lastpass` into the browser profiles, so that LastPass is finally able to write local backups from Firefox).

SYN-cook:
This seems to be related to #1095. My password database is 1) blacklisted and 2) not whitelisted for the browser, still I can use it with the extensions.

Apparently the extensions don't access the password database directly, but communicate with Keepass [via http](https://github.com/pfn/keepasshttp) or [via JSON-RPC](https://github.com/luckyrat/KeeFox/wiki/en-%7C-Technical-%7C-KeePassRPC-detail).

SYN-cook:
I have found only [one extension](https://chrome.google.com/webstore/detail/ckp-keepass-integration-f/lnfepbjehgokldcaljagbmchhnaaogpc) for Chrome that works differently, requires direct access to the password storage and has a notable userbase. 
If we unblacklist and whitelist $HOME/*.kdbx for browsers, it would be mostly for this extension.

netblue30:
Send a pull request, I'll merge it.

PR #1124, opened 2017-03-05
keepass browser integration, lastpass

SYN-cook:
Already outlined in #1120.

Further, according to the LastPass [documentation](https://lastpass.com/support.php?cmd=showfaq&id=425), only the extension for Firefox and its derivatives makes use of `~/.lastpass`... which is why I removed references to this folder from other profiles.

Except for LastPass in Firefox and Firefox derivatives, this is a pure tidy up mission.


SYN-cook:
In its current form, the CKP extension for Chrome is still not supported out of the box, which would require unblacklisting and whitelisting of $HOME/*.kdb(x). Convenience for eventual CKP users would be in conflict with the security interests of Keepass(X) users who don't employ browser extensions for their password management. It's the classic usability-security trade-off...

But if you wanna have this in, I'm happy to prepare a second pull request.

netblue30:
Merged. Let's leave CKP extension out for now. Thanks for the patch.

PR #1146, opened 2017-03-17
profile for engrampa

SYN-cook:
Forked from file-roller.profile. Tested and works well for me.

At first I had `no3d` in the profile, but then I was under the impression that `no3d` is already included in `nogroups` (membership in the video group should be a precondition for access to hardware acceleration), and so I removed it again.

Please correct as necessary.


netblue30:
All merged, thanks.

PR #1149, opened 2017-03-19
complete autostart blacklist for KDE

SYN-cook:
One more autostart folder used by Plasma ([ArchWiki](https://wiki.archlinux.org/index.php/KDE#Autostarting_applications))


netblue30:
Merged, thanks.

Issue #1150, opened 2017-03-19
terminix renamed to tilix


glitsj16:
Hi, just a FYI. Terminix is being renamed to Tilix (https://github.com/gnunn1/tilix). Adding 'blacklist ${PATH}/tilix' to /etc/firejail/disable-common.inc accordingly could help early adopters. No need to remove 'blacklist ${PATH}/terminix' just yet, the move could take a while to trickle-down in all distros.

Regards

netblue30:
Merged, thanks!

PR #1151, opened 2017-03-20
Handles #1150

Fred-Barclay:
Terminix is being renamed to tilix. This adds `${PATH}/tilix` to the blacklisted terminals in disable-common.inc without removing terminix (since there will still be users of terminix).
Thanks to @glitsj16 for pointing this out. :smile:


netblue30:
Merged, thanks.

PR #1152, opened 2017-03-21
blacklist X11 startup scripts

SYN-cook:
I tried to complete the list with some help from the [Debian manual](https://www.debian.org/doc/manuals/debian-reference/ch07.en.html#_starting_the_x_window_system). 

All scripts are now uniformly blacklisted. I guess having them all read-only would already be sufficient, but blacklisting doesn't do any harm, as these files are only run/accessed once in the X session startup process. EDIT: except potentially ~/.profile, which is left as read-only

Besides this I would like to suggest blacklisting a further location for systemd user units and just one more autostart folder (this time KDE4).

Finally there are some minor tidy up works in disable-programs.inc.

### DIFF
--- a/README
+++ b/README
@@ -99,6 +99,13 @@ Fred-Barclay (https://github.com/Fred-Barclay)
 	- compile/install scripts for --git-install/--git-uninstall commands
 	- tighten keepassx
 	- added Thunar profile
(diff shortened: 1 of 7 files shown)


netblue30:
All merged, thanks.

Issue #1117, opened 2017-02-28
--no-join (option to disallow firejail --join)


testbird:
For example, if networking tools (like khtml, wget, python, or even firefox) are run in a sandbox in order to only allow a limited set of tools access to the network (all other traffic blocked, only the sandbox with specific mac/ip has some allow rules), it is also important to prevent other applications from accessing the network by using --join and accessing the net thought the sandboxed tool.
--no-join ?

Starting a sandbox with an already reserved/existing mac/ip would also have to be blocked.
Specific mac/ip configured in a profile? and
--unique-net ?

netblue30:
I'll add --no-join. You'll be able to start a non-joinable sandbox.

The current code does an arp request before assigning the ip address. It will check if the address is in use. You can also specify a range of addresses for the new sandbox (--iprange). The mac address is assigned randomly by the kernel, unless it was specified by the user. Will this work?

testbird:
Thanks for explaining. I've tested creating a sandbox with an already existing mac address on the same machine and it also seems to fail properly: ioctl: network.c:217 net_if_up: Device or resource busy.

But the sandbox that is allowed to have network access may not always be running (and thus blocking the use of the addresses). This means blocking all network traffic for the host does not prevent anybody from gaining network access. Anyone/thing could start a firejail with the allowed network adresses.

I think my first idea, to have some configuration to reserve specific addresses for some profiles "unique or reserved-net" (so they can not be used with other configurations) may have been flawed.

Could it be a more general problem of the suid handling of firejail, that it effectively allows regular users to create new nodes on the network? May it be wise to only allow (non-local) --net configurations for root (or all users with preexisting permissions) and in root owned profiles? (Kind of executing part of firejail setup code under NO_NEW_PRIVS, exept when configured by root owned profiles?)

testbird:
It may not have been clear, when I wrote network access is blocked, I meant iptables on the host or the router. Which unprivileged users may circumvent by having firejail bring up an interface (with the address equal to a special profile that sets a single application up in a limited sandbox with an address that has network access granted).

netblue30:
> I meant iptables on the host or the router. 

There is no integration with the main netfilter setup running on the box.

testbird:
removed "dissallowing clone operations" from this issue (now #1123)

netblue30:
Ended up with a different implementation based on "join" flag in /etc/firejail/firejail.config.

[...]

PR #1155, opened 2017-03-23
enable/disable join support in /etc/firejail/firejail.config

startx2017:
I introduced a config param in /etc/fierjail/firejail.confg to disable --join command for non-root users.


netblue30:
Merged! I was planning to do it differently to fix https://github.com/netblue30/firejail/issues/1117, but this way is better. Thanks for the patch.

PR #1156, opened 2017-03-24
profile enhancements

SYN-cook:
- harden audacious profile
- added KFileDialog history to blacklist (~/.config/Audaciousrc, ~/.config/scribusrc)
- added pdf reader support to scribus profile (feature is not yet available in scribus stable though)
- `netlink` added to gwenview profile.
**EDIT/REPEATED UPDATE:** during startup there is a complaint `UdevQt: unable to create udev monitor connection`, which disappears when the netlink protocol is allowed. 
Google shows this string is in the [sources](https://github.com/KDE/solid/blob/master/src/solid/devices/backends/shared/udevqtclient.cpp) of [solid](https://techbase.kde.org/Development/Architecture/KDE4/Solid), and apparently it is printed when monitoring of the kernel uevent netlink socket fails. Btw, other programs that depend on the solid library and show exactly the same behavior include Dolphin and Amarok.
However, [here](https://cgit.kde.org/gwenview.git/commit/?id=8c6252178b6376d593f8645bae2e8156d2e009e3&context=1&ignorews=0&dt=0) it looks like solid is used only to display more precise device information and nicer icons in gwenviews import dialogue, which has been disabled anyways for gwenview5 (but will come back in the future). 
The sources also contain two device actions for automatically opening the import dialogue, e.g. when a camera is plugged in, but as far as I understand device actions hide all that netlink stuff from the target application. Guess we don't necessarily need netlink then :) I'll remove it again.
- `nosound` removes the sound from video previews in gwenview
- added gwenview KDE5 configuration files
- ~/.pki blacklisting moved from disable-programs.inc to disable-common.inc (it's the best place I found for it, although "top secret" doesn't really describe what it is)

### DIFF
--- a/etc/gwenview.profile
+++ b/etc/gwenview.profile
@@ -16,7 +16,7 @@ caps.drop all
 nogroups
 nonewprivs
 noroot
-protocol unix,netlink
+protocol unix
 seccomp
 tracelog
 
--- a/etc/gwenview.profile
+++ b/etc/gwenview.profile
@@ -5,6 +5,8 @@ include /etc/firejail/gwenview.local
 # KDE gwenview profile
 noblacklist ~/.kde/share/apps/gwenview
 noblacklist ~/.kde/share/config/gwenviewrc
+noblacklist ~/.config/gwenviewrc
+noblacklist ~/.config/org.kde.gwenviewrc
 include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-programs.inc
 include /etc/firejail/disable-devel.inc
@@ -14,13 +16,13 @@ caps.drop all
 nogroups
 nonewprivs
 noroot
-protocol unix
+protocol unix,netlink
 seccomp
-nosound
+tracelog
 
 private-dev
 
-#Experimental:
+# Experimental:
 #shell none
 #private-bin gwenview
 #private-etc X11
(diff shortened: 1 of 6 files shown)


etc/gwenview.profile
```diff
@@ -14,13 +16,13 @@ caps.drop all
 nogroups
 nonewprivs
 noroot
-protocol unix
+protocol unix,netlink
 seccomp
-nosound
+tracelog
```
SYN-cook: don't know though if tracelog is a sensible default for a program like gwenview. Please correct if you feel like.

SYN-cook:
@Fred-Barclay Sorry, I didn't mean to add commits after your review. i had in my mind to create and play with a new branch, but then I merged into my master branch by accident :stuck_out_tongue_winking_eye:

Fred-Barclay:
@SYN-cook No worries, mate.

SYN-cook:
netlink is taken out again (see first comment), sorry for confusing everyone

netblue30:
@Fred-Barclay - if the web tool allows you to merge it directly, instead of approving just go ahead and press the merge button! If there is something wrong, we can always fix it later.

Fred-Barclay:
@netblue30 will do, thanks! :smile:

PR #1159, opened 2017-03-25
Adds icedove directories in thunderbird profile

irregulator:
In debian stretch icedove is renamed to thunderbird. This happens
as of icedove version 1:45.7.1-1, see debian bug #816679 for
details.

Thunderbird debian package, as of stretch, contains a migration
script for user profiles. Namely /usr/bin/thunderbird is a wrapper
script, thunderbird-wrapper.sh. This scripts symlinks ~/.icedove
(if exists) to ~/.thunderbird thus ensuring Thunderbird will be
able to read old user's profiles.

That symlink breaks thunderbird when run with firejail since
firejail thunderbird.profile does not allow access to ~/.icedove.

This commit modifies accordingly the thunderbird.profile.

### DIFF
--- a/etc/thunderbird.profile
+++ b/etc/thunderbird.profile
@@ -14,6 +14,10 @@ noblacklist ~/.thunderbird
 mkdir ~/.thunderbird
 whitelist ~/.thunderbird
 
+noblacklist ~/.icedove
+mkdir ~/.icedove
+whitelist ~/.icedove
+
 # allow browsers
 ignore private-tmp
 include /etc/firejail/firefox.profile


irregulator:
I was pondering if that's the only to way to address this symlink issue. 

If you think it would be preferable to address this issue strictly in Debian (cause that's only where icedove users are found), I'd be happy to open a debian bug report and provide a patch there.

Let me know of your thoughts :)

netblue30:
Let's grab your version, icedove migration in Debian stretch is still under way. Maybe once stretch is released we'll revisit some of the profiles. Thanks for the patch.

PR #1163, opened 2017-03-27
blacklist more KDE files

SYN-cook:
More KDE files and folders I would suggest to blacklist.


Fred-Barclay:
Looks great - thanks!

Issue #1081, opened 2017-01-29
`firecfg --fix` does not ensure that `~/.local/share/applications` exists


ypid:
#### Actual behavior

```Shell
rm ~/.local/share/applications -rf && firecfg --fix
/home/user/.local/share/applications/python2.7.desktop skipped: No such file or directory
/home/user/.local/share/applications/python3.4.desktop skipped: No such file or directory
/home/user/.local/share/applications/cinnamon-settings-daemon.desktop skipped: No such file or directory
/usr/share/applications/cinnamon-polkit-gnome-authentication-agent-1.desktop - skipped, polkit-gnome-authentication-agent-1 not in PATH
/home/user/.local/share/applications/cinnamon2d.desktop skipped: No such file or directory
/home/user/.local/share/applications/cinnamon.desktop skipped: No such file or directory
/home/user/.local/share/applications/openjdk-7-policytool.desktop skipped: No such file or directory
/home/user/.local/share/applications/openjdk-8-policytool.desktop skipped: No such file or directory
/home/user/.local/share/applications/vlc.desktop skipped: No such file or directory
/home/user/.local/share/applications/chromium.desktop skipped: No such file or directory
```

#### Expected behavior

```Shell
mkdir -p ~/.local/share/applications && firecfg --fix
/home/user/.local/share/applications/python2.7.desktop created
/home/user/.local/share/applications/python3.4.desktop created
/home/user/.local/share/applications/cinnamon-settings-daemon.desktop created
/usr/share/applications/cinnamon-polkit-gnome-authentication-agent-1.desktop - skipped, polkit-gnome-authentication-agent-1 not in PATH
/home/user/.local/share/applications/cinnamon2d.desktop created
/home/user/.local/share/applications/cinnamon.desktop created
/home/user/.local/share/applications/openjdk-7-policytool.desktop created
/home/user/.local/share/applications/openjdk-8-policytool.desktop created
/home/user/.local/share/applications/vlc.desktop created
/home/user/.local/share/applications/chromium.desktop created
```

#### Version

```
firejail version 0.9.44.4

Compile time support:
    - AppArmor support is enabled
    - AppImage support is enabled
    - bind support is enabled
    - chroot support is enabled
    - file and directory whitelisting support is enabled
    - file transfer support is enabled
    - networking support is enabled
    - overlayfs support is enabled
    - private-home support is enabled
    - seccomp-bpf support is enabled
    - user namespace support is enabled
    - X11 sandboxing support is enabled
```

[...]

PR #1164, opened 2017-03-27
firecfg: create ~/.local/share/applications directory if it doesn't exist

startx2017:
Fix for bug #1081


netblue30:
Merged, thanks.

PR #1166, opened 2017-03-28
blacklist more KDE files

SYN-cook:
more KDE....

I removed the comment again that was introduced with my last commit, cause it doesn't tell the whole truth and my impression is we don't want to have long stories in profile files

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -15,7 +15,7 @@ read-only ${HOME}/.local/share/applications
 # X11 session autostart
 blacklist ${HOME}/.xinitrc
 blacklist ${HOME}/.xserverrc
-blacklist /etc/X11/Xsession.d/
+blacklist /etc/X11/Xsession.d
 blacklist ${HOME}/.Xsession
 blacklist ${HOME}/.xsession
 blacklist ${HOME}/.xsessionrc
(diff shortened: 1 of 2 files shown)


Fred-Barclay:
G'day @SYN-cook 
IMHO these look like things we might want to keep. I don't use kde though, so I don't know as much about this as you. :smile: Can you tell me why we need to remove these blacklists?

SYN-cook:
Hey @Fred-Barclay!
1. Create a custom profile in konsole and set it as default, which is what most users will do. Then go to the konsole folder that should be blacklistet for your system and insert a line `Command=/bin/bash /tmp/script.sh` into the default profile
2. Let's go to /usr/share/kde4/services/ServiceMenus and copy one of the files to the service folder that should be blacklistet for your system. It will show up as action in the Konqueror context menu (Dolphin should work also but somehow doesn't for me). Modify to `Exec=sh /tmp/script.sh`

Voilà. Sandbox escaped (with full user permissions)

SYN-cook:
And if that is an attack vector, Caja's configuration should be blacklistet, too (as was the one of Nautilus)

Fred-Barclay:
I'm not sure how, but I completely misread your first commit. I was thinking you were removing files from the blacklist.
This looks great. Merged. :+1:

Issue #1171, opened 2017-03-28
leftovers from 'tmpfs on ~/.cache' in /etc/firejail/firejail.config


glitsj16:
Hi, I noticed some recent commits removed the formerly introduced tmpfs on ~/.cache behavior. If this is indeed no longer in the code, there might be another leftover in firejail.config. Latest git checkout still references..

```
# Enable mounting a tmpfs on top of ~/.cache directory, default enabled.
# cache-tmpfs yes
```
I had to change the default to 'no' in relation to https://github.com/netblue30/firejail/issues/1141. Am I correctly assuming the tmpfs on ~/.cache feature is indeed removed from firejail?

Regards.

netblue30:
Fixed, thanks.

Issue #1170, opened 2017-03-28
removed man firejail-config still referenced in firejail.config


glitsj16:
Hi, https://github.com/netblue30/firejail/commit/e470524e20a5bf7c5907306b76d41adcb28ac490 removed man firejail-config. It is still referenced in /etc/firejail/firejail.config though. I'm throwing it in just in case it slipped under your radar.

Regards.

netblue30:
Fixed, thanks.

PR #1174, opened 2017-03-29
firejail.config cleanup

startx2017:
Fixing https://github.com/netblue30/firejail/issues/1171 and https://github.com/netblue30/firejail/issues/1170


netblue30:
Merged, thanks.

Issue #578, opened 2016-06-17
Thought: blacklist all .files and folders by default.


Fred-Barclay:
We already blacklist most hidden directories and files with disable-common, disable-passwdmgr, and especially disable-programs. Then we noblacklist the files in the appropriate profiles. 

This has the weakness of requiring multiple disable.inc files, which can lead to unnecessary duplication, complication, and confusion. It also opens a bit of a security risk in that hidden files that **aren't** blacklisted are visible to the sandbox. While currently firejail blacklists all the sensitive hidden files, sooner or later there will be one that isn't included in the sandbox that should have been.

Why not just disable them all in a single file? Something like _disable-hidden.inc_: `blacklist ${HOME}/.*`.
Then we can just add the necessary noblacklists to the profiles. We do this already since nearly all hidden files are blacklisted already, so most profiles would be ready-to-go as they are right now. A few profiles will have to be modified (esp. if they need .bashrc which is currently visible to the sandbox) but there is plenty of time for debugging before the next firejail release (right, @netblue30 ? :smile:)

And on the positive side, a bit of testing in terminal has shown no problems on my side. 

Thoughts, mates? My last big change wasn't welcomed by some so I thought I'd get responses before creating a pull request. A rough mental draft is ready to go depending on the feedback.

chiraag-nataraj:
Sounds like an improvement in security! :smile:

Fred-Barclay:
That's what I'm hoping. :)

Fred-Barclay:
Will probably conflict with #579

netblue30:
You guys are worse than a six year old :)

@Fred-Barclay: you are right, all the include system is a mess in this moment. What you propose is also somehow releated to https://github.com/netblue30/firejail/issues/504. Let's leave it open for now, I'll mark it as an enhancement and go for it in the release after the current one.

The thing is I've already started  adding private-bin and whitelisting /tmp directory all over the place. Also private-dev and private-etc are in plan. By the time I'm done, everything will be broken. So no messing around with the include files, instead just put in some pull requests with whatever you have in your private profiles. If is working on your setup, it should be fine, and we'll fix it when people start complaining.

Fred-Barclay:
"You guys are worse than a six year old :)"
How did you guess my age? :wink: I'm tellin' my mommy!

Will do. Will the next release be 0.9.41 or 0.9.42? I noticed you seem to use even-numbered point releases.

netblue30:
[...]

PR #1181, opened 2017-03-31
restrict more KDE files

SYN-cook:
some more KDE related configuration files. kdeglobals was read-only already in the Okular profile, but it makes a lot of sense to have it write protected for all Firejail profiles

So, I guess for this moment that was all about KDE :)


Fred-Barclay:
Merged, thanks!

I wonder if it would be a bad idea to just blacklist .kde/share?

SYN-cook:
I guess that would be worth a trial. One could actually try the same with ~/.config.

Fred-Barclay:
@SYN-cook... I suggested something similar in #578   Sooner or later I'm going to try it. :smile:

PR #1182, opened 2017-03-31
tidy up

SYN-cook:
1. `noblacklist ~/.lastpass` is redundant code, because there is no `include /etc/firejail/disable-passwdmgr.inc` in these profiles...... and I myself introduced it! I can't recall the details of how this happened, but please let's get rid of it.
2. minor reorganization in disable-common.inc


Fred-Barclay:
Great catch! That was redundant. :)

PR #1184, opened 2017-03-31
add new syscalls in default seccomp filter

startx2017:
Added the following: bpf, clock_settime, personality, process_vm_writev, query_module,  settimeofday, stime, umount, userfaultfd, ustat, vm86, vm86old


netblue30:
It passed all my automated testing. We'll keep an eye on it, and fix it as we go. Thanks!

PR #1186, opened 2017-03-31
Add Go, Rust, and OpenSSL to disable-devel.conf

SpotComms:


Fred-Barclay:
Looks good. :+1: 
I'm a little concerned that blacklisting openSSL could mess with the browsers a bit, but hopefully not. Anyways, we can easily fix it if there's an issue.

PR #1188, opened 2017-03-31
Allow fish_config

isalgueiro:
[fish_config](https://fishshell.com/docs/current/commands.html#fish_config) command launches the default web browser pointing to `~/.cache/fish/web_config-<six random chars>.html`. This allows to edit fish shell configuration with sandboxed firefox.


Fred-Barclay:
Hi @isalgueiro and thanks for the pull request!

I would personally be uncomfortable with this being added, since I'm not at all sure that Firefox, or any other browser, should be allowed to edit shell settings and configurations from within a sandbox. Let's see what @netblue30 thinks. 

Cheers!
Fred

PR #1195, opened 2017-04-05
blacklist attic and borg

SYN-cook:


Fred-Barclay:
Looks good, thanks! :+1:

Issue #421, opened 2016-04-09
okular and gwenview profiles


curiosity-seeker:
I suggest to add profiles for okular and gwenview. The following settings work for me:

```
include /etc/firejail/disable-common.inc
include /etc/firejail/disable-programs.inc
include /etc/firejail/disable-devel.inc
include /etc/firejail/disable-passwdmgr.inc
caps.drop all
seccomp
net none
noroot
shell none
```
I've chosen _net none_ as reportedly ransomware has also been found in pdf and image files. I think it's worth consideration to also adjust the evince profile accordingly.

avoidr:
Personally I think that global profiles that are distributed to everyone should not restrict legitimate uses. Allow whatever the program wants, and no more. It's better that each user tightens his profile to his own needs, than to cut functionality and have users bitch at you.
You can put your own profiles in `~/.config/firejail/` and include the global profile. That's how I do it, and I agree with you that some programs have no business accessing any network.

As for the profile, I would add a `noblacklist` at the top, with the program's config dir.
You could try `blacklist`ing `/etc` and `private-etc` any files that it needs.
Try `private-bin` with only the program's binary allowed.

curiosity-seeker:
>Personally I think that global profiles that are distributed to everyone should not restrict legitimate uses. Allow whatever the program wants, and no more. 

Well, I haven't seen any negative side-effects for both prgrams by doing this.

>You can put your own profiles in ~/.config/firejail/ and include the global profile.

I know, and it's something I'm doing, too.

>As for the profile, I would add a noblacklist at the top, with the program's config dir.

Why should I do this if that config dir is not blacklisted by one of the .inc files? It would be a different situation if a whitelist approach were applied - but as long as whitelist globbing (#216) is not yet available, it's difficult to implement this for those 2 programs.

>You could try blacklisting /etc and private-etc any files that it needs.
Try private-bin with only the program's binary allowed.

Yes, that's something I could try.

>Also protocol.

Not necessary with net none.

avoidr:
>Well, I haven't seen any negative side-effects for both programs by doing this.

You didn't, but maybe someone else would. What does the program need network access for?

>>As for the profile, I would add a noblacklist at the top, with the program's config dir.

>Why should I do this if that config dir is not blacklisted by one of the .inc files?

But it is. As far as I can tell, whenever a new profile is added, its config dir is also added to disable-programs.inc, hence having to noblacklist before.

>>Also protocol.

>Not necessary with net none.

[...]

PR #1198, opened 2017-04-06
Okular and Gwenview profiles, Baloo blacklist

SYN-cook:
- Gwenview: This patch is necessary for the KDE4 build, the KF5 build doesn't need any of it.
- Okular: It was already described in #421, and I can confirm that printing from okular is not possible without `private-bin lpr`

### DIFF
--- a/etc/disable-programs.inc
+++ b/etc/disable-programs.inc
@@ -96,6 +96,8 @@ blacklist ${HOME}/.config/mupen64plus
 blacklist ${HOME}/.config/nautilus
 blacklist ${HOME}/.config/nemo
 blacklist ${HOME}/.config/netsurf
+blacklist ${HOME}/.config/okularpartrc
+blacklist ${HOME}/.config/okularrc
 blacklist ${HOME}/.config/opera
 blacklist ${HOME}/.config/opera-beta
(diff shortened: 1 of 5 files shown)


Fred-Barclay:
Thanks @SYN-cook .
I wonder if we should uncomment the updated private-bin lines?

SYN-cook:
For Gwenview I personally don't see a problem with uncommenting shell none and private-bin, actually. It is a bit speculative, but if there is a new release of Firejail this year and Gimp 2.10 is not released yet (it has been announced for 2017), one could maybe consider appending the private-bin line already in advance.

Uncommenting private-bin in the okular profile is also perfectly fine for "normal" use, from my experience. The only thing that probably wouldn't work is editors (you can see in the settings that there is some kind of support in okular). I've looked it up and the editor support is only for LATeX, which I never use. 

Btw, the current profile also works with new okular version 1.x, I just tried it out. Only the blacklist/noblacklisting needs another little update...

SYN-cook:
So let's uncomment private-bin in the gwenview profile?

Fred-Barclay:
> So let's uncomment private-bin in the gwenview profile?
Yeah, let's try it. :smile: 

Thanks for the PR!

PR #1201, opened 2017-04-08
new baloo profile

SYN-cook:
seccomp needs to stay disabled, because baloo makes ioprio_set system calls, which are on the seccomp blacklist.

Tested both on KDE4 and KDE5.

### DIFF
--- a/etc/baloo_file.profile
+++ b/etc/baloo_file.profile
@@ -20,9 +20,13 @@ nonewprivs
 noroot
 nosound
 protocol unix
-# Baloo makes ioprio_set system calls, which are blacklisted by default. 
+# Baloo makes ioprio_set system calls, which are blacklisted by default.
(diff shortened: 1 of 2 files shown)


Fred-Barclay:
@netblue30 is there a way to keep both the default seccomp filter and add ioprio_set to the allowed list of syscalls? I see `seccomp.keep` but if I'm not mistaken it only whitelists the provided syscalls and doesn't kep the default filter.

@SYN-cook Thanks! I'd like to see if we can keep a seccomp filter around. Let's see what netblue30 says.

SYN-cook:
Since we are dealing with a daemon we can also isolate it from X11

netblue30:
All in. I re-enabled seccomp in the profile file - it is a huge list without ioprio_set.

Issue #772, opened 2016-09-10
--enable-apparmor option is lost with make deb


nick75e:
Hi!

Using `configure --prefix=/usr --enable-apparmor` shows this:
```
[...]
checking sys/apparmor.h usability... yes
checking sys/apparmor.h presence... yes
checking for sys/apparmor.h... yes
[...]
Configuration options:
   apparmor: -DHAVE_APPARMOR
```

But when I type `make deb` I get:
```
[...]
Configuration options:
   apparmor: 
```

The apparmor option is discarded at some point. _Everything is okay when using `make`_
Please fix.

netblue30:
I'll have to implement it. For now, change "./configure --prefix=/usr" into "./configure --prefix=/usr --enable-apparmor" in mkdeb.sh.

netblue30:
Fixed in git.

netblue30:
I had to roll it back. Use the change in mkdeb.sh.

laniakea64:
Was going to file a bug about this when I couldn't disable git-install in my .deb build of firejail.  After reading this issue, I wrote a Python 3 script to automate setting custom configure options in mkdeb.sh and then building.

Would it be useful for either of you?

netblue30:
Sure, send a pull request. Put the file in https://github.com/netblue30/firejail/tree/master/contrib, I'll modify the install script to have it installed automatically.

PR #1205, opened 2017-04-09
Add a script to build a .deb with custom configure options

laniakea64:
See https://github.com/netblue30/firejail/issues/772


netblue30:
All merged, thanks.

netblue30:
Install support is in!

laniakea64:
> Install support is in!

Hmm, it installs to `/usr/lib/firejail`.  Currently the script depends on being located in an immediate subfolder inside the firejail source tree.

I'll fix `fj-mkdeb.py` so that it doesn't really matter where it's located.

PR #1207, opened 2017-04-09
Fix fj-mkdeb.py not functional when installed

laniakea64:
This takes care of the problem mentioned in https://github.com/netblue30/firejail/pull/1205#issuecomment-292813894


Fred-Barclay:
Looks good - thanks! :+1:

PR #1208, opened 2017-04-10
fix baloo_file.profile (x11 isolation)

SYN-cook:
The last one of my commits from yesterday introduced X11 isolation but also a bug into the profile. Blacklisting /tmp/.X11-unix causes Baloo to hang after a while. I don't know why Baloo cares about X11, but we need to either remove this line or replace it with `x11 xorg`, which works for me.


Fred-Barclay:
Merged, thanks! :+1:

PR #1214, opened 2017-04-13
make Baloo experiment more meaningful

SYN-cook:
Baloo will fail to update its configuration file, unless the whole config folder is writable (it seems it wants to create a lock file). But if the whole config folder writable, having ${HOME} read-only doesn't make so much sense any more.

So I propose to remove not necessary lines and to add a note.


Fred-Barclay:
Merged, thanks! 👍

PR #1219, opened 2017-04-15
Add a profile for Dino

SpotComms:
A profile for the new Dino XMPP client https://github.com/dino/dino

### DIFF
--- a/etc/dino.profile
+++ b/etc/dino.profile
@@ -1,0 +1,35 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+include /etc/firejail/dino.local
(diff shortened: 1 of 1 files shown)


Fred-Barclay:
Looks good!
Do you want to see if we can get private-etc working? If so, try this: `private-etc group,resolv.conf,gtk-2.0,fonts`

SpotComms:
That does not fix the network issue, and it also breaks the gtk theme.

Fred-Barclay:
Okay, thanks for trying! Let's go ahead and get this merged and worry about adding private-etc later.
Cheers!

PR #946, opened 2016-11-29
Added 10 new profiles

SpotComms:
Added profiles for Bless, Gnome 2048, Gnome Calculator, Gnome Contacts, JD-GUI, Lollypop, MultiMC5, PDFSam, Pithos, and Xonotic
From https://github.com/SpotComms/FirejailProfileGenerator

PR #1220, opened 2017-04-15
Harden some profiles

SpotComms:
Hardens some profiles I added in #946 and some others.

Can we also get a per profile disable-mnt? Also what exactly are the side-effects of 'ipc-namespace'?

### DIFF
--- a/etc/bless.profile
+++ b/etc/bless.profile
@@ -17,6 +17,7 @@ include /etc/firejail/disable-devel.inc
 
 #Options
 caps.drop all
+net none
 netfilter
 nogroups
 nonewprivs
--- a/etc/bless.profile
+++ b/etc/bless.profile
@@ -18,7 +18,19 @@ include /etc/firejail/disable-devel.inc
 #Options
 caps.drop all
 netfilter
+nogroups
 nonewprivs
 noroot
-protocol unix,inet,inet6
+protocol unix           
 seccomp
+shell none
+
+private-dev
+private-etc fonts,mono
+private-tmp
+
+noexec ${HOME}
+noexec /tmp
+
+no3d
+nosound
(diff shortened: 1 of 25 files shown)


Fred-Barclay:
Hi @SpotComms 
Thanks for the pull request!

Are you aware that commenting out the protocol and seccomp lines actually weakens the sandbox? 😉

SpotComms:
Yes, those Java apps (JD-GUI, PDFSam, MultiMC5) don't appear to start anymore (they used to) with that, I couldn't figure out why. And seccomp is breaking MultiMC5 for some reason.

Fred-Barclay:
I just read through everything and it all looks really good! But just to check before it's merged, does it all work or are you still having some issues with certain computers?

SpotComms:
@Fred-Barclay so all of the profiles except for the 3 that use Java work. The 3 that use Java only seem to work on my desktop when using Wayland and not when using Xorg, which is odd because they work fine on other computers. But nothing that is broken was added in this commit, as in those protocol lines have been in master for a while now.

Fred-Barclay:
If you're having trouble with Wayland I wonder if `no3d` is the cause?

SpotComms:
@Fred-Barclay no thats the thing, its the exact opposite. The programs *are* working under wayland and not under xorg. And I've already tried the profiles without any of those changes, *and* they work just fine on other computers with wayland or xorg.

Fred-Barclay:
Oh, okay. That's weird.
One last question - did the program work on that computer with the old profiles? I think you said so but I want to be double-sure. 😄

SpotComms:
@Fred-Barclay to confirm, the 3 programs did work on the computer in question as of a ~month ago, and they've worked and are still working on others.

Fred-Barclay:
Great - let's merge 'em in.
Thanks!

PR #1221, opened 2017-04-15
Add a profile for Kodi

SpotComms:


Fred-Barclay:
Nice @SpotComms! 👍 
I wonder if we could use `tracelog` as a security filter?

Fred-Barclay:
Merged - thanks!

PR #1222, opened 2017-04-16
Add a profile for meld

SpotComms:


Fred-Barclay:
Hello again @SpotComms :)
I don't think we need the mkdir line (unless you're planning on adding a corresponding whitelist line, which probably wouldn't make sense for meld.)

BTW: meld looks pretty useful! How have I not heard of it 'till now?

SpotComms:
Yup, fixed.

And its 3 way diff is pretty nice. The directory diff is a bit janky sometimes though.

Fred-Barclay:
Awesome, thanks! 👍

PR #1232, opened 2017-04-17
Harden more profiles

SpotComms:
Hardens Chromium and Polari, and adds 'ipc-namespace' to all the others.


netblue30:
merged, thanks.

PR #1242, opened 2017-04-21
Add noexec ~/.local/share to all profiles using noexec ~

### LINKED ISSUE #1238: default bind mounts and noexec option
`noexec /tmp` and `noexec ${HOME}` don't necessarily block execution completely, I see few folders still allowing it:
- ~/.config/pulse
- /tmp/.X11-unix
- ~/.local/share

Since noexec options are being used in profiles, it would be nice to close loopholes for the standard configuration.

~/.local/share can be easily handled in disable-common.inc. And the .X11-unix and pulse folders should be also safe for remounting with a noexec flag.

SpotComms:
An alternative to #1239 to fix #1238

### DIFF
--- a/etc/arduino.profile
+++ b/etc/arduino.profile
@@ -26,4 +26,5 @@ shell none
 private-tmp
 
 noexec ${HOME}
+noexec ${HOME}/.local/share
 noexec /tmp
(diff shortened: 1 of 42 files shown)


netblue30:
I grabbed the version from @SYN-cook and made it global, thanks anyway!

SpotComms:
@netblue30 #1239 does break Steam and Gajim(?).

netblue30:
Keep an eye on it in case wee need to fix it. Thanks.

PR #1252, opened 2017-04-28
update k3b profile

SYN-cook:
- update noblacklist
- remove commented private-dev (it will always break k3b)
- remove nogroups (it should break k3b in some cases)


netblue30:
All merged, thanks.

PR #1253, opened 2017-04-28
Add overlay configuration to profiles

nl5887:
This commit will add the overlay configurations overlay, overlay-named and overlay-tmp to profile configuration files. 

Adding these configurations will allow profiles to be started by default within overlays.


netblue30:
Merged, thanks.

PR #1254, opened 2017-04-28
Prevent running shells recursively

nl5887:
When running a login shell, it occurred to me that the shell was being opened recursively. This commit will prevent recursive login shells.


netblue30:
Merged, thanks.

PR #1260, opened 2017-05-02
blacklist file-manager python scripts

SYN-cook:
and
- permit Nautilus to access Trash
- remove `noblacklist ~/.local/share/caja` from profile, as folder is not used by Caja


Fred-Barclay:
Merged, thanks!

Issue #1261, opened 2017-05-03
File Manager Sandboxing


SpotComms:
> Symlinks for the common file managers are installed in /usr/local/bin by firecfg. File managers are usually started by default at login time, and will be sandboxed. Clicking on a file in the file manager will start the corresponding program in the same sandbox as the file manager. For example, clicking on a video file will start a sandboxed VLC running the video. 

Does that not defeat the purpose of Firejail? Any sandboxed program spawned under an already existing sandbox will not have its profile set, correct? So then won't that give programs more permissions then they'd normally have? And at the same time falsely leading users to believe that their apps are sandboxed?

For example gedit has 'net none', but if you open up a txt file from within Thunar, it'll end up with networking access. This issue becomes even worse when you end up spawning an application with a whitelist only profile which then ends up with full access to your files. Demonstration here: https://gist.github.com/SpotComms/55c8372cef381439cd9a90476e175f93

This point was clearly brought up in #1178 so I don't understand how this happened.

cryzed:
That is correct, see also #1256, the last 2 comments especially. To quote @netblue30:

> Unfortunately that's not possible, the first sandbox will not allow the second sandbox to rise privileges and set up the profile.

If you are really motivated there is a way to work around it: Have one "firejail-launcher" process listening on a socket unsandboxed for incoming requests. Then from within the Sandbox that you want to start, you can configure the opener (xdg-open and variants) using desktop files or using your file manager's file association settings, to instead open the text file using something like: `firejail-launcher run gedit`. When you then attempt to edit the text file, the firejail-launcher process instead contacts the instance running outside the sandbox and launches gedit with the correct profile and arguments.

I did something like this, but I haven't used it much since, see [here](https://gist.github.com/cryzed/1fff0a41309f4555efbb48cf28cd7b65#file-glassbox-py-L356). So this Python script has a daemon command (starts the firejail-launcher outside the sandbox) and a run command, which is used from within the sandbox. To contact the daemon, XMLRPC is used. There's also a whitelist feature, so not every process in a firejail sandbox can launch every other process using the launcher -- however this is still incomplete: there's currently no easy way to create specific whitelists for specific sandboxes.

[...]

PR #1263, opened 2017-05-03
Disable file managers from firecfg by default

SpotComms:
See #1261


cryzed:
You missed KDE's dolphin.

SpotComms:
@cryzed Thanks, fixed.

PR #1265, opened 2017-05-04
blacklist keepass plugins

SYN-cook:
write access to keepass plugins is a security risk, and on the keepass homepage stated as such


Fred-Barclay:
Looks good, thanks!

PR #1266, opened 2017-05-05
harden baloo_file

SYN-cook:
but loosen write-protections in commented section. Stricter settings for ~/.local/share won't work, due to write-locks set by KDE


netblue30:
Merged, thanks!

PR #1270, opened 2017-05-11
completing noexec

### LINKED ISSUE #1238: default bind mounts and noexec option

SYN-cook:
- fixes #1238 
- little layout improvement (I think) for disable-common.inc

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -292,6 +292,6 @@ blacklist ${PATH}/urxvtcd
 blacklist /vmlinuz*
 blacklist /initrd*
 
-# completing noexec ${HOME} and noexec /tmp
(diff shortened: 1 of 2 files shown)


netblue30:
Thanks for the patch. I'll do some testing and merge it next week.

SYN-cook:
No worries! Take your time

netblue30:
merged, thanks.

PR #1278, opened 2017-05-17
make seccomp optional again

Duncaen:
This patch allows to build firejail again for architectures that don't support seccomp yet.


netblue30:
Merged, thanks.

PR #1284, opened 2017-05-18
allow python and python3 for youtube-dl in MPV profile

pigmonkey:
This is required for MPV to successfully use youtube-dl on Arch. I'm not sure if other distros still ship youtube-dl with python2, or if python2.7 can be removed.


Fred-Barclay:
Merged, thanks!

PR #1315, opened 2017-05-26
pulseaudio with whitelisting

### LINKED ISSUE #1312: No sound in Firefox
Sound in Firefox worked fine until recently (basic command `firejail firefox`, Firejail version from Git repository, Ubuntu 16.04 LTS).

I have already tried suggested workarounds with no success:
https://firejail.wordpress.com/support/known-problems/#pulseaudio

When I use `--noprofile` argument (`firejail --noprofile firefox`) sound works. Maybe Firefox profile needs some adjustments.

SYN-cook:
fixes #1312 (and belongs to #1270)

### DIFF
--- a/etc/whitelist-common.inc
+++ b/etc/whitelist-common.inc
@@ -10,6 +10,7 @@ whitelist ~/.local/share/icons
 whitelist ~/.config/user-dirs.dirs
 read-only ~/.config/user-dirs.dirs
 whitelist ~/.asoundrc
+whitelist ~/.config/pulse/client.conf
 whitelist ~/.config/Trolltech.conf
 
 # fonts


netblue30:
#1312 was introduced in #1270 with the switch of whitelisting and sound code blocks in src/firejail/sandbox.c. 

This is how is working: we create a brand new ~/.config/pulse/client.conf after whitelisting/blacklisting in order to fix an ugly pulseaudio bug. Even if the user whitelists the file, it will be replaced later by firejail.

One thing we cannot do is let the user use whatever client.conf he has in his home directory. Basically all distros in this moment have a broken pulseaudio with a wrong config file. The last distro I've seen with a pulseaudio working out of the box was Ubuntu 14.04. After that it went downhill.

PR #1317, opened 2017-05-28
Fix lintian warning package-contains-timestamped-gzip

laniakea64:


reinerh:
Thanks!

laniakea64:
np.

Can/should this go in the 0.9.46-bugfixes branch as well?

reinerh:
I don't think that's necessary. It's not a bug that would affect any user.

PR #1357, opened 2017-06-29
Add profile for Liferea

BafDyce:
Note that profile comes with a small problem:
Liferea allows to import and export the current feed list.
However, I don't know which directory I should allow for this?
Maybe ~/Downloads?

With the current profile, the export/import feature does NOT work.


netblue30:
Merged, thanks.

PR #1358, opened 2017-06-29
fix empty-string assignment

jmullee:
was causing a SEGV when processing nonexistent directory


netblue30:
Merged, thanks for the fix.

PR #1359, opened 2017-06-30
Fix race condition when setting up /run/firejail files (#1013)

### LINKED ISSUE #1013: "Error mkdir" appears inconsistently
I regularly start two programs with firejail in my xinitrc.  Rarely, but without any pattern I can find, one or both of them will fail, printing this to stderr:
```
Error mkdir: fs.c:38 create_dir_as_root: File exists
```
The commands in the xinitrc file were originally `firejail syncthing-gtk --minimized &` and `firejail redshift-gtk -l [location] &`.  When a failure occurs, rerunning those commands will reliable start the program in a jail.  In order to capture this error message, I have changed the commands in the xinitrc to:
```
sh -c "firejail --debug syncthing-gtk --minimized > ~/sync_stdout 2>~/sync_stderr" &
sh -c "firejail --debug redshift-gtk -l [location] > ~/red_stdout 2>~/red_stderr" &
```
When this error is hit, nothing else appears on stdout or stderr.  Both of these commands are using custom profiles, but it doesn't look like firejail is even getting far enough to read them.

BafDyce:
This attempts to fix #1013 

Which is caused by a race condition in `create_empty_dir_as_root` between `stat` and `mkdir`.

Note 1: I'm not quite sure what the comment `/* coverity[toctou] */` does. Is it necessary that it stays in the line right above to the if statement? If yes, please tell me, then I'll reorder the comments.

Note 2: There are a bunch of calls to `mkdir` throughout the code, some of them behind a similar `stat`-gate. Should I also add a similar change to these?

Note 3: For testing, I forced race conditions by inserting a sleep right after the call to `stat`

### DIFF
--- a/src/firejail/util.c
+++ b/src/firejail/util.c
@@ -821,7 +821,9 @@ void create_empty_dir_as_root(const char *dir, mode_t mode) {
 		if (arg_debug)
 			printf("Creating empty %s directory\n", dir);
 		/* coverity[toctou] */
-		if (mkdir(dir, mode) == -1)
+		// don't fail if directory already exists. This can be the case in a race
+		// condition, when two jails launch at the same time. See #1013
+		if (mkdir(dir, mode) == -1 && errno != EEXIST)
 			errExit("mkdir");
 		if (set_perms(dir, 0, 0, mode))
 			errExit("set_perms");


netblue30:
Thanks for the fix, it is quite cool!

coverity[toctou] is because of the stat call before mkdir. I'll trough the code for more cases as you suggested.

I used to test it with a simple script:
`````
#!/bin/bash
sudo rm -fr /run/firejail
firejail sleep 10 &
firejail sleep 10 &
firejail sleep 10
`````
Then, in a different terminal I would run "firejail --list" to see how many of them are running.

PR #1360, opened 2017-06-30
Allow env for youtube-dl in mpv profile

announ:
youtube-dl uses `/usr/bin/env` as its interpreter.
If `env` is not available, mpvʼs execution of youtube-dl fails.


Fred-Barclay:
Merged, thanks!

PR #1363, opened 2017-07-04
Allow ~/.netrc for youtube-dl

announ:
youtube-dl can use `.netrc` for site authentication ([`netrc` option](https://github.com/rg3/youtube-dl/blob/master/README.md#authentication-with-netrc-file)).


netblue30:
Merged, thanks.

PR #1365, opened 2017-07-04
Harden 50 profiles

SpotComms:
Hardened many profiles using disable-mnt and novideo
Fixed gnome-font-viewer

Currently only somewhat tested, please review


netblue30:
All merged, thanks.

PR #1367, opened 2017-07-05
Harden profiles

SpotComms:
- Added 'disable-devel.conf' to many profiles
- Added 'disable-mnt' to many profiles
- Added 'noexec' to many profiles
- Added 'memory-deny-write-execute' to some profiles
- Removed 'netfilter' and 'net none' from profiles with 'protocol unix'
- Cleaned up profiles using defaults
- Partially synchronized Chromium-based profiles

**Only somewhat tested**. I've looked over it a few times, but would like some more eyes to go over it. Looking back I should've broken it up, but it got out of hand pretty quickly.

Commands used
```
List all profiles with incorrect .local: grep -l "default.local" *.profile
List all profiles with defaults: grep -l "depending on your usage" *.profile
List all 'disable-mnt' candidates: grep -L "disable-mnt" $(grep -l "whitelist-common" *.profile)
List all profiles with 'net none' and 'netfilter': grep -l netfilter $(grep -l "net none" *.profile)
List all profiles with 'disable-*.inc' but without 'disable-devel.inc': grep -L "disable-devel" $(grep -l "disable" *.profile)
List all profiles with 'net none' and 'protocol unix': grep -l "net none" $(grep -l "protocol unix" $(grep -L "protocol unix,inet" *.profile))
```

Also I'm not going to make a pull request just yet, but review on https://github.com/SpotComms/firejail/tree/pr/etc would be nice.

### DIFF
--- a/etc/0ad.profile
+++ b/etc/0ad.profile
@@ -38,3 +38,6 @@ tracelog
 private-dev
 private-tmp
 disable-mnt
+
+noexec ${HOME}
+noexec /tmp
--- a/etc/2048-qt.profile
+++ b/etc/2048-qt.profile
@@ -15,9 +15,11 @@ include /etc/firejail/disable-passwdmgr.inc
 
 caps.drop all
 #ipc-namespace
+netfilter
 nogroups
 nonewprivs
 noroot
+nosound
 novideo
 protocol unix
 seccomp
--- a/etc/2048-qt.profile
+++ b/etc/2048-qt.profile
@@ -7,24 +7,25 @@ include /etc/firejail/2048-qt.local
 
 noblacklist ~/.config/xiaoyong
 noblacklist ~/.config/2048-qt
+
 include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-programs.inc
 include /etc/firejail/disable-passwdmgr.inc
 
 caps.drop all
-netfilter
+#ipc-namespace
+nogroups
 nonewprivs
 noroot
-protocol unix,inet,inet6
+novideo
+protocol unix
 seccomp
-
-#
-# depending on your usage, you can enable some of the commands below:
-#
-nogroups
 shell none
-# private-bin program
-# private-etc none
-# private-dev
-# private-tmp
-nosound
+
+private-dev
+private-tmp
+disable-mnt
+
+noexec ${HOME}
+noexec /tmp
--- a/etc/Thunar.profile
+++ b/etc/Thunar.profile
@@ -16,7 +16,9 @@ include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-passwdmgr.inc
 
 caps.drop all
+netfilter
 no3d
+nogroups
 nonewprivs
 noroot
 nosound
--- a/etc/Thunar.profile
+++ b/etc/Thunar.profile
@@ -16,20 +16,12 @@ include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-passwdmgr.inc
 
 caps.drop all
-netfilter
-nogroups
+no3d
 nonewprivs
 noroot
 nosound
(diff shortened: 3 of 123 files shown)


Fred-Barclay:
Wow that's a lot! 😁 There are a couple of things I'd like to research and test, but overall I think it looks pretty good.

curiosity-seeker:
While we're at it, I think it would be also a good idea to investigate where we can add

```
noexec ${HOME}
noexec /tmp
```

which is still missing in many profiles. For example, it's added to the `evince.profile` but not to `gwenview.profile` and `okular.profile`. It's also missing, e.g., in `chromium.profile` and `google-chrome.profile`. I remember that there was an issue regarding gimp where adding these options would break plugins. But I guess for most applications this would be okay.

Another thought: We have several profiles for applications running as root like dnsmasq, dnscrypt-proxy and unbound. Wouldn't it make sense to add the following rules?
```
blacklist /root
blacklist /home
```

Fred-Barclay:
@SpotComms The only problem I see is the addition of `net none` to all profiles with `protocol unix`. Sometimes the programme does need to access the unix or dbus socket. `protocol unix` allows that but blocks all other network access, while `net none` blocks everything so the programme doesn't work properly.

If you have atril or xreader (or possibly evince) installed you can check this. Start them with the normal profile and you'll be able to open a pdf (hopefully 😁). If you add `--net-none` to the arguments, you'll get an error like this:
```
(xreader:5): EggSMClient-WARNING **: Failed to connect to the session manager: Could not open network socket


** (xreader:5): WARNING **: Couldn't connect to accessibility bus: Failed to connect to socket /tmp/dbus-C5KfbsfovW: Connection refused
Failed to get bus connection: Could not connect: Connection refused

** (WebKitWebProcess:12): WARNING **: Couldn't connect to accessibility bus: Failed to connect to socket /tmp/dbus-C5KfbsfovW: Connection refused
```
The pdf reader will start, but you won't actually be able to view any pdfs. 

Cheers!

Ferroin:
@curiosity-seeker  Adding `noexec /tmp` shouldn't break anything legitimate, and I actually think if possible that should be part of the default sandbox with the option to override it.  `nonexec ${HOME}` on the other hand has the potential to break quite a few things, considering that most programs with plugins have the option to store them somewhere under ${HOME} and may run them directly.

As far as stuff running as root, blacklisting home directories (including /root) should be safe for pretty much everything in terms of network services, with the possible exception of some web servers (it will break most ~user setups).

curiosity-seeker:
@Ferroin : I'm glad that you agree. Yes, for `nonexec ${HOME}` we have to test if it causes any problems.

These are the profiles where I added 
```
noexec ${HOME}
noexec /tmp
```
without running into problems (so far):
```
chromium
google-chrome
gwenview
okular
quiterss
rhythmbox
guayadeque
dragon
digikam
cherrytree
```

Ferroin:
FWIW, most of the stuff I can think of that it might break other than GIMP is either using some scripting language for it's plugins (and thus not executing them in a way that `noexec` would stop), or doesn't make sense for sandboxing with firejail (like for example some older desktop environments).  My point is more that that's one that should be tested carefully because it's liable to break legitimate usage, whereas blocking execution in /tmp is not (not just because running stuff out of /tmp is bad practice, but there are also quite a few systems that already have /tmp mounted with `noexec`).

SpotComms:
@Fred-Barclay if that second commit is fine, I'll squash them.

Fred-Barclay:
@SpotComms I think it's ready to go! Do you want to squash it? Otherwise I can just merge as-is.

I've added back some netfilters/net nones that look to have been inadvertently removed. 

Cheers!
Fred

SpotComms:
Yea I thought I clobbered some. I've rebased it, feel free to merge.

Fred-Barclay:
Merged, thanks!

PR #1372, opened 2017-07-10
Fix permission denied for chromium-flags.conf in Arch

rccavalcanti:
Hi,

Since a recent update, Chromium on Arch can't access ~/.config/chromium-flags.conf anymore (permission denied).
That is fixed simply adding a noblacklist line to the profile.

Thank you for Firejail.


netblue30:
Thanks, that's a nice bug! The file was whitelisted in the home directory, only to be later blacklisted in disable-programs.inc.

PR #1373, opened 2017-07-11
Add a profile for SILENTARMY

SpotComms:
Profile for https://github.com/mbevand/silentarmy a zcash miner
- private-bin prevents it from starting, probably due to python
- disable-devel.inc prevents it from accessing any GPUs, not sure why


Fred-Barclay:
Merged, thanks!

PR #1377, opened 2017-07-13
Fix .java after e2449ae7d25925cec444ac08bbfb9cbc7199e647

SpotComms:
Any other Java programs?

### DIFF
--- a/etc/libreoffice.profile
+++ b/etc/libreoffice.profile
@@ -7,6 +7,7 @@ include /etc/firejail/libreoffice.local
 
 # Firejail profile for LibreOffice
 noblacklist ~/.config/libreoffice
+noblacklist ${HOME}/.java
 noblacklist /usr/local/sbin
 include /etc/firejail/disable-common.inc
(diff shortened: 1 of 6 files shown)


etc/libreoffice.profile
```diff
@@ -7,6 +7,7 @@ include /etc/firejail/libreoffice.local
 
 # Firejail profile for LibreOffice
 noblacklist ~/.config/libreoffice
+noblacklist ${HOME}/.java
```
SpotComms: LibreOffice uses Java for certain operations

etc/steam.profile
```diff
@@ -6,6 +6,7 @@ include /etc/firejail/globals.local
 include /etc/firejail/steam.local
 
 # Steam profile (applies to games/apps launched from Steam as well)
+noblacklist ${HOME}/.java
```
SpotComms: There are some Steam games that use Java

Fred-Barclay:
Merged, thanks!

PR #1380, opened 2017-07-13
tidy up of disable-common.inc

smitsohu:
after the partial reversion of #1270

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -296,5 +296,4 @@ blacklist /vmlinuz*
 blacklist /initrd*
 
 # complement noexec ${HOME} and noexec /tmp
-noexec ${HOME}/.config/pulse
 noexec /tmp/.X11-unix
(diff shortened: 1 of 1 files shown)


SpotComms:
Are we sure we can't uncomment noexec for trash? I don't see why it would need that.

smitsohu:
Are you referring to `noexec ${HOME}/.local/share`?

I agree, it should be safe to uncomment this line by default. I will add it to the PR

smitsohu:
Let me add a thought:

Uncommenting `noexec ~/.local/share` makes sense mainly as a secure default when also `read-write ~/.local/share` is uncommented.
However, given how many applications store something in ~/.local/share and would be affected by this change, I admit there is a real risk for breaking (exotic?) features in some of them.
Maybe we can avoid potential trouble and instead better document in disable-common.inc that the two lines `read-write ~/.local/share` and `noexec ~/.local/share` belong and should be uncommented together?

smitsohu:
Will close it for now, may reopen an improved request

PR #1389, opened 2017-07-16
Fix wget breaking rkhunter

SpotComms:
Not the best way, suggestions?

### DIFF
--- a/etc/wget.profile
+++ b/etc/wget.profile
@@ -32,5 +32,6 @@ private-dev
 # private-etc resolv.conf
 # private-tmp
 
+read-write /var/lib/rkhunter/tmp
 noexec ${HOME}
 noexec /tmp


Fred-Barclay:
IMHO it would be better to set up a separate rkhunter profile. We probably don't want to expose /var/lib/rkhunter/tmp as read-write to all programmes using wget. 

Cheers and thanks!
Fred

SpotComms:
I'll try to make one, my worry is that the sandbox even with features all disabled might cause rkhunter to not be able to scan properly.

Fred-Barclay:
How do you normally run rkhunter in firejail?

SpotComms:
I don't? But firecfg adds wget to /usr/local/bin, which rkhunter uses to update its signatures. So when rkhunter runs it tries to download updates to /var/lib/rkhunter/tmp which breaks since wget can't access that.

Fred-Barclay:
Oh, I see. Well, a rkhunter.profile with only `read-write /var/lib/rkhunter/tmp` should do fine, then. That way, when rkhunter calls wget, it will allow writing to that location.

PR #1390, opened 2017-07-16
Fix #1383

### LINKED ISSUE #1383: Programs with periods in their name fail to use their profile when run via symlink method
Examples: gimp-2.8.profile, idea.sh.profile

Related to #1268

SpotComms:


netblue30:
It seems to be working, thanks.

PR #1393, opened 2017-07-19
Improve mount handling

topimiettinen:
noexec/read-only/read-write handling: copy underlying FS flags
before remounting. If the flags are already OK, don't remount unnecessarily.


netblue30:
Merged, thanks.

PR #1394, opened 2017-07-19
Fix typo usr->user

topimiettinen:
This typo may have prevented blacklists of /run/user/{uid}/gnupg and
/run/user/{uid}/systemd from working.


netblue30:
Merged, thanks.

PR #1397, opened 2017-07-22
Add access to trash for eog

Panzerfather:
Eog needs access to the users trash folder to delete files.


Fred-Barclay:
Merged, thanks!

PR #1399, opened 2017-07-24
Private /lib feature

topimiettinen:


topimiettinen:
The code makes some x86_64 assumptions but it works for me. Please comment.

SpotComms:
Could you add support so that if just "private-lib" is defined with no args that it automatically adds the libs for the program? And keep the extra args support for more complex programs.

topimiettinen:
Good idea. Likewise, private-bin programs should be automatically handled. Then the args to private-lib should just add further libraries (instead of programs) to be copied, for example NSS or PAM libraries which are not automatically detected.

topimiettinen:
I implemented the new way in 5604600.

src/firejail/fs_lib.c
```diff
@@ -0,0 +1,187 @@
...
+				dbuf++;
+			}
+		}
+		sbuf++;
+	}
+	goto close;
+
+ error_close:
+	perror("copy libs");
+ close:
```
topimiettinen: munmap()

reinerh:
Could you please also evaluate the LIBDIR constant? It is configured during build time and points to the platform-specific library directory (e.g. /usr/lib/x86_64-linux-gnu, /usr/lib/i386-linux-gnu or even /usr/lib/x86_64-linux-musl). This would greatly improve support for architectures other than amd64.

topimiettinen:
@reinerh: I added LIBDIR and simple Elf32 support (and missing munmap) in 0e9adf2.

topimiettinen:
I added config file support in 0052c91b.

SpotComms:
So passing "--private-lib=" on the command line does some autodetection for some libs and works no issue for basic programs. But adding "private-lib" to a profile doesn't have the same effect, as it expects libraries to be passed to it. And passing unknown libraries to either crash firejail. For ease of use can the options support passing both libraries and programs?

topimiettinen:
Sorry, I didn't test the profile version very well. I pushed 8d5b397 which should fix this, now you can use private-lib without args in the profile.

SpotComms:
Thanks for https://github.com/netblue30/firejail/commit/8d5b39766410b6d6eba7e6805691fceb88eca004 I've however found another issue. When running "firejail program" and program is symlinked into /usr/local/bin via firecfg, private-lib will fail to identify the actual libs necessary and will always result in the sandbox exiting instantly.

topimiettinen:
Thanks, I think the symlink and libtracelog problems can be fixed.

The third one is not so obvious. Maybe /usr/lib should be handled in a different way. Not mounting pretty much defeats the idea but the current method could make the feature useless for many programs.

Fred-Barclay:
G'day @topimiettinen - pretty cool feature you've got here. 👍 
I'm getting this error when building from the latest here:
```
fs_lib.c: In function ‘copy_libs_for_lib’:
fs_lib.c:133:2: error: ‘for’ loop initial declarations are only allowed in C99 or C11 mode
  for (int i = 0; lib_paths[i]; i++) {
  ^
fs_lib.c:133:2: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code
Makefile:37: recipe for target 'fs_lib.o' failed
```

PR #1408, opened 2017-07-27
Zoom cache dir

drzraf:
Zoom seems to use of a QT cache-disk feature which depends upon a `~/.cache/<app>/qmlcache` directory.
If it can not, Zoom will segfault with
`mprotect failed in ExecutableAllocator::makeExecutable: Permission denied`


Fred-Barclay:
Merged, thanks!

PR #1409, opened 2017-07-27
Fix typo for fnet moveif invocation on 2nd interface.

caoliver:
I noticed a fair bit of copy pasta here.  While that may not necessarily be bad in itself, is there a reason where the duplication can be easily parameterized to avoid using C preprocessor macros with the ## operator to avoid this sort of typo?


netblue30:
Merged, thanks.

PR #1411, opened 2017-07-29
Add a profile for arm

SpotComms:
profile for arm https://www.torproject.org/projects/arm.html.en

### DIFF
--- a/etc/arm.profile
+++ b/etc/arm.profile
@@ -1,0 +1,42 @@
+# Persistent global definitions go here
+include /etc/firejail/globals.local
+
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+include /etc/firejail/arm.local
+
+# Firejail profile for arm
+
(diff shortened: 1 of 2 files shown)


etc/arm.profile
```diff
@@ -0,0 +1,42 @@
...
+include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-passwdmgr.inc
+include /etc/firejail/disable-programs.inc
+
+mkdir ${HOME}/.arm
+whitelist ${HOME}/.arm
+include /etc/firejail/whitelist-common.inc
+
+caps.drop all
+ipc-namespace
```
Fred-Barclay: Are we sure ipc-namesepace will work for everyone?

SpotComms:
I thought ipc-namespace was only breaking some gui applications? arm is cli only.

Fred-Barclay:
I'm not sure. Let's merge and see what happens. :)

PR #1412, opened 2017-07-29
Improve loading of seccomp filter and memory-deny-write-execute feature

topimiettinen:
The seccomp patch also fixes a memory leak and double load.

Memory-deny-write-execute is similar to systemd directive with same name (sans CamelCase).


SpotComms:
Oooh MemoryDenyWriteExecute looks awesome, but shouldn't it be in a separate PR?

SpotComms:
Can you make memory-deny-write-execute a config option too, instead of cli only?

topimiettinen:
It was there but accidentally parser expected a space after the word, fixed in 8d5b397.

Issue #1405, opened 2017-07-26
Firejail X11 sandboxing unusable with latest xpra


chiraag-nataraj:
Piggybacking off of #1016, but now I'm seeing this with every program. It got so bad that I had to write a wrapper script as follows:
````
#!/bin/sh

DISP="$1"
COM="$2"

xpra start $DISP --start-child="$COM" --exit-with-children=yes;
xpra attach $DISP &
````

I'm using xpra v2.1-r16505.

chiraag-nataraj:
With newer xpra versions, 5 seconds isn't enough delay (at least on my system) - not sure why exactly. Neither is waiting for `/tmp/.X11-unix/<display>`. Here is a sample log:
````
X.Org X Server 1.19.3
Release Date: 2017-03-15
X Protocol Version 11, Revision 0
Build Operating System: Linux 4.9.0-3-amd64 x86_64 Debian
Current Operating System: Linux chiraag 4.11.11-chiraag #4 SMP PREEMPT Thu Jul 20 00:00:32 EDT 2017 x86_64
Kernel command line: initrd=\33ef9f252dda82b147eebe6f5324ddb4\4.11.11-chiraag\initrd root=/dev/mapper/Crypto-Root ro quiet
Build Date: 07 July 2017  06:22:09AM
xorg-server 2:1.19.3-2 (https://www.debian.org/support) 
Current version of pixman: 0.34.0
	Before reporting problems, check http://wiki.x.org
	to make sure that you have the latest version.
Markers: (--) probed, (**) from config file, (==) default setting,
	(++) from command line, (!!) notice, (II) informational,
	(WW) warning, (EE) error, (NI) not implemented, (??) unknown.
(++) Log file: "/run/user/1000/xpra/Xorg.:212.log", Time: Fri Jul 28 16:34:27 2017
(++) Using config file: "/etc/xpra/xorg.conf"
(==) Using system config directory "/usr/share/X11/xorg.conf.d"

Xpra server pid 4338, xpra client pid 4383, jail 4384
*** Attaching to xpra display 212 ***

Reading profile /home/chiraag/.config/firejail/x-terminal-emulator.profile
Parent pid 4384, child pid 4385
Child process initialized in 26.27 ms
urxvt: can't open display :212, aborting.

Parent is shutting down, bye...
2017-07-28 16:34:30,312 Xpra gtk2 client version 2.1-r16525 64-bit
2017-07-28 16:34:30,312  running on Linux Debian unstable sid
[...]

PR #1415, opened 2017-07-30
Tentative implementation for #1405

chiraag-nataraj:
Fix for recent versions of `xpra` not working properly.

### DIFF
--- a/src/firejail/x11.c
+++ b/src/firejail/x11.c
@@ -685,6 +685,12 @@ void x11_start_xpra(int argc, char **argv) {
 	char *start_child_prefix = "--start-child=";
 	char *start_child;
 	start_child = malloc(total_length + strlen(start_child_prefix) + fpos + 2);
+	if (start_child == NULL)
+	  {
+	    fprintf(stderr, "Error: unable to allocate start_child to assemble command\n");
+	    exit(1);
+	  }
+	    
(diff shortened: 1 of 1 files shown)


chiraag-nataraj:
I realized I never `free` `start_child`. Also, I commented out the exit code since `xpra` handles it now, but that means there isn't a good place to free `start_child`...

src/firejail/x11.c
```diff
@@ -870,6 +862,15 @@ void x11_start_xpra(int argc, char **argv) {
 
 	/* sleep(1); // adding a delay in order to let the server start */
 
+	// wait for server to end
+	while (1) {
```
SpotComms: What kind of perf impact will this have, won't it just peg a core at 100%? Can you add a `nanosleep(250)`?

netblue30:
Let's give it a try, thanks!

netblue30:
Some more things going in there.

We have two start_xpra() functions, the old one and the new one. I am looking around, most distos are still on Xpra v0.17. I added in /etc/firejail/firejail.config a "xpra-attach" setting. You'll have to put it to yes and it will use your version of start_xpra(). For now the setting is "no", meaning it will run the old version - we'll switch it as soon as xpra 2.x versions start appearing in distributions.

PR #1416, opened 2017-07-30
telegram is called telegram-desktop in Debian

rahiel:
The Telegram client is called `telegram-desktop` in Debian: https://packages.debian.org/sid/telegram-desktop. (It's also not an IRC client, but uses its own protocol.


netblue30:
merged, thanks.

PR #1426, opened 2017-08-02
Apparmor: add local configuration

VladimirSchowalter20:
This adds possibility for local enchantments of firejail apparmor profile configuration. It's designed same way as in most upstream apparmor profiles.


startx2017:
Merged!

Issue #1096, opened 2017-02-10
Firefox 51.0 segfaults with --private-dev option


acrobat1:
Firefox 51.0 segfaults when is launched and when it's closed if using the **--private-dev** option 

`firefox[4203]: segfault at 0 ip 00007f5281ee721a sp 00007ffc6a3e1670 error 6 in libxul.so`

To be precise, it does not crash but it throws this error multiple times (using --tracelog) when it starts and when it exits - which I've never seen up to this 51.0 version (64bit).

I've tested with **--noprofile** and it seems consistent in Firejail  0.9.44.8.

netblue30:
I have it running fine here (regular AMD driver from Debian stable). Maybe is because of --tracelog? Build a custom profile like this:
`````
$ mkdir -p ~/.config/firejail
$ echo "ignore private-dev" > ~/.config/firejail/firefox.profile
$ echo "include /etc/firejail/firefox.profile" >> ~/.config/firejail/firefox.profile
`````
I'll keep an eye on it if anybody is seeing the problem, in the worst case we remove private-dev from the profile.

acrobat1:
Hi, sorry for my late reply : in the meantime I've tried with a "vanilla" Firefox 51 installation (=no modifications whatsoever) and this very short command line **firejail   --noprofile --private-dev** and it segfaults , remove the **private-dev** argument and it doesn't.

I too was thinking about --tracelog, but as far as I can see it really is triggered by --private-dev, and it only started with Firefox 51 - I've never seen it before.

Would that test above be comparable to what you were suggesting ?

Do you have any ideas about to why Firefox would do this from version 51.0 but not before ?

BTW, I too am on Debian stable with no weird drivers or custom kernels.

acrobat1:
It just occurred to me that in firejail.config I've set **chroot-desktop no** : could this be the reason of this issue?

netblue30:
That chroot-desktop should make a difference only if you are using --chroot on the command line.

acrobat1:
Well, I'm not using --chroot on the command line .. so , until someone can figure this out, the only thing for me left to do is to run firefox without **--private-dev** option.

BTW, what we exactly lose by dropping **--private-dev** ? Is that a major drawback security-wise ?

netblue30:
Not too much, it is a secondary feature.

PR #1427, opened 2017-08-02
Unify all profiles

### LINKED ISSUE #793: Overhaul of Profiles
So profiles right now are a mess, they're all in different formats by different people and profiles get added without disable-programs.inc getting updated. So I made a program that handles all programs in a streamlined XML file and then generates .profiles from that. I have manually rewritten all current profiles into said XML, and the generator spits out nice and tidy .profiles. I plan on making a pull request soon that replaces all the old profiles with the new ones. But I'd like some people to take a look over to make sure I didn't forget anything too major. A few profiles were also added and some were made more strict.

Links:
[Repository](https://github.com/SpotComms/FirejailProfileGenerator)
[XML](https://github.com/SpotComms/FirejailProfileGenerator/blob/master/Firejail_Config.xml)
[Profiles](https://github.com/SpotComms/FirejailProfileGenerator/tree/master/generated)
[Program](https://github.com/SpotComms/FirejailProfileGenerator/blob/master/src/info/spotcomms/firejailprofilegenerator/Main.java)

[Help Test](https://github.com/SpotComms/FirejailProfileGenerator/issues/1)

### LINKED ISSUE #1434: app local profiles should be placed before global local profiles
Currently local profiles are sourced in this precedence:
```
# Persistent global definitions go here

# This file is overwritten during software install.
include /etc/firejail/ark.local

```

It means that when user enables some option in `global.local `profile it can't be overridden in `ark.local` profile. Therefore any global option which breaks one app (example:` apparmor`) can be overridden only by editing `ark.profile` directly which will be overwritten during next update or by creating new profile in `~/.config/firejail/ark.profile` which have to be then maintained solely by user.

Basing on the above I propose changing precedence to:
```
# This file is overwritten during software install.
include /etc/firejail/ark.local

# Persistent global definitions go here

```
This way user can override any `global.local` option in `ark.local` file easily. As both `global.local` and `ark.local` are used only for personal local administration this shouldn't create any security issues.

SpotComms:
- ~~88 profiles have clobbered comments that need to be manually restored~~
- ~~**DO NOT MERGE**~~
- For ease of review, please look at the individual profiles in the branch itself here https://github.com/SpotComms/firejail/tree/pr/etc
- ~~Please comment on the format, if everyone agrees I'll start fixing comments~~
- Closes #793 
- Fixes #1434 
- ~~There is also some minor weirdness that I'll have to fix~~
- Program here: https://gist.github.com/SpotComms/5bbef9c1297a0d1eb2fa268ed8041d7f

Format
```
HEADER
BLACKLIST
NOBLACKLIST
DISABLE-INCLUDES
WHITELIST
OPTIONS
PRIVATE-OPTIONS
MISC-OPTIONS
INCLUDES
```

### DIFF
--- a/etc/7z.profile
+++ b/etc/7z.profile
@@ -1,23 +1,22 @@
+# Firejail profile for 7z
+# This file is overwritten after every install/update
 quiet
-# Persistent global definitions go here
-include /etc/firejail/globals.local
-
-# This file is overwritten during software install.
-# Persistent customizations should go in a .local file.
+# Persistent local customizations
 include /etc/firejail/7z.local
-
-# 7zip crompression tool profile
-ignore noroot
-
-include /etc/firejail/default.profile
+# Persistent global definitions
+include /etc/firejail/globals.local
 
 blacklist /tmp/.X11-unix
 
-tracelog
+ignore noroot
 net none
+no3d
+nosound
 nosound
 novideo
 shell none
+tracelog
+
 private-dev
-nosound
-no3d
+
+include /etc/firejail/default.profile
--- a/etc/android-studio.profile
+++ b/etc/android-studio.profile
@@ -1,11 +1,9 @@
-# Persistent global definitions go here
-include /etc/firejail/globals.local
-
-# This file is overwritten during software install.
-# Persistent customizations should go in a .local file.
+# Firejail profile for android-studio
+# This file is overwritten after every install/update
+# Persistent local customizations
 include /etc/firejail/android-studio.local
-
-# Firejail profile for Android Studio
+# Persistent global definitions
+include /etc/firejail/globals.local
 
 noblacklist ${HOME}/.AndroidStudio*
 noblacklist ${HOME}/.android
@@ -25,13 +23,15 @@ netfilter
 nogroups
 nonewprivs
 noroot
-#nosound
 novideo
 protocol unix,inet,inet6
 seccomp
 shell none
 
 private-dev
-#private-tmp
+# private-tmp
 
 noexec /tmp
+
+# CLOBBERED COMMENTS
+# nosound
--- a/etc/android-studio.profile
+++ b/etc/android-studio.profile
@@ -32,6 +32,3 @@ private-dev
 # private-tmp
 
 noexec /tmp
-
-# CLOBBERED COMMENTS
-# nosound
--- a/etc/bibletime.profile
+++ b/etc/bibletime.profile
@@ -1,11 +1,13 @@
-# Persistent global definitions go here
+# Firejail profile for bibletime
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /etc/firejail/bibletime.local
+# Persistent global definitions
 include /etc/firejail/globals.local
 
-# This file is overwritten during software install.
-# Persistent customizations should go in a .local file.
-include /etc/firejail/bibletime.local
+blacklist ~/.Xauthority
+blacklist ~/.bashrc
 
-# Firejail profile for BibleTime
 noblacklist ~/.bibletime
 noblacklist ~/.config/qt5ct
 noblacklist ~/.sword
@@ -15,13 +17,10 @@ include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-passwdmgr.inc
 include /etc/firejail/disable-programs.inc
 
-blacklist ~/.bashrc
-blacklist ~/.Xauthority
-
 whitelist ${HOME}/.bibletime
 whitelist ${HOME}/.config/qt5ct
 whitelist ${HOME}/.sword
-
+include /etc/firejail/whitelist-common.inc
 
 caps.drop all
 netfilter
@@ -35,7 +34,7 @@ seccomp
 shell none
 tracelog
 
-#private-bin bibletime,qt5ct
-private-etc fonts,resolv.conf,sword,sword.conf,passwd
+# private-bin bibletime,qt5ct
 private-dev
+private-etc fonts,resolv.conf,sword,sword.conf,passwd
 private-tmp
--- a/etc/bitlbee.profile
+++ b/etc/bitlbee.profile
@@ -1,13 +1,13 @@
-# Persistent global definitions go here
+# Firejail profile for bitlbee
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /etc/firejail/bitlbee.local
+# Persistent global definitions
 include /etc/firejail/globals.local
 
-# This file is overwritten during software install.
-# Persistent customizations should go in a .local file.
-include /etc/firejail/bitlbee.local
-
-# BitlBee instant messaging profile
 noblacklist /sbin
 noblacklist /usr/sbin
+
 include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-devel.inc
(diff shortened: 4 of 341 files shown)


etc/firefox.profile
```diff
@@ -1,77 +1,74 @@
...
+netfilter
+nogroups
+nonewprivs
+noroot
+protocol unix,inet,inet6,netlink
+seccomp
+shell none
+tracelog
+
+# private-dev might prevent video calls going out
```
smitsohu: Do we actually still need this comment at all?

SpotComms: That's actually something I wanted to bring up. I think we should keep them, but there are also quite a few comments that could also be removed.

VladimirSchowalter20: I think it's useful hint for users showing them consequences of some options so they can make informative decisions for manually override them.

Fred-Barclay: Yes let's keep those type of comments. As @VladimirSchowalter20 says it gives users some idea of what can happen if they uncomment the option. Also it's nice when I'm reviewing profiles to see if/how they can be hardened - if an option has broken a profile for some users, I can read the comment and know why I shouldn't add it back, rather than having to dig through the git log.

smitsohu: My impression was this one became obsolete with cfbcbf2c95455373aa2570827c52b7b87d80cfef, but of course, if it continues to be useful let's keep it :)

Fred-Barclay: @smitsohu Haha. You're right, we can probably uncomment that particular line now and video calls should still work.

We have had a report that Firefox segfaults with private-dev (see #1096) but it's only one report and for an old version of Firefox - just something to keep in mind if we do uncomment private-dev.

Fred-Barclay: @SpotComms Pedantic note here - I rather like having a space after the # in the comments (i.e. `# Firejail profile for XYZ` instead of `#Firejail profile for XYZ`). Thoughts?

SpotComms: @Fred-Barclay updated

Fred-Barclay:
@SpotComms can you explain more about the difference between OPTIONS and MISC-OPTIONS?

etc/android-studio.profile
```diff
@@ -1,14 +1,12 @@
...
-# This file is overwritten during software install.
-# Persistent customizations should go in a .local file.
+# Persistent local customizations
 include /etc/firejail/android-studio.local
 
-# Firejail profile for Android Studio
-
-noblacklist ${HOME}/.AndroidStudio*
 noblacklist ${HOME}/.android
+noblacklist ${HOME}/.AndroidStudio*
```
Fred-Barclay: I think in our other files (like disable-programs.inc) upper-case is placed above lowercase: i.e. 
```
noblacklist ${HOME}/.AndroidStudio*
noblacklist ${HOME}/.android
```

SpotComms: That is just Java's arraylist.sort(String::compareToIgnoreCase); Is it necessary?

SpotComms:
MISC-OPTIONS is just 'noexec' and 'memory-deny-write-execute'.
PRIVATE-OPTIONS is 'private*' and 'disable-mnt'

Here is the program: https://gist.github.com/SpotComms/5bbef9c1297a0d1eb2fa268ed8041d7f

etc/firefox.profile
smitsohu: @SpotComms Is it on purpose that commented include statements are removed completely from the profiles?

SpotComms: @smitsohu I probably had a reason, but I can't remember it, I've added them back.

etc/7z.profile
```diff
@@ -1,23 +1,22 @@
...
 
-# 7zip crompression tool profile
-ignore noroot
-
-include /etc/firejail/default.profile
-
 blacklist /tmp/.X11-unix
 
-tracelog
+ignore noroot
```
Fred-Barclay: Note to self: check if `ignore` has to go before `include /etc/firejail/default.profile`

SpotComms:
I have fixed comments and I also fixed sorting of commented out options.

curiosity-seeker:
@SpotComms 
> MISC-OPTIONS is just 'noexec' and 'memory-deny-write-execute'.

Just a thought. _man firejail_ says:
> If there are more than one mount operation on the path of the file or directory, noexec should be applied to the last one.

So shouldn't MISC-OPTIONS be better placed at the bottom of the profile (below INCLUDES) to be on the safe side?

SpotComms:
@curiosity-seeker I don't think there are any instances currently where that occurs. Meaning that this pr shouldn't cause any regressions. I do plan to make like a proper standalone c++ program that formats profiles, for people to use before submitting theirs, so that would handle that for the future.

curiosity-seeker:
> I do plan to make like a proper c++ program that formats profiles, for people to use before submitting theirs

Cool!

Fred-Barclay:
> I do plan to make like a proper standalone c++ program that formats profiles, for people to use before submitting theirs

Haha, I was planning on doing the same in python. 😄

SpotComms:
I figured if it was in cpp it could be included alongside firejail itself, like just a "firejail-format [x.profile...]"

Fred-Barclay:
@SpotComms I can see a message in my dashboard but not here about including it with firejail. Definitely that sounds like a good idea. I don't know if @netblue30 will want to add an option within firejail itself for wrangling profiles into shape, but if nothing else we can ship it in contrib. 

Besides language, how would this differ from your Java program? Definitely using C++ would be a good idea since a lot of people (myself included) probably won't have Java or open-jdk installed on their machines.

etc/bibletime.profile
```diff
@@ -1,11 +1,12 @@
...
+# Firejail profile for bibletime
+# This file is overwritten after every install/update
+# Persistent local customizations
 include /etc/firejail/bibletime.local
+# Persistent global definitions
+include /etc/firejail/globals.local
 
-# Firejail profile for BibleTime
+blacklist ~/.Xauthority
+blacklist ~/.bashrc
```
Fred-Barclay: Can we get a newline between blacklists and noblacklists?

etc/bitlbee.profile
```diff
@@ -16,16 +16,16 @@ include /etc/firejail/disable-programs.inc
...
 no3d
 nonewprivs
-private
-private-dev
-protocol unix,inet,inet6
-seccomp
 nosound
 novideo
+protocol unix,inet,inet6
 read-write /var/lib/bitlbee
```
Fred-Barclay: Would `read-write` make more sense somewhere else in the profile? Maybe earlier like where we're putting blacklists (see the Bibletime profile for reference) since the format is similar to that of blacklists?

SpotComms:
Fixed those last two comments and rebased.

Fred-Barclay:
@SpotComms Is this ready to merge?

SpotComms:
@Fred-Barclay I think so, if anything it shouldn't cause any regressions.

Fred-Barclay:
Can you do them as another commit?

SpotComms:
@Fred-Barclay All 341 profiles have been unified.

Fred-Barclay:
Merged, thanks for all your work!

PR #1177, opened 2017-03-29
various profile fixes and enhancements

SYN-cook:
- various KDE related enhancements. Please note that services work in Dolphin despite `blacklist ${HOME}/.local/share/kservices5`, so I didn't unblacklist it
- `private-dev` unfortunately prohibits playing audio CD in Audacious
- more Nautilus/Nemo folders

PR #1428, opened 2017-08-03
Change ${HOME}/.local/share/kservices5 to read-only

VladimirSchowalter20:
`${HOME}/.local/share/kservices5` is blacklisted since https://github.com/netblue30/firejail/pull/1177 but there is no explanation for this. Some applications try to read this directory and it has legitimate use cases. Also `/usr/share/kservices5` isn't blacklisted so blacklisting only it's local variant isn't consistent. Changing it to read-only is both safe and don't trigger blacklist violations.

```
blacklist violation - sandbox 20831, exe okular, syscall opendir, path /home/mr/.local/share/kservices5
blacklist violation - sandbox 22050, exe gwenview, syscall opendir, path /home/mr/.local/share/kservices5
blacklist violation - sandbox 22386, exe kde-open5, syscall opendir, path /home/mr/.local/share/kservices5
```

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -70,7 +70,7 @@ blacklist ${HOME}/.config/krunnerrc
 blacklist ${HOME}/.config/plasma-org.kde.plasma.desktop-appletsrc
 blacklist ${HOME}/.local/share/kglobalaccel
 blacklist ${HOME}/.local/share/konsole
-blacklist ${HOME}/.local/share/kservices5
+read-only ${HOME}/.local/share/kservices5
 blacklist ${HOME}/.local/share/kwin
 blacklist ${HOME}/.local/share/plasma
 blacklist ${HOME}/.local/share/solid


smitsohu:
Should the equivalent KDE4 locations 
```
${HOME}/.kde/share/kde4/services
${HOME}/.kde4/share/kde4/services
```
then also be read-only?

VladimirSchowalter20:
Probably yes, although those paths are for legacy software so I didn't test them. We should think about dropping them in future as they make configs cluttered and less readable. I doubt that anyone will use newest firejail with outdated software especially as there is LTS version avalaible.

Fred-Barclay:
People with Linux Mint 17.x KDE and Kubuntu 14.04 will still be using KDE 4 for the next 2 years. Let's include those paths as well. 😉

Fred-Barclay:
@VladimirSchowalter20 Which programs cause these warnings? Is it only okular and gwenview?

VladimirSchowalter20:
I believe there is many more as it's common KDE feature. There shouldn't be any secrets in this directory as it's nature is similar to `${HOME}/.local/share/applications` which is [already set](https://github.com/netblue30/firejail/blob/master/etc/disable-common.inc#L194) as `read-only` path. You can look into `/usr/share/kservices5` to get an idea.

I think firejail shouldn't block directories which apps want to access during normal usage as it creates false positives and clutters logs.

I personally don't want to add changes which I didn't test for myself so changing any other dirs is up to you.

Please merge.

smitsohu:
I would agree here with @VladimirSchowalter20: While generally speaking blacklisting is always better than mere read-only, it effectively doesn't add much security in the case of kservices5. From my perspective, too, there is not much value in keeping these configurations secret, and we don't loose anything if we trade it in for cleaner logs.

Having said that, loosening a restriction actually shouldn't break features, so I guess it is very safe to extend the fix to the equivalent KDE4 locations.

Fred-Barclay:
Thanks for the explanation guys. :) 
Merged!

PR #1430, opened 2017-08-04
profile fixes

VladimirSchowalter20:
Added directories which are used for tracking recently opened files and dirs by `qBittorrent` and `gwenview` apps. Similar dirs are already included in many other profiles. Also added KDE 5 version of `kbuildsycoca` to private-bin for `gwenview`.

### DIFF
--- a/etc/gwenview.profile
+++ b/etc/gwenview.profile
@@ -12,6 +12,7 @@ noblacklist ~/.kde/share/apps/gwenview
 noblacklist ~/.kde/share/config/gwenviewrc
 noblacklist ~/.config/gwenviewrc
 noblacklist ~/.config/org.kde.gwenviewrc
+noblacklist ~/.local/share/gwenview
 noblacklist ~/.local/share/org.kde.gwenview
 include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-programs.inc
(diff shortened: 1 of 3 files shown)


Fred-Barclay:
Merged, thanks!

etc/gwenview.profile
```diff
@@ -27,7 +28,7 @@ seccomp
 shell none
 tracelog
 
-private-bin gwenview,kbuildsycoca4,gimp,gimp-2.8
+private-bin gwenview,kbuildsycoca4,kbuildsycoca5,gimp,gimp-2.8
```
smitsohu: I'm sorry, it's me again :)
For me Gwenview works fine as it is now, that is without kbuildsycoca5.... though I actually don't understand why it works.

The sycoca5 database seems to live in /home/user/.cache, where it is accessible from inside the sandbox. If the database is hidden (whitelisted profile, mkdir ~/.cache) it is recreated, but surprisingly this happens whether kbuildsycoca5 is available in the sandbox or not. This is on Debian.

VladimirSchowalter20: Probably you are right. I made request for drop it.

Fred-Barclay: Thanks for the review @smitsohu . The pull request from @VladimirSchowalter20 dropping kbuildsyscoca5 has just been merged.

PR #1431, opened 2017-08-04
Add 8 new profiles

SpotComms:
I was unable to fully test TrueCraft due to https://github.com/SirCmpwn/TrueCraft/issues/268 , but the launcher and world generation seems to work fine.


Fred-Barclay:
Merged, thanks!

PR #1433, opened 2017-08-04
various profile fixes

VladimirSchowalter20:
Here are various fixes for calibre, mpv, qbittorrent, wireshark, wine and vlc profiles. All are tested by me. Explanations are in comments.

### DIFF
--- a/etc/calibre.profile
+++ b/etc/calibre.profile
@@ -22,7 +22,7 @@ nonewprivs
 noroot
 nosound
 novideo
-protocol unix,inet,inet6,netlink
+protocol unix,inet,inet6
 seccomp
 shell none
 tracelog
--- a/etc/calibre.profile
+++ b/etc/calibre.profile
@@ -22,7 +22,7 @@ nonewprivs
 noroot
 nosound
 novideo
-protocol unix,inet,inet6
+protocol unix,inet,inet6,netlink
 seccomp
 shell none
 tracelog
--- a/etc/mpv.profile
+++ b/etc/mpv.profile
@@ -16,11 +16,15 @@ include /etc/firejail/disable-passwdmgr.inc
 
 caps.drop all
 netfilter
+nogroups
 nonewprivs
 noroot
 protocol unix,inet,inet6
 seccomp
+tracelog
 
 # to test
+ipc-namespace
 shell none
 private-bin mpv,youtube-dl,python,python2.7,python3.6,env
+private-dev
--- a/etc/mpv.profile
+++ b/etc/mpv.profile
@@ -24,7 +24,7 @@ seccomp
 tracelog
 
 # to test
-ipc-namespace
+# ipc-namespace
 shell none
 private-bin mpv,youtube-dl,python,python2.7,python3.6,env
 private-dev
--- a/etc/qbittorrent.profile
+++ b/etc/qbittorrent.profile
@@ -34,11 +34,12 @@ nogroups
 nonewprivs
 noroot
 nosound
-protocol unix,inet,inet6
+protocol unix,inet,inet6,netlink
 seccomp
 
 # there are some problems with "Open destination folder", see bug #536
 #shell none
 #private-bin qbittorrent
 private-dev
+#private-etc X11,fonts,xdg,resolv.conf
 private-tmp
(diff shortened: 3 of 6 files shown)


etc/calibre.profile
```diff
@@ -22,7 +22,7 @@ nonewprivs
 noroot
 nosound
 novideo
-protocol unix,inet,inet6
+protocol unix,inet,inet6,netlink
```
Fred-Barclay: Is the message about netlink purely cosmetic or does it actually break calibre from working properly?

etc/mpv.profile
```diff
@@ -16,11 +16,15 @@ include /etc/firejail/disable-passwdmgr.inc
...
 netfilter
+nogroups
 nonewprivs
 noroot
 protocol unix,inet,inet6
 seccomp
+tracelog
 
 # to test
+ipc-namespace
```
Fred-Barclay: Looks good but let's comment out `ipc-namespace` please. We've had some breakages on other systems.

etc/qbittorrent.profile
```diff
@@ -34,11 +34,12 @@ nogroups
 nonewprivs
 noroot
 nosound
-protocol unix,inet,inet6
+protocol unix,inet,inet6,netlink
```
Fred-Barclay: Does qbittorrent work without netlink (despite the error message) or does it break using qbittorrent?

etc/vlc.profile
```diff
@@ -23,7 +23,7 @@ protocol unix,inet,inet6,netlink
 seccomp
 shell none
 
-private-bin vlc,cvlc,nvlc,rvlc,qvlc,svlc,vlc-wrapper
+private-bin vlc,cvlc,nvlc,rvlc,qvlc,svlc
```
Fred-Barclay: Just curious - was vlc-wrapper not needed in private-bin?

Fred-Barclay:
@VladimirSchowalter20  Thanks for this and sorry for all the questions (hope it doesn't flood your inbox)! 😄

etc/qbittorrent.profile
VladimirSchowalter20: I'm not sure. I'm not using much networking features under calibre but it has options like running it's own server which can be affected by his.

I think we shouldn't block by default anything what applications try access by default and throwing errors as it's very possible that it break some usage which can be hard to debug.

etc/vlc.profile
VladimirSchowalter20: That's my mistake. I wanted to add it first but reconsidered after reading [manual](https://manpages.debian.org/stretch/vlc-bin/vlc-wrapper.1.en.html). It needs sudo so it won't work with firejail anyway and it's very special case usage which 99% users won't care about.

Unfortunately I committed changes with `vlc-wraper` added to private-bin mistakenly so I had  to remove it in next commit. I'm sorry for this ugliness.

etc/calibre.profile
VladimirSchowalter20: I'm not sure. I'm not using much networking features under calibre but it has options like running it's own server which can be affected by his.

I think we shouldn't block by default anything what applications try access by default and throwing errors as it's very possible that it break some usage which can be hard to debug.

etc/mpv.profile
VladimirSchowalter20: Those breakages were in mpv? I had issues with many other apps while using ipc-namespace but not in mpv

etc/calibre.profile
Fred-Barclay: Hi - let's leave netlink disabled for now. It's an easy fix to change this from a user's perspective, and I think it's better to err on the side of too much restriction here.

etc/mpv.profile
Fred-Barclay: I don't recall any problems with mpv specifically, but here I'd rather err on the side of stability since IIRC ipc-namespaces were significantly problematic for several people. Let's comment it.

etc/qbittorrent.profile
Fred-Barclay: I see - thanks!

etc/vlc.profile
Fred-Barclay: No worries mate.

VladimirSchowalter20:
@Fred-Barclay I made changes which you asked. Thanks for reviews.

Fred-Barclay:
Merged, thanks!

PR #1437, opened 2017-08-06
fix steam startup with >=llvm-4

soredake:
`libGL: dlopen /usr/lib32/dri/nouveau_dri.so failed (libLLVMObjCARCOpts.so.4: cannot open shared object file: No such file or directory)`

### DIFF
--- a/etc/steam.profile
+++ b/etc/steam.profile
@@ -5,6 +5,9 @@ include /etc/firejail/globals.local
 # Persistent customizations should go in a .local file.
 include /etc/firejail/steam.local
 
+# with >=llvm-4 mesa drivers need llvm stuff
+noblacklist /usr/lib/llvm*
+
 # Steam profile (applies to games/apps launched from Steam as well)
 noblacklist ${HOME}/.java
 noblacklist ${HOME}/.Steam


SpotComms:
Isn't LLVM only needed for OpenCL?

Fred-Barclay:
Hi @soredake 
When do these errors occur? Is it every time you start a program with firejail, or only for certain programs?
Does this keep you from being able to use firejail, or do programs still work but print out this error message?

Thanks!

soredake:
@Fred-Barclay 
>When do these errors occur?
On steam startup.
>Is it every time you start a program with firejail, or only for certain programs?
Only for steam.
>Does this keep you from being able to use firejail,
Yes, steam is not working without this libs with new llvm version.

Fred-Barclay:
Thanks, merged!

cleeus:
I currently have to run bleeding edge mesa-17.3.0 with llvm-4.0.1.
This seems to affect most OpenGL applications including kodi and vlc.
I added 6c8cfe14c58a60e2089098db1bc5a71cce0d0ca1 to my defaults.global to fix that stuff.

PR #1438, opened 2017-08-06
Change KDE4 services folder to read-only

smitsohu:
Parallels #1428

Configurations in this location don't have much value as a secret, but need to be protected from manipulation. Let's make them available to all KDE apps for legitimate use. 
Looking at Dolphin, it is actually a security improvement.


Fred-Barclay:
Merged, thanks!

PR #1442, opened 2017-08-08
various little profile fixes and enhancements

smitsohu:
- more novideo options, without any systematics. Works as in _worksforme_
- blacklist the clementine database  and configuration
- better privacy for the tor-browser

### DIFF
--- a/etc/ark.profile
+++ b/etc/ark.profile
@@ -18,6 +18,7 @@ nogroups
 nonewprivs
 noroot
 nosound
+novideo
 protocol unix
 seccomp
 shell none
--- a/etc/audacious.profile
+++ b/etc/audacious.profile
@@ -17,6 +17,7 @@ caps.drop all
 netfilter
 nonewprivs
 noroot
+novideo
 protocol unix,inet,inet6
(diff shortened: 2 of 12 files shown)


smitsohu:
@SpotComms Thanks for the hint! I'll do the same with geary.

Fred-Barclay:
Hi @smitsohu 
Thanks for the improvements!

For mediathekview, why does it need access to .java?
For the tor browser profile, adding `machine-id` certainly doesn't hurt... but since `private-etc` only contains fonts, machine-id won't end up having any effect (since both the real-life and spoofed /etc/machine-id file are invisible to the sandbox). 😄

SpotComms:
@Fred-Barclay https://github.com/mediathekview/MediathekView is written in Java

also as a note .java is used for caching fonts and for preference storage (not commonly utilized though)

smitsohu:
@Fred-Barclay Hahaha, you are right! I can drop the commit again.
Regarding the Mediathekview, also see #1377

Fred-Barclay:
Merged, thanks!

PR #1443, opened 2017-08-08
Automatically build each commit on Travis CI and upload

probonopd:
Currently, no binaries of recent versions are available for download.

This PR, when merged, will compile this application on [Travis CI](https://travis-ci.org/) upon each `git push`, and upload it to a temporary download URL on transfer.sh (available for 14 days). The download URL is toward the end of each Travis CI build log of each build (see below for how to set up automatic uploading to your GitHub Releases page).

For this to work, you need to enable Travis CI for your repository as [described here](https://travis-ci.org/getting_started) __prior to merging this__, if you haven't already done so.

__Please note:__ Instead of storing AppImage builds temporarily for 14 days each on transfer.sh, you could use GitHub Releases to store the binaries permanently. This way, they would be visible on the Releases page of your project. This is what I recommend. See https://docs.travis-ci.com/user/deployment/releases/. If you want to do this for continuous builds, also see https://github.com/probonopd/uploadtool.

If you would like to see only one entry for the Pull Request in your project's history, then please enable [this GitHub functionality](https://help.github.com/articles/configuring-commit-squashing-for-pull-requests/) on your repo. It allows you to squash (combine) the commits when merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,5 @@ sudo: yes
 script:
   - ( cd firejail ; ./configure && make && sudo make install-strip )
   - find /usr/local
+  - ( cd /usr/local , tar cfvj * firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2 . )
+  - ls
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: c
+dist: trusty
+sudo: yes
+
+script:
+  - ( cd firejail ; ./configure && make && sudo make install-strip )
+  - find /usr/local
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,6 @@ sudo: true
 script:
   - ( cd firejail ; ./configure --prefix=/usr && make && sudo make install-strip DESTDIR=$(readlink -f appdir) )
   - find appdir/
-  - ( cd appdir/ , tar cfvj firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2 * )
-  - curl --upload-file ./appdir/firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2 https://transfer.sh/firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2
+  - ( cd appdir/ , tar cfvj ../firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2 * )
+  - ls
+  - curl --upload-file ./firejail-*.tar.bz2 https://transfer.sh/firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: c
 dist: trusty
-sudo: false
+sudo: true
 
 script:
   - ( cd firejail ; ./configure --prefix=/usr && make && sudo make install-strip DESTDIR=$(readlink -f appdir) )
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 sudo: true
 
 script:
-  - ( cd firejail ; ./configure && make --prefix=/usr && sudo make install && make test )
-  - ( cd firejail ; ./configure --prefix=/usr && make && sudo make install-strip DESTDIR=$(readlink -f appdir) )
+  - ( cd firejail ; ./configure --prefix=/usr && make && sudo make install && make test )
+  - ( cd firejail ; sudo make install-strip DESTDIR=$(readlink -f appdir) )
   - ( cd appdir/ ; tar cfvj ../firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2 . )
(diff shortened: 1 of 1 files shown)


SpotComms:
> Currently, no binaries of recent versions are available for download.

fwiw there is --git-install

> Download, compile and install mainline git version of Firejail from the official repository on GitHub.  The software is installed in /usr/local/bin, and takes precedence over the (old) version installed in /usr/bin. If for any reason the new version doesn't work, the user can uninstall it using --git-uninstall command and revert to the old version.

probonopd:
> fwiw there is --git-install

Never heard of it. Is it documented somewhere?

SpotComms:
In the man pages since I think 0.9.46?

probonopd:
Don't think it's needed here.

reinerh:
Would it be possible to instead build .deb and .rpm packages? I'm not sure if there are users that just unpack tarballs in their root directory (which for example wouldn't clean up removed files, and overwrite configuration/profiles etc.).

As this is meant to be CI, could you also run the test suite after the build (make test)?

probonopd:
Sure it would be possible to create packages but then we have to think about the policies of the different distributions and I find that very cumbersome. However it should not be a problem for someone who knows how to package deb and rpm to change the script accordingly.

I am adding `make test`.

reinerh:
There are already "deb" and "rpms" make targets that you could use.

.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: c
+dist: trusty
```
Fred-Barclay: I'm not very familiar with the Travis CI build process, but would this limit the builds in any way to running well only on a Ubuntu 14.04 base?

probonopd:
This means that the resulting binaries should run on distributions from 2014 and later. It is unlikely to work on older distributions, but newer ones should be OK.

.travis.yml
probonopd: This means that the resulting binaries should run on distributions from 2014 and later. It is unlikely to work on older distributions, but newer ones should be OK.

probonopd:
It seems like `make test` works only if firejail is installed to `/`. Is that true?

reinerh:
It should also work if it's installed in a directory in $PATH.

.travis.yml
Ferroin: As @probonopd mentioned, it should run on almost anything post 2014.  I've only ever seen issues in cases like this when dealing with either 'enterprise' distros (where the user should just use the packaged version or build it themself locally), or tight embedded distros with really aggressive hardening like Alpine.

probonopd:
It seems to be looking for profiles in `/etc` though.

.travis.yml
Fred-Barclay: Ok,  so that would cut out users running CentOS 6 or lower (or possibly 7.0 or lower?) or Debian Wheezy, right?. That's not necessarily problematic as @Ferroin points out.

probonopd: Yes. I am also not proposing this as a general distribution method, more as something to test the latest build without having to compile it. I could also use it for https://github.com/AppImage/AppImageHub

Fred-Barclay:
I have to admit I'm a little uneasy about automated builds. Travis CI seems to be well-respected, but is it something we can rely on to produce stable, secure builds? If a faulty build is created and easily available, how to we communicate to the chaps who've installed it that it needs to be fixed?

What sort of security risks/advantages are there to opening the repo to a third-party build service? Could the building process be manipulated by someone who's not affiliated with either firejail or Travis CI? This is a sandbox project after all - build security is something we need to consider carefully.

If it's triggered on each `git push`, does that introduce overhead or prevent us from pushing again until the current build is finished?

Sorry to be a bit on the negative side. 😄 We can't merge anyway until @netblue30 enables the Travis CI.

probonopd:
It may be a matter of personal taste but I would trust a build from Travis CI more than a binary that I have no audit trail for, e.g., a deb from SourceForge. With Travis CI, at least I can exactly see what git commit was compiled in which environment, have a build log, etc.

Another option, of course, would be using the openSUSE Build Service.

probonopd:
Tests are running now.

Ferroin:
Also, with a bit of work, you can integrate Travis to get an auto build of newly created pull requests, which means you both have verification that a PR builds, and have an easy option to point bug reporters at something they can test without having to build it locally themselves.

As far as security, I'd personally agree with @probonopd msot of the time.  That's actually part of the reason I run Gentoo on my systems, because it gives me an easy to see audit trail back to the source level. 

For the specific aspect of potential bad builds, what I would recommend to users is to use your distro's packaged version, or keep up-to-date as much as possible if you're going to use the upstream version.  IOW, give the same advice that is given now.

probonopd:
> Also, with a bit of work, you can integrate Travis to get an auto build of newly created pull requests, which means you both have verification that a PR builds, and have an easy option to point bug reporters at something they can test without having to build it locally themselves.

No extra work needed, should work with my PR.

Just remember to enable Travis CI for your repository as [described here](https://travis-ci.org/getting_started) __prior to merging this__.

Ferroin:
Huh, I hadn't realized it was that easy.

netblue30:
All set, thanks!

PR #1444, opened 2017-08-08
Harden 18 profiles using private-bin

SpotComms:
What is the proper way to handle python versions?
- Chromium on Arch no longer launches via a perl script [1](https://git.archlinux.org/svntogit/packages.git/commit/trunk?h=packages/chromium&id=a2d61887394fd1fec63d9fe5ba0d2c07325164e) [2](https://github.com/foutrelis/chromium-launcher/commit/9d112db0d19e581ba816cb552a527b5b6a4d9f2f) [3](https://github.com/foutrelis/chromium-launcher/commit/44fa2a2b57dfd3f2bc9f2bf17e738ea964f2dce2)
- strings will be the first profile with private-lib

### DIFF
--- a/etc/pdfsam.profile
+++ b/etc/pdfsam.profile
@@ -25,6 +25,7 @@ protocol unix
 seccomp
 shell none
 
+private-bin pdfsam,sh,bash,java,archlinux-java,grep,awk,dirname,uname,which,sort,find,readlink,expr,ls,java-config
 private-dev
 private-tmp
 
--- a/etc/apktool.profile
+++ b/etc/apktool.profile
@@ -24,6 +24,7 @@ protocol unix
 seccomp
 shell none
 
+private-bin apktool,bash,java,dirname,basename,expr
 private-dev
 
 noexec ${HOME}
--- a/etc/arm.profile
+++ b/etc/arm.profile
@@ -32,7 +32,7 @@ shell none
 tracelog
 
 disable-mnt
-# private-bin arm,tor,sh,python2,python2.7,ps,lsof,ldconfig
+# private-bin arm,tor,sh,bash,python2,python2.7,ps,lsof,ldconfig
 private-dev
 private-etc tor,passwd
 private-tmp
--- a/etc/baobab.profile
+++ b/etc/baobab.profile
@@ -24,6 +24,7 @@ protocol unix
 seccomp
 shell none
 
+private-bin baobab
 private-dev
 private-tmp
 
--- a/etc/bless.profile
+++ b/etc/bless.profile
@@ -25,6 +25,7 @@ protocol unix
 seccomp
 shell none
 
+# private-bin bless,sh,bash,mono
 private-dev
 private-etc fonts,mono
 private-tmp
--- a/etc/chromium.profile
+++ b/etc/chromium.profile
@@ -31,6 +31,7 @@ nogroups
 notv
 shell none
 
+# private-bin chromium,chromium-browser,chromedriver
 private-dev
 # private-tmp - problems with multiple browser sessions
 
--- a/etc/chromium.profile
+++ b/etc/chromium.profile
@@ -11,8 +11,7 @@ noblacklist ~/.config/chromium-flags.conf
 noblacklist ~/.pki
 
 include /etc/firejail/disable-common.inc
-# chromium is distributed with a perl script on Arch
-# include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-programs.inc
 
 mkdir ~/.cache/chromium
--- a/etc/dex2jar.profile
+++ b/etc/dex2jar.profile
@@ -25,6 +25,7 @@ protocol unix
 seccomp
 shell none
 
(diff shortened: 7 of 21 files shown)


etc/pdfsam.profile
```diff
@@ -14,7 +14,7 @@ include /etc/firejail/disable-programs.inc
 
 caps.drop all
 net none
-no3d
+# no3d
```
Fred-Barclay: What happens when no3d is enabled?

VladimirSchowalter20:
I tested chromium on Archlinux even with private-bin option enabled and it seems to work.

Ferroin:
@SpotComms Unfortunately, handling of Python versions is dependent on both the distro (some default to Python 3, some to Python 2, and a few (Alpine for example) default to no Python at all) and software dependent (some stuff calls the required interpreter directly, some use `env` to ensure they get the system version, and some may use special distro specific methods (Gentoo for example has a piece of software called `python-exec` that handles all of this).  If you just copy in all of `python`, `python?`, and `python?.*` (shell expressions, not standard regexes) and `env` that should cover most cases.

etc/pdfsam.profile
SpotComms: The programs works fine it just throws some warnings, I've added it back.

Fred-Barclay:
Unfortunately I don't know of a good way to handle private-bin and python. There are just so many different configurations!

SpotComms:
@Fred-Barclay should I just comment out private-bin for all the python ones? I mean currently no profiles seem to really have it enabled. I do really want arm to have it however.

Fred-Barclay:
> I do really want arm to have it however.
I think since arm uses python 2 which won't advance beyond 2.7, we can keep private-bin if you add `python` to the current private-bin list.

etc/less.profile
```diff
@@ -16,6 +16,7 @@ novideo
 shell none
 tracelog
 
+private-bin less
```
reinerh: There is currently a thread (#1450), where someone was using external filters for less for syntax highlighting.
I would guess that also for viewing compressed files external tools (like gzip etc.) are necessary.
Perhaps add a comment that mentions this?

etc/pdfsam.profile
```diff
@@ -24,6 +24,7 @@ protocol unix
 seccomp
 shell none
 
+private-bin pdfsam,sh,bash,archlinux-java,grep,awk,dirname,uname,which,sort,find,readlink,expr,ls,java-config
```
reinerh: What is archlinux-java?
I would expect that for all other distributions `java` is required for it to work. Can you please add that?
Has archlinux really no plain java command?

Ferroin:
@Fred-Barclay The problem with that is that `python` is not universally a 2.x series python interpreter.  On at least Arch Linux and it's derivatives, it's a Python 3 interpreter (`python2` is the Python 2.x interpreter name, and that isn't installed by default either), and on Buildroot-based systems it's whatever version of Python was picked at build time (because they only support having one python version).

etc/less.profile
Ferroin: Appropriate compression tools are required if you're viewing compressed files directly.

SpotComms: I've disabled private-bin in less.

etc/pdfsam.profile
SpotComms: archlinux-java is merely a tool for managing multiple jvm installs (7, 8, 9), and I forgot to actually add 'java'. It still worked though because java is in /usr/lib.

Fred-Barclay:
@Ferroin right, but since arm only uses python2.7, including `python` along with `python2` and `python2.7` in private-bin should have no effect on systems where python defaults to v3. Or am I looking at it wrong?

Ferroin:
I think you're right, but depending on the system and the default Python version, `python` may end up being a dangling symlink inside the jail (for example, Gentoo lets you select your default Python version, and `python`, `python2`, and `python3` are all symlinks pointing to the respective defaults).  Worst case, you end up giving the jailed app another python version to probe for vulnerabilities.

Fred-Barclay:
My thought is that since private-bin doesn't follow symlinks by default, if python is a symlink to a non-existent (inside the sandbox) python version, it's useless for attack. Is this wrong?

Ferroin:
That is correct, but it is still an information leak (it implies that the system's default Python version is not the version in the sandbox).  The worst case I gave was accounting for the possibility that `python` is a hard link or separate copy of the preferred interpreter.  Either way, this has the potential to be problematic with any similar language that offers multiple concurrently installed versions.

netblue30:
I put a fix in git for #1450. It turns out we already have a --writable-var-log command line option and profile command to disable the sandboxing of /var/log. Also, let's not add private-bin to the profile, ~/.lessfilter seems to be quite popular with power users, and there is no way to tell what they have in the script.

SpotComms:
@Fred-Barclay anything blocking this from being merged?

Fred-Barclay:
@SpotComms yeah, laziness on my part. 😆 
Sorry about that - merged, and thanks!

PR #1445, opened 2017-08-09
comment private-dev in VLC

smitsohu:
it unfortunately breaks [DVB](https://en.wikipedia.org/wiki/Digital_Video_Broadcasting) (devices are registered under /dev/dvb/...)

### DIFF
--- a/etc/vlc.profile
+++ b/etc/vlc.profile
@@ -22,7 +22,8 @@ seccomp
 shell none
 
 private-bin vlc,cvlc,nvlc,rvlc,qvlc,svlc
-private-dev
+# DVB (Digital Video Broadcast) devices stop working with private-dev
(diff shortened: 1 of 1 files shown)


smitsohu:
Question: Shouldn't it be possible to replicate `private-dev` with whitelisting? /dev/dvb could then easily be added.....

VladimirSchowalter20:
I think it's better to modify `private-dev ` so it will accept additions like:

`private-dev dvb`

This way it will allow defaults + dvb and so on. Same as private-bin do.

Fred-Barclay:
With eabdc943c71efee1c1fd77d37122898227dbc6a8 by @startx2017, we shouldn't need this anymore, but thanks for your contributions!

PR #1448, opened 2017-08-09
Match RPM license tag with license set in COPYING

da2x:
Known license tag should be `GPLv2+` ([full list](https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#SoftwareLicenses)).


reinerh:
Merged. Thanks!

PR #1451, opened 2017-08-10
more novideo options, enhanced mediathekview

smitsohu:
- A couple of novideo additions
- While VLC is the default, other media players can be used in Mediathekview also. The current list might be not complete yet, but these are all players with a Firejail profile from which I know they work with Mediathekview.

### DIFF
--- a/etc/mediathekview.profile
+++ b/etc/mediathekview.profile
@@ -5,9 +5,14 @@ include /etc/firejail/mediathekview.local
 # Persistent global definitions
 include /etc/firejail/globals.local
 
+noblacklist ~/.config/mpv
+noblacklist ~/.config/smplayer
+noblacklist ~/.config/totem
 noblacklist ~/.config/vlc
 noblacklist ~/.java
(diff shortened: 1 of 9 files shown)


etc/mediathekview.profile
```diff
@@ -23,6 +28,7 @@ protocol unix,inet,inet6
 seccomp
 tracelog
 
+disable-mnt
```
SpotComms: This can save/playback video files yea? Won't some people have that all on another drive?

smitsohu: @SpotComms It's for TV on demand, so everything is supposed to happen via Internet. Additionally, the default VLC closes automatically at the end of each film/movie. 

But still: It is of course not impossible to take a sandboxed, crippled, TV playing media player and try to access a local movie database. So yes, this change is debatable and has its con. It would make the profile more robust if we dropped disable-mnt. Maybe I was not abstracting enough from the way I use it :)

Fred-Barclay:
Merged, thanks!

PR #1460, opened 2017-08-11
fix tuxguitar comments

smitsohu:
sorry for the commit mess, thanks for your patience. Like this it is right finally.


netblue30:
Merged, thanks.

PR #1461, opened 2017-08-11
Fix notv placement

SpotComms:
- Fix notv placement
- Fix "include include" in mate-calculator.profile a1cb05b0304c0f0fb68582a34fcfcfe1d4b505f0
- Fix "nosound\nnosound" in 7z.profile
- Fix options out of alphabetical order in default.profile

### DIFF
--- a/etc/7z.profile
+++ b/etc/7z.profile
@@ -12,7 +12,7 @@ ignore noroot
 net none
 no3d
 nosound
-nosound
+notv
 novideo
 shell none
 tracelog
@@ -20,4 +20,3 @@ tracelog
 private-dev
 
 include /etc/firejail/default.profile
-notv
--- a/etc/0ad.profile
+++ b/etc/0ad.profile
@@ -27,6 +27,7 @@ netfilter
 nogroups
 nonewprivs
 noroot
+notv
 novideo
 protocol unix,inet,inet6
 seccomp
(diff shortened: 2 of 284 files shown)


etc/7z.profile
```diff
@@ -13,11 +13,11 @@ net none
 no3d
 nosound
 nosound
```
reinerh: Just noticed here "nosound" is duplicated. Can you please remove the line?

etc/Cyberfox.profile
```diff
@@ -2,5 +2,4 @@
 # This file is overwritten after every install/update
 
 
-# Redirect
```
reinerh: Why remove those? Were they not recently added intentionally?

SpotComms: Alias profiles are already marked with "Firejail profile alias for x"

etc/7z.profile
SpotComms: Fixed

etc/Cyberfox.profile
netblue30: Add it back.  I need "redirect" as a marker to add new commands globally, and some redirected profiles are not aliases. For example thunderbird redirects to firefox. I'll be bringing in some more, thanks!

SpotComms: @netblue30 Added back.

Fred-Barclay:
Merged, thanks!

PR #1465, opened 2017-08-13
fix smplayer for mpv

pizzadude:
this fixes smplayer if you use the mpv backend


SpotComms:
Shouldn't you also add "noblacklist ${HOME}/.config/mpv" ?

pizzadude:
I don't have anything in ~/.config/mpv

netblue30:
merged, thanks.

PR #1473, opened 2017-08-17
Dino: Fix file downloads

SpotComms:


etc/dino.profile
```diff
@@ -5,6 +5,7 @@ include /etc/firejail/dino.local
 # Persistent global definitions
 include /etc/firejail/globals.local
 
+noblacklist ${DOWNLOADS}
```
smitsohu: Hey @SpotComms :-) Are you having plans for the corresponding `blacklist ${DOWNLOADS}`?

SpotComms: What do you mean?

PR #1475, opened 2017-08-18
some fixes and enhancements

smitsohu:
A note regarding Skanlite: The default seccomp filter seems to break device discovery for Skanlite on Debian Stretch. Also I had mixed results with network namespaces. What worked for me on all platforms including Fedora is `protocol unix,netlink`, which is why I suggest bringing the option back.

### DIFF
--- a/etc/skanlite.profile
+++ b/etc/skanlite.profile
@@ -13,6 +13,7 @@ include /etc/firejail/disable-programs.inc
 
 caps.drop all
 # net none
+netfilter
 nodvd
 nogroups
 nonewprivs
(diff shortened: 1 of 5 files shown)


etc/skanlite.profile
```diff
@@ -12,18 +12,20 @@ include /etc/firejail/disable-passwdmgr.inc
 include /etc/firejail/disable-programs.inc
 
 caps.drop all
-netfilter
+# net none
```
Fred-Barclay: Since net none is commented, should we keep in netfilter?

smitsohu: Not sure how much we can gain from it (it looks skanlite crashes on some setups if the abstract dbus socket is absent, so if it doesn't crash `net none` should be the safest option). On the other hand it certainly doesn't harm to keep it in.

Fred-Barclay:
Merged, thanks!

PR #1482, opened 2017-08-19
Update waterfox.profile

hawkeye116477:
Waterfox 55 has new profile location


SpotComms:
Does it automatically move it? If so shouldn't we leave the old path for a while?

hawkeye116477:
@SpotComms It move when user check that he want to move. Ok, fixed.

Fred-Barclay:
Merged, thanks!

PR #1483, opened 2017-08-19
fix simple-scan

smitsohu:
Fix for broken device discovery. 
Mirrors recent changes for skanlite (#1475), which uses the same backend (SANE).
As tested on Debian Jessie.

### DIFF
--- a/etc/simple-scan.profile
+++ b/etc/simple-scan.profile
@@ -20,9 +20,11 @@ nonewprivs
 noroot
 nosound
 notv
-protocol unix,inet,inet6
+novideo
+protocol unix,inet,inet6,netlink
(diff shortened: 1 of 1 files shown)


etc/simple-scan.profile
```diff
@@ -20,9 +20,11 @@ nonewprivs
 noroot
 nosound
 notv
-protocol unix,inet,inet6
+novideo
+protocol unix,inet,inet6,netlink
```
Fred-Barclay: Hi @smitsohu 
Is netlink necessary here?

smitsohu: It won't find the scanner without netlink. I guess that will be the same for all SANE frontends, irrespective of platform.

Fred-Barclay:
Merged, thanks!

PR #1485, opened 2017-08-20
Add pass to common blacklist

jelford:
`pass` is a password manager that keeps files under `~/.password-store` by default.
See http://www.passwordstore.org/ for more info


reinerh:
Merged. Thanks!

PR #1489, opened 2017-08-22
profile fixes and enhancements

smitsohu:
- an attempt to bring the forks gedit-pluma-xed, evince-atril-xreader, eog-eom-xviewer, totem-xplayer, engrampa-file-roller to the same level.
`net none` makes the settings of various programs immutable, which is both a bug and a feature (dconf isolation). This pull request would make the feature opt-in. If we instead leave it as it is now, maybe we should at least put a comment in the profiles.
- added `memory-deny-write-execute` to audacious, audacity, qpdfview. As it seems to work for some, may I also propose putting this option back (but commented) in the vlc profile?
- commented recently added `novideo` in skanlite, simple-scan profiles. Apparently these scanning tools can make snapshots with the webcam, too.
- fixed some more `nodvd`, `notv` options
- EDIT: vlc and cvlc profiles are harmonized again.

### DIFF
--- a/etc/vlc.profile
+++ b/etc/vlc.profile
@@ -25,6 +25,7 @@ private-bin vlc,cvlc,nvlc,rvlc,qvlc,svlc
 private-dev
 private-tmp
 
+# mdwe is disabled due to breaking hardware accelerated decoding
 # memory-deny-write-execute
 noexec ${HOME}
 noexec /tmp
--- a/etc/vlc.profile
+++ b/etc/vlc.profile
@@ -25,5 +25,6 @@ private-bin vlc,cvlc,nvlc,rvlc,qvlc,svlc
 private-dev
 private-tmp
 
+# memory-deny-write-execute
 noexec ${HOME}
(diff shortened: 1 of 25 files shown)


etc/vlc.profile
```diff
@@ -25,5 +25,6 @@ private-bin vlc,cvlc,nvlc,rvlc,qvlc,svlc
 private-dev
 private-tmp
 
+# memory-deny-write-execute
```
SpotComms: Could you add a note like:
`# mdwe is disabled due to breaking hardware accelerated decoding`
and if you want also add noexec to cvlc.profile

smitsohu: Sure! Should I also put a warning note about mdwe into the cvlc profile? I can't test it on my current machine.

SpotComms: Yes, cvlc is just vlc without an interface. It should be commented as well.

netblue30:
Merged, thanks.

Issue #1346, opened 2017-06-23
private-bin and shells


reinerh:
Hi,

while investigating a failing tar.exp test, I found out that there is an issue with private-bin and shells.
On Debian/Ubuntu tar.exp fails, because bzip2 can't be executed. This probably fails because it can't execute the default shell. In the tar profile it allows "sh", but not for example "bash", "dash" etc, so they are not available in the private bin.
And indeed, after I added dash (my /bin/sh points to dash) to the private-bin line, the test was passing.
Adding all shells to all private-bin lines that require a shell is probably no good solution.

I'm wondering now if the symlink should be resolved and the destination binary be also copied to the private bin?

netblue30:
I added sh,bash,dash in several profiles: https://github.com/netblue30/firejail/commit/81b61d55a3174189d3c810f645f81f0ef48f7db0

I think that's the easiest fix for now. Some time ago somebody put in a patch for private-bin to follow the links. I had to disable it after some time, apparently on some distributions firefox is a symbolic link in /usr/bin directory pointing to the place where firefox was installed, and it was breaking firefox.

I disabled the patch, currently it is a config option in /etc/firejail/firejail config:
`````
# Follow symlink for private-bin command.
# Disabled by default
# follow-symlink-private-bin no
`````

reinerh:
Okay, thanks. I wasn't aware of the config option.
Let's keep your fix then for now.

PR #1493, opened 2017-08-24
Fix #1492

### LINKED ISSUE #1492: Xonotic doesn't launch with firejail
I run Ubuntu MATE 17.04 with the latest firejail built from git.

I have Xonotic installed from playdeb, but I had the same issue with the "regular" version from Xonotic's website.

If I do "locate xonotic" it resides in:
/usr/bin/xonotic
/usr/games/xonotic
/usr/lib/games/xonotic
/usr/share/games/xonotic

I tried adding "noblacklist /usr/share/games/xonotic" to the firejail profile but that didn't do anything.
I also tried adding "xonotic" to private-bin.

This is the output of "firejail xonotic"
`Reading profile /etc/firejail/xonotic.profile
Parent pid 32521, child pid 32522
Warning: skipping asound.conf for private /etc
Child process initialized in 200.18 ms

`

And here's the output with the --debug flag.
http://paste.ubuntu.com/25380452/

I tried disabling seccomp for the xonotic profile, that did nothing.

Any ideas?

SpotComms:
Xonotic fixes
- Fix sound
- Fix for playdeb version
- Fix for official .zip version
- Probably fixes other weird packaged versions

also fix similar potential issues in #1444 as described in #1346

### DIFF
--- a/etc/arm.profile
+++ b/etc/arm.profile
@@ -33,7 +33,7 @@ shell none
 tracelog
 
 disable-mnt
-# private-bin arm,tor,sh,bash,python2,python2.7,ps,lsof,ldconfig
+# private-bin arm,tor,sh,bash,dash,python2,python2.7,ps,lsof,ldconfig
 private-dev
 private-etc tor,passwd
 private-tmp
--- a/etc/apktool.profile
+++ b/etc/apktool.profile
@@ -25,7 +25,7 @@ protocol unix
 seccomp
 shell none
 
-private-bin apktool,bash,java,dirname,basename,expr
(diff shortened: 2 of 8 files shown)


etc/arm.profile
```diff
@@ -33,7 +33,7 @@ shell none
 tracelog
 
 disable-mnt
-# private-bin arm,tor,sh,bash,python2,python2.7,ps,lsof,ldconfig
+# private-bin arm,tor,sh,bash,dash,python2,python2.7,ps,lsof,ldconfig,sh
```
reinerh: sh was already included in this line. :)

netblue30:
Merged, thanks.

PR #1494, opened 2017-08-24
Allow private-bin parameters to be absolute paths

### LINKED ISSUE #1133: [discussion] allow to specify full path to the binary in private-bin
`private-bin` doesn't allow to specify full path to the binary, but only it's name:
```Error: "/bin/env" is an invalid filename```
But it can lead to issues like this:
```
$ firejail --join=tg ls /bin/env -l
lrwxrwxrwx 1 65534 65534 8 мар  7 23:48 /bin/env -> /bin/env
```
Here, it is happened because firejail copied `env` from `/usr/bin`, where it resides for compatibility while actually it is installed (like entire coreutils) to /bin.

And, mostly ***destructive***: it can even be situation when it will copy the wrapper from `/usr/local/bin` (which have bigger priority than even `/usr/bin`), but not the target binary which the wrapper actually calls...

INB4: yes, it can (partially) be fixed by allowing to follow symlinks, but it:
1) will cause #1129
2) it can be situations, when user would like to directly specify which binary should be used (from less-priority PATH part, for example, or even custom wrapper)

### PR DESCRIPTION
**Pull request by LaurentGH**
With Ubuntu 16.04, /usr/bin/which is a symlink to /bin/which. So, using "private-bin which" finds "which" in /usr/bin and adds the symlink to "which" in /bin mapped directory. The /bin directory thus contains a symlink named "which" pointing to "/bin/which" (itself). This creates a symlink loop, and does not work. In order to solve this, the full path can now be used, such as "private-bin /bin/which". This fixes #1133.

### DIFF
--- a/src/firejail/fs_bin.c
+++ b/src/firejail/fs_bin.c
@@ -95,20 +95,34 @@ static char *check_dir_or_file(const char *name) {
 }
 
 static void duplicate(char *fname, FILE *fplist) {
-	if (*fname == '~' || *fname == '/' || strstr(fname, "..")) {
+	if (*fname == '~' || strstr(fname, "..")) {
 		fprintf(stderr, "Error: \"%s\" is an invalid filename\n", fname);
 		exit(1);
 	}
 	invalid_filename(fname);
 
-	char *path = check_dir_or_file(fname);
-	if (!path)
-		return;
-
-	// expand path, just in case this is a symbolic link
 	char *full_path;
-	if (asprintf(&full_path, "%s/%s", path, fname) == -1)
-		errExit("asprintf");
+	if (*fname == '/') {
+		// If the absolute filename is indicated, directly use it. This
+		// is required for the following three cases:
+		//  - if user's $PATH order is not the same as the above
+		//    paths[] variable order
+		//  - if for example /usr/bin/which is a symlink to /bin/which,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by netblue30** [comment]
All merged, thanks.

File: src/firejail/fs_bin.c [hunk 1]
```diff
@@ -95,20 +95,34 @@ static char *check_dir_or_file(const char *name) {
...
+		// is required for the following three cases:
+		//  - if user's $PATH order is not the same as the above
+		//    paths[] variable order
+		//  - if for example /usr/bin/which is a symlink to /bin/which,
+		//    because in this case the result is a symlink pointing to
+		//    itself due to the file name being the same.
+		//  - if user wants to add a binary, which is not in the above
+		//    paths[] variable
+		if (asprintf(&full_path, "%s", fname) == -1)
+			errExit("asprintf");
```
msva: I'd change the error messages (here and below) to something useful, so user can understand what's wrong on debugging.

LaurentGH: Hello,
A grep command {grep -rsF 'errExit("asprintf")' * | wc -l} returned 244 occurrences of errExit("asprintf") in 52 files of the project source code.
So, it would be appropriate to do it for the 244 occurrences. However, people who had forked any of these 52 files will have to update their fork. Such a big change should be done/approved by official Firejail maintainers.

PR #1496, opened 2017-08-25
Fix MulitMC5 and Xonotic

SpotComms:
- MultiMC5 now has a 3rd path it uses
- Fix sound in Xonotic mentioned in #1492 
- Fix Lollypop not having certificate access


netblue30:
Merged, thanks.

PR #1501, opened 2017-08-29
Tweak itch.io profile

SpotComms:
I've not used itch.io but should we also remove whitelist? Do games actually store their configs in ~/.config/itch, or is itch purely a launcher and games can store wherever like Steam?

### DIFF
--- a/etc/itch.profile
+++ b/etc/itch.profile
@@ -5,14 +5,18 @@ include /etc/firejail/itch.local
 # Persistent global definitions
 include /etc/firejail/globals.local
 
-noblacklist ~/.config/itch
+# itch.io has native firejail/sandboxing support bundled in
+# See https://itch.io/docs/itch/using/sandbox/linux.html
+
+noblacklist ${HOME}/.config/itch
 
 include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-passwdmgr.inc
 include /etc/firejail/disable-programs.inc
 
-whitelist ~/.config/itch
-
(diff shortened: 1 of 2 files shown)


etc/itch.profile
```diff
@@ -22,12 +23,13 @@ nogroups
...
+novideo
 protocol unix,inet,inet6,netlink
 seccomp
 shell none
 
 private-dev
 private-tmp
 
-noexec ${HOME}
+#noexec ${HOME}
```
Fred-Barclay: Looks good! I've not used itch.io either, but do we need to comment out `noexec ${HOME}`?

SpotComms: Assuming its a launcher, its downloading games and needs to be able to run them.

smitsohu:
I also have not used itch, but I looked a bit on their page and I found this: https://itch.io/docs/itch/using/sandbox/linux.html

If I understand that right, shouldn't there be a comment in the profile that itch.io already ships Firejail and a Firejail profile?

SpotComms:
@smitsohu oh wow, thats nice... except how many people have /home mounted **without** `nosuid`?  But yes we should probably add a comment and also not let firecfg enable it by default.

Ferroin:
@SpotComms Well, to start with, anybody who has `/home` as part of the root filesystem, which I would guess is a significant majority of desktop Linux users considering that that's default behavior in almost every major distribution's installer, and is trivial to do with those that don't have that behavior.

netblue30:
merged, thanks.

pizzadude:
I wrote the itch.io profile because the itch.io built-in firejail doesn't work for me.

PR #1503, opened 2017-08-29
enhance and fix profiles (mostly novideo additions)

smitsohu:
- a bunch of `novideo` additions.
- audacious fails to launch due to `private-bin`, and interestingly it is not always reproducible. I was not successful in debugging this further. Maybe we disable it for the time being? This is the error message:
```
ERROR plugin-init.cc:147 [start_required]: No output plugin found.
(Did you forget to install audacious-plugins?)

```
- feh uses jpegexiforient and jpegtran for rotating images.
- Firefox supported KF5 Okular better than KDE4 Okular. This pull request brings both to the same level.

### DIFF
--- a/etc/amarok.profile
+++ b/etc/amarok.profile
@@ -17,6 +17,7 @@ nogroups
 nonewprivs
 noroot
 notv
+novideo
 protocol unix,inet,inet6
 # seccomp
 shell none
--- a/etc/audacious.profile
+++ b/etc/audacious.profile
@@ -24,8 +24,10 @@ seccomp
 shell none
 tracelog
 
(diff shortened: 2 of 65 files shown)


smitsohu:
To make reviewing a bit easier: This pull request adds `novideo` to a manual selection of profiles which were released in 0.9.46 with `private-dev` enabled, that is before commit 1328d523ed62dfcdb205580ce7c126813b592cb7.
Exceptions are: caja, nautilus, cmus, curl, enchant, mupen64plus and vim

netblue30:
All set, thanks.

PR #1511, opened 2017-08-31
improve servers, harden musescore

smitsohu:
- harden server profiles
- the default `netfilter` seems to be not compatible with running a server in the sandbox, my proposal is to remove it from the dnsmasq profile
- fix copy/paste mistake in musescore profile

### DIFF
--- a/README
+++ b/README
@@ -409,7 +409,7 @@ smithsohu (https://github.com/smitsohu)
 	- lots of profile hardening and fixes
 	- added MuseScore profile
 	- fixed device discovery for simple-scan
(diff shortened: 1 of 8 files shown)


SpotComms:
Why that last commit? Why would dnsmasq or unbound need access to other drives?

smitsohu:
@SpotComms You are right, they shouldn't need access to other drives, that's why I had `disable-mnt` inside first... it also seems as if they work ok.
But I want to check again if I overlooked something regarding the chroot, that's all :)

netblue30:
Merged, thanks.

PR #1515, opened 2017-09-01
Fix Spotify #1513

### LINKED ISSUE #1513: [Spotify] Blacklisting too restrictive - firejail 0.9.48
Hi, spotify version 1.0.49.125 on ArchLinux with firejail 0.9.48 would not start and I got 
`/usr/bin/spotify: line 2: /opt/spotify/spotify: Permission denied`

commenting out `blacklist /opt` and appending a `whitelist /opt/spotify` to a local `~/.config/firejail/spotify.profile` did the trick.

SpotComms:
Fixes #1513

### DIFF
--- a/etc/spotify.profile
+++ b/etc/spotify.profile
@@ -6,10 +6,7 @@ include /etc/firejail/spotify.local
 include /etc/firejail/globals.local
 
 blacklist ${HOME}/.bashrc
-blacklist /boot
 blacklist /lost+found
-blacklist /opt
-blacklist /root
 blacklist /sbin
 blacklist /srv
 blacklist /sys


VladimirSchowalter20:
Maybe remove only `blacklist /opt`. There is very little probability that other dirs will be needed.

SpotComms:
Those are already blocked by default.

reinerh:
Is blacklisting /sbin needed with private-bin?

SpotComms:
@reinerh it doesn't seem so, but on my system /sbin is just a symlink to /bin
```
firejail --profile=spotify.profile bash
bash-4.4$ ls /sbin
bash  ls  sh
```

netblue30:
Merged, thanks.

Issue #1516, opened 2017-09-02
Can't start yandex-browser with firejail


larkvirtual:
yandex-browser is a browser based on chromium with additional functions. Located at https://browser.yandex.ru/

bash-4.4$ firejail yandex-browser
Reading profile /etc/firejail/default.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-passwdmgr.inc

** Note: you can use --noprofile to disable default.profile **

Parent pid 11332, child pid 11333
Child process initialized
Found ffmpeg: /usr/lib/chromium-browser/libffmpeg.so
	avcodec: 3756644
	avformat: 3752551
	avutil: 3616613
Ffmpeg version is OK! Let's use it.
[1:1:0902/092123.862079:ERROR:nacl_fork_delegate_linux.cc(316)] Bad NaCl helper startup ack (0 bytes)
Failed to generate minidump.
Parent is shutting down, bye...
bash-4.4$ firejail --version
firejail version 0.9.44.8

Compile time support:
	- AppArmor support is enabled
	- AppImage support is enabled
	- bind support is enabled
	- chroot support is enabled
	- file and directory whitelisting support is enabled
	- file transfer support is enabled
	- networking support is enabled
	- overlayfs support is enabled
	- private-home support is enabled
	- seccomp-bpf support is enabled
	- user namespace support is enabled
	- X11 sandboxing support is enabled

SpotComms:
Yandex Browser is Chromium based, so you should be able to just take chromium.profile, rename it to yandex-browser.profile and then update the paths.

```
mkdir -p .config/firejail
cd .config/firejail
cp /etc/firejail/chromium.profile yandex-browser.profile
nano yandex-browser.profile #update paths
firejail yandex-browser #should work
```

larkvirtual:
Thanks a lot!
Now it works.

PR #1519, opened 2017-09-02
Add a profile for Yandex Browser

SpotComms:
Closes #1517 and #1516 

The beta and stable versions seem to use the same executable name.


netblue30:
All merged, thanks!

PR #1524, opened 2017-09-03
firejail profile for smtube

pizzadude:
if you think private-bin needs to be there, feel free to edit the profile, and users would need to add their own players via the smtube.local file

### DIFF
--- a/smtube.profile
+++ b/smtube.profile
@@ -1,0 +1,34 @@
+# Firejail profile for smtube
+# This file is overwritten after every install/update
+# Persistent local customizations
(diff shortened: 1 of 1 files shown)


SpotComms:
I assume smtube just gets the video stream and passes it to a player? Could you add notv, nodvd, and novideo? and does nogroups need to be disabled?

pizzadude:
It's because I copied from the SMPlayer profile as a starting point, and that had nogroups for some reason. I fixed it now.

Fred-Barclay:
Merged, thanks!

PR #1526, opened 2017-09-03
tighten some capability sets further

smitsohu:


etc/wireshark.profile
```diff
@@ -12,7 +12,7 @@ include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-passwdmgr.inc
 include /etc/firejail/disable-programs.inc
 
-# caps.drop all
+caps.keep net_admin,net_raw
```
VladimirSchowalter20: On my system it needs `dac_override` as well.

smitsohu: Thank you! I will add it.

startx2017:
Thanks.

PR #1528, opened 2017-09-04
syscalls blacklisted twice

smitsohu:
Group `@module` is already included in group `@privileged`. Tiny fix for syscalls in `@module` being printed twice with `seccomp.print`


topimiettinen:
Thanks, merged.

PR #1538, opened 2017-09-07
Fix gitg diff not showing

### LINKED ISSUE #1537: gitg not showing diff
As the title.  
And how to exclude a command from jail?

SpotComms:
Fixes #1537. This should probably be backported to 0.9.50-bugfixes.

Old versions of gitg fail to display git diffs when mdwe is enabled.


Fred-Barclay:
Merged, thanks!

PR #1544, opened 2017-09-10
localhost mail fix for mutt

### LINKED ISSUE #1454: mutt crashes when using t-prot
For firejail version 0.9.48 in debian:

mutt is an email-client that allows for great extensibility which isn't accounted for in the default firejail profiles. In my case, I encountered this because I use an add-on package called 't-prot' which is a message display filter built around a perl script. Commenting out the line `include /etc/firejail/disable-devel.inc` from the default mutt profile fixed the crash. My guess is that the reason is that the include file was blacklisting the perl shell.

Because mutt is designed to be extended by any kind pf plugin, and because package `t-prot` is used by many mutt users, maybe the project should address this for the default mutt profile.

As a more general matter, if I've correctly understood the problem:

1] This will be an issue for other blacklisted shells eg. ruby, python

2] Maybe the `disable-devel.inc` file should be divided in two in order to account for run-time shell interpreters like perl. Another solution might be to keep the `disable-devel.inc` unchanged, but create a second include file to whitelist run-time shell interpreters.

smitsohu:
- ~~fixes #1454~~
- brings support for local mailbox
- makes blacklisting more robust

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -130,6 +130,7 @@ blacklist /var/run/screens
 blacklist /var/run/systemd
 blacklist /var/spool/anacron
 blacklist /var/spool/cron
+blacklist /var/spool/mail
 
 # etc
 blacklist /etc/anacrontab
--- a/etc/mutt.profile
+++ b/etc/mutt.profile
@@ -33,7 +33,7 @@ noblacklist ~/postponed
 noblacklist ~/sent
 
 include /etc/firejail/disable-common.inc
-# include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-passwdmgr.inc
 include /etc/firejail/disable-programs.inc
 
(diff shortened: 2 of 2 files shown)


Boruch-Baum:
Instead of including ALL of the contents of disable-devel.inc, why not divide that file at line #38, where it begins dealing with interpreted languages. That would maintain the blacklist for compilation and valgrind, etc.

Also, I see that disable-devel.inc does not include common development tools such as make, cmake, cscope. Also, while gdb is included, it has other forms for other environments, such as gdb-tui. For python, the list is missing ipython and pdb. I've probably left a lot out for those languages, and I don't know ruby or perl or go to know what's missing for those.

smitsohu:
@Boruch-Baum Thanks for your suggestion. While I agree that mutt and some other profiles could profit from splitting disable-devel.inc, I'd prefer waiting for some more opinions before writing a pull request.

Anyways, fixing local mail was my main motivation, if commenting disable-devel.inc turns out to be controversial, I take it out and leave it for later.

netblue30:
Merged, thanks.

PR #1548, opened 2017-09-14
goobox enhancements

smitsohu:
1) Permitting internet access fixes broken metadata retrieval (relevant for the rip functionality)
2) We can safely enable private-dev after the introduction of nodvd


Fred-Barclay:
Sounds good, thanks!

PR #1549, opened 2017-09-14
harden baloo, clementine

smitsohu:
updates the seccomp filter for baloo and clementine, to include all new blacklisted syscalls. 
seems to work well.


netblue30:
Merged, thanks.

PR #1554, opened 2017-09-16
harden corebird

smitsohu:


smitsohu:
worked with old but breaks newest version, needs more work

PR #1558, opened 2017-09-17
Set `shell none` for ssh-agent configuration

jelford:
It's common to invoke ssh-agent in shell login config, so if firejail launches it through a shell, that can lead to an infinite loop


netblue30:
Merged, thanks for the patch.

PR #1561, opened 2017-09-17
fix usage of STEAM_RUNTIME_PREFER_HOST_LIBRARIES=1 in steam.profile

soredake:
steam.sh executes `/sbin/ldconfig -XNv 2> /dev/null` when STEAM_RUNTIME_PREFER_HOST_LIBRARIES is not zero and fails in some cases (see https://github.com/ValveSoftware/steam-for-linux/issues/4768) because of blacklisted /sbin and missing ld.so.conf{,.d}

https://steamcommunity.com/groups/SteamClientBeta/announcements/detail/586991182161672256


netblue30:
Merged, thanks.

PR #1565, opened 2017-09-19
Some profiles

nyancat18:
* Fixed Profiles: Kdenlive, Shotcut

* New Profile: Cinelerra, Cliqz, Bluefish

* Some hardening for some profiles

### DIFF
--- a/etc/7z.profile
+++ b/etc/7z.profile
@@ -17,6 +17,7 @@ notv
 novideo
 shell none
 tracelog
+caps.drop all
 
 private-dev
 
--- a/etc/atom.profile
+++ b/etc/atom.profile
@@ -5,6 +5,8 @@ include /etc/firejail/atom.local
 # Persistent global definitions
 include /etc/firejail/globals.local
 
+noexec ${HOME}
+noexec /tmp
 noblacklist ~/.atom
 noblacklist ~/.config/Atom
 
@@ -23,6 +25,7 @@ notv
 novideo
 protocol unix,inet,inet6,netlink
 seccomp
+net none
 shell none
 
 private-dev
--- a/etc/calligra.profile
+++ b/etc/calligra.profile
@@ -21,9 +21,10 @@ novideo
 protocol unix
 seccomp
 shell none
+net none
 
 private-bin calligra,calligraauthor,calligraconverter,calligraflow,calligraplan,calligraplanwork,calligrasheets,calligrastage,calligrawords,dbus-launch
 private-dev
 
-noexec ${HOME}
+#noexec ${HOME}
 noexec /tmp
--- a/etc/evince.profile
+++ b/etc/evince.profile
@@ -28,6 +28,7 @@ protocol unix
 seccomp
 shell none
 tracelog
+net none
 
 private-bin evince,evince-previewer,evince-thumbnailer
 private-dev
--- a/etc/tar.profile
+++ b/etc/tar.profile
@@ -18,6 +18,7 @@ notv
 novideo
 shell none
 tracelog
+caps.drop all
 
 # support compressed archives
 private-bin sh,bash,dash,tar,gtar,compress,gzip,lzma,xz,bzip2,lbzip2,lzip,lzop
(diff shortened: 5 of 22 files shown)


etc/calligra.profile
```diff
@@ -21,9 +21,10 @@ novideo
 protocol unix
 seccomp
 shell none
+net none
 
 private-bin calligra,calligraauthor,calligraconverter,calligraflow,calligraplan,calligraplanwork,calligrasheets,calligrastage,calligrawords,dbus-launch
 private-dev
 
-noexec ${HOME}
+#noexec ${HOME}
```
SpotComms: What does calligra need to execute? Are there extensions/plugins?

etc/tar.profile
```diff
@@ -18,6 +18,7 @@ notv
 novideo
 shell none
 tracelog
+caps.drop all
```
reinerh: are the capabilities not required for setting certain file flags, owner etc.?

SpotComms:
- 7z, tar, unrar, and unzip include default.profile so adding `caps.drop all` isn't needed
- evince having `net none` is known to cause issues see https://github.com/netblue30/firejail/pull/1367#issuecomment-315770405
- Instead of creating a profile for openshot-qt why not create an alias/redirect to openshot.profile
- Also your ~/.gitconfig doesn't have your user/email set
- Could you please try and stick to the format and keep options in alphabetical order

etc/atom.profile
```diff
@@ -23,6 +25,7 @@ notv
 novideo
 protocol unix,inet,inet6,netlink
 seccomp
+net none
```
reinerh: does allowing inet(6) protocols still make sense with "net none"?

SpotComms: No, it would also break the extension/themes store

etc/7z.profile
```diff
@@ -17,6 +17,7 @@ notv
 novideo
 shell none
 tracelog
+caps.drop all
```
reinerh: are some capabilities not required for setting certain flags, file ownership etc.?

etc/atom.profile
nyancat18: we could put it optional :)

reinerh:
@SpotComms I wasn't aware that default.profile also drops caps.
I just confirmed that it breaks archiving tools run as root...
I think the setting is too invasive. Other opinions?

```
# firejail tar xf foo.tar 
tar: foo: Cannot change ownership to uid 1000, gid 1000: Operation not permitted
tar: Exiting with failure status due to previous errors
```

SpotComms:
@reinerh I'm actually surprised that hasn't been caught earlier. We can either add `ignore caps.drop` to them all or switch them all from default.profile to a generic compression one, the latter might be better.

Fred-Barclay:
@SpotComms @reinerh I rather like the idea of a generic compression profile. 👍

etc/7z.profile
netblue30: removed caps.drop all

etc/atom.profile
netblue30: optional net none (commented out)

etc/evince.profile
```diff
@@ -28,6 +28,7 @@ protocol unix
 seccomp
 shell none
 tracelog
+net none
```
netblue30: commented out, breaks AppArmor on Ubuntu systems

etc/tar.profile
netblue30: removed caps.drop all

netblue30:
Merged, thanks.

I'll bring some fixes in, I'll go with the list from @SpotComms: https://github.com/netblue30/firejail/pull/1565#issuecomment-330705836

And yes, we have to figure out some generic profile for compression tools!

SpotComms:
@netblue30 already fixed in def6f1ee2435a4ce4712fdd709ea9d14253547f0 and 04adc450151cc5107098ef2f555ad526ac9f632e
I think I addressed everything and a few others.

netblue30:
Cool! I also added some small fixes.

etc/evince.profile
nyancat18: could  you  add a comment?.

just like.

>#please uncomment if you dont have apparmor *archlinux*

PR #1571, opened 2017-09-22
Update quiterss.profile

PharmaceuticalCobweb:
QuiteRSS uses ${HOME}/.local/share/QuiteRss for writing to it's feeds.db file. Without it people lose all their RSS feeds as soon as the close QuiteRSS.


netblue30:
Merged, thanks.

PR #1582, opened 2017-09-29
Add profile for gnome-ring

PharmaceuticalCobweb:
Added profile for gnome-ring in /etc. Pretty much the same as the Skype profile.

### DIFF
--- a/etc/gnome-ring.profile
+++ b/etc/gnome-ring.profile
@@ -1,0 +1,32 @@
+# Firejail profile for gnome-ring
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /etc/firejail/default.local
+# Persistent global definitions
+include /etc/firejail/globals.local
+
+noblacklist ${HOME}/.local/share/gnome-ring
+
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-passwdmgr.inc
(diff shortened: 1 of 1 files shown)


etc/gnome-ring.profile
```diff
@@ -0,0 +1,32 @@
+# Firejail profile for gnome-ring
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /etc/firejail/default.local
```
SpotComms: default -> gnome-ring

etc/gnome-ring.profile
```diff
@@ -0,0 +1,32 @@
...
+# Persistent global definitions
+include /etc/firejail/globals.local
+
+noblacklist ${HOME}/.local/share/gnome-ring
+
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-passwdmgr.inc
+include /etc/firejail/disable-programs.inc
+
```
SpotComms: Does it work with `include /etc/firejail/whitelist-var-common.inc` ?

PharmaceuticalCobweb: Had to add /etc/firejail/whitelist-var-common.inc to my /etc/ folder since I didn't have it. It does seem to work.

SpotComms:
Merged, thanks!

PR #1593, opened 2017-10-07
Create signal-desktop.profile

onny:
Profile for signal-desktop (standalone electron app, see https://github.com/WhisperSystems/Signal-Desktop)

### DIFF
--- a/etc/signal-desktop.profile
+++ b/etc/signal-desktop.profile
@@ -1,0 +1,35 @@
+# Firejail profile for sinal-desktop
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /etc/firejail/signal-desktop.local
+# Persistent global definitions
+include /etc/firejail/globals.local
+
+noblacklist ~/.config/Signal
+
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-programs.inc
+
+mkdir ~/.config/Signal
+whitelist ${DOWNLOADS}
+whitelist ~/.config/Signal
+include /etc/firejail/whitelist-common.inc
+
+caps.drop all
+netfilter
+nodvd
+nogroups
+nonewprivs
+noroot
+notv
+protocol unix,inet,inet6,netlink
+seccomp
+shell none
+
+disable-mnt
+private-dev
+private-tmp
+
+noexec ${HOME}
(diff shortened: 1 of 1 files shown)


etc/signal-desktop.profile
```diff
@@ -0,0 +1,35 @@
...
+# Persistent local customizations
+include /etc/firejail/signal-desktop.local
+# Persistent global definitions
+include /etc/firejail/globals.local
+
+noblacklist ~/.config/Signal
+
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-programs.inc
```
SpotComms: What about `include /etc/firejail/disable-passwdmgr.inc`?

etc/signal-desktop.profile
```diff
@@ -0,0 +1,35 @@
+# Firejail profile for sinal-desktop
```
SpotComms: "sinal" -> "signal"

etc/signal-desktop.profile
```diff
@@ -0,0 +1,35 @@
...
+
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-programs.inc
+
+mkdir ~/.config/Signal
+whitelist ${DOWNLOADS}
+whitelist ~/.config/Signal
+include /etc/firejail/whitelist-common.inc
+
```
SpotComms: If you're on 0.9.51 can you test if `include /etc/firejail/whitelist-var-common.inc` works?

etc/signal-desktop.profile
onny: Yes, we can include this

etc/signal-desktop.profile
onny: Seems to work fine

etc/signal-desktop.profile
```diff
@@ -32,4 +34,3 @@ private-dev
 private-tmp
 
 noexec ${HOME}
-noexec /tmp
```
Fred-Barclay: Should we keep `noexec /tmp`?

onny: Signal doesn't seem to work with noexec /tmp, don't know why

SpotComms:
Merged, thanks!

PR #1604, opened 2017-10-13
Addition of RLIMIT_AS

gosre:
Addition of RLIMIT_AS for limiting process's virtual memory. Kept the same format as your previous rlimit implementations.


startx2017:
merged, thanks.

PR #1613, opened 2017-10-23
Add Popcorn-Time profile

semente:
https://popcorntime.sh/

### DIFF
--- a/etc/Popcorn-Time.profile
+++ b/etc/Popcorn-Time.profile
@@ -1,0 +1,36 @@
+# Firejail profile for Popcorn-Time
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /etc/firejail/Popcorn-Time.local
+# Persistent global definitions
+include /etc/firejail/globals.local
+
+noblacklist ~/.Popcorn-Time
+noblacklist ~/.pki
+
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-programs.inc
+
+mkdir ~/.cache/Popcorn-Time
+mkdir ~/.config/Popcorn-Time
+mkdir ~/.pki
+whitelist ${DOWNLOADS}
+whitelist ~/.cache/Popcorn-Time
+whitelist ~/.config/Popcorn-Time
+whitelist ~/.pki
+whitelist ~/.Popcorn-Time
+include /etc/firejail/whitelist-common.inc
+
+caps.keep sys_chroot,sys_admin
+netfilter
+nodvd
+nogroups
+notv
+shell none
+
+private-dev
+private-tmp
+
+noexec ${HOME}
+noexec /tmp
--- a/etc/Popcorn-Time.profile
+++ b/etc/Popcorn-Time.profile
@@ -10,6 +10,7 @@ noblacklist ~/.pki
 
 include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-passwdmgr.inc
 include /etc/firejail/disable-programs.inc
 
 mkdir ~/.cache/Popcorn-Time
@@ -27,6 +28,8 @@ netfilter
 nodvd
 nogroups
 notv
+novideo
+seccomp
 shell none
 
 private-dev


etc/Popcorn-Time.profile
```diff
@@ -0,0 +1,36 @@
...
+# Persistent local customizations
+include /etc/firejail/Popcorn-Time.local
+# Persistent global definitions
+include /etc/firejail/globals.local
+
+noblacklist ~/.Popcorn-Time
+noblacklist ~/.pki
+
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-devel.inc
```
SpotComms: disable-passwdmgr.inc?

etc/Popcorn-Time.profile
```diff
@@ -0,0 +1,36 @@
...
+whitelist ~/.pki
+whitelist ~/.Popcorn-Time
+include /etc/firejail/whitelist-common.inc
+
+caps.keep sys_chroot,sys_admin
+netfilter
+nodvd
+nogroups
+notv
+shell none
```
SpotComms: novideo? seccomp?

Fred-Barclay:
I have to admit I'm a bit concerned about merging this. Popcorntime is a tool to torrent (both downloading and seeding) copyrighted movies and Tv shows without payment. That falls outside the law in many (most?) places. Their FAQ says as much.

I recognize we already include profiles for torrenting, but they're like our Tor and wireshark profiles - they can be used for many good and legitimate purposes. PCTime doesn't fall under this umbrella of legitimacy. I also recognize it's never our responsibility or even the right to decide how firejail can or cannot be used. But I **am** hesitant about including and shipping this particular profile with firejail.

Just my 2c. 😄

SpotComms:
While I basically agree with @Fred-Barclay ... people will use these programs no matter what, I'd at least rather them do it securely.

etc/Popcorn-Time.profile
semente: Yea!

etc/Popcorn-Time.profile
semente: Good, I'm testing with these new options. Looks everything fine.

semente:
hey, @Fred-Barclay! Thanks for the comment. I completely understand your point of view, even I don't agree with it. That is not only about copyright laws and "pirating" but also economy, society and human nature.

I would love to share my view on this subject with you. That is a important discussion. But perhaps at another opportunity and because @SpotComms has a good point: no one can stop a person to run `firejail --noprofile /path/to/popcorn-time`.

Also because there is at least two legitimate purpose for running Popcorn-Time: a person may own the content but for technical reasons (stream a video to a remote device? watch Netflix on non-proprietary platform?) chose to watch it on Popcorn - it is not a copyright infringement in some jurisdictions afaik; and a person may run it to watch authorized content. I'm not saying that is my case but it may be someone's else so, technically speaking, I don't see this software differently than others.

And because I think that is the classic case that is up to a person be aware of what she is doing and not to a software or we may commit some injustices. :)

semente:
@SpotComms by the way, I made this profile based on Chromium's since Popcorn-Time software is based on it. I may be missing more important stuff. I'm a new firejail user.

SpotComms:
Adding on to @semente's points
Another good one is that in some countries some content simply isn't legally available for purchase. Eg. music albums being locked to a specific country forever or movies coming out many months later then they did in US.

And even then many *paid* streaming services still include ads and even if they don't they are nearly guaranteed to sell your data. 

Netflix
> For example, we use cookies and other technologies to make it easy to access our services by remembering you when you return, to provide and analyze our services, to learn more about our users and their likely interests, and to deliver and tailor marketing or advertising.

Hulu
> We may share information collected from or about you with third parties as explained further below, including business partners, social networking services, service providers, advertisers, and other companies that are not affiliated with Hulu.

HBO GO
> (f) Third Party Ad Servers. As discussed in detail below in paragraph 5, on our sites or as part of our services, third party ad servers may present or serve advertisements, provide us with data collection, reporting, ad response measurement, and site analytics, and assist with delivery of relevant marketing messages and advertisements.

Its non-stop double-dipping.

gosre:
I live in the U.S. and have used popcorn time a lot in the past - ISP's definitely pick up quite fast that you're using popcorn-time and will quickly send you a notice if you don't use a VPN 100% of the time (even if you just click on a trailer through their app). It hasn't proven to be a dangerous program at all for the time its been online, but has gone through quite a bit of legal hurdles. 

While I agree with @Fred-Barclay about not wanting to ship it with firejail by default, it definitely is beneficial to have it as an optional download nonetheless to offer a secure experience for those more concerned.

netblue30:
Unfortunately, we cannot ship it. There is this request to include Popcorn-Time in Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=741972 - let's watch it, and if Debian grabs it, we reopen the pull request and merge it.

@semente: On the main page here on github in Security profiles section we keep links to profile repositories build by users (such as this one: https://github.com/chiraag-nataraj/firejail-profiles). From time to time we even merge them in. If you build one, send us the link, we'll be happy to add it to our list.

semente:
@netblue30 ok! Thanks!

l29ah:
@netblue30 why cannot you?

Issue #1633, opened 2017-11-06
blacklist /run/user/*/bus breaks keepassxc on gentoo


soredake:
with `blacklist /run/user/*/bus`:
```
❯ LC_ALL=C keepassxc-run
Reading profile /home/bausch/.config/firejail/keepassxc.profile
Reading profile /etc/firejail/keepassxc.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Parent pid 19868, child pid 19869
Private /etc installed in 32.12 ms
Child process initialized in 95.39 ms
qt5ct: using qt5ct plugin
process 7: D-Bus library appears to be incorrectly set up; failed to read machine uuid: Failed to open "/etc/machine-id": No such file or directory
See the manual page for dbus-uuidgen to correct this issue.
  D-Bus not built with -rdynamic so unable to print a backtrace

Parent is shutting down, bye...
```

with `noblacklist /run/user/*/bus` (successfully launched):
```
❯ LC_ALL=C keepassxc-run
Reading profile /home/bausch/.config/firejail/keepassxc.profile
Reading profile /etc/firejail/keepassxc.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Parent pid 20140, child pid 20141
Private /etc installed in 33.98 ms
Child process initialized in 91.06 ms
qt5ct: using qt5ct plugin
qt5ct: D-Bus system tray: no
libGL error: failed to open drm device: No such file or directory
libGL error: failed to load driver: nouveau
libGL error: unable to load driver: swrast_dri.so
libGL error: failed to load driver: swrast
```

smitsohu:
If you add the `machine-id` option to the profile and further append 'machine-id' to `private-etc` (see the keepassx profile for comparison), does it work again?

soredake:
Works, adding `machine-id` to `private-etc` is enough.

smitsohu:
Thank you! Just go ahead.

PR #1637, opened 2017-11-07
keepassxc.profile: fix crash on gentoo because of missing machine-id …

soredake:
…in /etc

https://github.com/netblue30/firejail/issues/1633


smitsohu:
Merged, thanks

PR #1643, opened 2017-11-12
Update franz.profile to work with version 5 upwards

onny:
Starting with version 5.0.0 I had to remove this line, otherwise Franz would not start properly


SpotComms:
Merged, thanks!

Issue #1660, opened 2017-11-28
firefox profile leaking filesystem access


Boruch-Baum:
The firefox profile is said to limit the browser's access to the user's filesystem to ~/Downloads and certain other specific files. However, my firefox file dialog (version 52 esr) includes at the bottom of the left-hand bookmark / location bar, an item labeled "other locations", and clicking on that shows that firefox has access to all filesystems / devices through that method.

Fred-Barclay:
G'day Boruch-Baum,
The firefox profile limits any access **in your home directory** to Downloads and a few necessary config files. It also strictly limits or prevents access to select other directories like /boot, /dev, /root, and /tmp.
Also, it prevents lots of potential attacks thanks to the seccomp filter and noroot/nonewprivs filters.

Other system files, such as those in /etc or /lib, are visible, though read-only, inside the sandbox. This is by design.

@netblue30 or some other contributors can certainly explain it better than I can, but in the meantime these links may be of interest to you. 😄 
https://github.com/netblue30/firejail/issues/1352
https://github.com/netblue30/firejail/issues/354
https://github.com/netblue30/firejail/issues/970

Cheers!
Fred

Boruch-Baum:
@Fred-Barclay: Your comment doesn't reflect my experience and seems to ignore what I actually reported, so I should be more specific.

1] Open an instance of firefox using firejail.
2] Open any URL, local or remote.
3] Type C-s (Control-s).
4] Within the save dialog that appears, click on the "Other Locations" line at the bottom of the bookmark bar on the left-hand side of the dialog.
5] Notice that you can now view all mount-points system-wide, including other unmounted partitions, USB drives, network connections, etc.

Nothing to do with access to /etc or /lib, as you commented, and in fact I don't see that I actually have access to those directories at all.

SpotComms:
@Boruch-Baum if you want to block access to other drives you can run this command:
`echo "disable-mnt" >> /etc/firejail/firefox.local`

smitsohu:
@Boruch-Baum for clarification: Is this issue about information leaking into the sandbox?

I tried to reproduce and I can _see_ some forbidden fruit, but I am not able to actually _access_ it.

Boruch-Baum:
@SpotComms: That did it! Thanks. Should this be made standard? Or has it been, and my version (0.9.50) is just not bleeding edge?

[...]

PR #1662, opened 2017-11-29
In Testing: (Re)add disable-mnt to common browser profiles.

Fred-Barclay:
Let's don't merge this just yet. 
This adds disable-mnt to most (all?) of our browser profiles as per the discussion in #1660, blocking access to mount points. This does mean some users (relatively few) will have some problems if their ${DOWNLOADS}, etc, are on external media, but I think it's an acceptable tradeoff for the increased sandbox tightness that will benefit the majority of users. 

Thoughts/comments/suggestions/"we-shouldn't-do-this-because-xyz"? :grin:

### DIFF
--- a/etc/brave.profile
+++ b/etc/brave.profile
@@ -35,4 +35,4 @@ notv
 # protocol unix,inet,inet6,netlink
 # seccomp
 
-# disable-mnt
+disable-mnt
--- a/etc/chromium.profile
+++ b/etc/chromium.profile
@@ -32,6 +32,7 @@ nogroups
 notv
 shell none
 
+disable-mnt
 # private-bin chromium,chromium-browser,chromedriver
 private-dev
 # private-tmp - problems with multiple browser sessions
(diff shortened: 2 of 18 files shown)


reinerh:
You could also add it to the surf browser profile.

Fred-Barclay:
Surf added, thanks @reinerh

Saren-Arterius:
Unfortunately I am one of the user who has download dir on external mount because my HDD is way bigger then my OS SSD partition in terms of space. Months ago I noticed /media would not mount after firejail upgraded, so I downgraded it and excluded it.

Until now I found the cause is this commit, so is there anything like `no-disable-mnt` so that I could add that line to .local files instead of removing all that line from firejail browser profiles and executing `chattr +C` on them? Thanks

SkewedZeppelin:
@Saren-Arterius you can run the following command to disable `disable-mnt` in all profiles
`echo "ignore disable-mnt" >> /etc/firejail/globals.local`

Saren-Arterius:
@SkewedZeppelin Thank a lot!

Issue #1320, opened 2017-05-30
0.9.44.10 -> 0.9.46: libstdc++.so cannot be loaded


uzytkownik:
I run into multiple programs failing to load C++ library under the firejail. Downgrading fixed the problem.

netblue30:
What distribution are you using?

uzytkownik:
Gentoo with gcc 6.3.0.

netblue30:
We had this problem about a year ago on Gentoo and we ended up commenting out "blacklist /usr/bin/gcc" line in /etc/firejail/disable-devel.inc - see #674 and #824. Go in this file and comment out (add a #) to all lines in gcc section of the file. One of them is creating the problem. If you run " find / -name libstdc++.so.6" what does it say?

uzytkownik:
```
# grep libstdc++.so.6 /var/db/pkg/sys-devel/gcc-6.3.0/CONTENTS  | awk '{print $2}'
/usr/lib/gcc/x86_64-pc-linux-gnu/6.3.0/32/libstdc++.so
/usr/lib/gcc/x86_64-pc-linux-gnu/6.3.0/32/libstdc++.so.6
/usr/lib/gcc/x86_64-pc-linux-gnu/6.3.0/32/libstdc++.so.6.0.22
/usr/lib/gcc/x86_64-pc-linux-gnu/6.3.0/libstdc++.so
/usr/lib/gcc/x86_64-pc-linux-gnu/6.3.0/libstdc++.so.6
/usr/lib/gcc/x86_64-pc-linux-gnu/6.3.0/libstdc++.so.6.0.22
/usr/lib/debug/usr/lib/gcc/x86_64-pc-linux-gnu/6.3.0/32/libstdc++.so.6.0.22.debug
/usr/lib/debug/usr/lib/gcc/x86_64-pc-linux-gnu/6.3.0/libstdc++.so.6.0.22.debug
/usr/share/gdb/auto-load/usr/lib64/gcc/x86_64-pc-linux-gnu/6.3.0/libstdc++.so.6.0.22-gdb.py
/usr/share/gdb/auto-load/usr/lib64/gcc/x86_64-pc-linux-gnu/6.3.0/32/libstdc++.so.6.0.22-gdb.py
```

I commented out all lines in GCC section and it didn't helped.

netblue30:
Does it work if you start the sandboxes with --noprofile? (firejail --noprofile program)

bitthief:
[...]

PR #1664, opened 2017-11-29
qtox needs libstdc++.so.6

soredake:
`qtox: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory`
Related https://github.com/netblue30/firejail/issues/1320


SpotComms:
Merged, thanks!

PR #1687, opened 2017-12-21
Added environment variable QML_DISABLE_DISK_CACHE=1 to okular.profile.

floxo:
Without it, recent okular versions (here 17.12.0-1 on Arch Linux) crash with

mprotect failed in ExecutableAllocator::makeExecutable: Permission denied

due to the noexec constraints in the firejail profile.


smitsohu:
Merged, thanks!

I have seen the same with kwin_x11, probably we will need to add it to a number of profiles. Only for the record: We are working around a [Qt bug](https://bugreports.qt.io/browse/QTBUG-58508) here.

PR #1691, opened 2017-12-22
Fix Deluge

DiGitHubCap:
Deluge needs access to more than the deluge binary if it runs as a daemon (or if you want to access it via the web or command line)

### DIFF
--- a/etc/deluge.profile
+++ b/etc/deluge.profile
@@ -32,6 +32,6 @@ seccomp
 shell none
 
 # deluge is using python on Debian
-private-bin deluge,sh,python*,uname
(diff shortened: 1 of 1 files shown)


Fred-Barclay:
Thanks! I don't have deluge installed - would you be able to remember off the top of your head what the other deluge* binaries are?

DiGitHubCap:
On my system, they are: deluge deluge-console deluged deluge-gtk deluge-web.

Fred-Barclay:
@DiGitHubCap If it's not too much trouble, can you add those instead?

DiGitHubCap:
Done.

Fred-Barclay:
Awesome, thanks!

PR #1697, opened 2017-12-27
Blacklist the monero wallets directory

sgtpep:
~/Monero/wallets is the default path suggested by the official wallet application, but it can be changed by user.


netblue30:
Merged, thanks.

PR #1701, opened 2017-12-30
tor flavours

bn0785ac:
all tor-browser langpacks

bugfixes:
* its posible run it with tor launcher, but i cant see my downloads
* personalized profiles (if we use a shared profile  we may get a data leak)


SpotComms:
I've added a commit that simplifies the profiles and bases off of the existing Tor browser profile. I still wonder if there is a better way to do this.

netblue30:
All merged, thanks.

PR #1708, opened 2018-01-01
Fix #1702 - Couldn't start 'minetest' in Debian Testing

### LINKED ISSUE #1702: Couldn't start 'minetest' in Debian Testing
With the current state of Debian Testing (with firejail 0.9.52) I couldn't start the game "minetest".

The output for _"firejail minetest"_ is the following:
> Reading profile /etc/firejail/minetest.profile
> Reading profile /etc/firejail/disable-common.inc
> Reading profile /etc/firejail/disable-devel.inc
> Reading profile /etc/firejail/disable-passwdmgr.inc
> Reading profile /etc/firejail/disable-programs.inc
> Reading profile /etc/firejail/whitelist-common.inc
> Parent pid 1258, child pid 1259
> Warning: skipping asound.conf for private /etc
> Private /etc installed in 24.43 ms
> Child process initialized in 76.66 ms
> minetest: error while loading shared libraries: libGL.so.1: cannot open shared object file: No such file or directory
> 

bitfreak25:
This removes the "private-etc" line from the "minetest"-profile for a successfully start of the game.

### DIFF
--- a/etc/minetest.profile
+++ b/etc/minetest.profile
@@ -32,6 +32,8 @@ shell none
 disable-mnt
 private-bin minetest
 private-dev
+# private-etc needs to be updated, see #1702
+#private-etc asound.conf,ca-certificates,drirc,fonts,group,host.conf,hostname,hosts,ld.so.cache,ld.so.preload,localtime,nsswitch.conf,passwd,pulse,resolv.conf,ssl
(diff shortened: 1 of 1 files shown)


SpotComms:
Merged, thanks!

PR #1710, opened 2018-01-01
Add profile for "playonlinux"

bitfreak25:
This profile have been successfully tested by starting a windows application through it.

"wine.profile" has been used as template for this. Only "noblacklist ${PATH}/nc" has been added because playonlinux needs it to run.

**Please note** that this is currently not tested due to security aspects, so it may need a rework later on. Because opening a unknown windows application through it could possibly be a security risk.

### DIFF
--- a/etc/playonlinux.profile
+++ b/etc/playonlinux.profile
@@ -1,0 +1,28 @@
+# Firejail profile for playonlinux
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /etc/firejail/playonlinux.local
+# Persistent global definitions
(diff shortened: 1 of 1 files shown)


SpotComms:
Looks good, but is it possible to instead replace all the parts from the wine profile with "include /etc/firejail/wine.profile"?

bitfreak25:
I thought about it. But playonlinux uses `${HOME}/.PlayOnLinux` and not `${HOME}/.wine`. (Sorry, I forget about that change.)

SpotComms:
Merged, thanks!

PR #1713, opened 2018-01-04
Apparmor: fix broken file dialogs in kde plasma

Vincent43:
For some time apparmor started breaking file dialogs in kde plasma (gwenview, calibre, qbittorrent, etc). typical audit report below:

`AVC apparmor="DENIED" operation="open" profile="firejail-default" name="/run/user/1000/#28520" pid=1997 comm="qbittorrent" requested_mask="w" denied_mask="w" fsuid=1000 ouid=1000`

`AVC apparmor="DENIED" operation="link" profile="firejail-default" name="/run/user/1000/qBittorrentZcaeTi.1.slave-socket" pid=3679 comm="qbittorrent" requested_mask="l" denied_mask="l" fsuid=1000 ouid=1000 target="/run/user/1000/#79965"`

This commit fixes this issue. Tested on Archlinux (linux 4.14.11, kde 5.11.5)


etc/firejail-default
```diff
@@ -30,7 +30,8 @@ profile firejail-default flags=(attach_disconnected,mediate_deleted) {
 /{,var/}run/user/**/dconf/user rw,
 /{,var/}run/user/**/pulse/ rw,
 /{,var/}run/user/**/pulse/** rw,
-/{,var/}run/user/**/*.slave-socket rw,
+/{,var/}run/user/**/*.slave-socket rwl,
+/{,var/}run/user/**/#@{PID} rw,
```
smitsohu: If I'm not mistaken everything behind '#' is just a comment. How much do we need this line, can we do without?

smitsohu: correcting myself. It seems that '#' can be part of a pathname! But this is undocumented as far as I can see, so I'd still suggest to escape the '#'

Fred-Barclay: I'm not that familiar with apparmor, but since it appears that any PID can be substituted here, would this allow any program to have rw privs in /run/user or /var/run/user?  
If so, does that have any security implications?

Maybe I'm completely off-base here... :smile:

Vincent43: It isn't related to "Process ID", it just catches string of random numbers, see example denial messages I posted in description. I used existing alias https://github.com/netblue30/firejail/blob/master/etc/firejail-default#L11 because it fits for the job.

Vincent43: It catches things like /run/user/1000/#79965 which are needed. I modified it as you requested.

smitsohu:
@Vincent43 thanks for your pull request. Unless someone else is faster than me, I'll take a look and merge it the coming week...

smitsohu:
Thanks, merged!

Issue #1737, opened 2018-01-17
Add support for syscalls 329-332 (pkey_mprotect, pkey_alloc, pkey_free, statx)


vn971:
These are new syscalls supported by linux kernel. See for example strace: https://sourceforge.net/p/strace/code/ci/master/tree/linux/x86_64/syscallent.h#l332
And also local `build/include/linux/syscalls.h` of your kernel of course.

Please add support for those syscalls to firejail? Thanks!

netblue30:
all set, thanks.

Issue #1723, opened 2018-01-10
ktorrent 5.1.0 doesn't run with firejail 0.9.52


paolomi:
ktorrent --version
ktorrent 5.1.0


firejail --version                                                                                                                                   
firejail version 0.9.52                                                                                                                                                            
                                                                                                                                                                                   
Compile time support:                                                                                                                                                              
        - AppArmor support is disabled                                                                                                                                             
        - AppImage support is enabled                                                                                                                                              
        - bind support is enabled                                                                                                                                                  
        - chroot support is enabled                                                                                                                                                
        - file and directory whitelisting support is enabled                                                                                                                       
        - file transfer support is enabled                                                                                                                                         
        - git install support is disabled                                                                                                                                          
        - networking support is enabled                                                                                                                                            
        - overlayfs support is enabled                                                                                                                                             
        - private-home support is enabled                                                                                                                                          
        - seccomp-bpf support is enabled                                                                                                                                           
        - user namespace support is enabled                                                                                                                                        
        - X11 sandboxing support is enabled 



Artix distro: https://artixlinux.org/


firejail --debug ktorrent
Autoselecting /bin/bash as shell
Building quoted command line: 'ktorrent' 
Command name #ktorrent#
Found ktorrent profile in /etc/firejail directory
Reading profile /etc/firejail/ktorrent.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc
Username test, groups 1002, 
Reading profile /etc/firejail/whitelist-common.inc
Reading profile /etc/firejail/whitelist-var-common.inc
DISPLAY=:0 parsed as 0
Using the local network stack
Parent pid 23083, child pid 23085
Initializing child process
Host network configured
PID namespace installed
Mounting tmpfs on /run/firejail/mnt directory
Creating empty /run/firejail/mnt/seccomp.protocol file
Creating empty /run/firejail/mnt/seccomp.postexec file
Build protocol filter: unix,inet,inet6
sbox run: /usr/lib/firejail/fseccomp protocol build unix,inet,inet6 /run/firejail/mnt/seccomp.protocol (null) 
sbox file descriptors:
total 0
lrwx------ 1 test test 64 Jan 10 21:19 0 -> /dev/null
lrwx------ 1 test test 64 Jan 10 21:19 1 -> /dev/pts/5
[...]

PR #1738, opened 2018-01-18
add new syscalls from glibc 2.26-10

vn971:
File generated by ../tools/extract_syscall, as per instructions

### DIFF
--- a/src/include/syscall.h
+++ b/src/include/syscall.h
@@ -71,6 +71,11 @@
 	{"alarm", __NR_alarm},
 #endif
 #endif
+#ifdef SYS_arch_prctl
+#ifdef __NR_arch_prctl
+	{"arch_prctl", __NR_arch_prctl},
+#endif
+#endif
 #ifdef SYS_bdflush
 #ifdef __NR_bdflush
 	{"bdflush", __NR_bdflush},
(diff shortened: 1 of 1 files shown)


vn971:
This PR fixes https://github.com/netblue30/firejail/issues/1737

vn971:
This pull is urgent as without it, all Qt applications _do not work_ with a seccomp whitelist. As far as I understood, Qt makes a `statx` call very early.
@netblue30 @Fred-Barclay

Vincent43:
glibc 2.26-10 is Archlinux specific package version. I suppose those changes come from newer linux kernel api  - 4.14.
https://git.archlinux.org/svntogit/packages.git/log/trunk?h=packages/glibc

netblue30:
All merged, thanks!

Fred-Barclay:
Thanks @vn971!
Interesting bit about Qt. Does someone know if it's related to the trouble we've seen with `QML_DISABLE_DISK_CACHE` and `QTWEBENGINE_DISABLE_SANDBOX`? #1723 for instance. 
cc @netblue30 @SpotComms @chiraag-nataraj @smitsohu

vn971:
@Fred-Barclay I had hopes for that once you wrote it, but unfortunately, it doesn't work. I did the following:
1. reverted back to stock firejail on my system
2. used `--env=QML_DISABLE_DISK_CACHE=1 --env=QTWEBENGINE_DISABLE_SANDBOX=1` to start qBittorrent

Unfortunately, I still got:
```
Parent pid 25023, child pid 25024
Private /etc installed in 5.12 ms
Post-exec seccomp protector enabled
Warning fseccomp: syscall "statx" not available on this platform
Child process initialized in 58.51 ms

```

Fred-Barclay:
@vn971 I was thinking the other way around. 😁 Would this PR mean that we don't have to set the Qt environmental variables?

Vincent43:
Qt qml cache bug is unrelated and will be fixed only in qt 5.11.

PR #1744, opened 2018-01-21
fixes for the keepassxc 2.2.5 version

soredake:


etc/keepassxc.profile
```diff
@@ -11,6 +11,8 @@ noblacklist ${HOME}/*.kdb
 noblacklist ${HOME}/*.kdbx
 noblacklist ${HOME}/.config/keepassxc
 noblacklist ${HOME}/.keepassxc
+# 2.2.4 needs this path when compiled with "Native messaging browser extension"
+noblacklist ${HOME}/.mozilla
```
SpotComms: How needed is this?

soredake: Without whitelist it can't create folder `$HOME/.mozilla/native-messaging-hosts/` and place `org.keepassxc.keepassxc_browser.json` in it

SpotComms: Ah so its just a file with a token to give Firefox addon access to the HTTP API? Did they remove the old style popup for when the HTTP plugin requests access? Like will it still fallback to the popup if it can't access .mozilla?

soredake:
It is not using http https://github.com/keepassxreboot/keepassxc/pull/608

etc/keepassxc.profile
soredake: It is not using http keepassxreboot/keepassxc#608

PR #1745, opened 2018-01-21
Apparmor: restrict access to writable files

Vincent43:
Access to writable files can be restricted to their owner only.


Vincent43:
I reverted `/proc` changes. They are read-only anyway.`/run/user/**` restrictions should be fine in any case.

smitsohu:
Thanks, merged!

PR #1764, opened 2018-02-07
Add seccomp filters for remmina, from an strace session connecting via RDP

jelford:
To derive the filters I ran: `strace -qcf remmina` then connected to a remote host via RDP.

The previous filter (just the default list) allowed Remmina to start, but would fail to establish the connection (it got as far as validating credentials; bad creds gave a different error).

The specific problematic sys call when I checked the audit log was `access` - but I didn't try incrementally building up from there, I just took the whole strace output.

### DIFF
--- a/etc/remmina.profile
+++ b/etc/remmina.profile
@@ -22,7 +22,8 @@ noroot
 notv
 novideo
 protocol unix,inet,inet6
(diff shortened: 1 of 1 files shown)


Fred-Barclay:
Looks good, thanks! 👍 

What happens if you only add the `access` syscall without the others, I wonder?

smitsohu:
Hmm. But is it possible we don't blacklist `access` in the default seccomp filter?

smitsohu:
@jelford sorry, I should have elaborated a bit more... please excuse the brevity. The problem is that seccomp whitelists are not necessarily portable. From [Wordpress](https://firejail.wordpress.com/documentation-2/seccomp-guide/):

> The exact list of syscalls depends on the kernel running the system, the version of the program and all the libraries the program is linking in.

In any case your whitelist is a great addition to the profile :+1: But we can only ship it as an opt-in unfortunately.

With regard to the original issue I wondered whether `firejail --ignore=blacklist remmina` actually also solves it (because, if I see it right, there is none of the forbidden syscalls in your whitelist).

jelford:
@smitsohu sorry, I've just been away from my workstation for a few days. I'll take a look at your suggestions over the weekend - I would definitely prefer to push something more portable. Thanks for the feedback.

jelford:
To answer questions above:
* Using just `seccomp.keep access` leads to a startup-crash (seccomp violation for a call to `brk`)

I'm now unable to replicate the original failure, so I'm going to close this off. I'm not sure whether there was a remmina update in the meantime, but I don't think this PR is useful. I'll certainly come back to it if I have problems again.

In the meantime, I think I need to understand better how `blacklist` works - I'll take a look through the code before the next seccomp related patch. Thanks for your help both.

smitsohu:
@jelford Let me emphasize that your pull request is highly welcome! We are very interested in the whitelist :smile: If you don't mind I reopen the pull request, and we merge it commented....

Issue #1773, opened 2018-02-11
rewrite browser profiles with redirections


smitsohu:
We are shipping Firejail with a large number of browser profiles. Many of these browsers are just forks off Firefox and Chromium, with only minor deviations from the original codebase.

In order to ease maintenance, I would like to propose rewriting these profiles with redirections to something like firefox-common.inc and chromium-common.inc. The browser profiles themselves could keep their specific paths (and `private-bin` if necessary).

Should projects begin to deviate more strongly (like Moon browser does already), it is still always possible to switch to a dedicated profile.

SkewedZeppelin:
Here are all the browsers we have profiles for. If I missed any just edit it in.

# Chromium Based
- bnox
- brave
- chromium (chromium-browser)
- dnox
- flashpeak-slimjet
- google-chrome (google-chrome-stable)
- google-chrome-beta
- google-chrome-unstable
- inox
- iridium (iridium-browser)
- opera
- opera-beta
- vivaldi (vivaldi-beta, vivaldi-stable)
- yandex-browser

# Gecko Based
- abrowser
- cliqz
- cyberfox (Cyberfox)
- firefox (firefox-developer-edition, firefox-esr, firefox-nightly, iceweasel)
- icecat
- palemoon
- torbrowser-launcher
- waterfox

# WekKit2Gtk
- epiphany
- surf
- uzbl

# QtWebEngine (cef?)
- qupzilla
- qutebrowser

# Text Based
- elinks
- lynx

# Unknown
- dillo
- netsurf

SkewedZeppelin:
Of note there are 4 email clients that all include firefox.profile for whatever reason:
fossamail, geary, icedove, and thunderbird

SkewedZeppelin:
Here I gave it a go and unified all of the Chromium-based profiles (except brave)

https://github.com/SkewedZeppelin/firejail/commit/0160a93ee20135930de968679811cb2e3e89481c

the Firefox ones might be more tedious, we should also evaluate whether or not we want to keep all of the paths that we currently do in Firefox

Fred-Barclay:
On the note about the email clients, I for one think we ought to not include firefox in their profiles. We already have to do some workarounds (like `ignore private-tmp` and commenting out `machine-id` in the thunderbird profile, all because we also include firefox's profile.)

I'll take a look at the firefox profiles and see if I can get something similar to @SkewedZeppelin

SkewedZeppelin:
[...]

PR #1774, opened 2018-02-11
Unify all Chromium and Firefox based browser profiles

SkewedZeppelin:
See #1773

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,15 @@ Use this issue to request new profiles: [#1139](https://github.com/netblue30/fir
 `````
 # Current development version: 0.9.53
 
+## Browser profile unification
+
+All Chromium and Firefox browsers have been unified to instead extend
+chromium-common.profile and firefox-common.profile respectively.
+This allows for reduced maintenance and ease of adding new browsers.
+NOTE: All users of Firefox-based browsers who use addons and plugins
+that read/write from ${HOME} will need to uncomment the includes for
+firefox-common-addons.inc in firefox-common.profile.
+
 ## New profiles
 
 Basilisk browser, Tor Browser language packs, PlayOnLinux, sylpheed, discord-canary,
--- a/etc/abrowser.profile
+++ b/etc/abrowser.profile
@@ -13,7 +13,8 @@ mkdir ${HOME}/.mozilla
 whitelist ${HOME}/.cache/mozilla/abrowser
 whitelist ${HOME}/.mozilla
 
-# private-etc ca-certificates,ssl,machine-id,dconf,selinux,passwd,group,hostname,hosts,localtime,nsswitch.conf,resolv.conf,xdg,gtk-2.0,gtk-3.0,X11,pango,fonts,abrowser,firefox,mime.types,mailcap,asound.conf,pulse,pki,crypto-policies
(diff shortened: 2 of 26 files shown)


README.md
```diff
@@ -98,6 +98,15 @@ Use this issue to request new profiles: [#1139](https://github.com/netblue30/fir
...
 # Current development version: 0.9.53
 
+## Browser profile unification
+
+All Chromium and Firefox browsers have been unified to instead extend
+chromium-common.profile and firefox-common.profile respectively.
+This allows for reduced maintenance and ease of adding new browsers.
+NOTE: All users of Firefox-based browsers who use addons and plugins
+that read/write from ${HOME} will need to uncomment the includes for
+firefox-common-addons.inc in firefox-common.profile.
```
smitsohu: you call it firefox-common-addons.inc here, but firefox-common-addons.profile everywhere else :)

smitsohu:
Thanks, awesome!

The only question is if we should enable firefox-common-addons by default or not. As you say in #1773, it will probably disrupt setups that are working fine with 0.9.52 and before. 
It is only IMHO, but I would rather think the commented firefox-common-addons should be the advanced setting, and the "everything just works" uncommented firefox-common-addons the default, although it is a little less secure.

etc/abrowser.profile
```diff
@@ -7,42 +7,14 @@ include /etc/firejail/globals.local
...
-caps.drop all
-netfilter
-nodvd
-nonewprivs
-noroot
-notv
-protocol unix,inet,inet6,netlink
-seccomp
-tracelog
+# private-etc ca-certificates,ssl,machine-id,dconf,selinux,passwd,group,hostname,hosts,localtime,nsswitch.conf,resolv.conf,xdg,gtk-2.0,gtk-3.0,X11,pango,fonts,abrowser,firefox,mime.types,mailcap,asound.conf,pulse,pki,crypto-policies
```
Fred-Barclay: I'm noticing some of the firefox-based profiles have different private-etc lines (for instance, this line is different than cyberfox which is different than iceweasel). Can we maybe have a unified line in firefox-common, and then for each profile add its individual etc folders? For instance, we can probably use something like
```
private-etc ca-certificates,ssl,machine-id,dconf,selinux,passwd,group,hostname,hosts,localtime,nsswitch.conf,resolv.conf,xdg,gtk-2.0,gtk-3.0,X11,pango,fonts,mime.types,mailcap,asound.conf,pulse,pki,crypto-policies
```
in firefox-common.
Then in abrowser, just do a `private-etc abrower` before we include firefox-common.

I did a simple test and it looks like this should work.

SkewedZeppelin: Will do, I wasn't sure if a second private-etc would extend or override.

README.md
SkewedZeppelin: No? There is firefox-common.profile and firefox-common-addons.inc

smitsohu: Yes! Sorry

Fred-Barclay:
Ship it!

SkewedZeppelin:
Merged. We'll still need to decide whether to leave https://github.com/netblue30/firejail/blob/master/etc/firefox-common.profile#L9 commented or not before next release.

PR #1775, opened 2018-02-14
Apparmor: don't duplicate userspace /run/user restrictions

Vincent43:
Currently userspace firejail do blacklist approach to `/run/user/` directory.  By default it blacklists `/run/user/**/systemd` and `/run/user/**/gnupg`. Additional restrictions can be enabled in profiles like blacklisting `/run/user/**/bus` , etc. The blacklist can be extended or degraded by profile which allows for fine grained hardening.

In apparmor we do whitelist approach instead. It means we have to explicitly enable access to every file which firejail already allow access. This duplicates functionality and amount of work to do. Moreover we end up with same list of allowed files as every one of them is used by some app and apparmor profile is global. It's even worse as firejail blacklist can be disabled with `writable-run-user` command which means we have to whitelist literally everything under `/run/user/` to not cause breakages when using apparmor.

The solution for all above is to leave handling of `/run/user` to userspace firejail which is better tool to do this. In apparmor we should only handle things which firejail can't do on its own.


Vincent43:
@netblue30 are those changes and https://github.com/netblue30/firejail/pull/1776 ok for you?

netblue30:
Sure, merge them in!

PR #1780, opened 2018-02-20
Add a profile for bitcoin-qt

baryluk:
Works just fine, including remembering all settings.

libGL / mesa-loader errors out when trying to detect device or trying to load drivers, but it still works fine with no3d.

It even works just fine with private-lib, but qt then complains about openssl incompatibilities. So for now, I left it commented out.

It appears also removing machine-id does not make any harm either, but I have seen it in ktorrent.profile (which was my starting point somehow), so I am not sure.

In syslog only thing I see is:

Feb 20 18:13:05 xyz firejail[9]: blacklist violation - sandbox 30207, exe bitcoin-qt, syscall opendir, path /dev/dri

but that is expected due no3d thingy. It is fine. Would be nice to be able to ignore that from logging.

### DIFF
--- a/etc/bitcoin-qt.profile
+++ b/etc/bitcoin-qt.profile
@@ -15,8 +15,6 @@ include /etc/firejail/disable-programs.inc
 mkdir ${HOME}/.bitcoin
 whitelist ${HOME}/.bitcoin
 
-whitelist ${HOME}/.bitcoin-testnet
-
 mkdir ${HOME}/.config/Bitcoin
 whitelist ${HOME}/.config/Bitcoin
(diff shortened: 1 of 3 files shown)


baryluk:
Ah. Sorry .bitcoin-testnet is uneeded, testnet blockchain is in ~/.bitcoin/testnet3/

Fred-Barclay:
Merged, thanks!
`machine-id` is actually a good thing to leave it - it spoofs the file `/etc/machine-id` with a random value so that the sandbox can't read the real ID of your machine.

PR #1783, opened 2018-02-21
viewnior needs access to X

joelazar:
Hi all,

Viewnior needs access to X so I think it would be useful not to blacklist .Xauthority, otherwise the application cannot open properly. What do you think about this change?

Br, Joe

### DIFF
--- a/etc/viewnior.profile
+++ b/etc/viewnior.profile
@@ -6,7 +6,6 @@ include /etc/firejail/viewnior.local
 include /etc/firejail/globals.local
 
 blacklist /run/user/*/bus
-blacklist ${HOME}/.Xauthority
 blacklist ${HOME}/.bashrc
 
 noblacklist ${HOME}/.Steam


Fred-Barclay:
Hi @joelazar and welcome to firejail!
Does viewnior not work (or have errors?) when .Xauthority is blacklisted? It's working fine on my end, though with some cosmetic error messages (and I don't even **have** an .Xauthority file on my system - Fedora 27 Cinnamon :smile: ).

I think, if viewnior does work for you, we probably shouldn't unblacklist .Xauthority. There are some other profiles for programs that use X for which we blacklist .Xauthority (bibletime and xiphos) so I suspect .Xauthority access may not be needed here.
Cheers, and thanks!
Fred

Fred-Barclay:
Closing for now...

joelazar:
Hi @Fred-Barclay,

First of all, sorry for my late response. Viewnior did not work for me at all, when .Xauthority was blacklisted. Let me share the output of trying to open the application from a terminal:

```
$ viewnior 
Reading profile /etc/firejail/viewnior.profile
Parent pid 29420, child pid 29421
Private /etc installed in 4.83 ms
Blacklist violations are logged to syslog
Child process initialized in 92.66 ms
No protocol specified
Cannot open display: 
Run 'viewnior --help' to see a full list of available command line options.

```

And the generated log item to syslog is:
```
febr 26 10:52:17 thinkpad firejail[29426]: blacklist violation - sandbox 29420, exe viewnior, syscall access, path /home/jlazar/.Xauthority
```

I think the missing .Xauthority file is because Fedora is using Wayland as a default display server protocol, instead of x11. I'm currently using Manjaro 17.1.5 with XFCE and I have x11:
```
$ loginctl show-session c2 -p Type
Type=x11
```

Would you accept this pull request if I remove xauthority blacklisting from bibletime and xiphos profile too?

Thanks in advance, Joe

joelazar:
read-only works fine too. Should I create a new pull request with the following [change](https://github.com/joelazar/firejail/commit/f3e129ef59f3aa1695c354a85ee84fc459abfcdb) ?

Vincent43:
Did you tested if bibletime and xiphos are broken with blacklisted `.Xauthority` also?

joelazar:
I did tested bibletime and the result was the same:
```
$ firejail bibletime 
Reading profile /etc/firejail/bibletime.profile
Parent pid 27532, child pid 27533
Warning: skipping sword for private /etc
Private /etc installed in 10.60 ms
Blacklist violations are logged to syslog
Child process initialized in 93.05 ms
No protocol specified
qt.qpa.screen: QXcbConnection: Could not connect to display :0.0
Could not connect to any X display.

```
but I did not tested xiphos though, because I came across some AUR issue here: 
```
pacaur -S xiphos
...
:: Checking biblesync integrity...
==> Making package: biblesync 1.1.2-3 (2018. febr. 26., hétfő, 14:35:01 CET)
==> Retrieving sources...
  -> Found biblesync-1.1.2.tar.gz
==> Validating source files with sha512sums...
    biblesync-1.1.2.tar.gz ... FAILED
==> ERROR: One or more files did not pass the validity check!
```
Do you think it's necessary to test this program too?

I think it would be a nice solution to add readonlyness to a common settings.

Vincent43:
You can download [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/plain/PKGBUILD?h=xiphos) and run `updpkgsums && makepkg -src` in same location as the downloaded file is the install created package.

Anyway let's wait for @Fred-Barclay opinion about adding `read-only ${HOME}/.Xauthority` globally.

Issue #1801, opened 2018-03-06
The "less" command can't view files when run as root, using firejail.


ThePythonicCow:
If I try to view a file, while root, using the firejailed "less" command, it either can't find the file to be viewed, or lacks permission to view it.

The firejailed "less" command works fine for me, when I am not su'd to root.

If the file to be viewed, using less as root, is some /var/log/example.log file, then the firejailed less command fails with EPERM Permission denied

If the file to be viewed, using less as root, is some file owned by an ordinary non-root user, such as /home/example_user/example.txt, then the firejailed less command fails ENOENT No such file or directory.

If I remove the /usr/local/bin/less symlink to firejail, then the "real" /usr/bin/less command runs and works fine.

I have only recently converted to Gentoo, and I have just started running firejail in this setup.

**My Gentoo version**: Gentoo Base System release 2.4.1
**My less -V version**: less 529 (PCRE regular expressions)
**The "firejail --version"**: firejail version 0.9.50

netblue30:
Let's close #1800 and #1801and move the discussion to https://github.com/netblue30/firejail/pull/1802 until we find a fix.

Issue #1800, opened 2018-03-06
Gentoo portage "emerge" cant find what it just downloaded with wget


ThePythonicCow:
The Gentoo portage command to install/update software "emerge --ask somepackage" fails for me. It seems to download the appropriate version of "somepackage" from the appropriate Gentoo server, but then fails immediately when it can't find what it just downloaded.

I have only recently converted to Gentoo, and I have just started running firejail in this setup, so I don't have a prior history of this working.

**My Gentoo version**: Gentoo Base System release 2.4.1
**The "emerge --version"**: Portage 2.3.19 (python 3.5.4-final-0, default/linux/amd64/17.0/desktop, gcc-6.4.0, glibc-2.25-r10, 4.15.2 x86_64)
**The "firejail --version"**: firejail version 0.9.50

I diagnosed the problem by using strace on the session within which I ran the failing emerge invocations.  I could see the emerge command invoke /usr/local/bin/wget to download of the desired package. This wget download would run to successful completion, but then immediately thereafter, the invoking emerge code would fail, complaining that it could not see what it just had wget download.

If I remove /usr/local/bin/wget (the symlink to /usr/bin/firejail), then emerge uses /usr/bin/wget, the non-firejailed instance of wget, and emerge successfully installs and upgrades all packages that I have tried, and otherwise behaves as expected.

Ferroin:
So, this is just me, but this sounds like a classic 'it hurts if I do X' problem.

Emerge should already be sandboxing downloads to a certain degree (make sure you have the `userfetch` feature enabled in make.conf), and sandboxing beyond that is of questionable value because you already functionally have to trust the systems you're downloading the source archives for the packages from, otherwise you can't trust those packages once built (and you can't just run everything in firejail, there's core system components at stake here that if compromised make any kind of sandboxing mostly irrelevant (think stuff like openssl or the Python interpreter)).

Vincent43:
I guess this an https://github.com/netblue30/firejail/issues/1801 are working as intended. I think the problem is that certain tools used for system administration shouldn't be included in default firecfg config. I'll make a PR which remove them.

netblue30:
Let's close #1800 and #1801and move the discussion to https://github.com/netblue30/firejail/pull/1802 until we find a fix.

PR #1802, opened 2018-03-06
Don't enable profiles which may be used for system administrtion

Vincent43:
Enabling those may break system update services on Arch and Gentoo. Also those tools are very often used for general system administration which means they shouldn't be enabled by default.

See https://github.com/netblue30/firejail/issues/1800 and https://github.com/netblue30/firejail/issues/1801 (possibly more)

### DIFF
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -91,7 +91,7 @@ display
 dnox
 dnscrypt-proxy
 dnsmasq
-dolphin
+# dolphin - may be used for system administration
 dooble
 dooble-qt4
 dosbox
@@ -192,7 +192,7 @@ jitsi
 k3b
 kaffeine
 karbon
-kate
+# kate - may be used for editing system files
 kcalc
 # kdeinit4
 kdenlive
@@ -214,7 +214,7 @@ ktorrent
 # kwin_x11
 kwrite
 leafpad
-less
+# less - may be used for system administration
 libreoffice
 liferea
 linphone
@@ -390,7 +390,7 @@ waterfox
 weechat
 weechat-curses
 wesnoth
-wget
+# wget - may be used in package management (Arch, Gentoo)
 wine
 wire
 wireshark


SkewedZeppelin:
less and wget are fine to remove. and we should still fix https://github.com/netblue30/firejail/pull/1565#issuecomment-330707453

do we really need to remove dolphin and kate though?

Vincent43:
I assume users may want to manage their system with graphical tools (not everyone want to do everything in terminal). Kate (and dolphin soon) has ability to safely edit any system file through polkit integration which is pretty neat.

I think we should enable by default only profiles which almost everyone want to enable without drawbacks. It would be nice to provide a way for users to override firecfg config as they like (add and delete profiles) but as for know IMO we should stick to "don't break userspace" rule :smile:

netblue30:
Maybe we can try something like this:

We build a new list of sysadmin apps, similar to  src/firecfg/firecfg.config. We modify the code handling the symlinks in /usr/local/bin and if the user is root we start the sandbox with a /etc/firejail/generic-sysadmin.profile instead of the regular application profile. In this profile we allow full filesystem access, and we add a default seccomp filter and drop some of the capabilities.

Vincent43:
I'm skeptical of this. Sysadmin apps don't always run as root (kate and dolphin never do). Full system access means access to firejail related configs too which means there are no any security guarantees anymore so why bother? Also apparmor support isn't flexible so it would make it incompatible with those profiles.

To be clear - this PR isn't about preventing to use those profiles with firejail but to prevent enabling them without user consent. We should still try to make them work in different usecases but sometimes disabling them is the only option.

ThePythonicCow:
Thanks for the rapid resolution of the two reports I filed, #1800 and #1801.  I agree with the conclusions reached and decisions made.  Well done - thanks!

Vincent43:
@ThePythonicCow actually there are not any conclusions or decisions made yet. :smile:

Ferroin:
Just my thoughts on this:

* I see near zero value to globally sandboxing wget.  Unlike a full web browser it doesn't parse any of what it fetches by default (you can do such parsing, but it's not widely used), and unlike things such as cURL or Aria2, it doesn't handle things like metalinks (which can be forged to trigger connection attempts to arbitrary systems) or torrent files (which have their own issues), so the only realistic attacks against it involve either some serious social engineering (which firejail can't really protect against) or MITM-style attacks (which firejail is of debatable use in defending against).
* I similarly see very little value in sandboxing `less` by default.  It does do some parsing, but it's not really a case where it's realistic to expect it to be exploitable through any means other than social engineering or scripting simply because it doesn't touch the network itself and is ridiculously well reviewed code.
* Regarding Kate and Dolphin, I'm not really sure here.  Sandboxing a file manager beyond blocking network access without just restricting the whole user account with something like SELinux seems a bit odd to me.  Similar for a text editor.

reinerh:
I think wget should continue to be protected by default.
There have already been a bunch of security vulnerabilities in the past (https://security-tracker.debian.org/tracker/source-package/wget).

Vincent43:
@reinerh it breaks updating on some systems which is much worse than any vuln it may have. Users can still enable protection at any time they want.

reinerh:
Is it known what exactly is breaking updating? Is the package manager downloading the files to a blacklisted location or so?
IMO fixing the profile to be more graceful to package managers would be better than disabling it.

Vincent43:
I think we can do both. Enable it only when it's tested on various systems. Breaking it and then wait for users to open issues is IMO wrong approach. It will discourage users to use firejail.

netblue30:
Ok guys, thanks for your input. Let's reset and start over.

We are breaking root user badly: wget, less, most text editors... The easy way out is to disable /usr/local/bin symlink processing in firejail code, if the user is root. Instead, we start the original command without any sandbox. In other words, no desktop integration for root user - root will have to run the regular "firejail application-name" to get it sandboxed.

For regular users the symlink processing will still be in place, so we can keep wget, less etc. in /usr/lib/firejail/firecfg.config, maybe add some other programs there that were taken out because of root.

I'll put in the fix over the weekend.

BTW, I've seen something interesting on CentOS yesterday. Look how they set $PATH by default for regular and root users:
`````
$ env | grep PATH
PATH=/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/netblue/.local/bin:/home/netblue/bin
$ sudo env | grep PATH
PATH=/sbin:/bin:/usr/sbin:/usr/bin
`````
They don't include /usr/local/bin in the root PATH.

Vincent43:
What if application elevates privileges only for certain commands, i.e. uses wget as non-root but then want to switch to root?

And what about kate and dolphin which elevate privileges only on demnand through polkit?

ThePythonicCow:
netblue30 wrote:
>> The easy way out is to disable /usr/local/bin symlink processing in
>> firejail code, if the user is root.
I rather like the idea of having firejail protection whether I'm running
as root or not, except when itbreaks something, of course <grin>.

Also, in my particular case (which is not the normal case, so perhaps
not worth your coding for)I run with mostly the same environment and PATH whether root or not.
I've been running thatway for 40 years now, and I am accustomed to arranging my system so that
I have securitysufficient for my needs while doing this.

I don't understand what you mean by disabling "/usr/local/bin symlink
processing in firejail code".
Does that mean that still having the /usr/local/bin firejail symlinks at
the head of my $PATH willbreak every command so symlinked?

Or perhaps that means that root can still exec some
/usr/local/bin/whatevercommand ->  /usr/bin/firejailsymlink, but that the result of doing so will be (in the case geteuid()
== 0) the same as if that root userhad invoked "whatevercommand" without /usr/local/bin in their $PATH.

Or perhaps it means something else?

In any case it means that I will be, if necessary, looking to construct
some sort of workaround or wrapper,so that I can:
  1) enjoy the benefits of firejail whether root or not,
  2) enjoy the benefits of the automatic firejail symlink on all
     commands configured for it, except      perhaps for those commands (such as wget when invoked by emerge,
      and less when invoked      as root) that I have determined don't work as root for my setup
      and usage patterns, and  3) run with the same PATH and other environment, whether ROOT or not,
     except perhaps as      part of the special case wrappers needed to make (1) and (2)
      work for me.
So ... I would encourage you to leave the door sufficiently open, and
the chosen implementationsufficiently well documented, that crufty old hackers with idiosyncratic
habits such as myself canmake firejail work for them, without extraordinary effort.

I was already thinking of coding some shell script wrappers for emerge,
wget, and/or less, toaccomplish this with the current firejail, better than my current
workaround of just removingthe wget and less symlinks from /usr/local/bin, but then when I realized
these design issueswere being given significant thought, I figured I'll just wait for the
firejail dust to settle first,before hacking whatever I need to hack to fit my situation.

Thanks for the considerable attention being given to these details, and
for a quite useful tool!
-- Paul Jackson pj@usa.net

Vincent43:
@Fred-Barclay Hm, I tested `firejail kate /etc/firejail/kate.profile` and it works! Changes are saved, even as `/etc` is mounted read-only. It escapes sandbox in some way.

That means we can relax the profiles instead of disabling them. `~/.config/firejail` is the only case which stays inaccessible even with `--noprofile`.

Considering above, I propose this:
1. Disable blacklists/whitelists in kate/dolphin profiles which will make them usable for system administration
2. Deal with wget/less the way which @netblue30 proposed (they will be unconstrained for root user)

What do you think? I can make PR for 1. I'm closing this as I don't want to implement it anymore but feel free to continue discussion here.

Ferroin:
Just handling wget as @netblue30 proposed may not fix the issue with Gentoo's package manager universally.  Whether or not it gets run as root is up to the user, it may istead be run as `portage` (the user name for the package manager itself, so dependent on what exactly is breaking (and I'm still not quite sure myself even having gone through the wget profile), it may still be an issue (and firejail **should not** be hard coding user names or UID's other than root).

ThePythonicCow:
Austin wrote:
>> Just handling wget as @netblue30 proposed may not fix the issue with
>> Gentoo's package>> manager universally. Whether or not it gets run as root is up to the
>> user, it may istead be>> run as portage

From what I can tell, looking at the Gentoo portage code for a
few minutes,fetching packages (which is where it uses wget) normally requires
superuser(geteuid() == 0), except in the special case of using the Gentoo
Project Prefix(https://wiki.gentoo.org/wiki/Project:Prefix) profile.  "Prefix" is
primarily usedon "non-standard" operating systems such as Mac OS, Windows and Sparc.

My impression is that the default configurations of emerge
require superuserpermissions when running emerge to fetch (download using wget) packageson Linux, which is where firejail would be encountered.

If my impressions are correct, then anyone determined enough to hack
the portage emerge command so that it could fetch packages on Linux
even if not superuser would be someone who would not require a default
firejail configuration to handle such an unusual use case.

-- Paul Jackson  pj@usa.net

Ferroin:
@ThePythonicCow While correct that you need to be running as root to fetch packages, wget may not be run as the root user  If the `userfetch` feature is enabled (I do not remember if it is by default or not, but enabling it is considered standard practice for properly securing a Gentoo system), then the wget command will not be run as the root user, but as the significantly restricted `portage` user that also gets used for other low-privilege operations that don't require root (notably, used for actually building packages if the `userpriv` feature is enabled, which is also considered standard practice for system security), so simply disabling firejail if wget gets invoked with an effective user ID of 0 may not be enough to fix the interaction with Portage.

ThePythonicCow:
Thanks for the explanations, Austin.

You inspired me to look at little further into my Gentoo Linux box, and
I find that"emerge --info" shows that my Portage has both userpriv and
userfetch set.
These are enabled in the /usr/share/portage/config/make.globals
file.  That'sthe way they came "from the factory", so far as I recall.

If, as root, I ask to install something that I don't already have
installed, such as
  emerge --ask dev-perl/Pango

then I can see from strace that (thanks to the "userfetch", I
presume) theprocess uid and gid are changed from root (0) to portage (250)
just beforerunning wget.

Then, as I reported in the earlier bugs, from which this bug is
descendant,the emerge fails if wget is a symlink to firejail, because the
file that wgetdownloads, which ends up below /var/tmp/portage/ thanks to firejail,
ends up not being available to the invoking emerge code, after the wgetcompletes successfully.

-- Paul Jackson pj@usa.net

netblue30:
@ThePythonicCow

The fix is in git, you can give it a try. This is the commit: https://github.com/netblue30/firejail/commit/37bc54cef3d156a9e9c0363c11c93e9de0b9f3e0

SkewedZeppelin:
@netblue30 Should that be optional via firejail.config?

ThePythonicCow:
Hmm ... that commit 37bc54c doesn't seem to work for me :).

The following code in that commit assumes that wget is being run as
root,when invoked from my Portage emerge installation command.

+ // desktop integration is not supported for root user; instead, the
  original program is started+ if (getuid() == 0) {
+    argv[0] = program;
+    execv(program, argv);
+    exit(1);
+ }

However, Portage emerge changes uid:gid from root:root to
portage:portage,just before calling execve of wget.

Then (if I have the wget symlink to firejail first on my $PATH),
firejail doesits thing, wrapping the real wget, and by the time that the real wget
is called,the directory it wants to download into,
"/usr/portage/distfiles", is notwritable, and the open() call to write the downloaded file into
that directoryfails with EROFS (Read-only file system).

Sorry :)

-- Paul Jackson pj@usa.net

PR #1820, opened 2018-03-18
Fix compilation with musl

g3ngr33n:
https://github.com/gentoo/musl/issues/109


src/libtrace/libtrace.c
```diff
@@ -30,7 +30,9 @@
 #include <arpa/inet.h>
 #include <sys/un.h>
 #include <sys/stat.h>
-#include <dirent.h>
+#include <syslog.h>
+ #include <dirent.h>
```
SkewedZeppelin: remove the erroneous space?

Issue #1822, opened 2018-03-18
Recalibration of D-Bus access


smitsohu:
We have three different options that govern whether an app is allowed access to D-Bus or not....

- `apparmor`: reliably shuts down all D-Bus access in Ubuntu and its derivatives
- `net none`: denies access to the session bus in Ubuntu 16.04 (Mint!), Debian Jessie.... in general distributions that were released some time ago. Also kills connection to the accessibility bus.
- `blacklist /run/user/*/bus`: denies access to the session bus in newer distribution releases.

Now that we are adding `apparmor` to many profiles (184ecb305e81942fa9218f14a93f584fbeaff230, 38ffc3f02575ee91fe0dbf92e6b8ebc93095e667, 971c62aa569d9161190705a0012b9ad02546822c, d8b4a633202a13a13c75779d1f40a99d6cc51dfb), I thought it might be a good timepoint to take a step back and reconsider how we want to address the D-Bus issue.

In a number profiles right now only one or two or these three options are enabled, which seems to be, at first glance at least, an inconsistent state. The rigorous approach would be to go through these profiles and either enable them all or comment them all (`net none` obviously can't go to some profiles).... that's IMHO the only way to ensure a smooth and consistent user experience _everywhere_. 

But then, as many Ubuntu (and Mint) users are expected to switch to upcoming Ubuntu 18.04 LTS, the number of people who may run into problems with `net none` should decrease considerably. Also the KDE file chooser works without D-Bus since recently. And finally, given how big a security hole D-Bus is, the ability to change application settings could also be considered minor in comparison (thinking of apps like eog or gedit).

Personally I have a lot of sympathies for a stricter approach to D-Bus. The question is if we are willing to move ahead even if it will cause problems for users with older software. Also I think we should try to combine `apparmor` with `blacklist /run/user/*/bus` (and `net none` where possible) to ensure consistently working profiles.

Vincent43:
Some thoughts:
`net none` affecting dbus is coincidental. It can't be used when application need network access.
`apparmor` isn't available in all distributions so we can't rely on it alone.

smitsohu:
@Vincent43 
> net none affecting dbus is coincidental. It can't be used when application need network access.

That's right, I've updated my original post.

Fred-Barclay:
@smitsohu and @Vincent43 Do you know why net none doesn't stop dbus access on newer distros?

smitsohu:
Replying to myself @smitsohu
[...]

PR #1825, opened 2018-03-22
WIP: Recalibration of D-Bus access, see #1822

smitsohu:
A few notes:

- Atril I found broken without D-Bus, probably because it can't talk to its server process atrild anymore. Same for Xreader.
- SMPlayer uses the KDE file dialog, so we need to treat it like other KDE apps.
- Audacity, Kodi are not yet included here.

Still I am a bit unsure if we really can freeze application settings in some cases. Looking at the rich settings that gedit or rhythmbox expose (for a reason), can we make exceptions case by case?

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -75,7 +75,7 @@ distclean: clean
 	for dir in $(APPS) $(MYLIBS); do \
 		$(MAKE) -C $$dir distclean; \
 	done
-	rm -fr Makefile autom4te.cache config.log config.status config.h uids.h
+	rm -fr Makefile autom4te.cache config.log config.status config.h uids.h dummy.o
 
 realinstall:
 	# firejail executable
@@ -195,7 +195,7 @@ uninstall:
 	rm -f $(DESTDIR)/$(datarootdir)/bash-completion/completions/firemon
 	rm -f $(DESTDIR)/$(datarootdir)/bash-completion/completions/firecfg
 
-DISTFILES = "src etc platform contrib configure configure.ac Makefile.in install.sh mkman.sh mketc.sh mkdeb.sh mkuid.sh COPYING README RELNOTES"
+DISTFILES = "src etc platform contrib configure configure.ac dummy.c Makefile.in install.sh mkman.sh mketc.sh mkdeb.sh mkuid.sh COPYING README RELNOTES"
 DISTFILES_TEST = "test/apps test/apps-x11 test/apps-x11-xorg test/root test/fcopy test/environment test/profiles test/utils test/compile test/filters test/network test/arguments test/fs test/sysutils test/chroot"
 
 dist:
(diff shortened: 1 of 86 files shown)


Fred-Barclay:
@smitsohu What about if we added a new file, disable-dbus.inc, with the lines:
```
apparmor
blacklist /run/user/*/bus
(*possibly* net none)
```
that way if a user needs to enable dbus for an application, they can just comment out one line instead of two to three, depending on which distro and version they're running?

smitsohu:
@Fred-Barclay Yep. You are right, we should try to make it simpler.  We could do as you suggest, or alternatively maybe introduce a new profile option `nodbus`. For example we could parse the `protocol` line in order to figure out when installing a new disconnected network namespace is possible (basically all profiles have this line). Then we could check if $DBUS_SESSION_BUS_ADDRESS is a file and blacklist it. If $DBUS_SESSION_BUS_ADDRESS is not set we try to blacklist /run/user/$UID/bus, the default location.

startx2017:
Voting for --nodbus, I'll give it a try this weekend.

startx2017:
I have --nodbus on command line and "nodbus" in profiles. It will look for network namespaces and "unix" on protocol line and it will print a warning:
`````
$ firejail --nodbus

Parent pid 6313, child pid 6314
Child process initialized in 46.72 ms
`````
Commit: https://github.com/netblue30/firejail/commit/ae008e5fa9e8a901fbf255664f3de775415a39a3

startx2017:
New commit: https://github.com/netblue30/firejail/commit/1ff6851a1a8dcb89f6cb4097b765167b0949a982

Fred-Barclay:
@startx2017 Nicely done! Chrome and gpredict, at least, are working fine on my system with `--nodbus`.

netblue30:
Quite cool, thanks! Let's start adding nodbus in the profiles, if something comes up we fix it in the code.

I tested it with gnome-terminal. gnome-terminal sends all its commands over dbus to be executed outside the sandbox - we blacklisted /usr/bin/gnome-terminal for this reason in disable-common.inc.

This is how it looks on Debian 7 (2013), Debian 9 (2017) and Arch:
`````
$ firejail --nodbus --noblacklist=/usr/bin/gnome-terminal --noblacklist=/bin/gnome-terminal gnome-terminal

Parent pid 2254, child pid 2255
Child process initialized in 40.03 ms
Error constructing proxy for org.gnome.Terminal:/org/gnome/Terminal/Factory0: Unknown or unsupported transport 'DBUS_SESSION_BUS_ADDRESS=unix' for address 'DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus'

`````

Fred-Barclay:
@smitsohu I meant to add just one commit (001c7e62c7703e80024b00d9dab7206d1deda985) to your branch but it looks like I added a whole bunch by mistake!

smitsohu:
@Fred-Barclay :smile: No worries! I can just delete the branch, once you give me green light. No need to tidy up anything.

Fred-Barclay:
Gopher it!

Issue #1711, opened 2018-01-02
Evince does not run due to --private-lib configuration option


elvetemedve:
### The problem
The **Evice Document Viewer** application does not start, because **private-lib** option in **evince.profile** prevents reading the file **/usr/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache**.

See [evince_log.txt](https://github.com/netblue30/firejail/files/1598834/evince_log.txt) for details.

### Environment
OS: Arch Linux 4.14.9-1-ARCH
Firejail version: 0.9.52
Evince version: 3.26.0

netblue30:
I cannot reproduce it on Arch running LXDE. Do you have  /usr/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache in your filesystem? They say in the log to create it by running
`````
gdk-pixbuf-query-loaders > /usr/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache
`````

**Comment by elvetemedve** [comment]
Yes, the file is present in the filesytem. But the private-lib switch makes it invisible for Evince.

**Comment by netblue30** [comment]
I'll try again to reproduce it, so far no luck. What window manager are you using?

**Comment by elvetemedve** [comment]
I'm using Gnome Shell (3.26.2) desktop environment. Window manager is Mutter.

**Comment by netblue30** [comment]
I have removed private-lib from the profile until we figure out what's going on. Thanks for the bug.

**Comment by netblue30** [comment]
Re-enabling private-lib temporarily in evince profile.

We found some problem with 32bit libraries being copied by private-lib instead of the regular 64bit libraries. @elvetemedve, can you give it a try please - we will take it out if it doesn't work. Thanks!

**Comment by Fred-Barclay** [comment]
I'm getting this error now when trying to open a file with evince on Fedora 27 amd64:
```
 firejail evince
Reading profile /etc/firejail/evince.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Parent pid 10632, child pid 10633
Private /etc installed in 5.42 ms
Standard C library installed in 83.23 ms
Program libraries installed in 236.50 ms
GdkPixbuf installed in 79.83 ms
GTK3 installed in 145.03 ms
Pango installed in 0.01 ms
[...]

PR #1828, opened 2018-03-24
gnome-recipes profile

glitsj16:
Not a very widely used app perhaps, but tasty nonetheless... Leaving out the `private-lib` part, as per the comment inside. I noticed some issues regarding `private-lib` and Gnome apps in general, [evince](https://github.com/netblue30/firejail/issues/1711) in particular. Will open a new issue on that shortly, I *think* I made some progress in that regard. More eyes needed though.


SkewedZeppelin:
Merged, thanks!

PR #1829, opened 2018-03-24
evince fixes

glitsj16:
Seems the current evince profile has issues with dconf, user icons and themes. These can all be easily fixed by including `whitelist-common.inc`. Also adding 2 redirect profiles for evince-previewer and evince-thumbnailer, both executables are included in evince as packaged on Arch Linux (and probably other distro's too).

After seeing [#1711](https://github.com/netblue30/firejail/issues/1711), I took another crack at trying to fix the `private-lib` condition, something I had fiddled with before without succes. But this time it looks as if things are settling nicely in that regard. I added what works for me after extensive testing. Leaving commented, so other people can take a closer look.

For what it's worth, these all work, ordered from less to more granular/restrictive:

- private-lib dconf,evince,GConf,gdk-pixbuf-2.0,gio,gvfs,libgconf-2.so.4,libpoppler-glib.so.8,librsvg-2.so.2,libxml2.so.2
- private-lib evince,gdk-pixbuf-2.0,gio,gvfs,libgconf-2.so.4,libpoppler-glib.so.8,librsvg-2.so.2,libxml2.so.2
- private-lib evince,gdk-pixbuf-2.0,gio,gvfs,libgconf-2.so.4,libpoppler-glib.so.8,librsvg-2.so.2
- private-lib evince,gdk-pixbuf-2.0,gio,gvfs/libgvfscommon.so,libgconf-2.so.4,libpoppler-glib.so.8,librsvg-2.so.2

### DIFF
--- a/etc/evince.profile
+++ b/etc/evince.profile
@@ -14,7 +14,6 @@ include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-passwdmgr.inc
 include /etc/firejail/disable-programs.inc
 
-include /etc/firejail/whitelist-common.inc
 include /etc/firejail/whitelist-var-common.inc
 
 caps.drop all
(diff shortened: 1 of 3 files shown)


etc/evince.profile
```diff
@@ -14,6 +14,7 @@ include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-passwdmgr.inc
 include /etc/firejail/disable-programs.inc
 
+include /etc/firejail/whitelist-common.inc
```
SkewedZeppelin: Including this would completely prevent any files (PDFs) from being accessed.
Remove this and I'll merge

glitsj16: I'll take another look at what I did here regarding whitelist-common.inc. Thanks for enlightening me on this. Have to take a break for a few hours, but I'll pick things up later today.

netblue30:
Merged, thanks.

PR #1830, opened 2018-03-24
whitelist-common.inc fixes for several profiles

glitsj16:
Some apps I use regularly are showing user icon, theme-related and dconf settings issues. Had things fixed in local profiles (gathering dust), time to clean those out and make a contribution. These apps were all lacking the generic `whitelist-common.inc` file.


SkewedZeppelin:
None of those profiles whitelist anything, including whitelist-common.inc will actually turn them into a whitelist profile and then break them.

glitsj16:
Looking at this the wrong way it seems, apologies for that. Will go over what I did in user profiles and try to fully understand. Really appreciate the insights. Thank you.

PR #1831, opened 2018-03-24
private-lib fix

glitsj16:
Using Arch Linux (Gnome Shell 3.26.2, Mutter WM, as in [#1711](https://github.com/netblue30/firejail/issues/1711)). After playing with several Gnome apps and `private-lib` conditions, it looks like there's progress to report. I made a few PR's today on the same topic, usually leaving things commented as to leave room for more eyes to double-check. In this case I took the liberty to throw in an uncommented one for eog. Please respond or rectify if this was uncalled for.

### DIFF
--- a/etc/eog.profile
+++ b/etc/eog.profile
@@ -37,7 +37,7 @@ shell none
 private-bin eog
 private-dev
 private-etc fonts
-private-lib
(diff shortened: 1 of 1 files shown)


Fred-Barclay:
Merged, thanks!

PR #1840, opened 2018-03-26
fix immutable settings

glitsj16:
On Arch Linux I'm seeing the below with latest firejail from git:

$ firejail gimp
Reading profile /home/glitsj16/.config/firejail/gimp.profile
Reading profile /etc/firejail/gimp.profile
Reading profile /etc/firejail/globals.local
Reading profile /etc/firejail/disable-passwdmgr.local
Parent pid 22315, child pid 22316

Child process initialized in 317.16 ms
Could not connect: Permission denied
Failed to connect to socket /run/user/1001/bus: Permission denied

### DIFF
--- a/etc/gimp.profile
+++ b/etc/gimp.profile
@@ -5,7 +5,7 @@ include /etc/firejail/gimp.local
 # Persistent global definitions
 include /etc/firejail/globals.local
 
-blacklist /run/user/*/bus
+# blacklist /run/user/*/bus - makes settings immutable
 
 noblacklist ${HOME}/.gimp*
 


SkewedZeppelin:
D-Bus access is currently being overhauled, see here #1822 
and @startx2017 just made progress by adding a new --nodbus option

Fred-Barclay:
Nicely done! But as @SkewedZeppelin said, we're actually in the middle of overhauling all that. :smile:

Vincent43:
It looks like dbus access is needed for gimp so `nodbus` won't help here.

smitsohu:
@glitsj16 Coming back to your issue, could I ask which platform exactly you are running. Only so I can try to replicate the problem. I see the warning message you are citing on different distros, but saving settings seems to work fine for me. Thanks :smile:

glitsj16:
@smitsohu No problem. As stated I'm running Arch Linux as my main distro. Also using Ubuntu 16.04. And that was probably where I tripped up. On Arch linux I have gimp 2.8.22-1. And I can `confirm` that saving settings does indeed work fine, regardless of the warning messages. On Ubuntu I tested gimp 2.9+ from a PPA recently. In an attempt to save a few bytes I shared some cache dirs and my firejail user profile between both systems via a symlinked dir. This exercise in futility left a stale `whitelist ${HOME}/foo` entree in ~/.config/firejail/gimp.profile that broke most of the settings in /etc/firejail/gimp.profile.

Returning to sanity and cleaning that out turns everything back to work as expected. My apologies for the confusion. This PR was closed legitimately and rightly so :smile:. Thanks for looking into this and for reminding me to not jump the gun when doing PR's without preparing a clean context.

smitsohu:
I see, thanks for the update :+1:

PR #1843, opened 2018-03-28
recalibrate dbus access, deploy nodbus option

smitsohu:
see #1822 and #1825.

- systematically replaces `blacklist /run/user/*/bus` with `nodbus` in all profiles
- enables `nodbus` in profiles with `apparmor` (except the app uses the KDE file dialog)
- moving forward, the "settings immutable" warning is omitted for `net none`. As less and less users are affected by the side-effects of `net none`, the hope is this helps to reduce confusion (rather than add to it).
- comments `apparmor` in the kate profile, based on the assumption that the execution restrictions break workflows relying on the embedded compiler or terminal window
- with contributions from @Fred-Barclay (001c7e62c7703e80024b00d9dab7206d1deda985)

### DIFF
--- a/etc/audacity.profile
+++ b/etc/audacity.profile
@@ -5,8 +5,6 @@ include /etc/firejail/audacity.local
 # Persistent global definitions
 include /etc/firejail/globals.local
 
-blacklist /run/user/*/bus
-
 noblacklist ${HOME}/.audacity-data
 
 include /etc/firejail/disable-common.inc
@@ -18,8 +16,9 @@ include /etc/firejail/whitelist-var-common.inc
 
 apparmor
 caps.drop all
-#net none
+net none
 no3d
+# nodbus
 nodvd
 nogroups
 nonewprivs
--- a/etc/audacity.profile
+++ b/etc/audacity.profile
@@ -18,7 +18,7 @@ apparmor
 caps.drop all
 net none
 no3d
-# nodbus
+# nodbus - problems on Fedora 27
 nodvd
 nogroups
 nonewprivs
--- a/etc/eog.profile
+++ b/etc/eog.profile
@@ -17,13 +17,11 @@ include /etc/firejail/disable-programs.inc
 
 include /etc/firejail/whitelist-var-common.inc
 
-# following line makes settings immutable
-apparmor
+# apparmor - makes settings immutable
 caps.drop all
-net none
+# net none - makes settings immutable
 no3d
-# following line makes settings immutable
-# nodbus
+# nodbus - makes settings immutable
 nodvd
 nogroups
 nonewprivs
--- a/etc/eog.profile
+++ b/etc/eog.profile
@@ -23,7 +23,7 @@ caps.drop all
 net none
 no3d
 # following line makes settings immutable
-nodbus
+# nodbus
 nodvd
 nogroups
 nonewprivs
--- a/etc/eog.profile
+++ b/etc/eog.profile
@@ -5,8 +5,6 @@ include /etc/firejail/eog.local
 # Persistent global definitions
 include /etc/firejail/globals.local
 
-# blacklist /run/user/*/bus - makes settings immutable
-
 noblacklist ${HOME}/.Steam
 noblacklist ${HOME}/.config/eog
 noblacklist ${HOME}/.local/share/Trash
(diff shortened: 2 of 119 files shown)


etc/scribus.profile
```diff
@@ -48,3 +47,6 @@ tracelog
 # private-bin scribus,gs,gimp*
 private-dev
 private-tmp
+
+# noexec ${HOME}
```
SkewedZeppelin: does it really need exec from home?

smitsohu: Well spotted! Just confirmed that scripting engine works with noexec, I will change it. Thanks.

etc/gimp.profile
```diff
@@ -18,6 +16,7 @@ include /etc/firejail/whitelist-var-common.inc
 apparmor
 caps.drop all
 net none
+nodbus
```
Vincent43: If I understand this correctly nodbus makes settings immutable here - https://github.com/netblue30/firejail/pull/1840

Vincent43:
I'm a little concerned about `nodbus` and immutable settings. Being unable to change app preferences will break usability. I think in such cases `nodbus` should be commented out by default.

etc/gimp.profile
smitsohu: I'll add a comment there. Thanks.

Fred-Barclay:
I normally go for a little extra security at the cost of some minor usability, but in this case I think I agree with @Vincent43 . I'm a bit concerned about immutable settings.

What if we cherry-picked the profiles from here that are good to go, and left all the ones where `nodbus` causes immutable settings, to think about and discuss for a bit before making a decision? I'd like to see the ones that are fine in the main branch asap. :smile:

smitsohu:
@Vincent43 @Fred-Barclay But shouldn't we then also disable `apparmor` in those profiles? `nodbus` is part of the `apparmor` feature.

Anyways, I'll take everything out that wants to write into dconf, so we can think about a bit longer.

Vincent43:
Yeah, all things that make app settings immutable should be commented out IMO. Apparmor blocks dbus only in Ubuntu but it's quite popular distro.

smitsohu:
All programs that write settings to dconf are now the same as before the patch. The only thing that changes is that `blacklist /run/user/*/bus` is still replaced with `nodbus`.

I left the immutability warning in to find the profiles back more easily.

etc/audacity.profile
```diff
@@ -18,8 +16,9 @@ include /etc/firejail/whitelist-var-common.inc
 
 apparmor
 caps.drop all
-#net none
+net none
 no3d
+# nodbus
```
Fred-Barclay: Should we uncomment this since /run/user/*/bus wasn't commented?

etc/eog.profile
```diff
@@ -19,10 +17,13 @@ include /etc/firejail/disable-programs.inc
 
 include /etc/firejail/whitelist-var-common.inc
 
+# following line makes settings immutable
 apparmor
 caps.drop all
-# net none - makes settings immutable
+net none
```
Fred-Barclay: Should we leave `net none` commented?

etc/gedit.profile
```diff
@@ -18,10 +16,14 @@ include /etc/firejail/disable-programs.inc
 
 include /etc/firejail/whitelist-var-common.inc
 
+# following line makes settings immutable
+apparmor
 caps.drop all
-# net none - makes settings immutable
 machine-id
+net none
```
Fred-Barclay: Should we comment out `apparmor` and `net none` to avoid immutable settings?

etc/audacity.profile
smitsohu: It also makes settings immutable, but only in Fedora Gnome. I think @SkewedZeppelin observed this first. I can add a comment there to make it clearer.

etc/gedit.profile
smitsohu: Oops! I overlooked `net none` when I claimed that everything is reverted to the situation before. Sorry!

To summarize: `net none` will continue to affect Mint 18, Ubuntu 16.04, CentOS... I guess that's about it. Debian Jessie will reach EOL soon. With Mint 19 and the next Ubuntu LTS waiting, I would find it defensible to enable `net none` by default in these profiles, but we can also comment it again, then truly reverting to the original situation.

Regarding apparmor I personally would tend to a yes, also because, as @Vincent43 has pointed out in another thread, patches are finally in the pipeline to bring D-Bus mediation into the mainline kernel. Any other opinions regarding apparmor?

smitsohu: So do I understand it right that consensus is emerging to disable all of `apparmor`, `nodbus`, `net none` when they freeze application settings?

(in case yes, how about an exception for the archive managers file-roller and engrampa? Because they have hardly any setting to change....)

Fred-Barclay: I think we should do that with this release and then re-think our approach afterwards.
Let's do that exception for the archive managers!

PR #1848, opened 2018-03-31
Add a section for cloud-providers to the default block-list

jelford:
AWS and GCP store credentials in local directories as part of project setup.

Configuration for cloud providers is sensitive information; it should be
in the default block list. I didn't see profiles for gcloud or awscli,
so haven't added any exclusions.
    
boto and kubectl are not provider-specific, but also store credentials for
whichever platforms they happen to be being used with.


Vincent43:
Seems legit, thank you.

PR #1849, opened 2018-03-31
Added a basic profile for gcloud

jelford:
Hi, another quick one hopefully. I tested the profile out with the following workflows:
* `gcloud init`
* `gcloud docker -- push ...`
* `gcloud compute instances list`

All seemed okay; I think the docker integration is the only thing that requires any "interesting" permissions, and the rest is just speaking to an API.

gcloud is the Google Cloud Platform's cli. It understands how to interact with GCP's endpoint, and it need to read config and creds from the user's $HOME, plus it integrates with docker-cli.

### DIFF
--- a/etc/gcloud.profile
+++ b/etc/gcloud.profile
@@ -1,0 +1,40 @@
+# Firejail profile for gcloud
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /usr/local/etc/firejail/gcloud.local
+# Persistent global definitions
+include /usr/local/etc/firejail/globals.local
+
+noblacklist ${HOME}/.boto
+noblacklist ${HOME}/.config/gcloud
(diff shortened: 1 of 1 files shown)


etc/gcloud.profile
```diff
@@ -0,0 +1,40 @@
+# Firejail profile for gcloud
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /usr/local/etc/firejail/gcloud.local
```
Vincent43: Please drop `/usr/local` prefix here and everywhere below.

SkewedZeppelin:
Merged, thanks!

jelford:
Thanks, and thanks for doing the `/usr/local` prefix fixup too. Noted
for next time.

PR #1852, opened 2018-04-01
Fix private-lib again

glitsj16:
During conversation on an [earlier attempt at fixing evince](https://github.com/netblue30/firejail/pull/1829) and the entailing [revert-commit](https://github.com/netblue30/firejail/pull/1829/commits/45732a22d1ea4ec0ade0775be7243e8669b7f850) this slipped through the cracks. The fix is tested on Gnome Shell 3.26.2 and 3.28.0 (on Arch).

I noticed the original issue #1711 is closed and tagged `in testing`. Feel free to postpone/close/disregard this fix untill testing is finished. Just wanted to remind people that a fix is available.

### DIFF
--- a/etc/evince.profile
+++ b/etc/evince.profile
@@ -38,6 +38,8 @@ private-etc fonts
 
 #private-lib - seems to be breaking on Gnome Shell 3.26.2, Mutter WM, issue 1711
 #private-lib evince,libpoppler-glib.so.8
+# private-lib fix slipped through the cracks of earlier, reverted commit
+private-lib evince,gdk-pixbuf-2.0,gio,gvfs/libgvfscommon.so,libgconf-2.so.4,libpoppler-glib.so.8,librsvg-2.so.2
 
 private-tmp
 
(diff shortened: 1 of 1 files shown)


etc/evince.profile
```diff
@@ -38,6 +38,8 @@ private-etc fonts
 
 #private-lib - seems to be breaking on Gnome Shell 3.26.2, Mutter WM, issue 1711
 #private-lib evince,libpoppler-glib.so.8
+# private-lib fix slipped through the cracks of earlier, reverted commit
```
SkewedZeppelin: should this comment be in here?

glitsj16: Hi @SkewedZeppelin. That comment is indeed superfluous and was more of a note-to-self kind of thing. An old habit I'll have to shake. Thanks for being thorough and raising the question here, exactly the 'spirit' I've come to appreciate as a day-to-day user of firejail.  I'll take it out asap.

etc/evince.profile
```diff
@@ -38,7 +38,6 @@ private-etc fonts
 
 #private-lib - seems to be breaking on Gnome Shell 3.26.2, Mutter WM, issue 1711
 #private-lib evince,libpoppler-glib.so.8
```
Fred-Barclay: You can delete this line since it's being replaced by the new private-lib filter. ;)

PR #1853, opened 2018-04-01
Create gnome-logs.profile

glitsj16:
New profile. A few harmless warnings do show, but nothing that doesn't also show without firejail.

### DIFF
--- a/README
+++ b/README
@@ -246,7 +246,7 @@ geg2048 (https://github.com/geg2048)
 	- kwallet profile fixes
 glitsj16 (https://github.com/glitsj16)
 	- evince-previewer, evince-thumbnailer profiles
-	- gnome-recipes profile
+	- gnome-recipes, gnome-logs profiles
 graywolf (https://github.com/graywolf)
 	- spelling fix
 greigdp (https://github.com/greigdp)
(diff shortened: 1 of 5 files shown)


SkewedZeppelin:
I had to make some tweaks to get it to actually show logs, but other then that it is fine.

In the future, /var/log is empty unless `writable-var-log` is specified. And netfilter isn't needed when you have `net none`.

Merged, thanks

glitsj16:
Hi @SkewedZeppelin, thanks for your review and for sharing those insights. Added to the notes. :+1: Regarding  /var/log. Your comment triggered the old memory bank. Quite some time ago I made a few commits to gnome-logs, basically adding support for [volatile](https://github.com/GNOME/gnome-logs/commit/1dffafc58fc8114c7d5e93f7e1341364ff1b9a7e) systemd journals. As my main work machine is a laptop, that is still the way I use systemd up untill the present (for better or worse).

Those volatile journal files are stored at `/run/log/journal` instead of /var/log, probably why I didn't pay attention to this as much as I should have. Another lesson learned. I'll go over firejail's code in that context, to actually check how /run/log is being treated exactly. Thanks again for pointing this all out so kindly and patiently.

PR #1854, opened 2018-04-01
add --noautopulse arg for complex pulse setups

### LINKED ISSUE #1028: ~/.config/pulse cannot be whitelisted (always overriden unless nosound)
Currently, there doesn't seem to be a way to tell firejail that I don't want it to mess with ~/.config/pulse. Either I can totally disable anything sound related using nosound, or it will just build it's own ~/.config/pulse in pulseaudio_init(). 

It would be really nice to have an option to whitelist/noblacklist that directory and let me shoot myself in the foot. I understand the defaults might work for most cases, but don't work for my case (remote playback, no spawning of pulseaudio, set up locally for the specific user) and I currently don't have any recourse to fix it.

melvinvermeeren:
such as remote pulse servers or non-standard socket paths

closes #1028

### DIFF
--- a/RELNOTES
+++ b/RELNOTES
@@ -1,5 +1,6 @@
 firejail (0.9.53) baseline; urgency=low
   * work in progress
+  * add --noautopulse to disable automatic ~/.config/pulse (for complex setups)
   * modif: support for private-bin, private-lib and shell none has been
      disabled while running AppImage archives in order to be able to use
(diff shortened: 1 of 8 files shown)


netblue30:
Merged, thanks.

netblue30:
I will also add support today to disable it globally in /etc/firejail/firejail.config, thanks for the patch.

PR #1859, opened 2018-04-04
Allow perl

glitsj16:
Now there's  `disable-interpreters.inc` atool profile needs allowing `perl` access.


SkewedZeppelin:
Merged, thanks!

PR #1860, opened 2018-04-04
syntax fixup

glitsj16:
Using `PKGBUILD` instead of `pckgbuild` might be slightly more familiar for Arch users.


SkewedZeppelin:
Merged, thanks!

PR #1861, opened 2018-04-04
mpd config modernizing

glitsj16:
Current mpd profile uses outdated ${HOME}/.mpdconf noblacklist location. Not sure that was ever actually used (I remember ~/.mpd). Bring in support for freedesktop location `${HOME}/.config/mpd`, as per [official documentation](https://www.musicpd.org/doc/user/config.html#config_file).


SkewedZeppelin:
Merged, thanks!

PR #1864, opened 2018-04-05
Update enchant profile

glitsj16:
Added `nodbus` and brushed-up the profile some more. Arch uses `enchant-2` & `enchant-lsmod-2` executables, while Debian/Ubuntu have `enchant` & `enchant-lsmod`. Introduced new redirect files to cope with those differences.


smitsohu:
Merged, thanks!

Issue #1847, opened 2018-03-29
Firefox 60.0b (dev edition) can't display webpages when 'seccomp' or 'shell none' enabled


yourcelf:
Using firejail 0.9.52 on Ubuntu 16.04.4.

It looks like Firefox 60b Developer Edition (currently at 60.0b7) isn't working if `seccomp` or `shell none` are enabled (they are enabled by default in both `firefox.profile` and  `firefox-developer-edition.profile`).  Firefox will launch and display the window, but is unable to load any webpages or display tabs.

With `shell none` enabled, firefox prints the following errors in the console:
```
###!!! [Parent][MessageChannel] Error: (msgtype=0x15007F,name=PBrowser::Msg_Destroy) Channel error: cannot send/recv

IPDL protocol error: Handler returned error code!

###!!! [Parent][DispatchAsyncMessage] Error: PBackgroundStorage::Msg_ClearMatchingOriginAttributes Processing error: message was deserialized, but the handler returned false (indicating failure)
```

With just `seccomp` enabled but `shell none` disabled, no obvious errors are printed, but the same behavior applies (the main window shows, but is unable to load webpages or tabs).

Launching with `firejail --noprofile` or using a profile with `seccomp` and `shell none` removed works fine.

netblue30:
Works fine for me on Debian stable. They are enabling 3D acceleration by default in the browser. 

@yourcelf what video card and video drivers do you have? Also try the following:
`````
$ firejail --no3d firefox  ----> in this case Firefox should revert to regular no-3D browsing
$ firejail --ignore=shell --allow-debuggers firefox ---> some Nvidia drivers are getting kill making a call to ptrace
`````

`````

**Comment by yourcelf** [comment]
```
$ lspci | grep VGA
00:02.0 VGA compatible controller: Intel Corporation Sky Lake Integrated Graphics (rev 07)
00:13.0 Non-VGA unclassified device: Intel Corporation Device 9d35 (rev 21)
```

After re-enabling `seccomp` and `shell none`, launching with `--no3d` and `ignore=shell --allow-debuggers` doesn't seem to fix the problem.  Every attempt to load a tab displays "Gah. Your tab just crashed."  With the `--ignore=shell --allow-debuggers` options, the crash message includes a feedback collector requesting that I submit a report to mozilla.

Vincent43:
Can you try if setting `media.cubeb.sandbox = false` in firefox config helps?

yourcelf:
[...]

Issue #1765, opened 2018-02-07
Firefox 60+ fails to run content processes


johnp:
Firefox Nightly doesn't seem to be able to start child processes / no tabs loading websites (except `about:`-stuff, which AFAIK runs in the parent process), crashing them instead:

https://crash-stats.mozilla.com/report/index/288d3c2b-c0be-4a85-af97-b294f0180207

I suspect [bug 1401062](https://bugzilla.mozilla.org/show_bug.cgi?id=1401062) to be the issue, but don't know how to debug further.

firejail latest master / 0.9.53 on Fedora 27, works with `--noprofile`.

smitsohu:
Thanks for your report. Could you please check briefly if it works when you replace `seccomp` with the long `seccomp.drop` line from [qutebrowser profile](https://raw.githubusercontent.com/netblue30/firejail/ca3964fed3f141e75bf396defaec3dca543ce49f/etc/qutebrowser.profile)?

johnp:
No, doesn't work, but I just saw that a fresh profile works fine. I'll check my preferences to see what's triggering the issue...

johnp:
I verified via mozregression that bug 140162 is the issue. ([pushlog](https://hg.mozilla.org/integration/autoland/pushloghtml?fromchange=cba9a29be139a663869a2c26557fd527a6981407&tochange=bd7ff5744eb29e105b7b3c37cb5f46164fa11ef4)) 
I haven't been able to identify what exactly in my profile is causing the issue to manifest, still trying though.

SpotComms:
We'll probably eventually need to remove` seccomp` and add `caps.keep sys_chroot,sys_admin` like chromium.profile as mentioned in #1414.

smitsohu:
It seems that three options are causing problems: `noroot`, `seccomp` and `apparmor`. `seccomp` works just fine on some distributions (arch, debian) and is breaking firefox on others (fedora, ubuntu), depending on how distributions handle unprivileged user namespaces.

@johnp Can you please comment out these lines in your firefox profile and give it a try? Instead of disabling `seccomp` altogether, you can also replace it with this long line:
```
[...]

PR #1865, opened 2018-04-06
Potential fixes for Firefox 60

Fred-Barclay:
See #1765 and #1847 for reference.
This comments out `shell none` and replaces `seccomp` with `seccomp.drop`. Palemoon and Basilisk get their own profiles again. 

I think we may need to give firefox-esr its own profile temporarily, until it's fully based on Firefox 60 sometime between May and August (60 ESR comes out on 9th of May, but 52 ESR will stick around until 21st of August I believe).

### DIFF
--- a/etc/basilisk.profile
+++ b/etc/basilisk.profile
@@ -12,26 +12,6 @@ include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-programs.inc
 
-# These are uncommented in the Firefox profile. If you run into trouble you may
-# want to uncomment (some of) them.
-#whitelist ${HOME}/dwhelper
-#whitelist ${HOME}/.zotero
-#whitelist ${HOME}/.vimperatorrc
-#whitelist ${HOME}/.vimperator
-#whitelist ${HOME}/.pentadactylrc
-#whitelist ${HOME}/.pentadactyl
-#whitelist ${HOME}/.keysnail.js
-#whitelist ${HOME}/.config/gnome-mplayer
-#whitelist ${HOME}/.cache/gnome-mplayer/plugin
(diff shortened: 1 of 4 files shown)


Fred-Barclay:
@Vincent43 Added, thanks!

smitsohu:
For a few weeks Palemoon and Basilisk were now based on firefox-common.profile, which comes with a number of additional options like `private-dev`, `nodbus` and so on. Would it be an option to keep it like that?

smitsohu:
corresponding blacklisting fixed in 63e6d734a55ad2505c7ad462f046296d15bd346e

Fred-Barclay:
Good thought @smitsohu . How does it look now?
I also removed the commented whitelists for extensions. They weren't being used anyway, and we can always just tell folks to `include /etc/firejail/firefox-common-addons.inc` if necessary.

smitsohu:
@Fred-Barclay Looks good! The only thing I'm still wondering is how to deal with pivot_root. Personally I would actually prefer to leave it out until we understand better when it's necessary.

SkewedZeppelin:
disable-interpreters.inc should be fine to include

but why are we breaking out palemoon and basilisk instead of just adding the following to them?
```
ignore seccomp.drop
seccomp
shell none
```

and why would we need to give firefox-esr it own?

Fred-Barclay:
@SkewedZeppelin I think that should work. :+1: I'll close this and just commit directly since it won't be a big change.

I was thinking we temporarily give firefox-esr it's own profile (although we really don't need to do any more than what you suggest above) because it can still use `seccomp` and `shell none`.  Once ESR moves to firefox 60, though, we could set it back to its current behavior since it will not be able to use `seccomp` or `shell none`.

Issue #1869, opened 2018-04-07
atool private-etc issue


glitsj16:
Hi, atool profile from latest git has private-etc `none`, resulting in zero-size extracted files:

```
$ grep private-etc /etc/firejail/atool.profile
private-etc none

$ firejail atool -x test.gz
Reading profile /etc/firejail/atool.profile
Reading profile /etc/firejail/globals.local
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-passwdmgr.local
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-programs.local
Parent pid 19111, child pid 19112

Warning: skipping none for private /etc
Private /etc installed in 0.13 ms
Blacklist violations are logged to syslog
Child process initialized in 471.84 ms
Error getpwuid: main.c:213 init_cfg: No such file or directory
test.txt.gz: extracted to `test'

Parent is shutting down, bye...

$ ls -l test
-rw-r--r-- 1 glitsj16 glitsj16 0 Apr  6 02:24 test.txt

= = =

$ firejail --private-etc=passwd,group atool -x test.gz

Reading profile /etc/firejail/atool.profile
Reading profile /etc/firejail/globals.local
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-passwdmgr.local
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-programs.local
Parent pid 15880, child pid 15881

Warning: skipping none for private /etc
Private /etc installed in 3.15 ms
Blacklist violations are logged to syslog
Child process initialized in 505.12 ms
test.txt.gz: extracted to `test'

Parent is shutting down, bye...

$ ls -l test
-rw-r--r-- 1 glitsj16 glitsj16 70 Apr  6 02:28 test.txt

```

Easily fixed by changing to `private-etc passwd,group`. I'll do a PR with the fix.
The profile might also benefit from 2 additional 'redirect' profiles too, for
`bunzip2` and `gunzip`. Both of these seem to be called by default.

Regards

glitsj16:
[...]

PR #1870, opened 2018-04-07
atool fixes

glitsj16:
Adding 'private-etc' `passwd,group` to unbreak the profile, fixes [issue 1869](https://github.com/netblue30/firejail/issues/1869).
Add 2 'redirect' profiles for gzip: `bunzip2` and `gunzip` (which atool uses to handle .gz archives).


Fred-Barclay:
Merged, thanks!

PR #1874, opened 2018-04-09
fix settings - disable nodus

glitsj16:


Fred-Barclay:
G'day @glitsj16 !
file-roller is one of the few profiles that, for now, we've decided to keep nodbus even though it makes settings immutable (see https://github.com/netblue30/firejail/pull/1843#discussion_r178156148). 
Thanks anyways!
Fred

PR #1881, opened 2018-04-11
Allow firefox to access proxychains' local config.

ratm-github:
This allows people to use proxychains in the firefox sandbox, e.g.:
$ firejail --profile=/etc/firejail/firefox.profile \
    proxychains /usr/bin/firefox

### DIFF
--- a/etc/firefox.profile
+++ b/etc/firefox.profile
@@ -7,11 +7,14 @@ include /etc/firejail/globals.local
 
 noblacklist ${HOME}/.cache/mozilla
 noblacklist ${HOME}/.mozilla
+noblacklist ${HOME}/.proxychains
 
 mkdir ${HOME}/.cache/mozilla/firefox
 mkdir ${HOME}/.mozilla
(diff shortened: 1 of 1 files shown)


Fred-Barclay:
Does exposing ~/.proxychains to firejail present any security risks for users when they're not using proxychains at the moment? I.e. if I have ~/.proxychains on my system, but I'm not using it for firefox, can an attacker access secrets or change important configurations in ~/.proxychains?

If so then it might be better not to add this to the default profile, and just let users add it to their customised profiles as needed. :wink:

ratm-github:
That's a very good point. ~/.proxychains/proxychains.conf might indeed contain some plaintext passwords for SOCKS5 proxies which support authentication (which are utterly rare, but still).

Write access might allow an attacker to redirect traffic to a proxy they control (which, if they managed to compromise firefox, they could probably do anyway).

All in all I think your point about leaving this out of the default profile is well taken, I'll withdraw the pull request.Thanks for the insightful comment!

PR #1886, opened 2018-04-12
allow java in some more profiles

smitsohu:
based on the presence of `noblacklist ${HOME}/.java`. I would merge if no one spots a false positive (Steam?)


smitsohu:
I'm planning to remove Steam from the list. It seems that games per Steam policy must come with their own JRE on Linux.

SkewedZeppelin:
@smitsohu we should leave Java in Steam for the case where a user adds a non-Steam game to it.

PR #1888, opened 2018-04-13
Amend Discord profile

flacks:
Add a few missing programs that Discord wants to private-bin, and change
'fonts' for 'machine-id' in private-etc, so that sound works. 'fonts'
doesn't seem to be needed.


SkewedZeppelin:
-1 on adding the browsers to private-bin. That won't work very well if one is launched in the context of this.

flacks:
OK. Is there a better solution than removing the browsers from private-bin?

SkewedZeppelin:
> Is there a better solution than removing the browsers from private-bin?


Users of Chromium based browsers can copy and paste the links. Users of Firefox-based browsers can just click them as long as an instance is already running, otherwise they need to copy and paste the link.

PR #1891, opened 2018-04-16
Fix firefox common addons

glitsj16:
`GNOME Shell integration` relies on dbus and python 3. When users opt to include firefox-common-addons.inc (in firefox-common.profile), chrome-gnome-shell is currently broken. Opening its preferences shows ... 'Although GNOME Shell integration extension is running, native host connector is not detected. Refer to documentation for instructions about installing connector.'. Easily fixed by adding `ignore nodbus` and `allowing python 3` inside `firefox-common-addons.inc`, so users that don't include that file (the default) get an unchanged firejail experience.

NOTE: I suspect that chromium-based browsers would need a similar profile change when using GNOME Shell integration , but I neither use nor know these well enough to propose a PR.

Additionally, users using the `Flash` plugin could benefit from adding `adobe` to `private-etc`. I left that commented, to stay in sync with what's in firefox-common.profile.

### DIFF
--- a/etc/firefox-common-addons.inc
+++ b/etc/firefox-common-addons.inc
@@ -53,3 +53,12 @@ whitelist ${HOME}/.wine-pipelight
 whitelist ${HOME}/.wine-pipelight64
 whitelist ${HOME}/.zotero
 whitelist ${HOME}/dwhelper
+
+# GNOME Shell integration (chrome-gnome-shell) needs dbus and python 3 (blacklisted by disable-interpreters.inc)
+ignore nodbus
+noblacklist ${PATH}/python3*
+noblacklist /usr/lib/python3*
+
+# Flash plugin
+# private-etc below works fine on most distributions. There are some problems on CentOS.
+#private-etc ca-certificates,ssl,machine-id,dconf,selinux,passwd,group,hostname,hosts,localtime,nsswitch.conf,resolv.conf,xdg,gtk-2.0,gtk-3.0,X11,pango,fonts,mime.types,mailcap,asound.conf,pulse,pki,crypto-policies,ld.so.cache,adobe


etc/firefox-common-addons.inc
```diff
@@ -53,3 +53,12 @@ whitelist ${HOME}/.wine-pipelight
...
 whitelist ${HOME}/dwhelper
+
+# GNOME Shell integration (chrome-gnome-shell) needs dbus and python 3 (blacklisted by disable-interpreters.inc)
+ignore nodbus
+noblacklist ${PATH}/python3*
+noblacklist /usr/lib/python3*
+
+# Flash plugin
+# private-etc below works fine on most distributions. There are some problems on CentOS.
+#private-etc ca-certificates,ssl,machine-id,dconf,selinux,passwd,group,hostname,hosts,localtime,nsswitch.conf,resolv.conf,xdg,gtk-2.0,gtk-3.0,X11,pango,fonts,mime.types,mailcap,asound.conf,pulse,pki,crypto-policies,ld.so.cache,adobe
```
SkewedZeppelin: You don't need to define the whole line, specifying most of the private- options will automatically append to the previous ones. Just replace it with something like

```
# Flash plugin
# private-etc must first be enabled in firefox-common.profile and in profiles including it
#private-etc adobe
```

etc/firefox-common-addons.inc
```diff
@@ -53,3 +53,12 @@ whitelist ${HOME}/.wine-pipelight
 whitelist ${HOME}/.wine-pipelight64
 whitelist ${HOME}/.zotero
 whitelist ${HOME}/dwhelper
+
+# GNOME Shell integration (chrome-gnome-shell) needs dbus and python 3 (blacklisted by disable-interpreters.inc)
```
SkewedZeppelin: you should move the above commented `noblacklist ${HOME}/.local/share/gnome-shell/extensions` and its associated whitelist option down here

SkewedZeppelin:
I'm sure we talked about it somewhere, but I think we were keeping GNOME extension downloading support disabled even if firefox-common-addons.inc was enabled/included.

As it could potentially allow any malicious browser addon to write directly to the GNOME extensions folder which from there could be used to compromise the entire shell.

glitsj16:
@SkewedZeppelin Thanks for your suggestions. Learning something everyday.

> I'm sure we talked about it somewhere, but I think we were keeping GNOME extension downloading support disabled even if firefox-common-addons.inc was enabled/included.
> 
> As it could potentially allow any malicious browser addon to write directly to the GNOME extensions folder which from there could be used to compromise the entire shell.

Good point. I guess this means these commits are useless. Leave things commented and add a warning note to that effect? Safer to install/update gnome-shell extensions via the package manager.

Fred-Barclay:
@SkewedZeppelin What if we made ${HOME}/.local/share/gnome-shell/extensions read-only?

Issue #1893, opened 2018-04-16
Lobase failing again :(   


bn0785ac:
Commit: f8b64047 (apr 9)

```
firejail lobase
Reading profile /etc/firejail/lobase.profile
Reading profile /etc/firejail/libreoffice.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Parent pid 9122, child pid 9123
Warning: An abstract unix socket for session D-BUS might still be available. Use --net or remove unix from --protocol set.
Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Blacklist violations are logged to syslog
Child process initialized in 105.14 ms
LibreOffice(16)/kdeui (kdelibs): Session bus not found 
To circumvent this problem try the following command (with Linux and bash) 
export $(dbus-launch) 

Parent is shutting down, bye...

```

bn0785ac:
it affects all lo progs (writer, calc, impress, math too)

bn0785ac:
Also



Lobase cant close pacefully with firejail

bn0785ac:
FIXES:


please comment nodbus (for libreoffice **global**)

and nonewprivs  for lobase

Issue #1879, opened 2018-04-11
Lighter profile for cinelerra


bn0785ac:
If you try to use x265  cin hangs  

with this x265 works (with it i got 4-6 threads instead of <4, and as i've and 8t cpu its fine)
https://p.teknik.io/pYzl9

bn0785ac:
Extra!

i've run a small bench

With default profile: 2 minutes 40 seconds

Now: 40 seconds

smitsohu:
Hi @bn0785ac,

do you use some proprietary graphics driver? With regards to seccomp, run `sudo journalctl | grep syscall` after cinelerra has been killed by firejail, it should show the violation.

In order to identify the problem with private-bin, easiest is to start the program as `firejail --build cinelerra`, then play a bit with it and close it again. In your terminal you'll find some profile suggestions, including for private-bin. Note however that the `--build` option is only available in firejail 0.9.52 or higher.

smitsohu:
> Its more a caps issue (if i remove seccomp works), its rare that caps.drop.all doesnt block anything.

In your profile `protocol unix` is enabled, can you confirm it works with this option?

bn0785ac:
works fine here (i've uploaded music videos to youtube with that)

smitsohu:
@bn0785ac Thanks. Can you please run `firejail cinelerra` and paste here what you get in the terminal (error messages, warnings)? Do you get anything from `sudo journalctl | grep syscall`, or does it return nothing?

If you want we find the problematic system call(s) together. Then we could consider doing something like in /etc/firejail/clementine.profile and selectively allow only the needed system call(s), in order to keep most of the seccomp filter in place.

smitsohu:
Some time ago I came across this [wiki article](https://wiki.tizen.org/Security:Seccomp), which seems to indicate that seccomp.keep is lighter.

Issue #1703, opened 2017-12-30
Couldn't start 'libreoffice' in Debian Testing


bitfreak25:
With the current state of Debian Testing (with firejail 0.9.52) I couldn't start the program "libreoffice".

The output for _"firejail libreoffice"_ is the following:
> Reading profile /etc/firejail/libreoffice.profile
> Reading profile /etc/firejail/disable-common.inc
> Reading profile /etc/firejail/disable-devel.inc
> Reading profile /etc/firejail/disable-passwdmgr.inc
> Reading profile /etc/firejail/disable-programs.inc
> Reading profile /etc/firejail/whitelist-var-common.inc
> Parent pid 1131, child pid 1132
> Blacklist violations are logged to syslog
> Child process initialized in 118.68 ms
> Warning: an existing sandbox was detected. /usr/bin/libreoffice will run without any additional sandboxing features
> Warning: failed to launch javaldx - java may not function correctly
> ERROR 4 forking process
> 
> Parent is shutting down, bye...

SpotComms:
Can you edit /etc/firejail/libreoffice.profile, disable all the options, and re-enable them one by one until it crashes? Thanks.

bitfreak25:
It crashed until I re-enable one of the following lines:
> nonewprivs
> noroot
> protocol unix,inet,inet6
> seccomp

All other could be re-enabled without problems.

The thrown errors looks like the same as above but with the exception of re-enabling line "noroot" with the following output:
> Reading profile /etc/firejail/libreoffice.profile
> Reading profile /etc/firejail/disable-common.inc
> Reading profile /etc/firejail/disable-devel.inc
> Reading profile /etc/firejail/disable-passwdmgr.inc
> Reading profile /etc/firejail/disable-programs.inc
> Reading profile /etc/firejail/whitelist-var-common.inc
> Parent pid 3634, child pid 3635
> Blacklist violations are logged to syslog
> Child process initialized in 54.45 ms
> No protocol specified
> Failed to open display
> [Java framework] Error in function createSettingsDocument (elements.cxx).
> javaldx failed!
> Warning: failed to read path from javaldx
> No protocol specified
> 
> (process:19): dconf-CRITICAL **: unable to create directory '/run/user/1000/dconf': Not a directory.  dconf will not work properly.
> 
> Parent is shutting down, bye...

SpotComms:
I'm unable to reproduce under Arch or under Debian Buster both with LibreOffice 5.4.4.2.

[...]

PR #1894, opened 2018-04-16
Repairing programs

bn0785ac:
Troubles:
* Cinelerra is single-threaded (*with an i3/i5* **may be acceptable**,  with a fx 8350/ryzen *1400,2400G,1500X/ ryzen 3*   its bad, but with a 1600/2600/1700/2700X/1800, i9, i7, i5 or threadripper* **is really annoying**)

* Libreoffice doesn't open *global*

* Lobase can't close *only lobase*


bn0785ac:
This pull fixes #1893 #1879

etc/natron.profile
```diff
@@ -12,7 +12,7 @@ noblacklist /opt/natron
 
 include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-devel.inc
-include /etc/firejail/disable-interpreters.inc
+#include /etc/firejail/disable-interpreters.inc
```
Fred-Barclay: Hi @bn0785ac Do you know which language(s) natron needs access to? We can allow access to those while still blocking access to the others in disable-interpreters.

etc/cin.profile
```diff
@@ -18,15 +18,15 @@ ipc-namespace
...
 nodvd
-nogroups
-nonewprivs
+#nogroups
+#nonewprivs
 notv
 noroot
 protocol unix
-seccomp
+#seccomp
```
Fred-Barclay: I'm not sure we should disable some of these (especially `seccomp`). @SkewedZeppelin any thoughts?

etc/libreoffice.profile
```diff
@@ -27,13 +27,13 @@ apparmor
...
-nodbus
+#nodbus
 nodvd
 nogroups
-nonewprivs
+#nonewprivs
 noroot
 notv
-protocol unix,inet,inet6
+#protocol unix,inet,inet6
```
Fred-Barclay: Why does this line need commenting?

etc/cin.profile
SkewedZeppelin: Agreed.

etc/cin.profile
```diff
@@ -18,15 +18,15 @@ ipc-namespace
...
 shell none
 
-private-bin cin,ffmpeg
+#private-bin cin,ffmpeg
```
SkewedZeppelin: This is okay

etc/libreoffice.profile
bn0785ac: i got crazy errors at libreoffice

etc/cin.profile
bn0785ac: I'll run a small bench with seccomp enabled & disabled

etc/natron.profile
bn0785ac: python :)

Fred-Barclay: Thanks!

etc/libreoffice.profile
smitsohu: @bn0785ac We still try to find a different solution to the problem with Libreoffice in #1703. `nodbus` (for KDE in general) and `apparmor` (for Kubuntu only) probably needs to be disabled, yes.

Vincent43: What breaks apparmor?

etc/cin.profile
bn0785ac: rendertime 

13:55  with     seccom
13:05  without it

etc/libreoffice.profile
bn0785ac: REAL errors :dagger:

etc/cin.profile
bn0785ac: rendertime 

13:55  with     seccom
13:05  without it

smitsohu: `seccomp` is indeed not for free performance-wise. The penalty is small, but when you do heavy computing, like rendering videos in your case, it could be measurable.

etc/libreoffice.profile
smitsohu: @Vincent43 D-Bus. Means issues are probably limited to Kubuntu, see #1703.

etc/cin.profile
bn0785ac: yeah :)   

we need say if this 7% gap  its acceptable (i've a fx 8350),  MAY be small (2-3)  at an i5,  <1 at an i3 or memecpus,  huge (14) with a 2700X  and INSANELY BAD (24-36) with a TR/i9

etc/libreoffice.profile
bn0785ac: no,  it happens here (manjaro *unstable* kde)


manjaro unstable = arch but with a 24 hours delay for update

Fred-Barclay: Can you paste the errors you get and describe what you see?
Thanks!

etc/cin.profile
smitsohu: @bn0785ac Since `seccomp` is an important component of the sandbox, I'm not sure if disabling it is the right way. But probably there should be a comment in the profile, so other users can find the solution easily when they run into this issue.

I assume `nogroups` has a similar effect?

bn0785ac: i like this **accepted**  

with nogroups i found a harder gap  
the same file spent 16:05  (same setup, fx 8350-4.1) **idk if with an i3-i5 you'll notice a ligher gap, or if with a  r5-2600/600/ ryzen 7/TR/i9 you'll notice an insanely evil gap**

>15%

bn0785ac: **edit**: i've added this.

* re-enable seccomp

* add a comment: its an estimated due to  my benchs,  **as i dont have other pcs** i can't confirm that an i3 really has a 2% gap, a 2700X  has a 14-28% gap,  or a 1950X/7980XE really has **OFFENSIVE gaps** (+50%)

etc/natron.profile
bn0785ac: Fixed (natron runs but i got crazy logs)


TESTING warning: noblacklist /usr/lib/python3* not matched by a proper blacklist command in disable*.inc

bn0785ac:
@smitsohu I've made all that you've asked 

* reduced gap + fixing multithread issues, and i'm still keeping seccomp enabled

* i've added a small warning about   the seccomp performance gap 

* fixed natron (with a better fix)

smitsohu:
@bn0785ac Thanks!

bn0785ac:
@smitsohu @SkewedZeppelin @Fred-Barclay 
can somebody accept it?

* cin keeps seccomp  (but i've recovered multithreading **its mandatory for video editing**)

* libreoffice its fixed (linux-kde)

* natron works (multithread )

Fred-Barclay:
@bn0785ac Have you been able to make any progress getting that info for LibreOffice? :smiley:

Fred-Barclay:
@smitsohu As far as I know we may not need `#nodbus` in this PR. Other than that we should be good to go!

Fred-Barclay:
Alright, here we go!

PR #1900, opened 2018-04-21
Add WebStorm profile

flacks:


etc/webstorm.profile
```diff
@@ -0,0 +1,36 @@
...
+noblacklist ${HOME}/.gitconfig
+noblacklist ${HOME}/.jack-server
+noblacklist ${HOME}/.jack-settings
+noblacklist ${HOME}/.java
+noblacklist ${HOME}/.local/share/JetBrains
+noblacklist ${HOME}/.ssh
+noblacklist ${HOME}/.tooling
+
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-passwdmgr.inc
```
Fred-Barclay: Webstorm is for Javascript development, right? We can probably also have `include /etc/firejail/disable-devel.inc` and possibly `include /etc/firejail/disable-interpreters.inc`.

SkewedZeppelin:
As a note the following are only in android-studio.profile and idea.sh.profile for Android app development/APK compilation support.

```
noblacklist ${HOME}/.jack-server
noblacklist ${HOME}/.jack-settings
```

I've removed the references from clion.profile here 13ad9ad7da8b7bab9e4df340467556ec1127efef, you should do the same here.

PR #1903, opened 2018-04-22
priv tweaks

smitsohu:
Lowering privileges for setting cpu affinity was no primary objective, it just turned out to be possible (from `man sched_setaffinity`: The caller needs an effective user ID equal to the real user ID or effective user ID of the thread identified by pid, or it must possess the CAP_SYS_NICE capability)


netblue30:
All merged, thanks!

PR #1910, opened 2018-04-25
Add XMind profile

flacks:


SkewedZeppelin:
Does this not need the following?

```
noblacklist ${HOME}/.java
```
and
```
# Allow access to java
noblacklist ${PATH}/java
noblacklist /usr/lib/java
noblacklist /etc/java
noblacklist /usr/share/java
```

PR #1911, opened 2018-04-26
Enable KVM on Android Studio

rccavalcanti:
Current profile denies access to /dev/kvm.

### DIFF
--- a/etc/android-studio.profile
+++ b/etc/android-studio.profile
@@ -32,7 +32,6 @@ protocol unix,inet,inet6
 seccomp
 shell none
 
-private-dev
 # private-tmp
 
 # noexec /tmp breaks 'Android Profiler'


Fred-Barclay:
G'day @rccavalcanti !
What happens with the current profile - that is, why does AndroidStudio need access to /dev/kvm?
Thanks!
Fred

rccavalcanti:
Hi @Fred-Barclay,
KVM is used for hardware acceleration on virtual devices (AVD Manager). Info [here](https://developer.android.com/studio/run/emulator-acceleration#vm-linux).

With the current profile, AVD Manager throws "/dev/kvm is not found". The device I have can't even be started, as I get "KVM is required to run this AVD".

Fred-Barclay:
Merged, thanks!

saddy001:
For me it throws permission denied for /dev/kvm in the current version. When I look at permissions in the jail:
```
ls -hla /dev/kvm 
crw-rw---- 1 nobody nogroup 10, 232 Okt 22 12:10 /dev/kvm
```

Is there a workaround?

PR #1914, opened 2018-04-27
Update Gajim profile

flacks:
- Version 1 ported to Python 3, so remove 2.7-related settings
- Plugins get installed in ~/.local/share/gajim/plugins
- Enable private-{etc,tmp}


SkewedZeppelin:
Looks good!
I would replace private-etc with the following however
```
private-etc alsa,asound.conf,ca-certificates,crypto-policies,fonts,group,hostname,hosts,ld.so.cache,ld.so.conf,localtime,machine-id,passwd,pki,pulse,resolv.conf,ssl
```

to not break sound and certificates.

SkewedZeppelin:
Merged, thanks!

PR #1916, opened 2018-04-28
Amend WebStorm profile

flacks:
- Add Android development-related paths
- Don't blacklist node/nvm for e.g. Node.js-related development
- Remove noexec from /tmp for e.g. same Node.js case


netblue30:
Merged, thanks.

PR #1918, opened 2018-05-01
priv tweak

smitsohu:
one more


netblue30:
All set. Check them in directly, any problems we'll deal with them when people start complaining. TravisCI will catch most problems anyway.

PR #1922, opened 2018-05-04
Allow GNOME Shell integration in unzip

glitsj16:
Without granting access to `${HOME}/.local/share/gnome-shell` unzip fails (silently) when updating GNOME Shell extensions. As removing outdated extension(s) prior to unzipping works as expected, this leaves users with crippled extensions.


SkewedZeppelin:
Merged, thanks!

PR #1923, opened 2018-05-04
add disable-interpreters.inc to gnome-logs

glitsj16:
Besides adding `include /etc/firejail/disable-interpreters.inc`, enabling both `private-etc` and `private-lib` (tested with systemd default storage and volatile journal).

### DIFF
--- a/etc/gnome-logs.profile
+++ b/etc/gnome-logs.profile
@@ -7,6 +7,7 @@ include /etc/firejail/globals.local
 
 include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-interpreters.inc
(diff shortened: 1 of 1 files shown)


Fred-Barclay:
@glitsj16 Thanks! I'm not sure we should include `private-lib` so close to the release of firejail 0.9.54 within the next few weeks, but the rest looks great.

@SkewedZeppelin Do you remember why `private-lib` was commented in d340cb30e7181b722a7be9330bca70ca70ae6651?

SkewedZeppelin:
With private-etc enabled log viewing works and whatnot, but timestamps are all UTC.

Fred-Barclay:
Alright, merged! Thanks!

PR #1924, opened 2018-05-04
add note for 'volatile' storage support

glitsj16:
@Fred-Barclay @SkewedZeppelin Did some further testing, especially when using systemd's `volatile` storage (which is stored only in memory, i.e. below /run/log/journal). In that case `machine-id` needs to be added to private-etc, as well as commenting both `nogroups` and `noroot`. Realizing this is not a default setup and the release of 0.9.54 is approaching, it's probably better to not mess with nogroups/noroot now. Perhaps adding `machine-id` to private-etc is not that problematic. Adding a reminder for users who run a volatile storage setup might be doable. Thoughts?

### DIFF
--- a/etc/gnome-logs.profile
+++ b/etc/gnome-logs.profile
@@ -19,6 +19,8 @@ net none
 no3d
 nodbus
 nodvd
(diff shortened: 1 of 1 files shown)


Fred-Barclay:
Sounds good, thanks!

Issue #1931, opened 2018-05-09
Blender with AMD GPU fails under firejail


RDProjekt:
When trying to run Blender with the supplied profile and have it use AMD card via OpenCL an error is encountered. The last part of the trace is as follows:

```
9:blender:opendir /sys/module/amdgpu:(nil)
Writing: /tmp/blender.crash.txt
9:blender:fopen64 /tmp/blender.crash.txt:0x7fb669776300
```
Apparently, amdgpu driver or OpenCL libraries try to access the `/sys/module/amdgpu` directory and fail since `/sys/module`  is permanently blacklisted (disabled).

netblue30:
all set, thanks

PR #1932, opened 2018-05-09
Fixes to make Blender with AMD GPU work under firejail (#1931)

RDProjekt:
This pull request allows whitelisting of directories under `/sys/module`. It's necessary for Blender to work with AMD graphics cards.

`/sys/module` is blacklisted by default, to keep compatibility with the current version, but its subdirectories can now be whitelisted.

There is also a new version of `blender.profile` making use of those capabilities.

### DIFF
--- a/etc/blender.profile
+++ b/etc/blender.profile
@@ -19,6 +19,11 @@ include /etc/firejail/disable-interpreters.inc
 include /etc/firejail/disable-passwdmgr.inc
 include /etc/firejail/disable-programs.inc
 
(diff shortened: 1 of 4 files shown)


SkewedZeppelin:
Is this for AMDGPU or AMDGPU Pro or ROCm?

RDProjekt:
We needed it for AMDGPU Pro and have not tested it on others.

netblue30:
Merged, thanks.

Issue #1930, opened 2018-05-09
Tor and Pale Moon Browsers hang and won't display.


Amii-Leigh601:
I have several browsers installed: Palemoon, Iridium, Tor, Waterfox, and Seamonkey, just to have lots to choose from. 
Just recently I updated the Firejail from what I had which worked at the time with everything I had installed, until I installed Waterfox, so I updated Firejail and now Waterfox works just peachy. Only thing is, now that I've done that Tor will hang: [code] $ firejail /usr/bin/tor-browser-en
Reading profile /etc/firejail/tor-browser-en.profile
Reading profile /etc/firejail/torbrowser-launcher.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-common.inc
Parent pid 1672, child pid 1673
Private /etc installed in 43.80 ms
Blacklist violations are logged to syslog
Child process initialized in 437.93 ms
Error: no suitable /usr/bin/tor-browser-en executable found

Parent is shutting down, bye...
[/code]

It runs fine without firejail, but that's hardly the point in having a sandboxing application, am I right? 
I also cannot start palemoon using firejail. I've posted about that issue on the Palemoon forums since it's only started this since I updated Palemoon to version 27.9.1:
[code]$ firejail palemoon
Reading profile /etc/firejail/palemoon.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-common.inc
Parent pid 3085, child pid 3086
TESTING warning: noblacklist /home/amii/.moonchild productions/pale moon not matched by a proper blacklist command in disable*.inc
Blacklist violations are logged to syslog
Child process initialized in 76.50 ms
[/code] 
It just hangs there until I close it.
I can run palemoon also without firejail so since I don't know how to write or modify code, I have to leave it to someone who can. Just so you can see what it looks like without firejail:[code]$ palemoon

(pale moon:3634): GLib-GObject-WARNING **: Attempt to add property GnomeProgram::sm-connect after class was initialised

(pale moon:3634): GLib-GObject-WARNING **: Attempt to add property GnomeProgram::show-crash-dialog after class was initialised

[...]

PR #1935, opened 2018-05-09
disable tracelog in firefox-common.profile

pizzadude:
tracelog breaks firefox 60+ catastrophically in my testing


Fred-Barclay:
G'day @pizzadude 
What distro are you using?

pizzadude:
also causes issues in pale moon and other firefox based browsers, see https://github.com/netblue30/firejail/issues/1930 , commenting out tracelog confirmed to work

Vincent43:
Maybe add reference (#1930) in profile why it's disabled.

pizzadude:
@Vincent43 ok, updated

Fred-Barclay:
I'd like to test this before merge just so I can get more insight, but thanks @pizzadude !

I wonder, though, if disabling tracelog is a viable long-term solution. Tracelog is a pretty valuable feature (in my opinion) and I wonder if we can or should modify tracelog instead so that it can be enabled on firefox profiles?

pizzadude:
this should be fixed somehow, whether it be disabling tracelog or fixing the problem in another way, before the new firejail release comes out so new users don't get confused and stop using firejail because of this

SkewedZeppelin:
I'm not seeing mass breakage as mention in #1930, but it does hang after close as in #1937.

This is with Firejail at 694e2027c5b6d03919bac4b5b305f6d3d834786c on Fedora 28 with Firefox 60 via
`sudo dnf --enablerepo=updates-testing update firefox`

Fred-Barclay:
Merged, thanks!

PR #1936, opened 2018-05-10
add whitelist-common.inc to qmmp profile

pizzadude:
this fixes the issue of it not importing the qt theme


SkewedZeppelin:
That would also prevent it from loading music. whitelist-common should only be used when the applications paths are also whitelisted.

pizzadude:
My bad. closing.

PR #1943, opened 2018-05-13
Krita: Enable python and KDE   (ksyscoca)

bn0785ac:
Without this,  krita crashes


Vincent43:
Fixed with https://github.com/netblue30/firejail/commit/634b2de7b7d71d1250c6d59361b65f876a55890e . Thx for reporting.

PR #1969, opened 2018-05-30
Allow Telegram to use Netlink sockets

flacks:
It complains in a console when they are unavailable.


flacks:
Hi @Fred-Barclay, there was only complaining, as far as I could tell. Either way, probably not going to be using tdesktop much more, so won't be investigating how it uses those sockets. Thanks anyway!

Fred-Barclay:
No worries, mate. Thanks as always!

PR #1976, opened 2018-06-01
Fix DjVu opening in Evince

announ:
Access to `libdjvulibre.so.*` is required for opening DjVu documents.
Otherwise, Evince does open, but fails to load the file with the error: `Failed to load backend for 'image/vnd.djvu+multipage': libdjvulibre.so.21: cannot open shared object file: No such file or directory`


SkewedZeppelin:
Merged, thanks!

PR #1977, opened 2018-06-04
Fix installing skins on qmmp.profile

pizzadude:
qmmp requires these unzipping programs in order to install skins, so i've added them to private-bin


Fred-Barclay:
Looks good, thanks!

PR #1996, opened 2018-06-14
Add gnome-mpv profile

flacks:


SkewedZeppelin:
Can either of the following be added?
```
include /etc/firejail/whitelist-var-common.inc
nodbus
```

flacks:
Both!

SkewedZeppelin:
Merged, thanks!

PR #1998, opened 2018-06-14
Amend Wire profiles

flacks:


netblue30:
OK, so you are basically removing etc/Wire.profile and you are renaming etc/wire.profile to etc/wire-desktop.profile. Is this correct?

flacks:
Yes - the latest binary you either get off wire.com or build yourself is named `wire-desktop`, so might as well remove/change older profile names. The new profile also adds a bunch of restrictions - the latest patch is working fine on my end.

netblue30:
all merged, thanks!

PR #1999, opened 2018-06-14
Create tor-browser-en directory, add missing bin programs

flacks:
tar and xz are needed on first run.


SkewedZeppelin:
Merged, thanks!

PR #2003, opened 2018-06-14
Revert private cache enabled by default

SkewedZeppelin:
Commit caa7ad8714206a158123773ddcaca6ef219a5501 enabled private-cache by default globally.

Should this really be enabled by default? Preventing programs from having persistent cache directories overall has many negative impacts.

- Most email clients cache emails to allow various functionality and downloading them can take many minutes
- Caching makes webpages load faster
- Most music players cache album art from libraries
- Mapping programs (gnome-maps, viking) don't need to constantly redownload tiles
- Mesa shader cache improves graphics load performance
- General waste of bandwith of clients/servers
- General increase in recomputing things
- General potential increase in disk I/O

I think it'd be better if we just manually add `private-cache` only to suitable programs.

private-cache will need to be added to more programs as identified by
```
grep 'whitelist ${HOME}/.cache' -L $(grep "private-cache" -L $(grep "redirect" -RiL))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,11 +100,6 @@ We also keep a list of profile fixes for previous released versions in [etc-fixe
 `````
 # Current development version: 0.9.55
 
-## Mounting a temporary filesystem on top of ~/.cache directory by default
-
-To disable it globally, set "private-cache no" in /etc/firejail/firejail.config.
-
-
 ## New commands:
 `````
       (wireless support for --net)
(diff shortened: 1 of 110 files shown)


Fred-Barclay:
I agree. For those of us who want it globally, we can just add `private-cache` to globals.local.

netblue30:
Reverted!

Issue #2006, opened 2018-06-20
globals.local included twice in profiles using x-common.profile format


glitsj16:
Profiles using an _include x-common.profile_ format (like {chromium,discord,firefox}-common.profile do) are seeing `globals.local` included twice. Probably more confusing than harmful, but then again I don't know what users put inside that file. This happens for the following profiles:

1. `chromium-common.profile`
- bnox.profile
- brave.profile
- chromium.profile
- dnox.profile
- enox.profile
- flashpeak-slimjet.profile
- google-chrome.profile
- google-chrome-beta.profile
- google-chrome-unstable.profile
- inox.profile
- iridium.profile
- opera.profile
- opera-beta.profile
- vivaldi.profile
- vivaldi-snapshot.profile
- yandex-browser.profile

2. `discord-common.profile`
- discord.profile
- Discord.profile
- discord-canary.profile
- DiscordCanary.profile

3. `firefox-common.profile`
- abrowser.profile
- basilisk.profile
- cliqz.profile
- cyberfox.profile
- firefox.profile
- icecat.profile
- iceweasel.profile
- palemoon.profile
- waterfox.profile

Luckily we don't need to change all affected profiles. Commenting `include /etc/firejail/globals.local` in `{chromium,discord,firefox}-common.profile` should be enough to avoid multiple inclusion of globals.local.

netblue30:
Thanks for the patch. It will speed up the setup.

PR #2007, opened 2018-06-20
include globals.local only once

glitsj16:
This should fix https://github.com/netblue30/firejail/issues/2006.


netblue30:
merged, thanks.

PR #2010, opened 2018-06-23
don't include globals.local twice (take two)

glitsj16:
Similar to https://github.com/netblue30/firejail/issues/2006. This time we avoid including globals.local twice in profiles that include default.profile.


Fred-Barclay:
Merged, thanks!

PR #2015, opened 2018-06-29
i'm modifying inox-hard (aur)  

bn0785ac:
can you add this for  support my changes?

snox comes from https://en.wikipedia.org/wiki/SNOX_process


bn0785ac:
second commit just remove undesired/unuesful info  

* why enox need acces to dnox dirs?
* why snox needs dnox dirs (my error copying enox profile)

netblue30:
Merged, thanks.

PR #2022, opened 2018-07-02
discord: added localtime and ld.so.cache to private-etc to fix wrong time and "discord: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory"

soredake:


Fred-Barclay:
Merged, thanks!

Issue #2018, opened 2018-06-30
Cannot run Geary 0.12.2-1


Bundy01:
Hi,
I can't launch Geary with Firejail:

```
Reading profile /etc/firejail/geary.profile
Reading profile /etc/firejail/whitelist-common.inc
Reading profile /etc/firejail/firefox.profile
Reading profile /etc/firejail/firefox-common.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-common.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Parent pid 5039, child pid 5040
Warning: An abstract unix socket for session D-BUS might still be available. Use --net or remove unix from --protocol set.
Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Post-exec seccomp protector enabled
Warning fseccomp: syscall "ni_syscall" not available on this platform
Warning fseccomp: syscall "umount" not available on this platform
[...]

PR #2025, opened 2018-07-04
Creating the gradio profile and editing the geary profile

Bundy01:
For geary, look at issue #2018. **Ignore dbus** does not work, it is necessary to **ignore nodbus** for launch geary.
Thanks.


Vincent43:
Thank you.

PR #2033, opened 2018-07-06
additional whitelist hardening

smitsohu:
Two commits with a relatively large diff, take your time to review. Also contains quite some nitpicking.


netblue30:
Seems to be fine, thanks!

PR #2057, opened 2018-07-23
Make ktorrent actually work

ltworf:
Restricting the network protocols makes it not work at all.

### DIFF
--- a/etc/ktorrent.profile
+++ b/etc/ktorrent.profile
@@ -45,7 +45,6 @@ noroot
 nosound
 notv
 novideo
-protocol unix,inet,inet6
 seccomp
 shell none
 


veloute:
qbittorrent and rtorrent both have `protocol unix,inet,inet6` in their profiles and they work fine.
Also, after testing ktorrent, it works fine for me with the current profile - are you sure it's not just on your end?

ltworf:
It claimed no connection with that, and then worked fine without it.

Are you sure you are using magnets and all that stuff?

veloute:
You're right - with the current profile magnet links don't work.
Adding `netlink` to the protocol line seems to fix it for me, try that out and see if it fixes your problem.

ltworf:
Seems to work now.

ltworf:
Seems to work now

Fred-Barclay:
@ltworf If you like, you're welcome to open a new PR with the `protocol unix,inet,inet6,netlink` line. :wink:

PR #2060, opened 2018-07-24
WIP: Add new disable-xdg.inc

SkewedZeppelin:
Now that @chiraag-nataraj added support for more XDG user dirs in fff306c991ff1f33ccb56313e4c166a169555656, we can blacklist these in many profiles that don't utilize whitelist.

I created a disable-xdg.inc. Downloads and Desktop are commented as they are usually a catch-all for most people.

And I included it in a small amount of profiles identified using the following command:
```
grep "disable-xdg" -RL $(grep 'whitelist \$' -RL $(grep "redirect" . -RiL))
```

there are still ~270 remaining, and strong discretion is needed, so I figured I'd get some discussion going before I went at it. Feel free to push to the branch at will.

### DIFF
--- a/etc/2048-qt.profile
+++ b/etc/2048-qt.profile
@@ -14,6 +14,11 @@ include /etc/firejail/disable-interpreters.inc
 include /etc/firejail/disable-passwdmgr.inc
 include /etc/firejail/disable-programs.inc
 
+mkdir ${HOME}/.config/2048-qt
+mkdir ${HOME}/.config/xiaoyong
+whitelist ${HOME}/.config/2048-qt
+whitelist ${HOME}/.config/xiaoyong
+include /etc/firejail/whitelist-common.inc
 include /etc/firejail/whitelist-var-common.inc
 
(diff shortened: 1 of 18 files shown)


netblue30:
I would say is quite cool, thanks!

SkewedZeppelin:
Ooh, I didn't want this merged, I was using it for staging. I didn't test any of them in any way.

I guess I'll just start doing the rest.

Fred-Barclay:
I have to admit I'm a bit confused by fff306c991ff1f33ccb56313e4c166a169555656 . Does it allow access to more directories by default? Or is it just add support for the xdg pathnames -- i.e. `firejail --whitelist=$MUSIC` would be the same as `firejail --whitelist=~/Music`?

SkewedZeppelin:
@Fred-Barclay The latter, it just allows use of the aliases to easily white/blacklist those paths, like ${DOWNLOADS}.

It makes it easier for us to further restrict the blacklist-only profiles.

PR #2063, opened 2018-07-25
created flameshot profile

veloute:


etc/flameshot.profile
```diff
@@ -0,0 +1,38 @@
...
+include /etc/firejail/disable-interpreters.inc
+include /etc/firejail/disable-passwdmgr.inc
+include /etc/firejail/disable-programs.inc
+include /etc/firejail/disable-xdg.inc
+
+caps.drop all
+ipc-namespace
+netfilter
+no3d
+# nodbus
```
Fred-Barclay: What happens if `nodbus` is enabled?

etc/flameshot.profile
```diff
@@ -0,0 +1,38 @@
...
+include /etc/firejail/flameshot.local
+# Persistent global definitions
+include /etc/firejail/globals.local
+
+include /etc/firejail/disable-common.inc
+include /etc/firejail/disable-devel.inc
+include /etc/firejail/disable-interpreters.inc
+include /etc/firejail/disable-passwdmgr.inc
+include /etc/firejail/disable-programs.inc
+include /etc/firejail/disable-xdg.inc
```
Fred-Barclay: Maybe add `noblacklist ${PICTURES}` so screenshots can be saved in the pictures directory? If so, please add a blank line and then this line under `include /etc/firejail/globals.local`

etc/flameshot.profile
```diff
@@ -0,0 +1,38 @@
...
+no3d
+# nodbus
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+novideo
+protocol unix,inet,inet6
```
Fred-Barclay: What happens with just `protocol unix`? This would disable internet access, is that something flameshot needs?

Alternatively, what if you removed the protocol line entirely and replaced `netfilter` with `net none`? Does it still work?

etc/flameshot.profile
```diff
@@ -0,0 +1,38 @@
...
+seccomp
+shell none
+
+disable-mnt
+private-cache
+private-dev
```
Fred-Barclay: Does the line `private-etc fonts` still work?

etc/flameshot.profile
```diff
@@ -0,0 +1,38 @@
...
+noroot
+nosound
+notv
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+
+disable-mnt
+private-cache
```
Fred-Barclay: Does the line `private-bin flameshot` work?

etc/flameshot.profile
veloute: some important things stop working.

etc/flameshot.profile
veloute: Flameshot has the option to upload screenshots to https://imgur.com

etc/flameshot.profile
veloute: Smart - will add.

etc/flameshot.profile
veloute: Getting SSL handshake errors when trying to upload images to imgur; adding hostname, hosts and resolv.conf doesn't seem to fix it - any idea?

veloute: Works fine now; added and committed.

Fred-Barclay: Excellent! Did it require all the etc/ directories (that is, can you remove any of the private-etc options and it still work)?

veloute: All that it needed was  'private-etc fonts,ca-certificates,ld.so.conf,resolv.conf,ssl'

PR #2081, opened 2018-08-14
Add descriptions to profiles

SkewedZeppelin:
This adds descriptions to ~290 profiles.
The descriptions were automatically pulled from the output of `apt show $package` and `pacman -Si $package`.

This will help when doing mass changes to profiles so we don't have to go searching for a package we don't know.

Opening this as a PR for discussion:
- license of the descriptions?
- format ok?
- placement ok?
- require all new profiles to have a description?

Program here: https://gist.github.com/SkewedZeppelin/ce7dd57140bbbd35c876f240a6850ba7

### DIFF
--- a/etc/0ad.profile
+++ b/etc/0ad.profile
@@ -1,4 +1,5 @@
 # Firejail profile for 0ad
+# Description: Real-time strategy game of ancient warfare
 # This file is overwritten after every install/update
 # Persistent local customizations
 include /etc/firejail/0ad.local
--- a/etc/2048-qt.profile
+++ b/etc/2048-qt.profile
@@ -1,4 +1,5 @@
 # Firejail profile for 2048-qt
+# Description: mathematics based puzzle game
 # This file is overwritten after every install/update
 # Persistent local customizations
 include /etc/firejail/2048-qt.local
--- a/etc/2048-qt.profile
+++ b/etc/2048-qt.profile
@@ -1,5 +1,5 @@
 # Firejail profile for 2048-qt
-# Description: mathematics based puzzle game
+# Description: Mathematics based puzzle game
 # This file is overwritten after every install/update
 # Persistent local customizations
 include /etc/firejail/2048-qt.local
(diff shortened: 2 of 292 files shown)


chiraag-nataraj:
So I guess I don't quite understand the necessity or purpose of the descriptions...

SkewedZeppelin:
@chiraag-nataraj It is just for reference and to ease maintenance.

When I did e5aba00d010e9e3af3e626bedb8acf8ae37b3b75 I spent a fair bit of time just trying to find out what a program is/does.

chiraag-nataraj:
@SkewedZeppelin Ah, I see, thanks.

> is there another distro with more/better descriptions that is also easy to parse?

I know Debian's might also be relatively easy (at the very worst, just `grep`ping for `^Description:` from the output of `aptitude show`).

SkewedZeppelin:
Okay here is it with descriptions from Ubuntu
https://github.com/SkewedZeppelin/firejail/commit/84a498a6ad1d1f3677e26f1ec555f9110891fae4

SkewedZeppelin:
Okay the latest version now has most descriptions from Ubuntu, and missing ones from Arch.

Commits with just descriptions from each are available under the descriptions-ubuntu and descriptions-arch branches here https://github.com/SkewedZeppelin/firejail/branches/active

Fred-Barclay:
I like it! I've just glanced it over but it seems the Ubuntu descriptions are more direct, less "fluffy" than the Arch ones so I'd prefer those when possible. :smile: 

One small nitpick -- sometimes the description begins with a capital letter and sometimes it doesn't. Can we make sure they're consistent? IMHO making sure the first letter is capitalised makes more sense, though it might not always be correct grammar; but really either way is fine as long as they're all consistent. :wink:

SkewedZeppelin:
I've gone through and cleaned up most of the descriptions. I also spotted one or two descriptions that I know aren't right, and removed those. If anyone else wants to make a pass through them, feel free to: `$(grep -Rl "Description" .)`

netblue30:
Merged!

PR #2092, opened 2018-08-22
Create pybitmessage.profile

1dnrr:
tested on fedora-28 with pybitmessage 0.6.3.2

### DIFF
--- a/etc/pybitmessage.profile
+++ b/etc/pybitmessage.profile
@@ -41,10 +41,9 @@ shell none
 disable-mnt
 private-bin pybitmessage,python*,sh,ldconfig,env,bash,stat
 private-dev
-private-etc PyBitmessage,PyBitmessage.conf,Trolltech.conf,fonts,gtk-2.0,hosts,ld.so.cache,ld.so.preload,localtime,pki,resolv.conf,selinux,sni-qt.conf,system-fips,xdg
+private-etc PyBitmessage,PyBitmessage.conf,Trolltech.conf,fonts,gtk-2.0,hosts,ld.so.cache,ld.so.preload,localtime,pki,resolv.conf,selinux,sni-qt.conf,system-fips,xdg,ca-certificates,ssl,pki,crypto-policies
 #private-lib
 private-tmp
-ca-certificates,ssl,pki,crypto-policies
 
 memory-deny-write-execute
 noexec ${HOME}
--- a/etc/pybitmessage.profile
+++ b/etc/pybitmessage.profile
@@ -47,7 +47,4 @@ ca-certificates,ssl,pki,crypto-policies
 
 memory-deny-write-execute
 noexec ${HOME}
-noexec /sbin
-noexec /usr/local/sbin
-noexec /usr/sbin
 noexec /tmp
--- a/etc/pybitmessage.profile
+++ b/etc/pybitmessage.profile
@@ -1,10 +1,10 @@
-# Firejail profile for default
+# Firejail profile for pybitmessage
 # This file is overwritten after every install/update
 # Persistent local customizations
 include /etc/firejail/pybitmessage.local
+# Persistent global definitions
+include /etc/firejail/globals.local
 
-# generic gui profile
-# depending on your usage, you can enable some of the commands below:
 noblacklist /sbin
 noblacklist /usr/local/sbin
 noblacklist /usr/sbin
--- a/etc/pybitmessage.profile
+++ b/etc/pybitmessage.profile
@@ -5,22 +5,23 @@ include /etc/firejail/pybitmessage.local
 # Persistent global definitions
 include /etc/firejail/globals.local
 
-noblacklist /sbin
-noblacklist /usr/local/sbin
-noblacklist /usr/sbin
+include /etc/firejail/whitelist-var-common.inc
 
 include /etc/firejail/disable-common.inc
 include /etc/firejail/disable-devel.inc
 include /etc/firejail/disable-passwdmgr.inc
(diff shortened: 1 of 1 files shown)


etc/pybitmessage.profile
```diff
@@ -0,0 +1,53 @@
+# Firejail profile for default
```
Vincent43: change `default`  for `pybitmessage`

etc/pybitmessage.profile
```diff
@@ -0,0 +1,53 @@
...
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /etc/firejail/pybitmessage.local
+
```
Vincent43: Add:
```

```

etc/pybitmessage.profile
```diff
@@ -0,0 +1,53 @@
...
+# generic gui profile
+# depending on your usage, you can enable some of the commands below:
```
Vincent43: delete those two lines

etc/pybitmessage.profile
```diff
@@ -0,0 +1,53 @@
...
+noblacklist /sbin
+noblacklist /usr/local/sbin
+noblacklist /usr/sbin
```
Vincent43: I don't see the reason for those three. Delete them.

etc/pybitmessage.profile
```diff
@@ -0,0 +1,53 @@
...
+private-etc PyBitmessage,PyBitmessage.conf,Trolltech.conf,fonts,gtk-2.0,hosts,ld.so.cache,ld.so.preload,localtime,pki,resolv.conf,selinux,sni-qt.conf,system-fips,xdg
+#private-lib
+private-tmp
+ca-certificates,ssl,pki,crypto-policies
+
+memory-deny-write-execute
+noexec ${HOME}
+noexec /sbin
+noexec /usr/local/sbin
+noexec /usr/sbin
```
Vincent43: Delete above three noexec rules from sbin

etc/pybitmessage.profile
1dnrr: I get the error and crush pybitmessage start if I comment all or any this three lines.
```
Traceback (most recent call last):
  File "bitmessagemain.py", line 19, in <module>
    import depends
  File "/usr/share/pybitmessage/depends.py", line 5, in <module>
    import pyelliptic.openssl
  File "/usr/share/pybitmessage/pyelliptic/__init__.py", line 16, in <module>
    from .openssl import OpenSSL
  File "/usr/share/pybitmessage/pyelliptic/openssl.py", line 549, in <module>
    loadOpenSSL()
  File "/usr/share/pybitmessage/pyelliptic/openssl.py", line 547, in loadOpenSSL
    raise Exception("Couldn't find and load the OpenSSL library. You must install it.")
Exception: Couldn't find and load the OpenSSL library. You must install it.
```
```
/etc/firejail/disable-common.inc:blacklist /sbin
/etc/firejail/disable-common.inc:blacklist /usr/local/sbin
/etc/firejail/disable-common.inc:blacklist /usr/sbin
```

Vincent43: Does it work if you add "openssl*" to private-bin option?

1dnrr: No, I checked it does not work. The same error occurs.

etc/pybitmessage.profile
```diff
@@ -0,0 +1,50 @@
...
+seccomp
+shell none
+
+disable-mnt
+private-bin pybitmessage,python*,sh,ldconfig,env,bash,stat
+private-dev
+private-etc PyBitmessage,PyBitmessage.conf,Trolltech.conf,fonts,gtk-2.0,hosts,ld.so.cache,ld.so.preload,localtime,pki,resolv.conf,selinux,sni-qt.conf,system-fips,xdg
+#private-lib
+private-tmp
+ca-certificates,ssl,pki,crypto-policies
```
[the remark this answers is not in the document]
1dnrr: pybitmessage does not work with plain private-etc (without PyBitmessage,PyBitmessage.conf,Trolltech...)
Do you think that the private-etc line should be removed?

etc/pybitmessage.profile
```diff
@@ -0,0 +1,50 @@
...
+# This file is overwritten after every install/update
+# Persistent local customizations
+include /etc/firejail/pybitmessage.local
+# Persistent global definitions
+include /etc/firejail/globals.local
+
+
+noblacklist /sbin
+noblacklist /usr/local/sbin
+noblacklist /usr/sbin
```
Vincent43: I'm still interested if noblacklist sbin is needes after noblacklist python was put in the right place

1dnrr: Just checked. Pybitmessage doesn't working with the "noblacklist python" lines and without this "/sbin" lines. These "/sbin" lines are necessary.

Vincent43: ok, thx

Vincent43:
Merged, thank you.

PeterSurda:
Thank you. I'm not a firejail expert, but don't you also need

    mkdir ~/.config/PyBitmessage
    whitelist ${HOME}/.config/PyBitmessage

?

Vincent43:
@1dnrr @PeterSurda 

Those can be added for `$HOME` hardening but someone would have to confirm it doesn't break anything. They would block access to `${HOME}/PyBitmessage` for example. I'm totally unfamiliar with PyBitmessage and I don't know what is needed for it to work. It would be great if you can test those rules and maybe open PR here.

PR #2093, opened 2018-08-23
Add automatically generated private-etc lines to all profiles

### LINKED ISSUE #1734: private-etc templates
I've been testing many of the profiles under Fedora 27 and quite a few are broken due to private-etc. See [0], [1], [2]. I think we should try and create some standardized templates for private-etc, test that they work across many distros, and then update all of the profiles.

To start:
- Base: localtime
- Internet: ca-certificates,crypto-policies,nsswitch.conf,pki,resolv.conf,ssl
- Sound: asound.conf,machine-id,pulse
- Window Frameworks: fonts,gtk-2.0,gtk-3.0,pango,xdg

We also need to figure out when the following are necessary
- group,host*,passwd,selinux

[0] - 68ccf1efee030470bf3f1666429e31374f2ae3a6
[1] - 4efa2d8ee13bc2b7e802f628d8b953a4fcfd9eec
[2] - 4efa2d8ee13bc2b7e802f628d8b953a4fcfd9eec

SkewedZeppelin:
**DO NOT MERGE!**

This adds automatically generated `private-etc` lines to ~380 profiles.
This basically closes #1734 

Program source here: https://gist.github.com/SkewedZeppelin/12d5bd24fca41f3c2a4b54d919c8ab44

This shoudn't be merged until `private-etc` supports globbing.
~~And we should also wait until after 0.9.56 is released.~~

Any feedback is appreciated.


Fred-Barclay:
Very nice!
Can you break down your program a bit for those of us who don't speak Java? :laughing: 

Does there need to be some fine-tuning still? I'm seeing a lot of things being added (like X11) that I don't recall being needed very much in `private-etc`?
Also maybe a bit too much is being added? I can't imagine keepassxc actually needed /etc/os-release, for instance.

Cheers!
Fred

SkewedZeppelin:
Yes, I don't like how much is added either, but 40 items in /etc is way better then 200+.

Feature "detection" is handled by
https://gist.github.com/SkewedZeppelin/12d5bd24fca41f3c2a4b54d919c8ab44#file-firejail-privateetcgenerator-java-L58-L101

and that info gets passed here
https://gist.github.com/SkewedZeppelin/12d5bd24fca41f3c2a4b54d919c8ab44#file-firejail-privateetcgenerator-java-L156-L212

and there is also some special handling here
https://gist.github.com/SkewedZeppelin/12d5bd24fca41f3c2a4b54d919c8ab44#file-firejail-privateetcgenerator-java-L107-L121

Vincent43:
I see some obvious issues like dnsmasq not containing `dnsmasq.conf` and `dnsmasq.conf.d`. I don't know if this automatic approach is realistic without manually checking app functionality across distros.

SkewedZeppelin:
@Vincent43 after I finish tweaking the program I plan to go back through and add any missing specific ones.

I have a list of other profiles too:
ardour5, ark, file, seamonkey, etc.

SkewedZeppelin:
Okay, I've tested working ~50 profiles so far.
It should be stable enough for you all to test. Feel free to comment working profiles or just push.

There is still the issue of the large number of added files/paths, but looking at it a bit:
- it should be safe, most aren't sensitive
- for profiles that already have a `private-etc`, it might actually fix minor quirks
- most of these profiles never had `private-etc` protections in the first place
- once globbing is supported, it look like less :smiley: 

Once more profiles have their `private-etc` enabled then:
1. we have to wait until after 0.9.56 release
2. someone still has to add globbing support to private-etc
3. we can rebase and merge

smitsohu:
I think the kde4rc, kde5rc are easier to add manually. It is now in large number of profiles, but only some KDE apps (maybe less than 20) need them.

SkewedZeppelin:
@smitsohu 20 is still a lot. The program does try to determine if it is a kde/qt app and exclude those if it isn't. But if it is determined to be a "gui" app, then it defaults to including gtk/kde/qt files.

See:
https://gist.github.com/SkewedZeppelin/12d5bd24fca41f3c2a4b54d919c8ab44#file-firejail-privateetcgenerator-java-L76-L84
and
https://gist.github.com/SkewedZeppelin/12d5bd24fca41f3c2a4b54d919c8ab44#file-firejail-privateetcgenerator-java-L240-L253

smitsohu:
It's cool, I like our program and the systematic approach. I was only speaking for myself when I said this way round is easier than that way, maybe we can clean up false positives later on if there are any

SkewedZeppelin:
@smitsohu I have refined the gtk/qt/kde detection. Only 13 profiles now include kde*rc, compared to 278 before.

Vincent43:
@SkewedZeppelin can you tell why `os-release` and `lsb-release` are needed for each profile? Currently os-release is used only in one profile (steam) and lsb-release in only two. Even if app want to read those it may work without them as well.

Vincent43:
Thx. protocols,services and rpc aren't privacy sensitive so they can stay. As for hostname I wonder if we can add it simultaneously with `hostname` firejail option.

reinerh:
A few thoughts/suggestions...

To figure out additional files/directories a program could be accessing, try running `strings $binary` and grep for patterns matching "^/etc/". It is likely those paths will at some point be accessed by the program.

If you have a Debian system, you can also check `/var/lib/dpkg/info/$PACKAGE.conffiles` for conffiles (files in /etc) that are shipped by the package.
For packages that are not currently installed on your system, apt-file is a nice tool (`apt-file list $PACKAGE | grep /etc/`).

SkewedZeppelin:
Okay, I have added support to get files from both `strings` and `apt-list`.
Their both vary in their output, and strings can output some useless lines sometimes.

I am going to make a VM with Ubuntu with as many supported programs as possible installed and let it run and hopefully it should get most of our bases covered.

reinerh:
Btw as mentioned, for apt-file you don't need to have the package installed. :-)

SkewedZeppelin:
Okay, so my solution: create two programs.
The first gathers results using multiple methods (strings, dpkg, rpm, pacman, apt-file)
and
the second that combines, processes, and outputs the profiles (rewrite of existing program).

I have created the gatherer program and now just need to rewrite the current program to take its output.

An example output of the gatherer
```
$ tree dnsmasq
dnsmasq
├── pacman.1535584341604.list
└── strings.1535584341604.list
$ cd dnsmasq
$ cat pacman*
dnsmasq.conf
$ cat strings*
ethers
dnsmasq.conf
hosts
resolv.conf
```

SkewedZeppelin:
Okay, I have implemented the two program approach.
The repo is here https://github.com/SkewedZeppelin/FirejailEtcGenerator

The latest version is generated using that new version.
There are still some issues, but it seems to be working okay.

vutny:
This PR is being superseded by #2745 , isn't it?

rusty-snake:
@vutny #2745 updates the templates in the profile template. This PR adds the template to all profiles.

vutny:
Aha. Looks like this one is really no merge candidate... Some profiles already have very specific `private-etc` value. Should we close it?

SkewedZeppelin:
@vutny I've been busy, but I can rebase it sometime soon probably.

Issue #2090, opened 2018-08-22
Our documentation


Fred-Barclay:
Hi folks
While @netblue30 has done a great job with [documentation](https://firejail.wordpress.com/documentation-2/), I’m wondering if we can do better?

(Also, man pages can be frightening for newbies – let’s forget we have those for a moment!)

In my opinion, there are a few issues with our current docs system:
1. It depends on a single person and either creates a lot of work for that person, or falls behind.
2. It’s not comprehensive and (to me) a bit confusing. Where, for example, in this sidebar screenshot would I go to find out how to use `--cgroup`? (Which, by the way, is only documented in the man pages and the 0.9.20 release notes.)
![docs](https://user-images.githubusercontent.com/11165995/44439964-81ca9100-a5b5-11e8-8674-df3f448528e8.png)
- What about how to configure firejail with firejail.config vs globals.local vs <profile_name>.local vs custom profiles in ~/.config/firejail?
- What is /etc/firejail/firejail.users? Should someone edit it by hand?
- What about learning how to use `--net=` and `--protocol=`? It's there, just not obvious IMHO.

3. It’s not particularly robust. What I mean is, it doesn’t reflect new features or the deprecation of old ones easily. What might have been a good way to do something in firejail 0.9.36 is quite possibly not the best way to do it in 0.9.56 – or may not even exist at all!

 `private-bin` for instance has gotten more powerful over time with globbing and so on, but some folks are still using older versions that are still supported, but don’t support private-bin globbing. If documents show globbing, it’ll mess up folks who use the older versions, but if they don’t, then a very useful feature is undocumented
(This is just an example, and not that great of one either!).

4. It’s a tad un-organised IMHO. This relates to my second point.
Also, while comments might be useful, they have the effect of spreading documentation thin. If someone asks how to do something, we might answer in the comments, but that’s **not** a good way of documenting features.
Also this leads to bugs and support requests being reported there, while IMHO we should have one, and only one, obvious spot for these – here! :smile: 

Instead, I propose that  we consider sites like [Read The Docs](https://docs.readthedocs.io/en/latest/index.html). This gives us multiple advantages:

1. Features are documented for all time! We can have [multiple versions up](https://docs.readthedocs.io/en/latest/versions.html#how-we-envision-versions-working) – for example, 0.9.44 docs and 0.9.56 docs at the same time. That way folks on different versions can always have the documentation that’s tailored to their specific version. And, it'll be really easy to link people to the exact docs they need.
[...]

Issue #2042, opened 2018-07-16
No ALSA sound in palemoon due to imported Firefox globals


dkavlakov:
The set of global Firefox-specific settings imported in the latest palemoon.profile includes a "nogroups" line, that prevents group-based device permissions from working and leaves Palemoon without access to  sound when ALSA is used. Unlike Firefox there is no forced Pulse Audio dependency in Palemoon (including the future 29 series) and it's profile should allow ALSA sound to be used.

chiraag-nataraj:
So I think the problem is that in the _most common_ usecase, `nogroups` provides some additional security. Removing it by default when it works for most people (and provides additional security) doesn't make much sense imo. At the same time, I recognize that it screws over people who are currently using ALSA (and will continue to do so). The problem is that `nogroups` doesn't _just_ allow ALSA to be used, right. It could enable other things which are far more nefarious. If you can document which specific groups are needed (by default, anyway), we might be able to bring in an `--enable-alsa` flag (and corresponding `enable-alsa` profile option) so that you can get ALSA sound back without having to disable `nogroups`.

netblue30:
@dkavlakov: can you run groups in a terminal, something like this:
`````
$ groups
netblue tty sudo autologin
`````

It could give us an idea what we need to add by default.

dkavlakov:
dino lp cdrom floppy tape audio dip video plugdev users netdev lpadmin scanner kvm sambashare wireshark vboxusers pulse-access bluetooth

This is a pretty typical example of the pre-systemd-takeover Debian group permissions model.

Is it security risk? It may become in case of serious misuse of the filesystem group permissions. Is it less secure than your current settings? No! Right now to have sound in Palemoon and Chromium one is supposed to install huge amount of bloatware, including entire self-sufficient privilege elevation system, managed by huge set of multiple third parties written policies. Getting rid of this _realy_ does provide additional security.

chiraag-nataraj:
@dkavlakov 
I hope you realize that pulseaudio is separate from systemd, was adopted well before systemd, and is a completely separate project. The fact that Lennart was a huge force behind both is the _only_ thing that connects the two.

[...]

PR #2096, opened 2018-08-24
Remove nogroups from audio player profiles

manevich:
`nogroups` breaks ALSA since many distros set 660 mode and `root:audio` as owner:group on `/dev/snd/*` ALSA files.

### DIFF
--- a/etc/audacious.profile
+++ b/etc/audacious.profile
@@ -23,7 +23,7 @@ apparmor
 caps.drop all
 netfilter
 nodbus
-nogroups
+# nogroups is unfeasible for ALSA since 660 mode is usualy set on /dev/snd/*
 nonewprivs
 noroot
 notv
--- a/etc/audacity.profile
+++ b/etc/audacity.profile
@@ -25,7 +25,7 @@ net none
 no3d
 # nodbus - problems on Fedora 27
 nodvd
-nogroups
+# nogroups is unfeasible for ALSA since 660 mode is usualy set on /dev/snd/*
 nonewprivs
 noroot
 notv
--- a/etc/deadbeef.profile
+++ b/etc/deadbeef.profile
@@ -19,7 +19,7 @@ include /etc/firejail/disable-xdg.inc
 caps.drop all
 netfilter
 no3d
-nogroups
+# nogroups is unfeasible for ALSA since 660 mode is usualy set on /dev/snd/*
 nonewprivs
 noroot
 notv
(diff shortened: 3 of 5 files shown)


chiraag-nataraj:
As I commented in #2042, I don't think this should be default, since `nogroups` provides a decent amount of security. We should probably document this somewhere (maybe once we get #2090 rolling?), but I don't think this is appropriate to do by default.

manevich:
But with `nogroup` audio/video player profiles are just broken. No sound for audio/video player means broken.
Is there some better way to fix it?
As far as I know most, if not all, major distros use permissions like I described on `/dev/snd/*` files.

SkewedZeppelin:
> Is there some better way to fix it?

`echo "nogroup" > /etc/firejail/globals.local`

Vincent43:
>  As far as I know most, if not all, major distros use permissions like I described on /dev/snd/* files.

Also all major distros use pulseaudio as default and this issue is irrelevant for them :smile:

manevich:
@SkewedZeppelin 
I already fixed it for me like this `echo "ignore nogroups" > /etc/firejail/audacious.local`, that not the problem.
Problem is that profiles shipped with firejail are still broken.
@chiraag-nataraj 
Are you sure **audio/video players** should be prevented from accessing ALSA device files **by default**?
Also many current profiles (like vlc) don't have `nogroups` set.

SkewedZeppelin:
`nogroups` was disabled in vlc in 4a1d906e89c0d0f8ebe6dce16b8b7c05f2c6084f
that was almost 2 years ago, we should enable it again after 0.9.56

chiraag-nataraj:
> Problem is that profiles shipped with firejail are still broken.

You're being disingenuous though. As @Vincent43 pointed out, all major distros ship with PulseAudio by default, and `nogroups` doesn't affect PulseAudio systems. So no, they're not broken on major distros by default.

> Are you sure **audio/video players** should be prevented from accessing ALSA device files **by default**?

No. But `nogroups` doesn't _just_ enable access to ALSA device files (again, as I pointed out in that issue). The real long-term solution here is to develop a `--groups`/`groups` set of options which lets the user specify which groups should be included.

And yeah, as @SkewedZeppelin said, we should really enable `nogroups` for VLC.

manevich:
OK, closed.

D-Nice:
Don't mean to bring this back from the dead, but what is the opinion of removing nogroups from browser profiles like chromium, which run into similar issues. Or should that be a user-based exception due to potential risks? And a better alternative be being able to specify groups eventually.

Vincent43:
I think it should be user-based exception, similarly. Of course the best solution would be to stop relying on legacy group ownership for audio access :smile:.

PR #2104, opened 2018-09-01
hardening evince, dbus not needed

matu3ba:
Upon testing evince, I figured out dbus may not be needed and my printer is still working.
In the initial commit 7a37dc31ab907d55eb88f2fa259f37046952a0c5 or issue #1843 
there is reasoning, however the profile still works for me.
The current apparmor profile does use dbus.

Upon use I found no side effects crashing the app when dbus was disabled.

### DIFF
--- a/etc/evince.profile
+++ b/etc/evince.profile
@@ -23,7 +23,7 @@ machine-id
 # net none breaks AppArmor on Ubuntu systems
 netfilter
 no3d
-# nodbus
+nodbus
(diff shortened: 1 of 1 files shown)


Fred-Barclay:
@matu3ba What distro are you using?

matu3ba:
@Fred-Barclay 
Ubuntu 18.04.1 LTS

PR #2115, opened 2018-09-15
Amend gnome-music profile

flacks:
- At least gnome-music 3.28.2 requires 'env'
- Enabling the disabled private-etc tested to be working


SkewedZeppelin:
Does gnome-music have any radio functionality?

PR #2123, opened 2018-09-26
Fix incorrect whitelist-path handling

miszr:
Fixed so that the check if whitelist-path is entire home directory is correctly handled.
Previously the check also matched all subdirectories and files under homedir, which basically broke all profiles whitelisting anything in homedir.

### Examples:
From "firefox.profile"
`whitelist ${HOME}/.cache/mozilla/firefox`

### DIFF
--- a/src/firejail/fs_whitelist.c
+++ b/src/firejail/fs_whitelist.c
@@ -507,7 +507,7 @@ void fs_whitelist(void) {
 			// both path and absolute path are under /home
 			if (strncmp(fname, cfg.homedir, strlen(cfg.homedir)) == 0) {
 				// entire home directory is not allowed
-				if (*(fname + strlen(cfg.homedir)) != '/') {
+				if (strlen(fname) == strlen(cfg.homedir)) {
 					free(fname);
 					goto errexit;
 				}


src/firejail/fs_whitelist.c
```diff
@@ -507,7 +507,7 @@ void fs_whitelist(void) {
 			// both path and absolute path are under /home
 			if (strncmp(fname, cfg.homedir, strlen(cfg.homedir)) == 0) {
 				// entire home directory is not allowed
-				if (*(fname + strlen(cfg.homedir)) != '/') {
+				if (strlen(fname) == strlen(cfg.homedir)) {
```
smitsohu: Hi @miszr, thank you for auditing the whitelisting sources! I'm having difficulties provoking the original bug this tries to solve. Also I tend to believe that the current logic is indeed right and has the little extra advantage of additionally catching cases like /home/user1/foo and /home/user123/foo... although this shouldn't be of much practical relevance, because fs_private() masks all other users later on, even if the sandbox was started with `--allusers`

smitsohu: @miszr please wait for a moment, I think I got an idea.

smitsohu:
Thank you for this pull request. I am going to close it, but please don't let this discourage you. Continue looking at the sources and let us know about everything you spot.

PR #2130, opened 2018-10-01
FIX-2045: Fix command name parsing for program paths with spaces.

### LINKED ISSUE #2045: Profile autodetection fails when path contains spaces
A Debian user [has reported](https://bugs.debian.org/903831) that firejail doesn't use the correct profile if the path contains spaces.
I was able to confirm that:
```
$ mkdir /tmp/nonexisting\ profile
$ cp /bin/ls /tmp/nonexisting\ profile/
$ firejail --debug /tmp/nonexisting\ profile/ls 
Autoselecting /bin/bash as shell
Building quoted command line: '/tmp/nonexisting profile/ls' 
Command name #nonexisting#
Attempting to find default.profile...
```
It tries to load the profile for the program "nonexisting" and falls back to the default profile.

crass:
This fixes #2045, which might be used to trick firejail into not auto-detecting a profile or auto-detecting the wrong profile.

### DIFF
--- a/src/firejail/util.c
+++ b/src/firejail/util.c
@@ -636,33 +636,33 @@ void extract_command_name(int index, char **argv) {
 	if (!cfg.command_name)
 		errExit("strdup");
 
-	// restrict the command name to the first word
-	char *ptr = cfg.command_name;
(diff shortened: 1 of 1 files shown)


netblue30:
Merged, thanks for the fix.

PR #2133, opened 2018-10-03
added keep-dev-shm so firefox-sync works

veloute:
keep-dev-shm allows [firefox-sync](https://wiki.archlinux.org/index.php/Firefox/Profile_on_RAM#The_script) to work

### DIFF
--- a/etc/firefox-common.profile
+++ b/etc/firefox-common.profile
@@ -43,6 +43,7 @@ shell none
 
 disable-mnt
 private-dev
+keep-dev-shm # for use with firefox-sync
 # private-etc below works fine on most distributions. There are some problems on CentOS.
 #private-etc ca-certificates,ssl,machine-id,dconf,selinux,passwd,group,hostname,hosts,localtime,nsswitch.conf,resolv.conf,xdg,gtk-2.0,gtk-3.0,X11,pango,fonts,mime.types,mailcap,asound.conf,pulse,pki,crypto-policies,ld.so.cache
 private-tmp


chiraag-nataraj:
There is absolutely no reason to have this as default in my opinion. We could document it here, but it's probably better if it's documented on that ArchWiki page (they could have written the temporary profile to `/tmp`, for example).

Fred-Barclay:
@chiraag-nataraj does this only affect Arch? IMHO I'd merge this -- it seems reasonable to make sure sync works (unless there are consequences to allowing access to /dev/shm I'm not aware of). :smile:

Fred-Barclay:
@chiraag-nataraj You're right -- I didn't even look at the link, just assumed it was the Firefox service.

@veloute Let's close this, but please don't let it discourage you from contributing in the future! But we really shouldn't allow Firefox to access /dev/shm without good reason. If it's something you need, and you're okay with the potential risks @chiraag-nataraj mentioned above, you can create a file `/etc/firejail/firefox.local` and put `keep-dev-shm` in it. IMHO it's better to edit the firefox-sync script to use a less sensitive location.

veloute:
> @chiraag-nataraj You're right -- I didn't even look at the link, just assumed it was the Firefox service.
> 
> @veloute Let's close this, but please don't let it discourage you from contributing in the future! But we really shouldn't allow Firefox to access /dev/shm without good reason. If it's something you need, and you're okay with the potential risks @chiraag-nataraj mentioned above, you can create a file `/etc/firejail/firefox.local` and put `keep-dev-shm` in it. IMHO it's better to edit the firefox-sync script to use a less sensitive location.

no worries :)

PR #2138, opened 2018-10-05
Fix incorrect --list and --tree output under certain circumstances

crass:
Firejail should look for processes with names exactly named "firejail" to avoid accounting for processes with a "firejail" prefix.  For example, I have a script named `firejail-bridged.sh` which forks a `firejail` child.  When it is running, `firejail --list` shows the script rather than the child firejail process and `firejail --tree` shows the process tree for the script and all its subprocesses.


netblue30:
All merged, thanks!

PR #2141, opened 2018-10-06
Update appimage size calculation to newest code from libappimage.

crass:
These changes specifically allow the `61.0.1.glibc2.7` version of the firefox appimage to run successfully.  Currently firejail errors out printing that the ELF size is 0 and its an appimage type 1, both of which are incorrect.  I've updated for both 32 and 64-bit appimages, but I've only tested with this 64-bit one (which loads the appimage successfully, but I can't get a successful connection).  But I believe my copy/paste skills are up to par.

This can be tested with the appimage here:
https://bintray.com/probono/AppImages/Firefox/61.0.1.glibc2.7#files

### DIFF
--- a/src/firejail/appimage_size.c
+++ b/src/firejail/appimage_size.c
@@ -18,6 +18,9 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 /*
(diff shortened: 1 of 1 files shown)


netblue30:
Merged, thanks.

PR #2152, opened 2018-10-09
Improve profile handling

crass:
The first commit is not strictly necessary, but it makes the code easier to understand because `expand_home` was really expanding all the macros.

The second commit allows specifying `--profile=firefox` on the command line and doing what you'd expect.  That is looking first in `.config/firejail` for `firefox.profile`, and if not there checking `${CFG}/firefox.profile`.

The third commit enhances the `include` directive to allow specifying just the filename component of a profile path and searching the userdir and then system config dir for the file to include.

And the fourth commit did `sed -i 's#include /etc/firejail/#include #' etc/*` to use the behavior of the third commit.  This makes it simple to "hook" various include directives without needing write access to `SYSCONFDIR`, which is especially useful for `globals.local`.


SkewedZeppelin:
So I agree that not having `/etc/firejail` hardcoded in profiles is needed, but I don't like how loose/fuzzy this makes it. I think we need something in between.

crass:
I'll note here that actually this is incomplete without the corresponding documentation changes that I forgot.

@SkewedZeppelin, I'm happy to discuss your concerns further.  I understand your uneasy feeling and I've thought about it, but I haven't come up yet with a specific case where this causes problems.  I don't think it makes `.local` files useless on a multi-user system, with a single user that it does seem to.  My problem with `.local` files is that they require root to edit, which is not nice.  Also, its nice to keep the configuration of unprivileged programs in `$HOME`.

As for the uneasy feeling, tell me if I'm wrong, but I think I stems from how with this change a root owned profile can load a user owned profile, which then loads a root owned profile.  It seems like the user owned profile might be able to do something malicious.  But I don't think it can do anything that it wouldn't be able to do by completely specifying a user owned profile that doesn't include any root owned profiles.

Also, I don't think it makes things necessarily any more confusing (of course it would be easy to come up with confusing examples, but you could have those under the current situation now too).  There are two main beneficial ways to using that new feature that I had in mind.
1. The clearest benefit is with `globals.local`.  Now the `globals.local` can be updated by an unprivileged user, making it easy for users to make profile-wide changes.
2. Similarly, if I want to add to `disable-interpreters.inc`, there's not a painless way for me to to this as a user.  I'd have to copy `disable-interpreters.inc` and all the profiles that I want to use that include that.  Which would then make it less maintainable.
3. The other use case is to try to maintain a DRY (Don't Repeat Yourself) config set.  So say as a user you want to modify `firefox-common.profile`. Currently you have to copy `firefox.profile` to `$HOME/.config/firejail`, then create a `firefox-common.profile` stub profile which adds the changes and then includes `${CFG}/firefox-common.profile`.  The `firefox.profile` doesn't need to be there, and potentially doing that way you won't get an update when the sysadmin upgrades firejail and thus the `${CFG}/firefox.profile` file.

I think this feature clearly has its benefits.  And I suppose someone could do something convoluted with it if they wanted to, but the real question is _How easy is it to shoot yourself in the foot_?  What we don't want are features that expose you to easily making yourself less secure.

crass:
@SkewedZeppelin would you feel more comfortable with #2155 being added?  I think it would go nicely with this pull.

SkewedZeppelin:
@crass #2155 would help and should be implemented either way.
I do agree with your benefits of this, I just think there need to be more handles on it.

crass:
Ok, I've decided to break this up in to two pull requests.  This is the first two commits.  The second two will be in another pull request.

chiraag-nataraj:
@SkewedZeppelin How is this different from the way `firejail` currently selects a profile to use if none is specified? So if you do `firejail firefox`, it uses the program name to search for the corresponding profile, and it looks in `.config/firejail` first and then `${CFG}`. Isn't this basically the same thing, but generalized to the `--profile` and (now in #2156) to the `include` directive in profiles? 

I guess what I don't get is why this wouldn't have the same issues that `firejail` already has when searching for which profile file to use. And local files aren't really useless since you can still reference the global `.local` files from your personal `.profile` files (or override the system `.local` files with your own without messing with the system files).

I think I agree with @crass that this is useful and makes firejail profiles more flexible.

SkewedZeppelin:
I guess I was just uneasy at first.
With the four commits plus the fifth to optionally disable user profiles the end result is much improvement.

smitsohu:
> I don't consider this a work in progress right now.

I guess we just don't have a "review in progress" badge :smile:

Fred-Barclay:
Alright, I'm happy with this one. :smile: @crass you want to merge (and welcome aboard, BTW!).

smitsohu:
@Fred-Barclay Shouldn't we wait with the merge until 0.9.56.1 is out?

> This is probably a bugfix release: fixes, small features, new profiles. If we end up implementing something major we'll switch to a regular 0.9.57 release.

from caec31e7af6771af04fa2b2b98850422d1604c2b

Fred-Barclay:
@smitsohu @crass Yep, didn't realise that. :)
Let's don't merge to master yet. Maybe merge to a different branch if we don't want to keep these open?

crass:
@Fred-Barclay @smitsohu should I make a `devel` or `next` branch for these merges?  It makes a little more work for the developers.  Personally, I think we should just merge these features when they are approved.  So if this is considered big enough of a new feature, then we merge it and bump the version to `0.9.57` and move on.  I would consider this a "small feature" and thus leave the version at `0.9.56.1`.  Anyone have thoughts on this?

crass:
@Fred-Barclay  This renames `expand_home` to `expand_macros` and allows profile detection for `--profile` arguments (eg. `--profile=firefox` will load `${CFG}/firefox.profile` unless there is a `${HOME}/.config/firejail/firefox.profile`.

I'm fine with waiting to merge #2156, as its less useful to me. However, philosophically, I don't think we should let version numbers ever get in the way of merging code.  Version numbers are cheap, updating a patch becomes more expensive as time goes by.  We could move to `0.9.60` tomorrow if we wanted to and it wouldn't matter a bit.  So this is a call for everyone to put their 2 cents on the pq's in and lets get the merge/modified/rejected...

Vincent43:
As I read https://github.com/netblue30/firejail#current-development-version-09561 , I think @netblue30 will switch to `0.9.57` as needed. If this is approved there is no need to wait.

crass:
I actually meant the doc to read `Reading profile /etc/firejail/icecat.profile`, to show that it worked for `${CFG}` profiles.  But this change is fine with me.  Who is expected to do the merge?  Me?  or someone else?

Fred-Barclay:
@crass normally we merge our own PRs except for when @netblue30 is faster and beats us to 'em. :laughing:

Fred-Barclay:
Only request -- the git history looks a bit of a mess here. Can we clean it up before merging to master?

crass:
Yeah, I've rebased pq onto current master, did a merge, then pushed upstream.

PR #2156, opened 2018-10-09
Improve include handling

crass:
This relies on pull #2152.  It should be referenced for the initial commentary on these commits.  So the last two pertain to this pull request.

The third commit enhances the `include` directive to allow specifying just the filename component of a profile path and searching the userdir and then system config dir for the file to include.

And the fourth commit did `sed -i 's#include /etc/firejail/#include #' etc/*` to use the behavior of the third commit. This makes it simple to "hook" various include directives without needing write access to `SYSCONFDIR`, which is especially useful for `globals.local`.


Fred-Barclay:
Please, folks, can we not merge this (and the others from @crass ) for a few days? I'd like to look at it in-depth and mull for a bit.
Thanks!

Sorry too @crass !

crass:
I've updated the branch to be rebased off the branch in #2152 and added documentation changes.

SkewedZeppelin:
@crass yes, you'll need to rebase for all the new profiles that were added.

crass:
Thanks @Fred-Barclay, I'll add that documentation as well while merging.

crass:
Actually, the documentation on `${CFG}` appears to be sufficiently clear under the `include` section of `firejail-profile`.  Let me know if there was somewhere specifically that you thought needed it.

Fred-Barclay:
@crass for some reason `/${CFG}` in `man firejail-profile` was returning "pattern not found". I see it now; it looks good. Thanks!

PR #2157, opened 2018-10-09
Fix 2155 - Add user-profiles configuration option.

### LINKED ISSUE #2155: Add firejail.config option to prevent loading of user profiles.
On systems where the unprivileged user does not have access to root, the sysadmin should have the ability to disallow custom profiles.  This will lower the attack surface on firejail itself.  In order for this to be effective, when user profiles are disabled, the `--profile` option should only allow loading files from `${CFG}`.  What should this option be named in `firejail.config`?  Perhaps `user_profiles`? And I suggest having them off by default (secure default).

crass:
This is based on the work in pull #2152.  The actual code for this pull is the last commit.  Its pretty straight forward.  This fixes #2155.

### DIFF
--- a/etc/firejail.config
+++ b/etc/firejail.config
@@ -5,6 +5,9 @@
 # Enable AppArmor functionality, default enabled.
 # apparmor yes
 
+# Enable per user profiles in ${HOME}/.config/firejail/, default enabled.
+# user-profiles yes
+
 # Number of ARP probes sent when assigning an IP address for --net option,
 # default 2. This is a partial implementation of RFC 5227. A 0.5 seconds
 # timeout is implemented for each probe. Increase this number to 4 if your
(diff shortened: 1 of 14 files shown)


reinerh:
I like the new setting, but I don't like disabling user profiles by default.
This would break the setups of some users (that are currently using user profiles) on the next upgrade.
I think upgrades should be more graceful, and the new option could be mentioned in the release notes / changelog.
Also I'm not sure if the security gain of disabling user profiles by default is big enough to break current behavior.

Fred-Barclay:
@crass What are your current thoughts on this? Do you still think it shouldn't be merged (if I understand your last comment correctly)? If so, we maybe should close now. :smile:

Fred-Barclay:
Closed for now -- let's reopen if needed.

PR #2159, opened 2018-10-11
allow overriding of disable-mnt with noblacklist

### LINKED ISSUE #2154: Split up disable-mnt
We should split `disable-mnt` into two different directives (`disable-mnt` and `disable-media`) to make things more granular.

smitsohu:
with an ambition to fix #2154


netblue30:
all set, thanks!

PR #2163, opened 2018-10-11
Harden strings.profile

glitsj16:
The strings profile seems to work just fine here when adding a few extra options.

Regards.


netblue30:
Merged, thanks.

PR #2164, opened 2018-10-11
New profile aria2c

glitsj16:
This program hasn't got a GUI (at least not in aria2c itself), so no need to add it to firecfg I guess.

Regards.


netblue30:
Merged, thanks.

PR #2165, opened 2018-10-11
Authenticator

glitsj16:
Not sure how/when the GNOME filenaming guru's came up with something so ugly (personal opinion), but that's the actual name of the .desktop file (https://gitlab.gnome.org/World/Authenticator/blob/master/data/com.github.bilelmoussaoui.Authenticator.desktop.in)...

Side-note: personally I don't use `firecfg`, but I assume the list entree should be pointing to the program's .desktop file... may need a double-check from someone that does use it.

Regards.

### DIFF
--- a/etc/authenticator.profile
+++ b/etc/authenticator.profile
@@ -1,0 +1,49 @@
+# Firejail profile for authenticator
+# Description: 2FA code generator for GNOME
+# This file is overwritten after every install/update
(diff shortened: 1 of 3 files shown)


netblue30:
Merged, thanks.

SkewedZeppelin:
firecfg just uses the regular binary name, in this case: authenticator

PR #2169, opened 2018-10-11
New profile devilspie

glitsj16:
This is the older version, not to be confused with devilspie2. Offers a few extra's that aren't in devilspie2, so it might be of interest to people still (especially because it has a GUI called gdevilspie, which I'm working on to add to the profiles).

Regards.


netblue30:
Merged, thanks.

PR #2178, opened 2018-10-13
Fix atool for UID/GID > 1000

glitsj16:


glitsj16:
@Fred-Barclay Good question. I only noticed it recently when building a package for Arch with makepkg, which uses fakeroot. Did some more testing and it looks as if it does more than just complain:

$ id -u glitsj16
1001

$ firejail /usr/bin/atool --list firejail-master.zip
Error: cannot read UID_MIN and/or GID_MIN from /etc/login.defs, using 1000 by default
Error unshare: main.c:884 main: Operation not permitted

$ firejail /usr/bin/atool --extract firejail-master.zip
Error: cannot read UID_MIN and/or GID_MIN from /etc/login.defs, using 1000 by default
Error unshare: main.c:884 main: Operation not permitted
firejail-master.zip: archive is empty

glitsj16:
@Fred-Barclay Follow-up: when I whitelist the path where firejail-master.zip resides, atool (and associates) seem to work just fine.

$ firejail --whitelist=~/Downloads /usr/bin/atool --list firejail-master.zip
Archive:  firejail-master.zip
ba4fc8f08c89d87061ab1bba0305b81c6af1bd7e
  Length      Date    Time    Name
        0  2018-10-13 16:23   firejail-master/
      592  2018-10-13 16:23   firejail-master/.gitignore
<output continues as expected>

So `login.defs` is not needed, but shouldn't atool work `everywhere`?

netblue30:
Merged!

PR #2181, opened 2018-10-13
New profile gnome pie

glitsj16:


SkewedZeppelin:
The consensus on programs that launch other programs was decided to not be added to firecfg (so it isn't enabled by default).

Does stuff launched launch under it? else it'll need to be removed.

SkewedZeppelin:
Merged, thanks!

PR #2182, opened 2018-10-13
New profile mencoder

glitsj16:


SkewedZeppelin:
Is mencoder part of mplayer? can it extend mplayer.profile?

glitsj16:
That might indeed be a better way to handle it. Redoing the profile now, need to do some testing first, but it should be fine extending mplayer.profile.

SkewedZeppelin:
Merged, thanks!

PR #2186, opened 2018-10-13
Fix issue #2148: Make sure firejail can find helper programs in sandbox regardless of options.

### LINKED ISSUE #2148: firejail should never fail to find its helper binaries for arbitrary install paths
I found this annoying as I wanted to install a development version in a sub-directory of `$HOME`.  But when I run firefox with firejail using the development binary and profile, I get the following debug lines:
```
sbox run: /home/user/development.forked/firejail.git/installed/lib/firejail/fseccomp drop /run/firejail/mnt/seccomp /run/firejail/mnt/seccomp.postexec @clock,@cpu-emulation,@debug,@module,@obsolete,@raw-io,@reboot,@resources,@swap,acct,add_key,bpf,fanotify_init,io_cancel,io_destroy,io_getevents,io_setup,io_submit,ioprio_set,kcmp,keyctl,mount,name_to_handle_at,nfsservctl,ni_syscall,open_by_handle_at,personality,pivot_root,process_vm_readv,ptrace,remap_file_pages,request_key,setdomainname,sethostname,syslog,umount,umount2,userfaultfd,vhangup,vmsplice allow-debuggers (null) 
Dropping all capabilities
Drop privileges: pid 7, uid 1000, gid 1000, nogroups 1
No supplementary groups
execvp: No such file or directory
Error: failed to run /home/user/development.forked/firejail.git/installed/lib/firejail/fseccomp
Error: proc 26027 cannot sync with peer: unexpected EOF
```

If I add `--whitelist=/home/user/development.forked/firejail.git/installed/lib/firejail`, things work as expected.

Also, if firejail is installed under some mount directory (ie `/media/extra_storage/local-install`), a similar error happens when firejail is run with the `--disable-mnt` option. A similar problem would likely happen if `$HOME` is on a different mount point under `/media` and symlinked in `/home`.  However, here the problem is not resolved with an added `--whitelist` option.  The issue appears to be that by the time firejail attempts to whitelist the firejail lib install directory, `/media` has already been blacklist and thus set to deny users read access.

Firejail should succeed in both these cases (and in all cases where it has been properly installed).  The first issue is fairly straight forward to resolve, but the second one is trickier.

crass:
These changes allow firejail to always be able to find and run its helper libraries even when they are installed in a location that would otherwise be inaccessible in the sandbox (eg. under `${HOME}` or `/media` when `disable-mnt` is enabled), which fixes #2148.  The basic idea behind the implementation is to first unshare firejail's mount namespace and make all mounts private.  Then early on we bind-mount `LIBDIR`, where the helpers programs are actually stored, to `/run/firejail/lib` and have all the helpers be run from there.

Despite there being a fair amount of lines to the changeset, the magic really happens in only a few lines.  The unsharing of the mount namespace and making mounts private happens in `src/firejail/main.c` line 883-886, and the bind-mount of `LIBDIR` in `src/firejail/preproc.c` lines 77-78.

### DIFF
--- a/src/firejail/bandwidth.c
+++ b/src/firejail/bandwidth.c
@@ -406,17 +406,17 @@ void bandwidth_pid(pid_t pid, const char *command, const char *dev, int down, in
 	if (devname) {
 		if (strcmp(command, "set") == 0) {
 			if (asprintf(&cmd, "%s/firejail/fshaper.sh --%s %s %d %d",
-				LIBDIR, command, devname, down, up) == -1)
+				RUN_FIREJAIL_LIB_DIR, command, devname, down, up) == -1)
 				errExit("asprintf");
 		}
 		else {
 			if (asprintf(&cmd, "%s/firejail/fshaper.sh --%s %s",
-				LIBDIR, command, devname) == -1)
+				RUN_FIREJAIL_LIB_DIR, command, devname) == -1)
 				errExit("asprintf");
 		}
 	}
 	else {
-		if (asprintf(&cmd, "%s/firejail/fshaper.sh --%s", LIBDIR, command) == -1)
+		if (asprintf(&cmd, "%s/firejail/fshaper.sh --%s", RUN_FIREJAIL_LIB_DIR, command) == -1)
 			errExit("asprintf");
 	}
 	assert(cmd);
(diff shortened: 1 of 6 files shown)


netblue30:
Looks fine, thanks!

PR #2190, opened 2018-10-13
Revert "Fix issue #2148: Make sure firejail can find helper programs in sandbox regardless of options."

netblue30:
Reverts netblue30/firejail#2186


netblue30:
reopening it, run into some problems with the interaction between mount and network namespaces. I'll bring it back in after some testing.

netblue30:
All fixed!

crass:
Hi @netblue30 
Thanks for taking a look at this.  I'm curious though, what issues did you run into with the original changeset, which had the `unshare` in the main process?  It looks like it might have revolved around `fnet`, but I had no issue with networking with my changes.  Doing the `unshare` in the sandbox seems fine, I'm just wondering what I missed.  BTW, thanks for all the great work!

crass:
I'm reopening this because, I've run into an issue where the path to `libpostexecseccomp.so` in `ld.so.preload` is to `LIBDIR`, which doesn't exist in the sandbox.  I believe we should just change it back to `RUN_FIREJAIL_LIB_DIR`, right?

crass:
Closing because it doesn't make much sense to reopen a pull request...

netblue30:
> I'm just wondering what I missed

You didn't miss anything. The patch was excellent. I run the tests on it ("make test") and it was clean. However, when I started clicking on menus in my desktop manager (LXDE), programs with a network namespace would not start. I still cannot explain it, starting the programs from command line with a network namespace would work fine.

So I moved the mount at the beginning of the sandbox process, and all was fine. I did run into some existing bug with private-lib. It looks like fnet, fseccomp, etc. are linked in with libapparmor, so I needed to add this library by default. All seems to be fine now, probably we'll fix some minor problems. Anyway, thanks for the patch!

crass:
Thanks for taking the time to write the explanation :smile:.  A side effect of that change though is more complexity in the code and issue #2197.  Also, a side-effect of the original code is that firejail didn't have to work about any mounts in the main process (eg. appimage).  Sometimes I get still mounted appimage mounts if I have to kill firejail, so I have to clean those up.  I think its nice to just not worry about the mounts.

PR #2192, opened 2018-10-13
The path in ld.so.preload should point to RUN_FIREJAIL_LIB_DIR, as LIBDIR may not exist.

crass:
This got broken in pq #2190, which partially reverted #2186.


startx2017:
fixed!

PR #2193, opened 2018-10-14
Make --join return exit code of the invoked program

grimskies:
This PR changes the behavior of --join command, making it return the exit code of the invoked program.

```
pawel@azath:~$ firejail --join=test exit 123
Switching to pid 10387, the first child process inside the sandbox
Child process initialized in 45.88 ms
pawel@azath:~$ echo $?
123
```


startx2017:
Thank you for the patch.

PR #2199, opened 2018-10-16
Fix #2142: Firefox appimage fails because it needs non-default seccomp

### LINKED ISSUE #2142: Firefox 61.0.1.glibc2.7 Appimage with multi-process windows enabled does not successfully load pages
Using up to date master with #2141 added, the `61.0.1.glibc2.7` firefox appimage displays, with saved tabs when using a pre-existing profile.  However none of the tabs load, including `file:///` urls.  Version `56.0.2.glibc2.3.4` works just fine.  I would presume all the latest profile fixes are in master, is that a wrong assumption?  Has anyone else gotten 61 non-appimage to run?

https://bintray.com/probono/AppImages/Firefox/61.0.1.glibc2.7#files

crass:
Do not override user provided seccomp lists when in chroot/overlay/appimage, but to use the default if none is provided.  This is a more general fix.  I see no reason we are forcing users to use a default seccomp filter with those options.  There should be a default if no seccomp filter is provided, but if one is provided, it should not be ignored in favor of the default.  This allows the firefox appimage to work, because its sandboxing need chroot, which is disabled in the @default seccomp filter (see #2142).

### DIFF
--- a/src/firejail/sandbox.c
+++ b/src/firejail/sandbox.c
@@ -530,14 +530,6 @@ static void enforce_filters(void) {
 #ifdef HAVE_SECCOMP
 	enforce_seccomp = 1;
 #endif
-	if (cfg.seccomp_list_drop) {
-		free(cfg.seccomp_list_drop);
-		cfg.seccomp_list_drop = NULL;
-	}
-	if (cfg.seccomp_list_keep) {
-		free(cfg.seccomp_list_keep);
-		cfg.seccomp_list_keep = NULL;
-	}
 
 	// disable all capabilities
 	if (arg_caps_default_filter || arg_caps_list)
@@ -547,8 +539,7 @@ static void enforce_filters(void) {
 	// drop all supplementary groups; /etc/group file inside chroot
 	// is controlled by a regular usr
 	arg_nogroups = 1;
-	fmessage("\n** Warning: dropping all Linux capabilities and enforcing  **\n");
-	fmessage("**                  default seccomp filter                 **\n\n");
+	fmessage("\n**     Warning: dropping all Linux capabilities     **\n");
 }
 
 int sandbox(void* sandbox_arg) {


chiraag-nataraj:
@netblue30 Is there any reason we override the provided seccomp list if using those options?

netblue30:
The user can include in appimage a SUID binary and take over the box. We do three things to stop him: set default seccomp, set nonewprivs, clear all capabilities. I guess any one of the three will do it. I would say we can leave seccomp out and depend only on nonewprivs and capabilities. So we merge it, and fix whatever comes after.

smitsohu:
But this is CVE-2016-10123 revisited. Enforcement of no_new_privs is broken now.

smitsohu:
Please disregard above comment. I'm asking myself though if no_new_privs could be enforced independently even without HAVE_SECCOMP compile flag.

crass:
> The user can include in appimage a SUID binary and take over the box.

@netblue30 Wow, so if I understand correctly appimage without any kind of jailing is a priv escalation vuln?  I suppose that might flow from allowing fuse mounted iso which can have a setuid binary and can be loaded by an under privileged program.  However, I would think this would be a more widely known bug.  Or am I reading this wrong?  Is there a CVE or other documentation for this?  I've been searching but haven't quite found anything.

Fred-Barclay:
@crass not appimage but possibly similar: https://www.cvedetails.com/cve/CVE-2017-9780/

smitsohu:
@crass As far as I understand, fuse addresses the problem in a different way by enforcing nosuid, nodev mounts. There _are_ indeed CVE's where it was possible to circumvent this restriction, a rather famous example being [CVE-2016-1576](https://www.openwall.com/lists/oss-security/2016/02/24/8).

Firejail doesn't use fuse, and it doesn't mount the appimage filesystem nosuid, nodev. Instead we are using nonewprivs, capabilities and seccomp, which is just a different (and IMHO better) way to achieve the same thing.
However, I'm wondering if it would be possible to just also add these two mount flags to firejail's appimage mounts, at least if the user is non-privileged. After all, all non-sandboxed appimages are running fine with nosuid, nodev. Not that it would be necessary, but as security in depth.

crass:
Yep, I knew firejail doesn't use fuse, I was/am just surprised/skeptical about the non-firejail scenario.  I like your idea of adding `nosuid` and `nodev` to the appimage loopback mounts. This could easily be done on [line 116 and 121 of appimage.c](https://github.com/netblue30/firejail/blob/d69e0cf1b35cac9185081bd6d95d82024868ae76/src/firejail/appimage.c#L116). If we find any appimages legitimately needing suid or devices in them, then we can revisit this.

PR #2201, opened 2018-10-17
Add nou2f to all profiles

### LINKED ISSUE #2194: add nou2f to all profiles with private-dev
We punched a small hole in `private-dev` to get yubikeys and such working, but it hasn't been closed yet in the profiles that we ship.

I would propose to add `nou2f` to all profiles with `private-dev`, including the browsers and keepassxc. The reason is that /dev/usb can contain devices such as printers and keyboards, and exposing them in the sandbox is usually neither necessary nor desirable.
Imho users who own a yubikey and want to use it with firejail should rather opt-in and disable nou2f themselves.

smitsohu:
- Closes #2194

### DIFF
--- a/etc/0ad.profile
+++ b/etc/0ad.profile
@@ -32,6 +32,7 @@ nogroups
 nonewprivs
 noroot
 notv
+nou2f
 novideo
 protocol unix,inet,inet6
 seccomp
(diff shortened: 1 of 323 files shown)


netblue30:
all set!

PR #2227, opened 2018-10-30
Expanded the comment about nodbus breaking stuff

alexara:
As I figured out nodbus may break the streaming experience with Firefox 63.0 on Plasma 5.14.2, Framework 5.51.0

* KDE Connect: Users won't be able to control a stream with KDE Connect anymore
* power management: During a stream the screen may dim or turn off


reinerh:
thanks, merged.

PR #2237, opened 2018-11-05
Update enpass.profile

glitsj16:
Added description and `# Firejail profile for enpass` to be consistent with existing profiles.


SkewedZeppelin:
Merged, thanks!

PR #2247, opened 2018-11-05
Add new config option to disable U2F in browsers, enabled by default

SkewedZeppelin:
Allows easily reenabling security dongles where needed.

Am I missing anything?

Also should we consider allowing U2F in supported browsers by default?
U2F dongles can be had for under $10 and add easy security to online accounts.
With Google's recent push again and the future FIDO2 potential we'll hopefully see increased adoption of it.


netblue30:
merged!

njfox:
What do I need to put in my `firefox.local` to re-enable U2F?

SkewedZeppelin:
@njfox Nothing, just edit `/etc/firejail/firejail.config` and change `#browser-disable-u2f yes` to `browser-disable-u2f no`

njfox:
Thanks, I think I misread the description--I thought it was saying that u2f would be disabled by default after this patch. My FIDO U2F key currently isnt working in firefox under firejail and I thought this might be the culprit.

SkewedZeppelin:
@njfox go to about:config and change `security.webauth.u2f` to true

njfox:
I did, the key works without firejail but doesn't work with firejail. I've noticed there are some previous issues about this that suggested something like `--ignore=privatedev` so I'll give that a try

SkewedZeppelin:
@njfox to confirm you did change `browser-disable-u2f` to no in `/etc/firejail/firejail.config`, correct?

njfox:
Just tried, didn't fix the problem

SkewedZeppelin:
@njfox to confirm, you are running firejail 0.9.58.2?

njfox:
```
$ firejail --version
firejail version 0.9.58
```

SkewedZeppelin:
@njfox try just removing the whole `nou2f` line from `/etc/firejail/firefox-common.profile`
if that doesn't work then try `firejail --ignore=private-dev /usr/bin/firefox` in addition to it

njfox:
Removing `?BROWSER_DISABLE_U2F: nou2f` from `/etc/firejail/firefox-common.profile` resolved the issue. Is this something I should open an issue about?

SkewedZeppelin:
@njfox you can if you want, I however cannot reproduce.

njfox:
Thanks for your help. Just to make sure I understand, does #2201 essentially turn off U2F for everything by default?

SkewedZeppelin:
@njfox yes, #2201 disabled it in all profiles by default. #2247 however made it easier to reenable in profiles that actually utilize it

PR #2253, opened 2018-11-08
Have appimage handling be the same with or with out special -- argument.

crass:
When not using `--` on the command line along with `--appimage`, `extract_command_name` is not run and instead `cfg.command_name` is set and `arg_shell_none` is set to `0`.  This is what one wants when running appimages where the `AppRun` is a shell script (eg. digikam).  However, this logic is not consistently applied when processing the `--` special option.

The effect of this is that using `--` with the digikam appimage and digikam profile fails to run digikam because the digikam profile sets `shell none`.  This can be overcome by adding `ignore=shell`, but that shouldn't be necessary.

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -2244,21 +2244,18 @@ int main(int argc, char **argv) {
 				return 1;
 			}
 		}
-		else if (strcmp(argv[i], "--") == 0) {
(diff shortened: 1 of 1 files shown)


netblue30:
thanks, all set.

Issue #2252, opened 2018-11-08
More complete documentation for -c option


crass:
The documentation for the `-c` option for firejail should be more detailed.  Its not clear to me why I would use it.  The manpage says `Execute command and exit.`, which isn't helpful because that's what I would expect firejail to do by default.  Knowing a bit more about firejail, I might expect that it means the firejailed process is forked and the monitor process exits straight away, instead of waiting for the child process to terminate.  However, upon reviewing the code this doesn't seem to be accurate either.

As far as I can tell by looking at `arg_command` usage, the `-c` switch just turns off some extra/debug messages.  So what was/is the intent of this option?  If I've estimated it correctly, then it should be renamed and the documentation should be more accurate/descriptive.  Anyone have ideas about this?

netblue30:
-c is a standard option for shells (bash, dash, etc), it is usually inserted on the command line by some of the login programs. You can use firejail as a restricted login shell (see man firejail-login) from a ssh or telnet server for example. There could be some other login applications,

I'll take a look at the code and clean some of it up. The code has been that way since forever.

crass:
Ahh, ok, makes sense.  I don't have any issue with the code per se, just didn't see its purpose.  The documentation should include these details.

PR #2254, opened 2018-11-09
Add better documentation for "-c" option.

crass:
I've taken a stab at updating the documentation on the `-c` option per the discussion in #2252.  Let me know if its incorrect or should be changed in some way.


src/firejail/usage.c
```diff
@@ -42,7 +42,7 @@ static char *usage_str =
 	"    --blacklist=filename - blacklist directory or file.\n"
 	"    --build - build a whitelisted profile for the application.\n"
 	"    --build=filename - build a whitelisted profile for the application.\n"
-	"    -c - execute command and exit.\n"
+	"    -c - login shell compatibility option (has no effect).\n"
```
reinerh: if it's not meant to be run by the user and has no effect, I wouldn't even show it in the usage list.

PR #2255, opened 2018-11-09
Allow prefixing colon to profile argument of --profile to for a profile search

crass:
I ran into this while executing digikam with firejail while in the current directory there was a file named `digikam`.  So firejail was trying to read that file as a profile file.  This patch would allow me to use `--profile=:digikam` to force digikam to do a profile search instead of looking in the current directory for profiles.


netblue30:
Merged.

PR #2260, opened 2018-11-15
Fix Evince profile

announ:
Evince 3.30.1 on Ubuntu 18.10 fails to open with the following error unless `private-etc machine-id` is allowed:
```
dbus[194]: D-Bus library appears to be incorrectly set up: see the manual page for dbus-uuidgen to correct this issue. (Failed to open "/var/lib/dbus/machine-id": No such file or directory; Failed to open "/etc/machine-id": No such file or directory)
```


Fred-Barclay:
Merged, thanks!

PR #2272, opened 2018-11-24
removing vim and ranger from firecfg

veloute:
ranger is a file manager - https://github.com/netblue30/firejail/issues/1261
vim is often used to edit system files - sandboxing breaks functionality.


SkewedZeppelin:
Merged, thanks!

Issue #2262, opened 2018-11-17
local/python for mpv and youtube-dl


pirate486743186:
the distribution provided youtube-dl is always seriously outdated, but it's important to be able to use the latest.
An easy way of installing youtube-dl is with pip. It will be the latest and installed in /usr/local . youtube-dl can be used by other apps, at least one of them is mpv.

So for youtube-dl and mpv and maybe other apps that depend on youtube-dl ,add:

```
noblacklist /usr/local/lib/python2*
noblacklist /usr/local/lib/python3*
```

PR #2275, opened 2018-11-26
allowing local python* in mpv and youtube-dl #2262

pirate486743186:
allowing local python* in mpv and youtube-dl #2262 .For use with latest youtube-dl installed with pip. Having the latest is very important here.

Maybe others are using it? grep only gives these two.


SkewedZeppelin:
Merged, thanks!

PR #2276, opened 2018-11-26
refactor private-cache and tmpfs

smitsohu:
has the immediate benefit that the result of combining `--noexec` and `--tmpfs` does not depend on the sequence of the options.


netblue30:
Merged, thanks.

PR #2278, opened 2018-11-27
Update kdenlive.profile

rusty-snake:
Add mlt-melt to private-bin, this is needed on Fedora-systems.

The programm `melt` is needed for rendering, because there are more than one programm that is called `melt`, Fedora (and maybe some other distros) install the `melt`-programm from MLT as `mlt-melt` under `/usr/bin`.


SkewedZeppelin:
Merged, thanks!

PR #2279, opened 2018-11-27
allowing youtube-dl and python in gnome-mpv

pirate486743186:
like the title says. Allowing youtube-dl.


SkewedZeppelin:
Merged, thanks!

PR #2281, opened 2018-11-28
restricting more, HOME and tmp in mpsyt.profile

pirate486743186:
Of course i got it wrong XD
Restricting HOME and tmp in mpsyt.profile

### DIFF
--- a/etc/mpsyt.profile
+++ b/etc/mpsyt.profile
@@ -13,6 +13,18 @@ noblacklist ${HOME}/.netrc
 noblacklist ${HOME}/mps
 noblacklist ${MUSIC}
 noblacklist ${VIDEOS}
+noblacklist ${DOWNLOADS}
+
+mkdir ${HOME}/.config/mps-youtube
+
+whitelist ${HOME}/.config/mpv
+whitelist ${HOME}/.mplayer
+whitelist ${HOME}/.config/mps-youtube
+whitelist ${HOME}/.netrc
+whitelist ${HOME}/mps
+whitelist ${MUSIC}
+whitelist ${VIDEOS}
+whitelist ${DOWNLOADS}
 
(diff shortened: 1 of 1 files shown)


glitsj16:
@pirate486743186 You might want to consider adding `ffmpeg` to `private-bin` for transcoding support. See $HOME/.config/mps-youtube/transcode for details.

etc/mpsyt.profile
```diff
@@ -13,6 +13,18 @@ noblacklist ${HOME}/.netrc
 noblacklist ${HOME}/mps
 noblacklist ${MUSIC}
 noblacklist ${VIDEOS}
+noblacklist ${DOWNLOADS}
+
+mkdir ${HOME}/.config/mps-youtube
+
+whitelist ${HOME}/.config/mpv
```
SkewedZeppelin: these should be grouped with the other whitelists

etc/mpsyt.profile
```diff
@@ -13,6 +13,18 @@ noblacklist ${HOME}/.netrc
...
+whitelist ${HOME}/.mplayer
+whitelist ${HOME}/.config/mps-youtube
+whitelist ${HOME}/.netrc
+whitelist ${HOME}/mps
+whitelist ${MUSIC}
+whitelist ${VIDEOS}
+whitelist ${DOWNLOADS}
```
SkewedZeppelin: you should also add the following above the other whitelist inc
`include /etc/firejail/whitelist-common.inc`

glitsj16: @SkewedZeppelin Shouldn't that be added without `/etc/firejail` as `include whitelist-common.inc`?

SkewedZeppelin:
Merged, and fixed up in cc9db57343c3df0a9fc191e31273c955314c8f72

PR #2285, opened 2018-12-03
add HAS_NODBUS conditional, ${RUNUSER} makro

smitsohu:
based upon which the KDE sockets can be restricted better


smitsohu:
will replace it with something better ...

Issue #1917, opened 2018-04-29
LibreOffice won't start on Ubuntu 18.04


derba:
$ libreoffice 
Reading profile /etc/firejail/libreoffice.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Parent pid 21258, child pid 21259
Blacklist violations are logged to syslog
Child process initialized in 77.09 ms
Warning: failed to launch javaldx - java may not function correctly
ERROR 4 forking process

Parent is shutting down, bye...

$ journalctl | tail
ápr 29 19:33:20 Lapi audit[20977]: AVC apparmor="ALLOWED" operation="open" profile="libreoffice-oopslash" name="/run/firejail/mnt/fslogger" pid=20977 comm="oosplash" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
ápr 29 19:33:20 Lapi kernel: audit: type=1400 audit(1525023200.547:90): apparmor="ALLOWED" operation="open" profile="libreoffice-oopslash" name="/run/firejail/mnt/fslogger" pid=20977 comm="oosplash" requested_mask="r" denied_mask="r" fsuid=1000 ouid=0
ápr 29 19:33:20 Lapi audit[20991]: AVC apparmor="ALLOWED" operation="exec" info="no new privs" error=-1 profile="libreoffice-oopslash" name="/usr/lib/libreoffice/program/javaldx" pid=20991 comm="osl_executeProc" requested_mask="x" denied_mask="x" fsuid=1000 ouid=0 target="unconfined"
ápr 29 19:33:20 Lapi audit[20993]: AVC apparmor="ALLOWED" operation="exec" info="no new privs" error=-1 profile="libreoffice-oopslash" name="/usr/lib/libreoffice/program/soffice.bin" pid=20993 comm="osl_executeProc" requested_mask="x" denied_mask="x" fsuid=1000 ouid=0 target="libreoffice-soffice"
[...]

PR #2293, opened 2018-12-09
enable apparmor in libreoffice profile

smitsohu:
originally this was tracked in #1917


smitsohu:
@Fred-Barclay From what I've seen, it works out of the box for Ubuntu/Mint/Debian users. Only if they comment the `apparmor` line there is a need for additional adjustments. 

Since Ubuntu's apparmor policy is not in enforce mode, the proposed profile should lead to better Libreoffice confinement for almost everyone.

PR #2294, opened 2018-12-10
Add a profile for thunderbird-wayland

rusty-snake:
Add a profile for thunderbird-wayland and enable ist for firecfg.


SkewedZeppelin:
Merged, thanks.

PR #2297, opened 2018-12-15
enforce nonewprivs instead of seccomp for chroot sandboxes

smitsohu:
Currently users are always able to specify a seccomp filter of their choosing, including for chroot, appimage and overlay sandboxes. For example, Firejail allows me to do something like:

`firejail --noprofile --chroot=/somedir --seccomp.drop=merry,christmas`

Since the requirement for seccomp looks more or less obsolete now, this PR proposes to remove it entirely and replace it with an explicit nonewprivs enforcement.

There is no security benefit (or harm, AFAICT) in this. It is only about the streamlining of requirements and keeping things easy to explain. While I would regard the patch as ready for merging, I was also hoping to inspire an exchange if it is actually heading in the right direction.

Cheers!

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -866,7 +866,6 @@ int main(int argc, char **argv) {
 	int lockfd_directory = -1;
 	int option_cgroup = 0;
 	int custom_profile = 0;	// custom profile loaded
-	int arg_seccomp_cmdline = 0; 	// seccomp requested on command line (used to break out of --chroot)
 	int arg_caps_cmdline = 0; 	// caps requested on command line (used to break out of --chroot)
 
 	// drop permissions by default and rise them when required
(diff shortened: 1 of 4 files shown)


startx2017:
I was under the impression we were already doing it! It greatly simplifies the way we can explain it to our users. Somebody from Debian brought this up and we ended with two different behaviors depending on seccomp declaration (command line vs profile file). This way is much better.

@smithosu - can you please push the same fix on LTSbase branch, thanks!

startx2017:
@netblue30 - we can also remove 0.9.38-LTS branch. By now nobody should be using it.

smitsohu:
> @smithosu - can you please push the same fix on LTSbase branch, thanks!

Sure, it will take a few days though.

netblue30:
I removed 0.9.38-LTS branch.

PR #2308, opened 2018-12-20
Update disable-common.inc, disable-programs.inc.

rusty-snake:
Add some missing paths to disable-programs.inc and add some paths to disable-common.inc.
<br><br>
By the way, what about the paths of programs for which there is no profile, can you simply add them? (e.g. zim, tilix, aqemu, eclipse,...)

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -272,6 +272,11 @@ read-only ${HOME}/.bin
 read-only ${HOME}/.local/bin
 read-only ${HOME}/.cargo/bin
 
+# cargo
+read-only ${HOME}/.cargo/env
+blacklist ${HOME}/.cargo/registry
+blacklist ${HOME}/.cargo/config
+
 # Write-protection for desktop entries
 read-only ${HOME}/.config/menus
 read-only ${HOME}/.local/share/applications
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -268,6 +268,9 @@ read-only ${HOME}/.gem
 read-only ${HOME}/.luarocks
 read-only ${HOME}/.npm-packages
 read-only ${HOME}/bin
+read-only ${HOME}/.bin
+read-only ${HOME}/.local/bin
+read-only ${HOME}/.cargo/bin
 
 # Write-protection for desktop entries
 read-only ${HOME}/.config/menus
--- a/etc/disable-programs.inc
+++ b/etc/disable-programs.inc
@@ -395,6 +395,7 @@ blacklist ${HOME}/.local/share/akregator
 blacklist ${HOME}/.local/share/apps/korganizer
 blacklist ${HOME}/.local/share/aspyr-media
 blacklist ${HOME}/.local/share/baloo
+blacklist ${HOME}/.local/share/bibletime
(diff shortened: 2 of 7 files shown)


etc/disable-common.inc
```diff
@@ -268,6 +268,9 @@ read-only ${HOME}/.gem
 read-only ${HOME}/.luarocks
 read-only ${HOME}/.npm-packages
 read-only ${HOME}/bin
+read-only ${HOME}/.bin
+read-only ${HOME}/.local/bin
+read-only ${HOME}/.cargo/bin
```
SkewedZeppelin: why not just make all of cargo readonly?

etc/disable-programs.inc
```diff
@@ -452,6 +453,7 @@ blacklist ${HOME}/.local/share/ocenaudio
 blacklist ${HOME}/.local/share/okular
 blacklist ${HOME}/.local/share/orage
 blacklist ${HOME}/.local/share/org.kde.gwenview
+blacklist ${HOME}/.local/share/rhythmbox
```
SkewedZeppelin: rhythmbox.profile will need to be updated

SkewedZeppelin:
> By the way, what about the paths of programs for which there is no profile, can you simply add them? 

With care, yes.

etc/disable-programs.inc
rusty-snake: Upps Forgotten, coming soon.

etc/disable-common.inc
rusty-snake: Idea:
```
$ ls .cargo
bin/  config  env  registry/
```
config and registry are only needed by cargo it self, so we can blacklist it.
env is maybe used by some shell config files (e.g. .profile, .bashrc,...). only read-only.

etc/disable-programs.inc
```diff
@@ -395,6 +395,7 @@ blacklist ${HOME}/.local/share/akregator
 blacklist ${HOME}/.local/share/apps/korganizer
 blacklist ${HOME}/.local/share/aspyr-media
 blacklist ${HOME}/.local/share/baloo
+blacklist ${HOME}/.local/share/bibletime
```
rusty-snake: Fix in bibletime.profile comming soon.

etc/disable-common.inc
SkewedZeppelin: I don't rust, so I'll trust your judgement.

etc/disable-programs.inc
rusty-snake: Sorry, but now is there

etc/disable-programs.inc
rusty-snake: Fixed.

etc/disable-common.inc
rusty-snake: OK, after a short look, if we blacklist `registry` we muss update atom, gnome-builder, bracktes, maybe vim and emacs, in the future geany. If this is ok i will update these profiles tomorrow. It's getting late for me ;)

rusty-snake: OK, I've made some changes.

SkewedZeppelin:
Merged, thanks!

PR #2313, opened 2018-12-23
Fix gajim.profile

rusty-snake:
On my system (F29) gajim only starts if it has access to the user default shell.

### DIFF
--- a/etc/gajim.profile
+++ b/etc/gajim.profile
@@ -43,7 +43,7 @@ protocol unix,inet,inet6
 seccomp
 
 disable-mnt
-private-bin python,python3,sh,gpg,gpg2,gajim
+private-bin python,python3,sh,gpg,gpg2,gajim,bash,zsh
 private-dev
 private-etc alsa,asound.conf,ca-certificates,crypto-policies,fonts,group,hostname,hosts,ld.so.cache,ld.so.conf,localtime,machine-id,passwd,pki,pulse,resolv.conf,ssl
 private-tmp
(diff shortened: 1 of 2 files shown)


rusty-snake:
Gajim also needs access to `paplay` to play notification tones.

rusty-snake:
Don't merge yet, I think I'll have a few more fixes in the next few days.

etc/gajim.profile
```diff
@@ -41,9 +41,15 @@ notv
 nou2f
 protocol unix,inet,inet6
 seccomp
+tracelog
+shell none
```
SkewedZeppelin: swap

rusty-snake: OK. Done.

etc/gajim.profile
```diff
@@ -41,9 +41,15 @@ notv
...
 
 disable-mnt
-private-bin python,python3,sh,gpg,gpg2,gajim
+private-bin python,python3,sh,gpg,gpg2,gajim,bash,zsh,paplay,gajim-history-manager
+private-cache
 private-dev
 private-etc alsa,asound.conf,ca-certificates,crypto-policies,fonts,group,hostname,hosts,ld.so.cache,ld.so.conf,localtime,machine-id,passwd,pki,pulse,resolv.conf,ssl
 private-tmp
+
+noexec ${HOME}
```
Fred-Barclay: Will `noxec ${HOME}` prevent plugins from running?

rusty-snake: Short: No.
Long: I was thinking, too, but I didn't notice any mistakes. I checked and all plugin files in `.local/share/gajim` have `rw-r-----`. All plugins are written in python.

SkewedZeppelin:
Merged, thanks!

PR #2319, opened 2018-12-27
Add a profile for ghostwriter

rusty-snake:
Add a profile for ghostwriter.
Ghostwriter is a text editor for Markdown.

### DIFF
--- a/etc/ghostwriter.profile
+++ b/etc/ghostwriter.profile
@@ -21,8 +21,8 @@ include whitelist-common.inc
 apparmor
 caps.drop all
 netfilter
-no3d
-nodbus
+#no3d
+#nodbus
 nodvd
 nogroups
 nonewprivs
@@ -31,17 +31,19 @@ nosound
 notv
 nou2f
 novideo
-protocol unix
+protocol unix,inet,netlink
 seccomp
 shell none
 tracelog
 machine-id
 
-private-bin ghostwriter
+# Breaks Translation
+#private-bin ghostwriter,pandoc
 private-cache
 private-dev
 private-etc cups,crypto-policies,localtime,drirc,fonts,xdg,gtk-3.0,dconf,machine-id
-private-lib
+# Breaks Translation
+#private-lib
 private-tmp
 
 noexec ${HOME}
--- a/etc/ghostwriter.profile
+++ b/etc/ghostwriter.profile
@@ -1,0 +1,48 @@
+# Firejail profile for ghostwriter
+# Description: Cross-platform, aesthetic, distraction-free Markdown editor.
+# This file is overwritten after every install/update
+# Persistent local customizations
+include ghostwriter.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/ghostwriter
+
+include disable-common.inc
+include disable-devel.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+
+mkdir ${HOME}/.config/ghostwriter
+whitelist ${HOME}/.config/ghostwriter
+include whitelist-common.inc
+
+apparmor
+caps.drop all
+netfilter
+no3d
+nodbus
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+shell none
+tracelog
+machine-id
+
+private-bin ghostwriter
+private-cache
+private-dev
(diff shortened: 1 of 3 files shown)


rusty-snake:
I hope everything works now.

etc/ghostwriter.profile
```diff
@@ -0,0 +1,50 @@
...
+
+include disable-common.inc
+include disable-devel.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+
+mkdir ${HOME}/.config/ghostwriter
+whitelist ${HOME}/.config/ghostwriter
+include whitelist-common.inc
```
Fred-Barclay: If whitelists are included then markdown files from Documents and other locations can't be accessed.

etc/ghostwriter.profile
```diff
@@ -0,0 +1,50 @@
...
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,netlink
+seccomp
+shell none
+tracelog
+machine-id
```
Fred-Barclay: Please move `machine-id` between `caps.drop.all` and `netfilter`.

etc/ghostwriter.profile
```diff
@@ -0,0 +1,50 @@
...
+
+# Breaks Translation
+#private-bin ghostwriter,pandoc
+private-cache
+private-dev
+private-etc cups,crypto-policies,localtime,drirc,fonts,xdg,gtk-3.0,dconf,machine-id
```
Fred-Barclay: What happens if `xdg` is removed?

rusty-snake: Nothing special. No differences.
This is there because its the outputline from `firejail --build`

etc/ghostwriter.profile
rusty-snake: Uhh good argument.
Is it better to whitelist ${Documents}, ${Downaloads}, ... or remove the whitelist?

etc/ghostwriter.profile
rusty-snake: Fixed, not pushed yet.

etc/ghostwriter.profile
SkewedZeppelin: For existing text editors we allow everywhere
just include disable-xdg and noblacklist documents

Vincent43: I think adding ${Documents}, ${Downloads} would be enough.

SkewedZeppelin: let me reword that
```
include disable-common.inc
include disable-passwdmgr.inc
include disable-programs.inc
```
are in all the current editors, and should be included here too

rusty-snake: My idea:
```
noblacklist ${HOME}/.config/ghostwriter
noblacklist ${DOCUMENTS}
noblacklist ${PICTURES}

include disable-common.inc
include disable-devel.inc
include disable-passwdmgr.inc
include disable-programs.inc
include disable-xdg.inc

mkdir ${HOME}/.config/ghostwriter
mkdir ${DESKTOP}
mkdir ${DOCUMENTS}
mkdir ${DOWNLOADS}
mkdir ${PICTURES}
whitelist ${HOME}/.config/ghostwriter
whitelist ${DESKTOP}
whitelist ${DOCUMENTS}
whitelist ${DOWNLOADS}
whitelist ${PICTURES}
include whitelist-common.inc
```

etc/ghostwriter.profile
rusty-snake: Thanks, but I still think a hint in `CONTRIBUTING.md` is good for newcomers.

etc/ghostwriter.profile
Vincent43: @SkewedZeppelin if you remove whitelist then all (non explicitly blacklisted) dotfiles will be accessible. Are we expecting users will store markdown files somewhere in dotfiles?

@rusty-snake why ${PICTURES}? The rest looks good to me.

rusty-snake: @Vincent43 For inserting screen photos, for example.

rusty-snake: For this I use `*.local` in `~/.config/firejail.profile`

Vincent43: Yeah, this is double edged sword. Certainly some users wouldn't want that every random app can access their custom home dirs and dotfiles. IMO it's better to fail-safe here rather than fail-open and use `*.local` for overrides.

rusty-snake: Now the whitelit is comment out.

Fred-Barclay:
Merged, thanks!

PR #2334, opened 2019-01-11
restrict audacious profile

pirate486743186:
maybe restrict it a bit more?


SkewedZeppelin:
we don't currently use whitelist in lollypop, rhythmbox, or gnome-music, and probably others.

pirate486743186:
beh :P

PR #2344, opened 2019-01-16
Update gajim.profile

rusty-snake:
Fix plugin-update and install.

Remove `private-cache`.
Reason: I saw that on my system (F29) it break the plugin install and update. (Gajim 1.1.1)


SkewedZeppelin:
Merged, thanks!

PR #2347, opened 2019-01-18
add google earth pro, update google earth profile

Lockdis:


etc/google-earth-pro.profile
```diff
@@ -0,0 +1,4 @@
+# Redirect
+include google-earth.profile
+
+private-bin google-earth-pro,sh,bash,grep,sed,ls,dirname
```
SkewedZeppelin: you just need `private-bin google-earth-pro` since the rest are already included

SkewedZeppelin:
Merged, thanks!

PR #2349, opened 2019-01-19
comment `machine-id` in libreoffice.profile

rusty-snake:
I made a presentation with impress. The presentation contains a video. The video also worked but when machine-id is on, the sound doesn't work.


Fred-Barclay:
Merged, thanks! `machine-id` blocks pulseaudio from working, good catch! :smile:

PR #2353, opened 2019-01-20
Harden gnome-calculator profile

glitsj16:


etc/gnome-calculator.profile
```diff
@@ -37,7 +37,9 @@ shell none
 
 disable-mnt
 private-bin gnome-calculator
+private-cache
 private-dev
+private-etc fonts
```
SkewedZeppelin: this will probably break currency conversion, which we do allow

PR #2358, opened 2019-01-21
Update peek.profile

Lockdis:
switched to whitelist profile + remove nodbus (breaks the dropdown menu for the format selection on arch based manjaro kde)

### DIFF
--- a/etc/peek.profile
+++ b/etc/peek.profile
@@ -9,6 +9,12 @@ noblacklist ${HOME}/.cache/peek
 noblacklist ${PICTURES}
 noblacklist ${VIDEOS}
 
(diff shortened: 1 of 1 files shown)


Lockdis:
also `net none` is giving this error :  dbind-WARNING **: 05:23:48.632: Couldn't connect to accessibility bus: Failed to connect to socket /tmp/dbus-8w6O3wUFuT: Connection refused

but the program seems to works fine.

Lockdis:
it break configuration saving,
probably /usr/share/glib-2.0/schemas/com.uploadedlobster.peek.gschema.xml

PR #2366, opened 2019-01-24
add nyx, crow, fix g earth pro

Lockdis:
added crow, nyx profile and blacklisted his folder (it contain the TOR remote control password), added missing reference to google earth pro


etc/nyx.profile
```diff
@@ -0,0 +1,58 @@
...
+noblacklist ${PATH}/python3*
+noblacklist /usr/include/python3*
+noblacklist /usr/lib/python3*
+noblacklist /usr/local/lib/python3*
+noblacklist /usr/share/python3*
+
+noblacklist ${HOME}/.nyx
+
+mkdir ${HOME}/.nyx
+
```
Fred-Barclay: No empty line here please

etc/nyx.profile
```diff
@@ -0,0 +1,58 @@
...
+protocol unix,inet,inet6
+seccomp
+shell none
+# tracelog
+
+disable-mnt
+private-bin nyx,python
+private-cache
+private-dev
+private-etc passwd,tor
```
Fred-Barclay: Maybe add `fonts`? Also, does nyx need internet connection?

SkewedZeppelin: nyx should basically be a copy of arm.profile, a lot changed but the profile still works iirc

etc/crow.profile
```diff
@@ -0,0 +1,57 @@
+# Firejail profile for crow
+# This file is overwritten after every install/update
+# Persistent local customizations
+include crow.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/crow
```
smitsohu: can you add a corresponding blacklist to disable-programs.inc or another place you deem appropriate?

etc/nyx.profile
Lockdis: arm profile is giving access to more binaries and etc, are they needed for nyx? it seems to work fine without

Lockdis: @Fred-Barclay without protocol unix,inet,inet6 it's not possible to connect to the localhost tor, should we block internet using a netfilter rule?

etc/crow.profile
Lockdis: i have removed them as they are not blacklisted, thanks

etc/crow.profile
```diff
@@ -23,35 +20,28 @@ include disable-xdg.inc
 
 include whitelist-common.inc
 
-# apparmor
 caps.drop all
-# ipc-namespace
+ipc-namespace
 netfilter
 no3d
-nodbus
```
Lockdis: breaks status bar icon

etc/nyx.profile
SkewedZeppelin: > arm profile is giving access to more binaries and etc

ah, it does not, I forgot nyx lost the ability to start Tor itself, and can only monitor now


for the protocol, most systems default to the control socket, so you can `protocol unix` and add a comment

glitsj16:
@Fred-Barclay I'm assuming you're questioning whether it's useful to add/keep commented lines in profiles is more general than these nyx/crow/google earth pro, correct? Personally I have always been contributing profiles like that. More out of habit than anything else, I like having sort of a format of the most generic options to start building on. But I wouldn't object if this workflow would change according to clear guidelines. I would dislike the comments inside profiles disappearing though, these I find very useful indeed. Just my 2 cents :-)

Please disregard if I read your remark out of context and the above makes no sense.

glitsj16:
@SkewedZeppelin Looks like a clear and workable guideline to me. I for one can definately live with that. Should we have a seperate thread on overhauling all profiles accordingly? Or is there such an issue already?

SkewedZeppelin:
@glitsj16 see #1427, there are probably only a handful of profiles that have deviated since then.

etc/nyx.profile
Fred-Barclay: I'm mostly surprised this works without adding the likes of `resolv.conf` and whatnot to private-etc -- see, for example, the spotify and firefox-common profiles.

SkewedZeppelin: @Fred-Barclay nyx only connects to Tor's control socket or port, it itself makes no connections to the Internet

etc/crow.profile
```diff
@@ -0,0 +1,47 @@
...
+include disable-devel.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+include whitelist-common.inc
+
+caps.drop all
+ipc-namespace
```
Fred-Barclay: ipc-namespace tends to break on certain distributions. Probably better to remove or at the very least comment out.

etc/nyx.profile
```diff
@@ -0,0 +1,52 @@
...
+
+include disable-common.inc
+include disable-devel.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+caps.drop all
+ipc-namespace
```
Fred-Barclay: ditto on ipc-namespace :wink:

etc/nyx.profile
```diff
@@ -0,0 +1,52 @@
...
+seccomp
+shell none
+
+disable-mnt
+private-bin nyx,python*
+private-cache
+private-dev
+private-etc passwd,tor,fonts
+private-opt none
+private-srv none
```
Fred-Barclay: Not a review, just a comment -- I wonder about the usefulness of using `none` for private-* filters by default? I *think* it's something I've tended to avoid by default, but maybe we should use it more frequently???

Fred-Barclay:
@Lockdis I'm good with merging once `ipc-namespace` is removed. :wink:

Fred-Barclay:
Oops ... was trying to push changes removing `ipc-namespace` and now github won't let me re-open this PR. :fearful: Anyhow, will manually merge these changes. Thanks @Lockdis !

PR #2373, opened 2019-01-27
Fix gnome-maps

rusty-snake:
Without `noblacklist ${HOME}/.local/share/flatpak` gnome-maps not start on Fedora 29.


Fred-Barclay:
Merged, thanks!

PR #2382, opened 2019-01-30
Create firefox-wayland.profile

rusty-snake:
Same as #2294 .


rusty-snake:
There is one. Upps. Sorry.

PR #2384, opened 2019-02-01
remove nodbus from MPRIS client profiles

carloabelli:
MPRIS is a D-Bus interface for controlling media players. The nodbus
option prevents these players from being controlled through MPRIS.
It seems that most of the other media players did not have the nodbus
option or it was already commented out.


Vincent43:
@carloabelli could you add comment above `nodbus` option that it conflicts with MPRIS?

Issue #2383, opened 2019-01-30
Issue with Streaming / Streamed YouTube Videos


aberja:
When I run `firejail vivaldi` I cannot watch YouTube videos that are streaming or were previously streamed.  I receive a message that states "An error occurred.  Please try again later..."

I can watch non streamed YouTube videos without any issue.  I can also watch the streamed and streaming YouTube vidoes in Vivaldi without firejail.

If I close Vivaldi and run `firejail --noprofile vivaldi` I can watch the streamed and streaming YouTube videos without any issue.

Any thoughts on why I need to use --noprofile and what I need to do to resolve this issue?

Setup:
Mint LMDE 3
Vivaldi version: 2.2.1388.37 (Stable) 64 bit
Firejail version: 0.9.59

SkewedZeppelin:
Are you using the proprietary NVIDIA drivers?

Using --noprofile, provides little to no security benefits.

aberja:
Yes, I'm using proprietary NVIDIA drivers for the GTX1070 video card on my ASUS G752vs laptop.

Fred-Barclay:
Does `firejail --ignore=noroot vivaldi` work?

rusty-snake:
Does `firejail --ignore=private-dev vivaldi` work?

Lockdis:
i'm on manjaro the first time i have launched vivaldi in firejail i was able to play a live stream on youtube then it stopped to works, to fix this i had to install the ffmpegs codecs, (vivaldi-codecs-ffmpeg-extra-bin on aur for precompiled bins) and it works again with firejail vivaldi-stable and default probiles

aberja:
> Does `firejail --ignore=private-dev vivaldi` work?

No that does not resolve the issue.

> it's a problem related to codecs, 

I hear what you are saying but if I run `firejail vivaldi` or any of the other varieties with firejail I receive the codecs message and streaming YouTube videos will not run.

if I just run `vivaldi` , I never receive the codecs issue and the streaming YouTube videos always work.

> edit not sure if it matter but i notice that you have installed version of codecs 70.* will mine are 71.0.3578.98-0

Thanks for the heads-up on the codecs 71 vs 70.  Although I noticed that your screenshot from vivaldi refers to the same 70 codec that I have installed.  I'm assuming your codecs 71.* is coming from the arch repository which I think is always on the bleeding edge :)

SkewedZeppelin:
[...]

Issue #2380, opened 2019-01-29
firejail 0.9.58 breaks many programs (execute permission denied)


odiferousmint:
I have no idea why, but I get:

```
Error: execute permission denied [...]
Error: no suitable X executable found
```

for many programs. 0.9.56 is the last version that works.

I have the directory whitelisted, so I have no idea why it would not work whereas 0.9.56 works just fine.

The programs in question are wire and vivaldi. Probably breaks many other programs which I have not tried yet.

Any ideas? What am I missing? Do I need to add something to my local profile to be able to run an executable inside ${HOME}?

Edit: it seems like it runs if I get rid of `noexec ${HOME}`. How come that it works in 0.9.56 but not in 0.9.58 with `noexec ${HOME}` present?

In the meantime I think I found a bug. If I include `disable-programs.inc` that blacklists 2 paths, I will get a permission denied error *even* after I have manually whitelisted and noblacklisted them at the end of the profile file. It seems like as if that `noblacklist` and `whitelist` does not overwrite the previous `blacklist`. This is a bug, right?

SkewedZeppelin:
See https://github.com/netblue30/firejail/issues/2375#issuecomment-458379200

netblue30:
> for many programs.

@odiferousmint what programs are you having problems with noexec home? So far we figured out the browsers, but there could be more.

Lockdis:
everything works for me now, not sure if it's because i have updated firejail today or something else was wrong with my pc, anyways it's fixed for me. thx

chiraag-nataraj:
I'll go ahead and close this, as it looks like the issue has been resolved.

odiferousmint:
> > for many programs.
> 
> @odiferousmint what programs are you having problems with noexec home? So far we figured out the browsers, but there could be more.

I only tested with `wire-desktop` and `vivaldi`.

Issue #2378, opened 2019-01-29
firejail 0.9.58 and youtube-dl issue


martywd:
OS:  Linux Mint 19.1 MATE 64-bit
youtube-dl v. 2019.01.27
upgrading firejail from 0.9.56 > 0.9.58

After downloading and installing 'firejail-apparmor_0.9.58_1_amd64.deb' from sourceforge, the 'firejail youtube-dl' <.....> command yields the follow errors:

> ~ $ firejail youtube-dl --help
> Error: execute permission denied for /home/marty/bin/youtube-dl
> Error: no suitable youtube-dl executable found
> ~ $ firejail youtube-dl --version
> Error: execute permission denied for /home/marty/bin/youtube-dl
> Error: no suitable youtube-dl executable found

Permissions for 'youtube-dl':

> -rwx------ 1 marty marty 1733727 Jan 28 14:43 /home/marty/bin/youtube-dl

Purged the 'firejail-apparmor' install and installed 'firejail_0.9.58~rc1_1_amd64.deb'.  No change, same errors when attempting to use 'youtube-dl' with firejail* 0.9.58.

Reverted back to 'firejail_0.9.56_1_amd64.deb' and now the 'firejail youtube-dl' command functions **_without_** errors once again.
.

SkewedZeppelin:
Please try
`firejail --ignore='noexec ${HOME}' /home/marty/bin/youtube-dl`

martywd:
> Please try
> `firejail --ignore='noexec ${HOME}' /home/marty/bin/youtube-dl`
> 

**Yes** that command works for me from the command line.  Actually, 
> `firejail --ignore='noexec ${HOME}' youtube-dl`
>

works as well.  Thanks for the tip!

> noexec before 0.9.58 didn't recursively apply to paths, allowing programs in ~ for example to run when they shouldn't of
> see: [a7164cb](https://github.com/netblue30/firejail/commit/a7164cb38171c0291fac7f3a0767fe8f1c69b02d) and #2375

I'll check those links out, thanks again.

Now to figure out _how_ to enter this 'ignore' into ~/.config/firejail/youtube-dl.profile' so I don't have to add the above switch all the time?

.

SkewedZeppelin:
`echo 'ignore noexec ${HOME}' > ~/.config/firejail/youtube-dl.local`
should work

martywd:
> `echo 'ignore noexec ${HOME}' > ~/.config/firejail/youtube-dl.local`
[...]

Issue #2375, opened 2019-01-28
LD_PRELOAD failed to map segment from shared object


damluk:
Use case:
```
firejail --env="LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libcurl.so.4:$HOME/.local/share/spotify/spotifywm.so" spotify
...
ERROR: ld.so: object '/home/damluk/.local/share/spotify/spotifywm.so' from LD_PRELOAD cannot be preloaded (failed to map segment from shared object): ignored.
```

This has worked with 0.9.56. If I can trust git bisect, the first bad commit was a7164cb3

damluk:
The ignore option helped, thanks!

PR #2386, opened 2019-02-01
Temporary fix for noexec ${HOME} breakage

### LINKED ISSUE #2385: Firefox cannot play Netflix videos if started in firejail
OS: Arch Linux 
 Kernel: x86_64 Linux 4.20.6-zen1-1-zen
 DE: XFCE4
 WM: Xfwm4
 CPU: Intel Core i5-4590 @ 4x 3.7GHz
 GPU: GeForce GTX 970

firejail version 0.9.58
Mozilla Firefox 65.0

I upgraded firejail (0.9.56-1 -> 0.9.58-1) on Jan 28th and since I cannot watch Netflix videos (DRM) in firefox. When I load one, just the gray background is shown. When I close firefox, the process(es) stay open in the background so I have to kill them.
firejail --debug or journalctl show nothing suspicious.
The config I use is the standard config minus `noroot` because of https://wiki.archlinux.org/index.php/Firejail#Proprietary_Nvidia_drivers

If I can do anything to help, I'll be glad to do so :-)

SkewedZeppelin:
We are probably missing more, feel free to commit directly to this branch

Closes #2385, #2383, #2380, #2378, #2375

### DIFF
--- a/etc/firefox-common.profile
+++ b/etc/firefox-common.profile
@@ -51,5 +51,6 @@ private-dev
 #private-etc ca-certificates,ssl,machine-id,dconf,selinux,passwd,group,hostname,hosts,localtime,nsswitch.conf,resolv.conf,xdg,gtk-2.0,gtk-3.0,X11,pango,fonts,mime.types,mailcap,asound.conf,pulse,pki,crypto-policies,ld.so.cache
 private-tmp
 
-noexec ${HOME}
+# breaks DRM binaries
+#noexec ${HOME}
 noexec /tmp
--- a/etc/kodi.profile
+++ b/etc/kodi.profile
@@ -41,5 +41,6 @@ tracelog
 private-dev
 private-tmp
 
-noexec ${HOME}
+# breaks plugins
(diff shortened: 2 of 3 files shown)


Fred-Barclay:
@SkewedZeppelin is it safe to assume that pretty much any program that can play video is broken? If so I'll add VLC, xplayer, totem,...

SkewedZeppelin:
@Fred-Barclay I don't think that is the case

Vincent43:
@Fred-Barclay only those which try to execute binaries from `$HOME` are affected

@SkewedZeppelin you may add comments why it's disabled. Also `apparmor` should have same effect as `noexec $HOME` so we either have to disable it as well or point users to [workaround](https://github.com/netblue30/firejail/blob/master/etc/firejail-default#L80).

smitsohu:
Does this qualify for a bugfix release?
@netblue30

netblue30:
Yes, bugfix release next Friday, Feb 8. Let's check it in!

PR #2391, opened 2019-02-06
Add a profile for klavaro

rusty-snake:


etc/klavaro.profile
```diff
@@ -0,0 +1,55 @@
...
+mkdir ${HOME}/.local/share/klavaro
+mkdir ${HOME}/.config/klavaro
+whitelist ${HOME}/.local/share/klavaro
+whitelist ${HOME}/.config/klavaro
+include whitelist-common.inc
+
+apparmor
+caps.drop all
+machine-id
+netfilter
```
SkewedZeppelin: netfilter -> net none

etc/klavaro.profile
```diff
@@ -0,0 +1,55 @@
...
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.local/share/klavaro
+mkdir ${HOME}/.config/klavaro
+whitelist ${HOME}/.local/share/klavaro
+whitelist ${HOME}/.config/klavaro
+include whitelist-common.inc
```
SkewedZeppelin: include whitelist-var-common.inc

etc/klavaro.profile
```diff
@@ -0,0 +1,55 @@
...
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+shell none
+tracelog
+
+disable-mnt
+private-bin klavaro
```
SkewedZeppelin: add tclsh
edit: more are missing: execvp: No such file or directory

etc/klavaro.profile
```diff
@@ -0,0 +1,55 @@
...
+private-cache
+private-dev
+private-etc fonts
+private-lib
```
SkewedZeppelin: remove

etc/klavaro.profile
rusty-snake: interesting, I don't have any problems or messages like this.
`firejail --profile=klavaro.profile --debug klavaro 2>&1 | grep execvp`
`execvp argument 0: klavaro`

SkewedZeppelin: wow, you're correct
using your profile it works
but I was just testing no profile with `private-bin klavaro` and it wouldn't launch

weird

adding tclsh and bash get it going without a profile

SkewedZeppelin:
Merged, thanks!

PR #2398, opened 2019-02-10
Fix snap.profile description

glitsj16:
Probably not the best of descriptions, suggestions welcome.


Fred-Barclay:
I think it's a great description! Certainly better than the DNA one... :rofl:

PR #2403, opened 2019-02-14
minor fixes to keepassxc, thunderbird and pluma

PalinuroSec:
We tested these patches in Parrot OS and we believe they could be useful in upstream too


Fred-Barclay:
Merged, thanks!

PR #2409, opened 2019-02-17
Added profile for webui-aria2

schtobia:
[webui-aria2](https://github.com/ziahamza/webui-aria2) is a popular UI
for the aria2 download manager.

### DIFF
--- a/etc/webui-aria2.profile
+++ b/etc/webui-aria2.profile
@@ -1,0 +1,31 @@
+# Firejail profile for webui-aria2
+# This file is overwritten after every install/update
+# Persistent local customizations
+include webui-aria2.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${PATH}/node
+
+include disable-common.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-devel.inc
+include disable-interpreters.inc
+
+caps.drop all
+netfilter
+nodvd
+nogroups
+nonewprivs
+noroot
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+
+private-cache
+private-dev
+private-tmp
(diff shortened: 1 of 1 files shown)


SkewedZeppelin:
how would a user use this profile?
`firejail --profile=webui-aria2 node node-server.js`?
maybe add that as a comment to the profile

does any distro ship it?

schtobia:
Added the comment. `webui-aria2` is AFAIK not packaged by any distro, but it is [recommended by the german wiki for ubuntu](https://wiki.ubuntuusers.de/aria2/#webui-aria2). I think it's the only reasonable maintained frontend for `aria2`.

glitsj16:
I'm not seeing how `aria2c` would be able to work with this profile when that's also being firejailed. Perhaps a redirect profile would be more suitable?

```
# Firejail profile for webui-aria2
include webui-aria2.local
# already included by redirect profile
#include globals.local

noblacklist ${PATH}/node

private-bin node

# Redirect
include aria2c.profile
```

glitsj16:
@schtobia Great. In that case I would merge this if you would make a few smallish changes.

> include disable-common.inc
> include disable-passwdmgr.inc
> include disable-programs.inc
> include disable-devel.inc
> include disable-interpreters.inc
> 
Can you order these alphabetically, cfr. our other profiles? Does webui-aria2 work when adding `include disable-xdg.inc` as well here?

> caps.drop all
> netfilter

Is it possible to add/did you test whether these options work or break webui-aria2?
`nodbus`
`nosound`

`noexec ${HOME}`
`noexec /tmp`

schtobia:
Done. May I remark that the order of the `include` statements came from the [webstorm profile](/netblue30/firejail/blob/master/etc/webstorm.profile), as this was the blueprint for my profile?

glitsj16:
@schtobia Thanks. Merged and fixed webstorm in the process via https://github.com/netblue30/firejail/pull/2418.

PR #2412, opened 2019-02-17
Add profiles for mypaint & mypaint-ora-thumbnailer

rusty-snake:


etc/mypaint.profile
```diff
@@ -0,0 +1,51 @@
...
+disable-mnt
+private-cache
+private-dev
+private-etc fonts,gtk-3.0,dconf
+private-tmp
+
+noexec ${HOME}
+noexec /tmp
+
+# vim:set syntax=sh:
```
SkewedZeppelin: remove

etc/mypaint.profile
```diff
@@ -0,0 +1,51 @@
...
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+shell none
+tracelog
+
+disable-mnt
```
SkewedZeppelin: remove
not in krita or gimp
users might store work on another drive

etc/mypaint.profile
rusty-snake: Upps, forgotten to remove.

SkewedZeppelin:
Merged, thanks!

Issue #2399, opened 2019-02-11
Flameshot - error while loading shared libraries


aberja:

aberja:
When I try to run flameshot with firejail I receive the following error:
`flameshot: error while loading shared libraries: libGL.so.1: cannot open shared object file: No such file or directory`

I tried adding `ld.so.cache` to the flameshot.profile, but that does not resolve the issue.

Commenting out the following line does resolve the issue:
`#private-etc fonts,ld.so.conf,resolv.conf,ca-certificates,ssl,pki,crypto-policies`

```
g752vs@g752vs:~$ firejail flameshot
Reading profile /usr/local/etc/firejail/flameshot.profile
Reading profile /usr/local/etc/firejail/disable-common.inc
Reading profile /usr/local/etc/firejail/disable-devel.inc
Reading profile /usr/local/etc/firejail/disable-interpreters.inc
Reading profile /usr/local/etc/firejail/disable-passwdmgr.inc
Reading profile /usr/local/etc/firejail/disable-programs.inc
Reading profile /usr/local/etc/firejail/disable-xdg.inc
Parent pid 22354, child pid 22355
Warning: skipping crypto-policies for private /etc
Private /etc installed in 13.34 ms
1 program installed in 2.00 ms
Child process initialized in 81.53 ms
flameshot: error while loading shared libraries: libGL.so.1: cannot open shared object file: No such file or directory
```

SkewedZeppelin:
@aberja is the following from #2383 still the same?

> Mint LMDE 3
> Firejail version: 0.9.59
> NVIDIA Driver version: 384.130

Fred-Barclay:
@aberja Also does this actively interfere with using Flameshot? I get the same message on my system (also LMDE 3) but flameshot works well as far as I can tell.

If it doesn't actually interfere it's likely this is just cosmetic.

Cheers!
Fred

aberja:
> @aberja is the following from #2383 still the same?
> 
> > Mint LMDE 3
> > Firejail version: 0.9.59
> > NVIDIA Driver version: 384.130

Yes

aberja:
> Also does this actively interfere with using Flameshot?

Yes, Unfortunately Flameshot will not start unless I comment out the line.

chiraag-nataraj:
Add `alternatives` to the `private-etc` line.

aberja:
> Add `alternatives` to the `private-etc` line. 

[...]

PR #2415, opened 2019-02-17
Add 'alternatives' to all private-etc lines

Fred-Barclay:
See #2399 for context/discussion. @chiraag-nataraj suggested that we make sure we add `alternatives` to all profiles with private-etc. In particular this should help avoid the error `<some program>: error while loading shared libraries: libGL.so.1: cannot open shared object file:` for systems with multiple versions of libGL.so.

For profiles with only `private-etc none` (see list below), none was replaced with alternatives. 
1. default
2. devilspie
3. devilspie2
4. display
5. enchant
6. etr
7. exiftool
8. frozen-bubble
9. github-desktop
10. highlight
11. mediainfo
12. odt2txt
13. open-invaders
14. pdf2txt
15. pingus
16. QMediathekView
17. server
18. simutrans
19. strings
20. supertux2
21. transmission-show

Also I added it to all firefox-based profiles, but it might be better to just add it to Firefox's profile and then let `include firefox-common` do its magic. Thoughts?

Cheers!
Fred

(Also I'm planning on alphabetising the private-etc lines.... but that's another commit for another day!)

### DIFF
--- a/etc/display.profile
+++ b/etc/display.profile
@@ -39,5 +39,6 @@ shell none
 
 private-bin display,python*
 private-dev
-# private-etc alternatives - on Debian-based systems display is a symlink in /etc/alternatives
+# On Debian-based systems, display is a symlink in /etc/alternatives
+private-etc alternatives
 private-tmp
--- a/etc/display.profile
+++ b/etc/display.profile
@@ -39,5 +39,5 @@ shell none
 
 private-bin display,python*
 private-dev
-# private-etc none - on Debian-based systems display is a symlink in /etc/alternatives
+# private-etc alternatives - on Debian-based systems display is a symlink in /etc/alternatives
 private-tmp
--- a/etc/firefox.profile
+++ b/etc/firefox.profile
@@ -17,7 +17,7 @@ whitelist ${HOME}/.mozilla
 # firefox requires a shell to launch on Arch.
 #private-bin firefox,which,sh,dbus-launch,dbus-send,env,bash
 # private-etc must first be enabled in firefox-common.profile
-#private-etc firefox
+#private-etc alternatives,firefox
 
 # Redirect
 include firefox-common.profile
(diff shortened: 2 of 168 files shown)


etc/display.profile
```diff
@@ -39,5 +39,5 @@ shell none
 
 private-bin display,python*
 private-dev
-# private-etc none - on Debian-based systems display is a symlink in /etc/alternatives
+# private-etc alternatives - on Debian-based systems display is a symlink in /etc/alternatives
```
chiraag-nataraj: Now we should (possibly) be able to uncomment this?

etc/firefox.profile
```diff
@@ -17,7 +17,7 @@ whitelist ${HOME}/.mozilla
 # firefox requires a shell to launch on Arch.
 #private-bin firefox,which,sh,dbus-launch,dbus-send,env,bash
 # private-etc must first be enabled in firefox-common.profile
-#private-etc firefox
+#private-etc alternatives,firefox
```
chiraag-nataraj: Shouldn't need this, since we're putting it in firefox-common.

etc/ping.profile
```diff
@@ -41,7 +41,7 @@ private
 #private-bin has mammoth problems with execvp: "No such file or directory"
 private-dev
 # /etc/hosts is required in private-etc; however, just adding it to the list doesn't solve the problem!
-#private-etc resolv.conf,hosts,ca-certificates,ssl,pki,crypto-policies
+#private-etc alternatives,resolv.conf,hosts,ca-certificates,ssl,pki,crypto-policies
```
chiraag-nataraj: Probably not needed in this profile. `ping` doesn't really depend on external utilities :)

Fred-Barclay:
Thanks mates. I'll leave this up for a bit more and then merge if no other changes are needed.

I went ahead and removed `alternatives` from profiles that include firefox-common since they'll inherit from there.

Issue #2411, opened 2019-02-17
How to allow kate to edit files in .config ?


Utini2000:
Hello everyone,

I can't figure out how to allow "kate" to edit files in /home/user/.config/ ?
I already allowed the folder itself but every file I open will be run as an empty "new file" in kate although it's an existing file with content.

Thanks in advance!

rusty-snake:
You can run kate without firejail for something like that: `/bin/kate` odr /usr/bin/kate` depending on your distro.

or by `noblacklist`ing everything in .config: `firejail --noblacklist="${HOME}/.config/*" kate`

Utini2000:
Already did noblacklist everything but it still doesnt work :S

rusty-snake:
@Utini2000  With or without the `${HOME}`.

Utini2000:
In my kate.local 

noblacklist ${HOME}/.config/*

When I open some in .config it is just an empty file and kate says "New file".

rusty-snake:
Maybe .config/firejail is special.

Utini2000:
Lol that is weird,

The folder for firejail,mpv,vlc in /.config/... wont work. 
cantana for example does work.

Vincent43:
`.config/firejail ` will be always blacklisted in firejail. For others you have to disable `include disable-common.inc`, `include disable-passwdmgr.inc` and `include disable-programs.inc` in `kate.profile`.

chiraag-nataraj:
@Utini2000 So I've actually run into this issue because I sandbox `emacs` (my editor of choice). My personal favorite solution has been to whitelist a specific directory (in my case, `~/.config/emacs/emacs_tmp`) and write a wrapper script which hardlinks the file into that directory, opens `emacs` for editing the file, and deletes the hardlink when I close the file. It's definitely a kludgy solution, but has the advantage that I can, for example, even edit firejail profiles in `emacs` using this method _even if_ `emacs` is jailed.

Utini2000:
Hmm but what is the logic behind disabling the whole .config folder for kate but not for other editors e.g. nano?

rusty-snake:
@Utini2000 look [here](https://github.com/netblue30/firejail/tree/master/etc/nano.profile).

or with words: nano doesn't have a firejail profile.

Utini2000:
Link doesnt work.

But if nano doesn't have a profile, why does kate? :D

rusty-snake:
[...]

PR #2422, opened 2019-02-19
Create nano.profile

rusty-snake:
See: #2411

Notes: I think it is still to strict.

### DIFF
--- a/etc/nano.profile
+++ b/etc/nano.profile
@@ -39,6 +39,7 @@ tracelog
 private-bin nano,rnano
 private-cache
 private-dev
+# Comment the next line if you want to edit files in /etc directly
 private-etc alternatives,nanorc
 
 memory-deny-write-execute
--- a/etc/nano.profile
+++ b/etc/nano.profile
@@ -1,0 +1,47 @@
+# Firejail profile for nano
+# Description: nano  is a easy text editor for the terminal
+# This file is overwritten after every install/update
+# Persistent local customizations
+include nano.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/nano
+noblacklist ${HOME}/.nanorc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+apparmor
+caps.drop all
+machine-id
+net none
+no3d
+nodbus
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+shell none
+tracelog
+
+disable-mnt
+private-bin nano
+private-cache
+private-dev
+private-etc nanorc
+private-tmp
+
+memory-deny-write-execute
+noexec ${HOME}
+noexec /tmp
--- a/etc/nano.profile
+++ b/etc/nano.profile
@@ -14,10 +14,10 @@ include disable-devel.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
-include disable-xdg.inc
 
 apparmor
 caps.drop all
+ipc-namespace
 machine-id
 net none
 no3d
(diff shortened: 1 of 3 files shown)


etc/nano.profile
```diff
@@ -0,0 +1,47 @@
...
+include disable-common.inc
+include disable-devel.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+apparmor
+caps.drop all
+machine-id
```
rusty-snake: Schould we add `ipc-namespace`?

etc/nano.profile
```diff
@@ -0,0 +1,47 @@
...
+seccomp
+shell none
+tracelog
+
+disable-mnt
+private-bin nano
+private-cache
+private-dev
+private-etc nanorc
+private-tmp
```
reinerh: by default I can't edit files in /tmp or /etc?

rusty-snake: right
>  I think it is still to strict.

etc/nano.profile
```diff
@@ -0,0 +1,47 @@
...
+
+noblacklist ${HOME}/.config/nano
+noblacklist ${HOME}/.nanorc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
```
rusty-snake: Will be removed because of DOCUMENTS, MUSIC/VIDOES (playlists)

etc/nano.profile
```diff
@@ -0,0 +1,47 @@
...
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+shell none
+tracelog
+
+disable-mnt
```
rusty-snake: Will be removed.

glitsj16:
Please add support for restricted nano, a.k.a. `rnano` in private-bin. Users might have that set as EDITOR/SUDO_EDITOR.

glitsj16:
Another request if you don't mind. To bring this in line with recent changes in all other profiles, would you mind adding `alternatives` to `private-etc` please?

etc/nano.profile
SkewedZeppelin: won't hurt, cli only

rusty-snake:
@glitsj16, @Fred-Barclay at the moment I'm tend to remove `private-etc`.

rusty-snake:
From my side it can be merged.

glitsj16:
@rusty-snake There's a double include for globals.local in rnano.profile (due to redirection to nano.profile, which already does that). You can just comment it cfr. https://github.com/netblue30/firejail/issues/2006.

rusty-snake:
@glitsj16 Done.

glitsj16:
@rusty-snake Ah, you commented include rnano.local instead of `include globals.local`. It is the latter that causes the double include, as redirecting to nano.profile in the last line already does that. Also, have you decided to keep the profile from editing files under /etc on purpose or is this restriction something you plan to drop? Personally I tend to do quick edits in /etc quite regularly. Which is just me though, I can always override `private-etc` in a local file. Just asking :smile:.

rusty-snake:
@glitsj16 Arrg, I have edit the wrong line. I wanted to comment globals, 😬 😡.

`private-etc`:
  * My idee was to not put it to firefcg.config like vim or emacs => By default there is no symlink.
  * You need root privilegs to edit in /etc.
    1. You are root or you use sudo => You have a Problem.
    2. You use sudoedit (or sudo -e) => than sudo creates a copy in /tmp
  * sudoedit is the better/more secure way

**==> I will leave `private-etc` and add a comment.**

Utini2000:
Although this profile exists, it will now be linked on Arch when running sudo firecfg?

rusty-snake:
If you think you found a bug, open an issue for it.

Also

https://github.com/netblue30/firejail/blob/e8b693c814a4e157bcae6f5c60d1624d185d8837/src/man/firecfg.1.in#L142-L162

Utini2000:
> If you think you found a bug, open an issue for it.

Yep sorry - I think this is not a bug with this profile but generally with my setup. 
I have several application that are not run in firejail although a profile exists. 
I will keep this problem in https://github.com/netblue30/firejail/discussions/6438

Thanks!

PR #2424, opened 2019-02-20
Add gconf-editor profile

glitsj16:


SkewedZeppelin:
doesn't dconf-editor do everything that gconf-editor does?

glitsj16:
@SkewedZeppelin Not quite. GConf is the older settings database format, dconf the newer. See [here](https://wiki.archlinux.org/index.php/GNOME_package_guidelines) for more info. Some apps still use the former, that's why I added a seperate profile for it.

PR #2428, opened 2019-02-20
Add assogiate profile

glitsj16:


SkewedZeppelin:
why does it need ${PICTURES}?

glitsj16:
@SkewedZeppelin I added in ${PICTURES} because assogiate leaves it up to the user to change/define a custom image for each known and newly created mimetype. Because the profile includes `disable-xdg.inc` I whitelisted the usual suspect where a user would store custom imagery. We could leave it out, but in my opinion that would make the firejail profile less user-friendly. Am I tackling this incorrectly here?

PR #2430, opened 2019-02-20
Update .travis.yml

zero77:
This PR builds a deb package using Travis CI, which should then automatically be uploaded to git after.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,19 @@ dist: trusty
 sudo: true
 
 script:
-  - sudo apt-get -y install expect csh xzdec lintian fakeroot
+  - sudo apt-get -y install expect csh xzdec lintian fakeroot checkinstall
+  - ./configure && make && sudo checkinstall --install=no -y --pkgversion=0 --pkgname=firejail-amd64.deb
   - ( ./configure --prefix=/usr && make && sudo make install && make test-travis )
   - ( sudo make install-strip DESTDIR=$(readlink -f appdir) )
-#  # If successful, build release tarball
-#  - ( cd appdir/ ; tar cfvj ../firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2 . )
-#  - curl --upload-file ./firejail-*.tar.bz2 https://transfer.sh/firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2
-#  - # Could use https://github.com/probonopd/uploadtool to upload to GitHub Releases instead
+  - #If successful, build release tarball
+  - #( cd appdir/ ; tar cfvj ../firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2 . )
+  - #curl --upload-file ./firejail-*.tar.bz2 https://transfer.sh/firejail-build$TRAVIS_BUILD_NUMBER.tar.bz2
+  - # Could use https://github.com/probonopd/uploadtool to upload to GitHub Releases instead
+  - # ls -lh out/* Assuming you have some files in out/ that you would like to upload
+  - wget -c https://github.com/probonopd/uploadtool/raw/master/upload.sh
+  - bash upload.sh *.deb
+
+branches:
+  except:
+    - # Do not build tags that we create when we upload to GitHub Releases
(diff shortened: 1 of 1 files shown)


reinerh:
Thanks for your suggestion!
Personally I think releasing a .deb after every commit is not a good idea.

Users might think this is something considered "release-worthy" and therefore stable and install it regularly.
And if they overwrite their distribution-supplied packages with the automatically built ones, that will also lead to many problems, as the packaging is quite different, and files will conflict with e.g. the firejail-profiles package (and in addition to that, the automatic .deb has no proper cleanup / transferral of conffiles).

And what I also find troubling is to wget a "random" shell script that will have the responsibility of making those releases. It's taken directly from a master branch, so it could change every time and in the worst case release malicious code to our users.

I'm not sure if there is a demand for regularly updated .deb snapshots.
I'm curious what others are thinking.

zero77:
> Users might think this is something considered "release-worthy" and therefore stable and install it regularly.

If it helps i can label the deb as unstable or development or similar.
I can also, look into adding a description that lets users know this is for testing purposes only.
Also, that it is really just a quicker way of building from source code and thay should use it at there own risk.




> And if they overwrite their distribution-supplied packages with the automatically built ones, that will also lead to many problems, as the packaging is quite different, and files will conflict with e.g. the firejail-profiles package (and in addition to that, the automatic .deb has no proper cleanup / transferral of conffiles).

This deb has been built using the Compile and install method from the README.md,  is there a better method to use one that will cover this.


> And what I also find troubling is to wget a "random" shell script that will have the responsibility of making those releases. It's taken directly from a master branch, so it could change every time and in the worst case release malicious code to our users.


I used that script because it is located on github and in the past i have seen it being widely used by other repositories who seem to trust it and the developer who created it.
But, that said it may be better to use a copy of it stored on this repository which will help with both stability and security.

> I'm not sure if there is a demand for regularly updated .deb snapshots.

The whole reason i created this, was as an alternative option to the Compile and install method, one that's quick, easy and arguably more stable.
Because, it's built on ubuntu 14.04 lts and tested on ubuntu 18.04.

Also, i picked a deb because, once merged it can make some things easier but at the same time not create any extra work for devs.

Fred-Barclay:
Thanks @zero77! I'm with @reinerh on this one though; I think it may introduce more issues. 

I was wondering, if this were merged, would it upload releases directly here to GitHub? If so, @netblue30 has mentioned in the past that this wasn't a good option for firejail: https://github.com/netblue30/firejail/commit/9e2319c136030192bf4aa9a738518ab2dc9a8367#r23741582

Cheers!
Fred

zero77:
@Fred-Barclay 
I appreciate your concerns about users updating to early and running into problems, yes this definitely should be prevented.

But, what's the difference between users installing from the archive (zip, tar.gz ) vs installing from a deb when a new release is out shourley, it's the same code but just different ways to install it.

Would it not be better to provide both a deb and the existing archives but, add a warning message.

> Could you elaborate on this? Our Travis CI runs Ubuntu 14.04 if I'm not mistaken, and I'm not sure where the 18.04 test you mention comes into play. 😉

I should have explained, I am running 18.04 and its working well for me after testing.

netblue30:
There is another problem. People are monitoring using atom feeds the release page here on github for the real releases.

@probonopd used to have some code in this file to push a binary package to www.transfer.sh. The website went down for awhile, but I see it is up again. Maybe we can bring the code back and push instead of the generic binary package a debian package, or maybe both of them. It was removed in this commit: https://github.com/netblue30/firejail/commit/35b7851dec8771fbb80112078d48e822a539ab3c#diff-354f30a63fb0907d4ad57269548329e3

zero77:
@netblue30 
I am not familiar with transfer.sh what does it do.
Also, is it worth me opening another PR that just creates the deb without uploading it so, transfer.sh can be implemented.

netblue30:
Sorry for the late response.

> is it worth me opening another PR that just creates the deb without uploading 

Yes.

rusty-snake:
Any progress here?

netblue30:
Let's close it down for now.

PR #2431, opened 2019-02-20
fix start-tor-browser.desktop.profile

smitsohu:
Updates paths to the pattern that is used nowadays. I am not sure however if the now deleted pattern was used at some point in the past, so there could be still reason to keep it in.

`mkdir` directives were taken out because basically there is no browser without these folders (also to avoid cluttering the home directory).

The profile should include now all available localizations.


smitsohu:
Changing the approach and only adding new restrictions on top of what we have now. No heavy tor user here, please give me a hint if I'm doing nonsense.

SkewedZeppelin:
Its a mess, but looks fine

smitsohu:
> Its a mess

This!

No idea how simple or difficult that would be to implement, but globbing for profile names would help a lot.

smitsohu:
> Its a mess

It is! Maybe we can leave it open for some time and see if someone comes with a better idea.

netblue30:
OK, let's give it one more week and we check it in. After that I'll do an rc1 release to get some more people testing it.

netblue30:
all set!

PR #2445, opened 2019-02-21
Add 'video' to profile options

iron-bun:
The novideo can be set globally and video permissions can be whitelisted on a per-application basis


Vincent43:
The usual way to handle this situation is to use `ignore novideo`. AFAIK we don't have a precedent for adding a positive option.

iron-bun:
Ignore command does the trick. I should have RTFM'd a little harder I guess. Thanks.

PR #2465, opened 2019-02-24
Harden inkscape.profile

glitsj16:


etc/inkscape.profile
```diff
@@ -41,5 +43,6 @@ shell none
 private-dev
 private-tmp
 
+memory-deny-write-execute
```
SkewedZeppelin: do extensions continue to work?

glitsj16: I must admit I only tested a few extensions. And after some more testing (at least on Arch) some need python. To be safe I'll remove memory-deny-write-execute and allow python.

glitsj16: Should be fixed in https://github.com/netblue30/firejail/pull/2473.

PR #2469, opened 2019-02-25
Harden eog.profile

glitsj16:


etc/eog.profile
```diff
@@ -17,11 +17,11 @@ include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 
-include whitelist-var-common.inc
```
SkewedZeppelin: removed?

glitsj16: Same comment as for gpicview. Both applications seem to run just fine without including whitelist-var-common.inc, on Arch and Ubuntu. But it wouldn't be much trouble to put it back in if so desired.

PR #2470, opened 2019-02-25
Harden gpicview.profile

glitsj16:


etc/gpicview.profile
```diff
@@ -14,9 +14,10 @@ include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 
-include whitelist-var-common.inc
```
SkewedZeppelin: removed?

glitsj16: Yes I tested quite extensively without including whitelist-var-common.inc on both Arch and Ubuntu. Gpicview works just fine without it (as does eog). Do you want me to put it back in?

SkewedZeppelin: yes please
there are lots of files under /var that programs shouldn't have access to

glitsj16: Fixed in https://github.com/netblue30/firejail/pull/2476.

PR #2475, opened 2019-02-25
Archivers

glitsj16:
Hi all. For several months I have been testing archivers succesfully with hardened and streamlined profiles.  I noticed there were some that used `nogroups` while others didn't. All of the tested archivers currently have `ignore noroot`, yet I can't see any reason for it (which of course doesn't mean there isn't one).

Additional hardening options I've enabled in all of the tested profiles:

`apparmor`
`ipc-namespace`
`machine-id`
`net none`
`nodbus`
`nogroups`
`noroot`
`protocol unix`

`private-cache`

`memory-deny-write-execute`
`noexec ${HOME}`
`noexec /tmp`

During the test period I have not experienced any errors or other weirdness. Again, that doesn't mean there aren't any problems, I just didn't see any on my test platforms (Arch Linux/Ubuntu LTS). Which makes me pose the question if people (especially running Fedora) would be interested to test these hardened profiles and provide feed-back here.

If this looks futile, not worth the trouble or otherwise superfluous, please feel free to point that out. No harm is done, this is a draft PR made for that purpose.

Looking forward to your input.

### DIFF
--- a/etc/7z.profile
+++ b/etc/7z.profile
@@ -9,18 +9,26 @@ include 7z.local
 
 blacklist /tmp/.X11-unix
 
-ignore noroot
+apparmor
+ipc-namespace
+machine-id
 net none
 no3d
 nodbus
 nodvd
+nogroups
 nosound
 notv
 nou2f
 novideo
+protocol unix
 shell none
 tracelog
 
 private-dev
+
+memory-deny-write-execute
+noexec ${HOME}
+noexec /tmp
 
 include default.profile
--- a/etc/7z.profile
+++ b/etc/7z.profile
@@ -25,6 +25,7 @@ protocol unix
 shell none
 tracelog
 
+private-cache
 private-dev
 
 memory-deny-write-execute
--- a/etc/atool.profile
+++ b/etc/atool.profile
@@ -22,10 +22,13 @@ include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 
+apparmor
 caps.drop all
-netfilter
+ipc-namespace
+machine-id
 net none
 no3d
+nodbus
 nodvd
 nogroups
 nonewprivs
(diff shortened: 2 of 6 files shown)


SkewedZeppelin:
Relevant: https://github.com/netblue30/firejail/pull/1565#issuecomment-330707453

caps.drop and noroot break root usage

there was this https://github.com/SkewedZeppelin/firejail/commit/e0d74178e2421c07aab7b0c7e41110b5098f9753 but it is very old

glitsj16:
@SkewedZeppelin Thanks for those references. I like the idea of a generic-compression.profile. What happened to it? Let me provide some context as to why I believe the current state of archiver profiles is rather a mess. Like you said, `caps.drop` + `noroot` break root usage. **Does that mean all current archiver profiles are broken in that sense?** Because none that include default.profile have `ignore caps.drop`, only `ignore noroot` (7z, gzip, tar, unrar, unzip and xzdec). From the ones that don't include default.profile, atool has both `caps.drop all` and `noroot`; bsdtar and cpio both have `caps.drop all`.

On another note (something I haven't accounted for in the current draft), is it intentional that archivers aren't supposed to work under /etc by including private-etc (like atool, bsdtar and tar do)?

Vincent43:
> On another note (something I haven't accounted for in the current draft), is it intentional that archivers aren't supposed to work under /etc by including private-etc (like atool, bsdtar and tar do)?

Did you saw any archives there?

glitsj16:
@Vincent43 Nope, but I regularly tar up the whole /etc tree for backup purposes. I assume I'm not the only one interested in doing that (see e.g. documentation for [Arch Linux](https://wiki.archlinux.org/index.php/Full_system_backup_with_tar) and [Ubuntu](https://help.ubuntu.com/community/BackupYourSystem/TAR)). When firejailing tar this produces a backup of around 2KB, without it the file is around 7MB. Just something I was bitten by when I first started firejailing archivers. Easily overridden in a tar.local file I know. Personally I think archivers shouldn't be crippled in such a way. At least not without properly informing the user. Just an opinion, nothing more, nothing less.

Vincent43:
System backup by definition needs full access which is against what firejail is used for. I would recommend doing system backups outside firejail, i.e. by specyfing full path like `/usr/bin/tar`.

glitsj16:
Closing this draft PR. Doesn't seem to make much sense after all.

PR #2483, opened 2019-02-28
Add new profile for redshift

glitsj16:


SkewedZeppelin:
Does this really work?
I tried adding a profile for Redshift long ago and it would repeatedly cause X to crash or hang at random intervals (a few seconds to a few hours).

glitsj16:
@SkewedZeppelin For me this works just fine. Tested on both ArchLinux and Ubuntu 16.04 LTS. Do note that this is the CLI redshift, and not redshift-gtk. If it breaks on Fedora (which I haven't tested) we can always drop the profile again. Would you be willing to do another test?

SkewedZeppelin:
@glitsj16 yes, it does seem to work fine no issues

glitsj16:
@SkewedZeppelin Thanks for confirming. I had a few stabs at redshift-gtk too, but that isn't working for me just yet. As the GUI will use the now sandboxed redshift, this is not a big deal. Maybe I'll look into that later.

PR #2484, opened 2019-02-28
Add new profile for netactview

glitsj16:


SkewedZeppelin:
`${HOME}/.netactview` needs to be whitelisted and created

glitsj16:
Done via https://github.com/netblue30/firejail/pull/2510.

PR #2487, opened 2019-03-01
Add new profile for hardinfo

glitsj16:


SkewedZeppelin:
Does it still display accurate information with no3d, notv, nou2f, novideo, private-dev, etc?

glitsj16:
Don't know what I was thinking when I added this, or whether I was thinking at all, period. Your remarks are spot on, and very much appreciated, as always. Should be fixed in https://github.com/netblue30/firejail/pull/2509.

PR #2488, opened 2019-03-01
Add new profile for regextester

glitsj16:


SkewedZeppelin:
you have whitelist enabled, but no files whitelisted

can you add private/read-only ?

glitsj16:
@SkewedZeppelin Should be fixed in https://github.com/netblue30/firejail/pull/2508. Thanks for your suggestions.

PR #2493, opened 2019-03-01
Add new profile for pavucontrol

glitsj16:


etc/pavucontrol.profile
```diff
@@ -0,0 +1,50 @@
...
+include disable-programs.inc
+include disable-xdg.inc
+
+include whitelist-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
```
SkewedZeppelin: remove machine-id

etc/pavucontrol.profile
```diff
@@ -0,0 +1,50 @@
...
+novideo
+protocol unix
+seccomp
+shell none
+
+disable-mnt
+private-bin pavucontrol
+private-cache
+private-dev
+private-etc alternatives,asound.conf,fonts,pulse
```
SkewedZeppelin: add machine-id

etc/pavucontrol.profile
glitsj16: Done via https://github.com/netblue30/firejail/pull/2511/.

etc/pavucontrol.profile
glitsj16: Done via https://github.com/netblue30/firejail/pull/2511/.

PR #2494, opened 2019-03-01
Add new profile for xfce4-mixer

glitsj16:


etc/xfce4-mixer.profile
```diff
@@ -0,0 +1,49 @@
...
+include disable-programs.inc
+include disable-xdg.inc
+
+include whitelist-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
```
SkewedZeppelin: machine-id can cause access issues to pulse

etc/xfce4-mixer.profile
```diff
@@ -0,0 +1,49 @@
...
+novideo
+protocol unix
+seccomp
+shell none
+
+disable-mnt
+private-bin xfce4-mixer,xfconf-query
+private-cache
+private-dev
+private-etc alternatives,asound.conf,fonts,pulse
```
SkewedZeppelin: add machine-id here to allow real one

PR #2498, opened 2019-03-02
Add new profile for geekbench

glitsj16:


etc/geekbench.profile
```diff
@@ -0,0 +1,50 @@
...
+# Persistent global definitions
+include globals.local
+
+include disable-common.inc
+include disable-devel.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
```
SkewedZeppelin: include whitelist-var-common.inc

etc/geekbench.profile
```diff
@@ -0,0 +1,50 @@
...
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+tracelog
+
+disable-mnt
+private-bin bash,geekbenc*,sh
```
SkewedZeppelin: will it work with `private`?

etc/geekbench.profile
```diff
@@ -0,0 +1,50 @@
...
+private-cache
+private-dev
+private-etc alternatives,groups,passwd,lsb-release
+private-lib libstdc++.so.*
+private-tmp
```
SkewedZeppelin: add `private-opt none`

PR #2501, opened 2019-03-02
Add new profile for sysprof

glitsj16:


SkewedZeppelin:
Won't the process namespace break attaching to existing debuggable programs?

glitsj16:
@SkewedZeppelin For me at least ipc-namespace doesn't seem to cause any breakage.

PR #2504, opened 2019-03-02
mdwx: block memfd_create

topimiettinen:
Some profiles may need adjusting if app uses memfd_create(2) and
memory-deny-write-execute was enabled.


RELNOTES
```diff
@@ -8,6 +8,7 @@ firejail (0.9.58,2) baseline; urgency=low
   * new profiles: klavaro, vscodium
   * browser profiles fixes
   * various other bugfixes
+  * memory-deny-write-execute now also blocks memfd_create
```
SkewedZeppelin: this needs to be moved into the 0.9.59 changelog above

topimiettinen: Thanks, fixed!

topimiettinen:
Updated and also added a test.

SkewedZeppelin:
did you mean to commit `test/filters/memwrexe` ?

topimiettinen:
It could make some sense for 32 bit versions if it would be difficult to compile them. I'd remove the binaries.

Overall, it would be nice to make the whole test setup more generic (for example `chown netblue:netblue` fails on my system), so it could be run as part of normal build process.

topimiettinen:
Applied as  59e3061, let's see if anything breaks.

PR #2510, opened 2019-03-04
Fix netactview

glitsj16:


SkewedZeppelin:
This commit just broke netactview by removing noblacklist and duplicated the mkfile/whitelist which was added in https://github.com/netblue30/firejail/commit/4d25fe5b7d804cf0ac72d96764451ccbc6b9739b#diff-0396d2c052c187b3dbf0d15d66da7569

PR #2512, opened 2019-03-04
Fix machine-id for xfce4-mixerXfce4 mixer

glitsj16:


SkewedZeppelin:
already fixed in 4d25fe5b7d804cf0ac72d96764451ccbc6b9739b

PR #2529, opened 2019-03-06
Update ffmpeg.profile

rusty-snake:
* harden ffmpeg
 * ffmpeg can be used for network-streams
 * alphabetical order (`noroot`, `nonewprivs`)

### DIFF
--- a/etc/ffmpeg.profile
+++ b/etc/ffmpeg.profile
@@ -7,36 +7,44 @@ include ffmpeg.local
 # Persistent global definitions
 include globals.local
 
+noblacklist ${MUSIC}
+noblacklist ${VIDEOS}
+
 include disable-common.inc
 include disable-devel.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
+include disable-xdg.inc
 
 include whitelist-var-common.inc
 
 apparmor
 caps.drop all
+ipc-namespace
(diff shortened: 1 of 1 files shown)


rusty-snake:
@glitsj16 I'm not quite sure, but it looks like it's break it. (I have only Intel -- no AMD or nvidia)

etc/ffmpeg.profile
```diff
@@ -7,36 +7,44 @@ include ffmpeg.local
...
-nonewprivs
-noroot
-# protocol none - needs to be implemented!
+protocol inet,inet6
 seccomp
 # seccomp.keep futex,write,read,munmap,fstat,mprotect,mmap,open,close,stat,lseek,brk,rt_sigaction,rt_sigprocmask,ioctl,access,select,madvise,getpid,clone,execve,fcntl,getdents,readlink,getrlimit,getrusage,statfs,getpriority,setpriority,arch_prctl,sched_getaffinity,set_tid_address,set_robust_list,getrandom
 shell none
 tracelog
 
+disable-mnt
```
SkewedZeppelin: remove this, people might store files on other drivers

glitsj16:
@rusty-snake Thanks. Let's be conservative and keep `no3d`. You might add a comment, something like '# no3d might break HW accelerated de/encoding - comment when appropriate'.

PR #2532, opened 2019-03-06
Add ffmpeg redirect profiles

glitsj16:


rusty-snake:
@glitsj16 Why not adding to firecfg.config?

glitsj16:
@rusty-snake Good question. To be honest I do not know the exact `guidelines` for adding to firecfg, if there are any. I also have zero experience with it, never used it (which isn't an excuse for the lack of knowledge about it, granted). Until now my working-assumption has always been to add to firecfg when an application comes with a .desktop file (or several). But that doesn't seem to make much sense looking at the firecfg list. I'll make note of this and will add the ffmpeg redirect profiles shortly. Thanks for making me re-think my firecfg workflow :smile: .

PR #2534, opened 2019-03-07
Harden gnome-clocks.profile

rusty-snake:


rusty-snake:
@glitsj16 interesting, which OS do you use? I have no issues with `nodbus` and `memory-deny-write-execute`.

glitsj16:
@rusty-snake I use Arch Linux and occasionally Ubuntu LTS systems.
Arch info:
pacman -Q firejail gnome-clocks
firejail-git 0.9.59.r4921.0e523b2f-1
gnome-clocks 3.30.1-1

There isn't a `nodbus` entree in the current gnome-clocks.profile, nor in your PR. I suggested adding the comment because (at least on Arch) when enabled gnome-clocks throws the below several times:
...
(gnome-clocks:16): dconf-WARNING **: 15:20:34.519: failed to commit changes to dconf: Unknown or unsupported transport “DBUS_SESSION_BUS_ADDRESS=unix” for address “DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1001/bus”
...

Can't get to a Ubuntu system at the moment but I'm quite convinced that will behave the same, dconf communication being implemented via D-Bus. Adding a comment for nodbus is quite standard in the profiles, which is why I suggested adding one, now you're looking at the gnome-clocks profile. Nothing crucial though, just a suggestion to keep profiles layout/syntax in check. I might be a tad neurotic here, wouldn't be the first/last time :smile: .

As for the `memory-deny-write-execute` remark we might see different behavior if you happen to run a release version of firejail. Recent [mdwx changes](https://github.com/netblue30/firejail/commit/59e30614ad1cd7a8d6f3c685472fada37d1ed2d7) could be in play in this context. But alas I confirm memory-deny-write-execute cripples gnome-clocks alarms on Arch, as much as I love to see the feature enabled. Hope this clarifies my earlier comment.

rusty-snake:
@glitsj16 ok that make sense.

rusty-snake:
Now I also have problems with `nodbus`, I could swear I didn't have it before. `ipc-namespace` cause some little issues.

PR #2535, opened 2019-03-07
Create nomacs.profile

rusty-snake:
Add a profile for nomacs. Requestet in https://github.com/netblue30/firejail/issues/1139#issuecomment-382608746 by @rekixex.

### DIFF
--- a/etc/disable-programs.inc
+++ b/etc/disable-programs.inc
@@ -435,6 +435,7 @@ blacklist ${HOME}/.local/share/data/Mendeley Ltd.
 blacklist ${HOME}/.local/share/data/Mumble
 blacklist ${HOME}/.local/share/data/MusE
 blacklist ${HOME}/.local/share/data/MuseScore
+blacklist ${HOME}/.local/share/data/nomacs
 blacklist ${HOME}/.local/share/data/qBittorrent
 blacklist ${HOME}/.local/share/dino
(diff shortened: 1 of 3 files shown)


glitsj16:
- add `blacklist ${HOME}/.config/nomacs` to disable-programs.inc and the counterpart `noblacklist ${HOME}/.config/nomacs` in nomacs.profile.
- comment nodbus, as it breaks creating a network proxy
Error creating proxy: Unknown or unsupported transport “DBUS_SESSION_BUS_ADDRESS=unix” for address “DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1001/bus” (g-io-error-quark, 13)
 --> add `# nodbus - Breaks creating proxy`
- app allows user to define 'Open With' applications (which we can't predict) that will break with private-bin nomacs
 --> add `# private-bin nomacs`
- support for users with UID > 1000
nomacs:12): GLib-WARNING **: 00:21:51.827: getpwuid_r(): failed due to unknown user id (1001)
--> add `login.defs` to private-etc

glitsj16:
@rusty-snake Looks good. If you're happy with the profile now we can merge it.

rusty-snake:
@glitsj16, yes, I'm happy with it. Even since yesterday, but the "Ready for Review" button was missing yesterday :(

Issue #2507, opened 2019-03-03
Profiles not in firecfg


SkewedZeppelin:
here for reference purposes, after I updated firecfg in f7925af7026fa90dd285407c0c869bfaeb3984ad

- commons: `electron, clamav, chromium-common, firefox-common`

- compressors: `7z, 7za, 7zr, acat, adiff, als, apack, arepack, aunpack, bsdcat, bsdcpio, bsdtar, bunzip2, cpio, gtar, gunzip, gzip, lbunzip2, lbzcat, lbzip2, lzcat, lzcmp, lzdiff, lzegrep, lzfgrep, lzgrep, lzip, lzless, lzma, lzmadec, lzmainfo, lzmore, p7zip, tar, unlzma, unrar, unxz, unzip, xz, xzcat, xzcmp, xzdec, xzdiff, xzegrep, xzfgrep, xzgrep, xzless, xzmore`

- editors: `emacs, nano, rnano, rview, rvim, vim, vimcat, vimdiff, vimpager, vimtutor`

- needs args: `webui-aria2, tor, spectre-meltdown-checker, makepkg, cower, itch, fetchmail, aosp`

- file managers: `thunar, Thunar, ranger, pcmanfm, nemo, caja`

- known issues: `gnome-ring, git`

- other: `gsettings, dconf`

- potential breakage: `gpg-agent, gpg, gnome-pie, gnome-keyring-3, gnome-keyring`

- special: `Xvfb, default, server`

- terminals: `x-terminal-emulator`

- window managers: `openbox, i3, fluxbox, devilspie, devilspie2, blackbox, awesome`

- profiles that could probably be enabled: `start-tor-browser.desktop, sftp, scp, file, mpd, gajim-history-manager, curl`

- profiles included but disabled that could probably be enabled: `pycharm-community, pycharm-professional, atom, atom-beta, cryptocat, Cryptocat`

Vincent43:
`curl` is used during package updates in Arch so I wold rather avoid it in firecfg.

For things which are already enabled in `firecfg` but IMO shouldn't are: `patch` which is commonly used during building something and `less` which breaks manpages in ubuntu due to conflicts with their AppArmor profile (which is enabled by default there).

SkewedZeppelin:
@Vincent43 updated for curl

I haven't seen any issues with patch (both on its own, and nested under aosp.profile), but less should be safe to remove.

SkewedZeppelin:
[...]

PR #2537, opened 2019-03-07
Add ffmpeg redirect profiles to firecfg

glitsj16:
Making this a draft PR because I'm **not exactly sure about when to add to firecfg/when not**. I think I've asked this before, or it came up in some other context. In any case, I could benefit from a refresher on the topic. Thanks to @rusty-snake for bringing this to my attention in https://github.com/netblue30/firejail/pull/2532. There are several other profiles I added recently that might need to be added to firecfg. I'll await reactions here and will provide the necessary PR's if needed.

### DIFF
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -156,6 +156,9 @@ fbreader
 feedreader
 feh
 ffmpeg
+ffmpegthumbnailer
+ffplay
+ffprobe
 file-roller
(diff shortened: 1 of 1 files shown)


glitsj16:
After some consideration it seems more appropriate to just add these in a proper PR. Main rationale being to ease the ongoing discussion in https://github.com/netblue30/firejail/issues/2507.

PR #2539, opened 2019-03-07
Add code-oss profile

njfox:
Adds a profile alias for `code-oss`, which is the open-source build of Visual Studio Code from the official Arch community repository.


SkewedZeppelin:
Merged, thanks!

PR #2549, opened 2019-03-09
Fix #2548

### LINKED ISSUE #2548: Flameshot raw screenshot issue
Executing `flameshot gui -r` which is supposed to just output back the raw screenshot, adds the command at the start when run in firejail, causing all images to be invalid, e.g. the header of the image file turns into `]0;firejail flameshot gui -r` or `]0;firejail /usr/bin/flameshot gui -r `.

7twin:


glitsj16:
@7twin To avoid any firejail output it's common practice to put that `quiet` option as close to the top of the profile as possible, so it gets executed early. Place it just below the line that reads
`# This file is overwritten after every install/update`
There are plenty of examples, e.g. 7z.profile. When done I'll merge the PR asap. Thanks for bringing this to our attention!

7twin:
@glitsj16 I wasn't sure of the order, thanks for pointing that out!

Issue #1331, opened 2017-06-10
Firefox crash


qazip:
Hello,

Running firefox with firejail (firejail firefox) makes it crash:

```
firejail firefox     
Reading profile /etc/firejail/firefox.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/whitelist-common.inc
Parent pid 10326, child pid 10327
Blacklist violations are logged to syslog
Child process initialized in 78.99 ms
Warning: cannot switch euid to root
Warning: cannot switch egid to root
Warning: cannot switch euid to root
Warning: cannot switch egid to root
Warning: an existing sandbox was detected. /usr/bin/firefox will run without any additional sandboxing features
[Parent 7] ###!!! ABORT: X_ShmPutImage: BadValue (integer parameter out of range for operation); 3 requests ago: file /build/firefox-IKSm1A/firefox-53.0.3+build1/toolkit/xre/nsX11ErrorHandler.cpp, line 147
[Parent 7] ###!!! ABORT: X_ShmPutImage: BadValue (integer parameter out of range for operation); 3 requests ago: file /build/firefox-IKSm1A/firefox-53.0.3+build1/toolkit/xre/nsX11ErrorHandler.cpp, line 147
ExceptionHandler::GenerateDump cloned child 71
ExceptionHandler::SendContinueSignalToChild sent continue signal to child
ExceptionHandler::WaitForContinueSignal waiting for continue signal...

(crashreporter:72): Gdk-WARNING **: crashreporter: Fatal IO error 11 (Resource temporarily unavailable) on X server :0.0.


Parent is shutting down, bye...

```

Firejail version:

`firejail version 0.9.46
`

netblue30:
Are you using Arch Linux?

This is a problem with newer versions of GTK3 library. We have a fix on mainline, or you can still use 0.9.46 and modify some of your profile files in /etc/firejail directory. As root user, open /etc/firejail/firefox.profile file in a text editor and comment out ipc-namespace line (add a #, like this: https://github.com/netblue30/firejail/commit/3d828459a18716b54fd91ddc36d38134e79aa81e).

There are some other programs with the same problem. The fix is the same for all of them: comment out ipc-namespace in the profile file. I'll have a new release out shortly.

qazip:
[...]

Issue #1300, opened 2017-05-22
Gnome-calculator not working in firejail 0.9.46


qazip:
This is the output after trying to run gnome-calculator:

```
gnome-calculator 
Reading profile /etc/firejail/gnome-calculator.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/whitelist-common.inc
Parent pid 9613, child pid 9614
Child process initialized in 46.05 ms

(gnome-calculator:7): Gdk-ERROR **: The program 'gnome-calculator' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadAccess (attempt to access private resource denied)'.
  (Details: serial 178 error_code 10 request_code 130 (MIT-SHM) minor_code 1)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the GDK_SYNCHRONIZE environment
   variable to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)

Parent is shutting down, bye...
```

I'm using Ubuntu 16.04 and I did "sudo firecfg".

Fred-Barclay:
I can't reproduce on my system (Debian Sid with the MATE desktop), either with or without firecfg. 

What happens if you remove the links created by firecfg with `sudo firecfg --clean` and then run `firejail gnome-calculator`? What about if you just run `gnome-calculator`?

qazip:
After removing the links, `gnome-calculator` runs fine, `firejail gnome-calculator` does not:

```
firejail gnome-calculator
Reading profile /etc/firejail/gnome-calculator.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/whitelist-common.inc
Parent pid 12407, child pid 12408
Child process initialized in 71.47 ms

(gnome-calculator:7): Gdk-ERROR **: The program 'gnome-calculator' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadAccess (attempt to access private resource denied)'.
[...]

Issue #1287, opened 2017-05-19
Firefox crashes with ipc-namespace (or something related)


msva:
Unfortunatelly, I've found, that after updating firejail yesterday my firefox started to crash on random events like calling context menu (RMB), moving backwards in history (`Alt`+`⇐`).
A bit of debugging showed that if I comment out `ipc-namespace` crashes disappears.
Also I've noticed, that previously (at the point where all worked) I was using firejail at the state of a1dde3456c4f249202ec9311590526e88a77c42b, but as I see, ipc-namespace was added in the profile in 4f238b75de05d91f200305335da1f019810ac149 (half-a-month before even that), so, I'm bit surprised, how can it brake anything.

Moreover, the commits after a1dde345 looks like unrelated (regarding to commit messages) to the problem.
Maybe, it is some newly introduced feature in the code, and maybe it is somehow related to some updated system libraries.

Anyway, as I said, disabling `ipc-namespace` solves the crashes, and uncommenting it back - also returns crashes.

I'd like to debug it more properly if you'll help me with directions :)

P.S. Firefox's crash looks like that:
```
(firefox:8): Gdk-ERROR **: The program 'firefox' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadValue (integer parameter out of range for operation)'.
  (Details: serial 4854 error_code 2 request_code 130 (unknown) minor_code 3)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the GDK_SYNCHRONIZE environment
   variable to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)

Parent is shutting down, bye...
```

SpotComms:
My apologies, this is the third program that I've added ipc-namespace thats causing issues. Its weird though because I've not actually been able to reproduce any of these issues. @netblue30 can you shed any light as to why ipc-namespace would work for some and not others?

msva:
Moreover, as I said, it worked even on my host, until yesterday's system update (a part of which was firejail update) :)

netblue30:
Fixed - commented out ipc-namespaces.

PR #2552, opened 2019-03-11
ipc-namespace causing problems with file-roller

veloute:


glitsj16:
@veloute Could you provide some more info on what you're seeing when launching file-roller? I'm trying to reproduce, but so far I haven't been able to. So far I've tested functionality with tar.bz2, tar.gz, tar.xz and zip. Works just fine here with the fire-roller.profile. I only see the warnings about dconf and D-Bus...

`
(file-roller:4): dconf-WARNING **: 12:25:26.073: failed to commit changes to dconf: Unknown or unsupported transport “DBUS_SESSION_BUS_ADDRESS=unix” for address “DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1001/bus”`

...which are cosmetic, as the produced archives look as expected. I'm not implying that `ipc-namespace` can't cause issues, just looking for reproducability.

veloute:
with ipc-namespace enabled, file-roller will not launch at all.
the error:
`(file-roller:4): Gdk-ERROR **: 04:37:46.138: The program 'file-roller' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadAccess (attempt to access private resource denied)'.
  (Details: serial 180 error_code 10 request_code 130 (MIT-SHM) minor_code 1)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the GDK_SYNCHRONIZE environment
   variable to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)

`

veloute:
also found ipc-namespace is causing issues with pavucontrol.

this has been an issue before for other profliles:
https://github.com/netblue30/firejail/issues/1300
https://github.com/netblue30/firejail/issues/1287
https://github.com/netblue30/firejail/issues/1331

SkewedZeppelin:
Merged, thanks!

PR #2566, opened 2019-03-12
Harden gucharmap

glitsj16:


etc/gucharmap.profile
```diff
@@ -35,8 +35,12 @@ seccomp
 shell none
 
 disable-mnt
+private-bin gucharmap
 private-cache
 private-dev
+private-etc alternatives,fonts
+private-home .config,.icons,.themes
```
SkewedZeppelin: eh, I don't like this
plus it already has whitelist-common

glitsj16: @SkewedZeppelin I was trying to find a way to have gucharmap use the configured user's theme, something that didn't work for me with the profile before. Got carried away I guess. Will rectify asap. Thanks for pointing this out here. Just to make sure, you're referring to the `private-home` line yes? Or both?

SkewedZeppelin: Just the private-home line
the theme should work fine with just whitelist-common

PR #2573, opened 2019-03-12
viewnior is completely broken with 'hostname viewnior'

veloute:
error message:

> Cannot open display: 
> Run 'viewnior --help' to see a full list of available command line options.
> monitoring pid 10
> 
> Sandbox monitor: waitpid 10 retval 10 status 256
> 

### DIFF
--- a/etc/viewnior.profile
+++ b/etc/viewnior.profile
@@ -20,7 +20,6 @@ include disable-programs.inc
 
 apparmor
 caps.drop all
-hostname viewnior
 ipc-namespace
 net none
 no3d


glitsj16:
@veloute Thanks for reporting this. Just went over my entire dev machine (running Arch Linux) to double-check if I had something exporting the DISPLAY env var in a firejail context (e.g. in globals.local) because (again) I cannot reproduce. Otherwise I hadn't put the hostname option in the profile. Haven't found anything yet though.

Merging because the profile needs to work, obviously :smile:.

veloute:
@glitsj16 so viewnior was working fine for you with the option enabled?
i also just realised that 'ipc-namespace' is also causing problems for me - viewnior opens, but doesn't actually show the image.

with ipc-namespace:
image

without:
image

glitsj16:
@veloute Correct, I made the profile after testing on Arch Linux and Ubuntu LTS, like I always do. It works just fine for me here, with both `hostname` and `ipc-namespace`. I remember https://github.com/netblue30/firejail/pull/2552, you mentioned `ipc-namespace` there too. Which is just fine, this is not a critique. It only made me wonder if something like a local firejail profile, a firejail-related shell script (I do have plenty of those as I'm not using firecfg to rely on stuff in /usr/local/bin) or an overlooked .bashrc setting could explain we seeing different things on the same platform. A rolling release one no less. As I have been contributing several profiles and intend to keep doing so, I just want to avoid breaking stuff for people because my dev box is misconfigured :smile:. Nothing for you to worry about, I appreciate your input very much, as I'm sure others do too.

I'll go ahead and fix ipc-namespace for viewnior in a moment. Thanks again for pointing these things out here.

glitsj16:
@veloute Hopefully https://github.com/netblue30/firejail/pull/2575 fixes your viewnior. Cheers.

smitsohu:
@glitsj16  Is it possible you are testing on Wayland? As far as I understand these ipc-namespace issues are limited to X11.

rusty-snake:
Viewnior < 2.0 (not released yet) use GTK+2 => with Wayland it is running under XWayland.

glitsj16:
@smitsohu Have touched Wayland only briefly. All my testing is done on X11, which is why I can't explain the aforementioned differences. For me options like `hostname` and `ipc-namespace` hardly ever break something. I do have a rather heavily symlinked setup in my ${HOME} dir, but I haven't been able to ascertain yet if that affects things. At least I'm aware of it now and take it into account when adding/editing/proposing profiles. Allthough not to be overlooked, both options are not the most crucial ones IMHO.

Issue #2505, opened 2019-03-02
new *-common includes


smitsohu:
How about two more includes, one collecting various noexec directives in one place, and the other collecting some common whitelisting in /run/user/$UID?

A noexec-common.inc (or whatever the name) could for example look like
```
noexec ${HOME}
noexec ${RUNUSER}
noexec /dev/shm
noexec /tmp
noexec /var
```
(adapted from a suggestion of @Vincent43) and go to all profiles that have an `apparmor` option currently. As this duplicates apparmor restrictions, there is obviously only limited value for people who run firejail with apparmor enabled. But for everyone else it could be an important addition.

A simple whitelist-run-common.inc (or whatever the name)
```
whitelist ${RUNUSER}/bus
whitelist ${RUNUSER}/pulse
```
should be sufficient for most of the programs. Some profiles (those with `writable-run-user` option) will need an extra line `whitelist ${RUNUSER}/gnupg`.

Btw, what usually doesn't work is `whitelist ${RUNUSER}/gvfs`, because of the broken Firejail - FUSE interaction.

Vincent43:
I wonder if we could sneak-in those in  `disable-common.inc` and `whitelist-common.inc`.

smitsohu:
> I wonder if we could sneak-in those in disable-common.inc and whitelist-common.inc.

Good question. `noexec ${HOME}` and `noexec /dev/shm` can be problematic, I was a bit worried about breaking many profiles.

Regarding the whitelisting in /run/user, I currently cannot really overlook the consequences of removing the gvfs mount point. From what I'm reading, Gnome media players or image managers might run into issues.

Vincent43:
Ok. Let's introduce `noexec-common.inc` then and use it to replace `noexec` rules in app profiles gradually.

smitsohu:
Or maybe `disable-exec.inc`? We can call it from anywhere in the profile, so it could go to the disable-* block as well.

Vincent43:
Yes, it may fit better.

Boruch-Baum:
Is it possible to have a sort-of `whitelist` command as an exception to `noexec`? For example, [here](https://github.com/netblue30/firejail/issues/2324#issuecomment-474517220) I have a solution for applying cpulimits to firejails, but it only works in that case because the default profile for firefox has the noexec line commented out; when the noexec line is active, the script in `~/.local/bin` can't be executed, even though it initiated the call to firejail!

[...]

PR #2576, opened 2019-03-12
add disable-exec.inc to all profiles with apparmor

smitsohu:
As discussed in #2385 and #2505. These restrictions already exist in the firejail-default `apparmor` policy, so there should be no regressions (only uncovered bugs).

### DIFF
--- a/etc/electron.profile
+++ b/etc/electron.profile
@@ -7,6 +7,7 @@ include electron.local
 include globals.local
 
 include disable-common.inc
+include disable-exec.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 
--- a/etc/electron.profile
+++ b/etc/electron.profile
@@ -7,7 +7,6 @@ include electron.local
 include globals.local
 
 include disable-common.inc
-include disable-exec.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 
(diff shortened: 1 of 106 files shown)


rusty-snake:
Now we have `noexec ${HOME}` and `noexec /tmp` duplicated in most profiles, IMHO it would make sense to remove `noexec ${HOME}` and `noexec /tmp` from these profiles.

smitsohu:
@rusty-snake Thanks, you're right. I clean it up later today.

smitsohu:
Done :smile:

etc/electron.profile
```diff
@@ -7,6 +7,7 @@ include electron.local
 include globals.local
 
 include disable-common.inc
+include disable-exec.inc
```
Fred-Barclay: This didn't already have noexec though, right?

smitsohu: Right! Is it possible electron.profile is a bit too restrictive anyway?

Vincent43: It could be. I think we should remove `disable-exec.inc` from it and do other changes in separate PR.

smitsohu: I'll remove it in a second commit.

Vincent43:
The next step would be to replace `noexec $HOME` and `noexec /tmp` with `disable-exec.inc ` in rest of profiles.

PR #2577, opened 2019-03-12
Harden meld.profile

rusty-snake:


glitsj16:
@rusty-snake Your additions look good. I'm not entirely convinced we'd ever want to limit a diff'ing tool's access to /etc. So you could just as well drop the commented private-etc line alltogether. Perhaps other people would want to voice an opinion on that before merging, let's wait a few days, ok?

rusty-snake:
Yes, I don't see any issues with this line but but some users want to compare in /etc others may not need it and prefer to restrict access, then such a working private-etc line is also quite helpful.

glitsj16:
@rusty-snake Point taken. Personally I like comments in profiles, learned a lot on the internals of firejail by reading them.

> ... some users want to compare in /etc others may not need it and prefer to restrict access, then such a working private-etc line is also quite helpful

 How about adding a smallish reminder just above the private-etc line to that effect?

rusty-snake:
@glitsj16 sounds good.

>Uncomment the next line if you don't need to compare in /etc.

PR #2581, opened 2019-03-12
exiftool needs access to the /usr/bin/vendor_perl directory in archlinux

veloute:
i couldn't get private-bin to work with directories, if there is a way, that'd be preferred over commenting out private-bin completely.

### DIFF
--- a/etc/exiftool.profile
+++ b/etc/exiftool.profile
@@ -40,6 +40,8 @@ seccomp
 shell none
 tracelog
 
+# To support exiftool in private-bin on Arch Linux (and derivatives), symlink /usr/bin/vendor_perl/exiftool to /usr/bin/exiftool and uncomment the below.
+# Users on non-Arch Linux distributions can safely uncomment the below to enable extra hardening.
 #private-bin exiftool,perl
 private-cache
 private-dev
--- a/etc/exiftool.profile
+++ b/etc/exiftool.profile
@@ -40,7 +40,7 @@ seccomp
 shell none
 tracelog
 
-private-bin exiftool,perl
+#private-bin exiftool,perl
 private-cache
 private-dev
 private-etc alternatives


glitsj16:
@veloute Due to [Arch Linux's perl policy](https://wiki.archlinux.org/index.php/Perl_Policy) firejail in its current form can't use its `private-bin` option on a path like /usr/bin/vendor_perl/exiftool if I at all read the [relevant code](https://github.com/netblue30/firejail/blob/master/src/firejail/fs_bin.c) correctly. The quick and dirty way to support `private-bin exiftool,perl` for Arch Linux is to instruct users to symlink /usr/bin/vendor_perl/exiftool into /usr/local/bin/exiftool. Maybe we could add a comment to such effect, so users are aware of this situation? We'd indeed have to disable private-bin like you suggest in this PR, but at least users are informed a bit more accurately. Thoughts?

glitsj16:
@smitsohu If you have some spare time, would you be willing to voice an opinion on wheter or not we can add private-bin support for /usr/bin/vendor_perl? Or should we live with this and inform users via a comment in the profile? I have been toying around a bit with that path, trying to add it in https://github.com/netblue30/firejail/blob/master/src/firejail/fs_bin.c (lines 30-to-40) without succes. Not a critical issue IMHO, just something that would be nice to have.

smitsohu:
I'm not particularly familiar with private-bin/private-lib to be honest, but directories are reliably excluded currently: 
* https://github.com/netblue30/firejail/blob/79073954ab3a4941b581b7ce7fd6682799a1f554/src/firejail/fs_bin.c#L61

* https://github.com/netblue30/firejail/blob/79073954ab3a4941b581b7ce7fd6682799a1f554/src/firejail/fs_bin.c#L112-L114

That doesn't mean it's not possible, but no idea how much effort that would be. Maybe someone who has written code there can chime in? 

Supporting `/usr/bin/vendor_perl` is potentially generally useful, but for the time being I would just leave it commented and add an explanation in the profile. Other ideas?

veloute:
> for Arch Linux is to instruct users to symlink /usr/bin/vendor_perl/exiftool into /usr/local/bin/exiftool. 
Doing that would interfere with firecfg symlinking to /use/local/bin/, wouldn't it?

glitsj16:
@smitsohu Thanks for your input, much appreciated. My initial idea was to ask the Arch packager of exiftool to add a symlink in a more conventional path :1st_place_medal:. Only then I remembered I'm actually to blame for @veloute's issue: I simply forgot having made such a symlink ages ago and that bit me when adding private-bin to the exiftool profile. Ah well, let's wait for @veloute's edit and proceed.

glitsj16:
@veloute Good point. Advising to put the symlink in /usr/bin as not to interfer with firecfg?

veloute:
so should we keep private-bin commented out or keep it as-is with the commented advisory above it?
if we keep it as-is, we should probably remove it from [/src/firecfg/firecfg.config](https://github.com/netblue30/firejail/blob/master/src/firecfg/firecfg.config) to prevent confusion from users.

glitsj16:
@veloute Keep private-bin commented out, and add something like:

```
# To support exiftool in private-bin on Arch Linux (and derivatives), symlink /usr/bin/vendor_perl/exiftool to /usr/bin/exiftool and uncomment the below.
# Users on non-Arch Linux distributions can safely uncomment the below to enable extra hardening.
#private-bin exiftool,perl
```
or whatever you prefer to that effect. Like that the profile should be firecfg-proof without causing confusion or you needing to touch any other file for this PR. Hope this helps, and apologies for breaking exiftool in the first place...

veloute:
> Hope this helps, and apologies for breaking exiftool in the first place...

lol, no worries

PR #2584, opened 2019-03-13
Harden youtube-dl.profile

rusty-snake:


glitsj16:
@rusty-snake I just noticed you saw my question in https://github.com/netblue30/firejail/commit/1862d24fd9990bfc61f9ae9710f089b3d8038427. Now you're already working on the youtube-dl profile in this PR, if you'd like you can go ahead and integrate the new disable-exec stuff. In the mean time I'm running some tests with the changed you propose here. I happen to have a rather extensive non-default/exotic youtube-dl usage with firejail so give me some time to respond on the results :smile:.

glitsj16:
@rusty-snake Thanks for adding disable-exec. The profile looks really nice now with your extra hardenings. Sadly `memory-deny-write-execute` breaks on Arch Linux for me. So if you would be willing to edit that in this can be merged.

glitsj16:
Great job, thank you! I'll let the travis check do its job and will merge in a minute.

etc/youtube-dl.profile
```diff
@@ -45,8 +52,11 @@ seccomp
 shell none
 tracelog
 
+disable-mnt
```
SkewedZeppelin: media on other drives is common enough use case?
most other media profiles don't have this

rusty-snake: ytdl is only for downloading.

glitsj16: @SkewedZeppelin Nice catch, fixed.

rusty-snake: I still don't see a reason, why ytdl should can access other drivers.

glitsj16: @rusty-snake 
> I still don't see a reason, why ytdl should can access other drivers.

Suppose a user wants to download to a USB-drive mounted at /mnt/my-youtube-usb-stick/foo that will fail with `disable-mnt`. Another reason why @SkewedZeppelin is correct would be that youtube-dl supports `--external-downloader foo`, which also should be able to save straight to a device mounted under /mnt. Or do you have something else in mind?

PR #2587, opened 2019-03-13
Fix incorrect parsing of --keep-var-tmp command

lskrejci:
The command was only recognized if it was passed as the first argument.
Passing it on any other position on the command line caused the following
error:

    Error: invalid --keep-var-tmp command line option

Supplying it as the first argument also resulted in other commands that are
parsed after it to be silently ignored.


smitsohu:
Merged, thanks!

PR #2600, opened 2019-03-15
Seahorse revisited

glitsj16:
Currently the seahorse{-tool} profiles include gpg.profile, which works fine. IMHO this could be improved upon. This draft attempts to do so by decoupling seahorse{-tool} from the gpg profile and by making seahorse profiles fully whitelist profiles instead.

Comments, opinions, etcetera very welcome.

### DIFF
--- a/etc/seahorse.profile
+++ b/etc/seahorse.profile
@@ -4,23 +4,60 @@
 # Persistent local customizations
 include seahorse.local
 # Persistent global definitions
-# added by included profile
-#include globals.local
+include globals.local
+
+blacklist /tmp/.X11-unix
 
 # dconf
 mkdir ${HOME}/.config/dconf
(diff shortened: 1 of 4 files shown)


etc/seahorse.profile
```diff
@@ -4,23 +4,60 @@
 # Persistent local customizations
 include seahorse.local
 # Persistent global definitions
-# added by included profile
-#include globals.local
+include globals.local
+
+blacklist /tmp/.X11-unix
```
rusty-snake: Remov, seahorse is graphical (or am I talking bullshit?)

glitsj16: Thanks guys. I'm prone to keep it in seahorse-daemon, which is CLI.

PR #2602, opened 2019-03-16
mount runtime seccomp files read-only

smitsohu:
In order to reduce exposure to locations that are both writable and executable (only processes with euid of the user):
* group user owned /run/firejail/mnt/seccomp.* files in a directory and
* mount this directory read-only
* remove user owned /run/firejail/mnt/libfiles once it is not needed any more

/run/firejail/mnt/pulse and /run/firejail/mnt/sec.Xauthority are noexec already, and the appimage mounts are read-only, so I think this PR should cover the last remaining paths. To make this possible it was necessary to drop capabilities a little bit later during sandbox setup.


netblue30:
@smithosu - looks fine, bring it in! It has a conflict in src/firejail/sandbox.c. Thanks!

PR #2614, opened 2019-03-22
Add kid3, kid3-cli, kid3-qt

rusty-snake:


glitsj16:
Merged. Thanks for adding these profiles.

PR #2615, opened 2019-03-22
Add VCS support to meld

rusty-snake:


etc/meld.profile
```diff
@@ -10,7 +10,7 @@ noblacklist ${HOME}/.local/share/meld
 noblacklist ${PATH}/python*
 noblacklist /usr/include/python*
 noblacklist /usr/lib/python*
-noblacklist /usr/local/lib/python*
+noblacklist /usr/local/lib/python*,git,svn,hg,bzr,cvs
```
SkewedZeppelin: find all and replace?

rusty-snake: not entered, but maybe a button accidentally pressed (using vim). will be removed.

glitsj16:
Personally I find the meld profile a bit too restrictive for a diff tool. Nothing that can't be handled via a local override, but I find myself diff'ing files blacklisted by `disable-common.inc` and `disable-programs.inc` quite often. Just a thought...

rusty-snake:
I would add it with a little more feedback.

rusty-snake:
I only tested git, because I don't use svn,hg,bzr,cvs / have no experience.

rusty-snake:
From my site it can be merged.

glitsj16:
@rusty-snake For consistency's sake, would you mind using the same format as other profiles do that need to allow python? I mean with reagards to commenting and spacing. Also would be nice to alphabetically order private-bin and private-etc. Here's an [example](https://git.io/fjTkt
).

rusty-snake:
@glitsj16 done, thanks for the preliminary work.

glitsj16:
Merged, thanks!

PR #2625, opened 2019-03-29
add gnuchess to play against computer

veloute:
see dependencies on https://www.archlinux.org/packages/extra/x86_64/gnome-chess/

### DIFF
--- a/etc/gnome-chess.profile
+++ b/etc/gnome-chess.profile
@@ -33,7 +33,7 @@ shell none
 tracelog
 
 disable-mnt
-private-bin fairymax,gnome-chess,hoichess
+private-bin fairymax,gnome-chess,hoichess,gnuchess
 private-dev
 private-etc alternatives,fonts,gnome-chess
 private-tmp


glitsj16:
@veloute The Arch repo page for [gnuchess](https://www.archlinux.org/packages/community/x86_64/gnuchess/files/) also mentions executables `gnuchessu` and `gnuchessx`. And then there's [xboard](https://www.archlinux.org/packages/community/x86_64/xboard/) as optional GUI for gnuchess. Do you happen to know if these have/need their own additional config or data files? Seems unlikely they all use ${HOME}/.local/share/gnome-chess, that's why I'm asking. Also, if some have `online` play functionality the profile needs to be changed to account for that too (netfilter, protocol).

It might make more sense to split-out the different executables in private-bin into seperate profiles, each redirecting to gnome-chess.profile. That way all of them can be integrated into [firecfg.config](https://github.com/netblue30/firejail/blob/master/src/firecfg/firecfg.config) as well. Would you like to take care of this?

veloute:
gnome-chess only needs gnuchess included in private-bin as it's used as a chess engine to make it possible to play against a computer - it doesn't need anything else added to be completely functional

gnuchess (and the gui, xboard) could use firejail profiles; will make them soon and make a new pr.

SkewedZeppelin:
Merged, thanks!

PR #2633, opened 2019-03-31
private-bin breaks --join for filezilla

veloute:
perhaps there's another solution, but this is the only fix i could find for fixing --join.

error i get:
```
Switching to pid 13327, the first child process inside the sandbox
Child process initialized in 10.86 ms
execvp: No such file or directory
```

### DIFF
--- a/etc/filezilla.profile
+++ b/etc/filezilla.profile
@@ -36,6 +36,7 @@ protocol unix,inet,inet6
 seccomp
 shell none
 
-private-bin filezilla,uname,sh,bash,python*,lsb_release,fzputtygen,fzsftp
+# private-bin breaks --join
+# private-bin filezilla,uname,sh,bash,python*,lsb_release,fzputtygen,fzsftp
 private-dev
 private-tmp


SkewedZeppelin:
what is your default shell?

veloute:
you're right, my bad; that didn't even cross my mind.
maybe we should add other common shells to the private-bin line?

SkewedZeppelin:
hmm, I think at one point if `sh` was there it would add the current default, but I'm not sure.

veloute:
could something maybe be done with the $SHELL variable?

rusty-snake:
On my System sh is a symlink to bash, but my user default shell is zsh.
This has already caused some issues with FJ.

netblue30:
Here is the deal with private-bin. The ugliest executable you can add on this line is a shell. An attacker after getting control of the program will want to start a shell session on your computer. So, if there is no shell available, bad luck! For various reasons, a small number of programs are using an external shell, usually /bin/sh. Some of them might even go for $SHELL env variable.

In the case of filezilla, we already know it is using /bin/bash, so the damage is already done. I don't see any reason not to add zsh on that line. @veloute if you just add zsh on private-bin line, is it working?

veloute:
yes, joining works after adding zsh to private-bin.

netblue30:
ok, let's go for adding zsh on the private-bin line - I'll check it in in a few minutes

Vincent43:
I guess every other profile with `private-bin sh,bash` should be fixed this way.

netblue30:
Let's wait for people to start complaining about it.

veloute:
> Let's wait for people to start complaining about it.

i think i'm ready to complain about it :P 
i've run into the problem of joining working about 40% of the time; out of the few i randomly just tested, keepassxc, mpv, hexchat and flameshot won't allow me to join (always with "execvp: No such file or directory" as the error) with zsh as the default shell.

PR #2641, opened 2019-04-09
Add cheese.profile

rusty-snake:
Add a profile for cheese.

**This profile is untested**, because of own hardware limitations, if someone could please test this.


glitsj16:
@rusty-snake Same hardware limitations, so can't confirm its working (allthough it looks good on first glance). Just wanted to inform you that `Cheese` should be added to firecfg.config too for the workaround.

netblue30:
Let's try it out.

PR #2649, opened 2019-04-13
Add a conditional to control DRM/noexec exception for browsers

SkewedZeppelin:
Browsers are large attack surface, `noexec ${HOME}` is fairly important.
Ignoring it just to allow proprietary (malware) DRM to run by default is silly.
Thoughts?


etc/firejail.config
```diff
@@ -8,6 +8,9 @@
 # Disable U2F in browsers, default enabled.
 # browser-disable-u2f yes
 
+# Allow DRM execution in browsers, default disabled.
+# browser-allow-drm no
```
SkewedZeppelin: also these two browser-* options are out of order (my fault)
but they are hard to move now because it might break user configs? or can they just be moved?

Vincent43: I don't think moving them break anything. Distro package mangers usually don't overwrite config files automatically and even if they did then changing a comma will have same effect as rewriting whole thing from scratch. They're commented out by default so in case users would miss them while merging configs then they won't override any old settings.

netblue30:
All set, thanks.

cddmp:
I want to leave this here just for the records:
While trying to debug the DRM issue myself when it first occurred under Arch, it turned out that 
`noexec ${HOME}/*`
actually still worked though
`noexec ${HOME} `
didn't.

PR #2651, opened 2019-04-14
Add .pythonrc.py to disable-common.inc

Senemu:
Python 2 executes ``~/.pythonrc.py`` when importing the [``user`` module](https://docs.python.org/2/library/user.html).

Users might also source it from the interactive startup file of their Python interpreter or directly point the [``PYTHONSTARTUP`` environment variable](https://docs.python.org/using/cmdline.html#envvar-PYTHONSTARTUP) to it.

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -227,7 +227,6 @@ read-only ${HOME}/.pgpkey
 read-only ${HOME}/.plan
 read-only ${HOME}/.profile
 read-only ${HOME}/.project
-read-only ${HOME}/.pythonrc.py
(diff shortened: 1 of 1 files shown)


SkewedZeppelin:
This should probably also be `noblacklisted` in geany, gnome-builder, pycharm*, etc.

Senemu:
I accidentally also changed the order of one other line with my last commit (by sorting the entire section).

I have deleted my fork, so I cannot make any further changes.

Fred-Barclay:
I can add to geany etc. Merged -- thanks!

PR #2660, opened 2019-04-22
Gajim: Allow reading of system-wide Flatpak locale

flacks:
Gajim will not start without being able to read this file, even if it
doesn't exist.

### DIFF
--- a/etc/gajim.profile
+++ b/etc/gajim.profile
@@ -32,6 +32,7 @@ whitelist ${HOME}/.cache/gajim
 whitelist ${HOME}/.config/gajim
 whitelist ${HOME}/.local/share/gajim
 whitelist ${DOWNLOADS}
+whitelist /var/lib/flatpak/exports/share/locale
 include whitelist-common.inc
 
 caps.drop all


rusty-snake:
Merged, Thanks.
Just out of interest, which distro?

flacks:
Thank you; Arch Linux

Vincent43:
I think access to this file was blocked by `blacklist /var/lib/flatpak` rule in `disable-common.inc`. Adding `whitelist /var/lib/flatpak/exports/share/locale` on top of above would mean whole `/var` would be empty in sandbox (as the only whitelisted entry was blacklisted before).

The proper and secure way to fix this would be adding:
```
noblacklist /var/lib/flatpak
...
whitelist /var/lib/flatpak/exports/share/locale
```

flacks:
Can you explain why this is more proper/secure? I don't understand why you need to allow more /var directories than Gajim seems to need.

rusty-snake:
@flacks `firejail --ignore=private-bin --profile=gajim.profile ls /var` /var is empty because of the blacklisting rule. Maybe its also works with only the `noblacklist` rule. So you can try it ;).

Vincent43:
> Can you explain why this is more proper/secure? I don't understand why you need to allow more /var directories than Gajim seems to need.

The _proper_ means it matches the description and intent of your changes. Your rule didn't allow to read `/var/lib/flatpak/exports/share/locale`, it made whole `/var` empty which accidentally solved your issue. Combination of:
```
noblacklist /var/lib/flatpak
whitelist /var/lib/flatpak/exports/share/locale
```
will do what you tried to achieve while still maintaining the same security level you accidentally achieved. `include whitelist-var-common.inc` adds few common, harmless dirs (like fontconfig cache) which may be useful for app at some point.

Vincent43:
Actually I think `include whitelist-var-common.inc` may be enough to fix the issue.

Vincent43:
I fixes this in https://github.com/netblue30/firejail/commit/78f1e3a03575ed3ddf55dc7f266274e52f05f979

rusty-snake:
Just for my System:
```
$ firejail --ignore=private-bin --profile=gajim.profile ls /var/lib
dbus
$ firejail --ignore=private-bin --profile=gajim.profile --whitelist=/var/lib/flatpak ls /var/lib
dbus flatpak
$ firejail --ignore=private-bin --profile=gajim.profile --whitelist=/var/lib/flatpak ls /var/lib/flatpak
Permission denied
$ firejail --ignore=private-bin --noblacklist=/var/lib/flatpak --profile=gajim.profile --whitelist=/var/lib/flatpak ls /var/lib/flatpak
app  appstream  exports  oci  repo  runtime
```

Vincent43:
> Gajim will not start without being able to read this file, **even if it
> doesn't exist.**

@rusty-snake `/var/lib/flatpak` shouldn't exist in sandbox after my change thus gajim should work. Do you see something different?

PR #2672, opened 2019-05-01
Support Enpass v6

glitsj16:
Added support for latest upstream version and refactored enpass.profile as a whitelisting profile.

`Note`: when sandboxing enpass, I couldn't get its companion webextension to work in browsers that already run firejailed. Not sure whether or not this was already the case with Enpass versions pre-dating v6. In all honesty I didn't spend all that much time testing different browsers/extensions as I personally moved to Bitwarden (which has a similar if not larger feature-set, is fully open-source and doesn't suffer from this dysfunctionality). Perhaps someone who actively uses Enpass can confirm all this and add a note to the profile accordingly.

### DIFF
--- a/etc/enpass.profile
+++ b/etc/enpass.profile
@@ -12,12 +12,6 @@ noblacklist ${HOME}/.config/Sinew Software Systems
 noblacklist ${HOME}/.local/share/Enpass
 noblacklist ${DOCUMENTS}
 
-whitelist ${HOME}/.cache/Enpass
-whitelist ${HOME}/.config/sinew.in
-whitelist ${HOME}/.config/Sinew Software Systems
-whitelist ${HOME}/.local/share/Enpass
-whitelist ${DOCUMENTS}
-
 include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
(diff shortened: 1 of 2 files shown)


etc/enpass.profile
```diff
@@ -19,10 +28,13 @@ include disable-xdg.inc
...
 include whitelist-var-common.inc
 
+# machine-id and nosound break audio notification functionality
+# comment both if you need that functionality or put 'ignore machine-id'
+# and 'ignore nosound' in your enpass.local
+
 caps.drop all
 machine-id
-net none
-no3d
```
SkewedZeppelin: no3d removed?

etc/enpass.profile
```diff
@@ -6,9 +6,18 @@ include enpass.local
 # Persistent global definitions
 include globals.local
 
+noblacklist ${HOME}/.cache/Enpass
+noblacklist ${HOME}/.config/sinew.in
 noblacklist ${HOME}/.config/Sinew Software Systems
+noblacklist ${HOME}/.local/share/Enpass
 noblacklist ${DOCUMENTS}
 
+whitelist ${HOME}/.cache/Enpass
```
SkewedZeppelin: mover after includes

etc/enpass.profile
glitsj16: @SkewedZeppelin Nice catch! I forgot to re-add it after testing. Without no3d Enpass throws some errors about not finding libGL (which are cosmetic ones). I'll fix it, thanks.

etc/enpass.profile
glitsj16: Done.

PR #2673, opened 2019-05-01
Add Bitwarden profile

glitsj16:


etc/bitwarden.profile
```diff
@@ -0,0 +1,56 @@
...
+# Persistent local customisations
+include bitwarden.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/Bitwarden
+ignore noexec /tmp
+noblacklist ${DOWNLOADS}
+
+whitelist ${HOME}/.config/Bitwarden
```
SkewedZeppelin: move after includes
also include whitelist-common.inc

etc/bitwarden.profile
```diff
@@ -0,0 +1,54 @@
...
+# Description: A secure and free password manager for all of your devices
+# This file is overwritten after every install/update.
+# Persistent local customisations
+include bitwarden.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/Bitwarden
+ignore noexec /tmp
+noblacklist ${DOWNLOADS}
```
rusty-snake: Not needed. [disable-xdg.inc](https://github.com/netblue30/firejail/blob/master/etc/disable-xdg.inc) has `blacklist ${DOWNLOADS}` commented out.

PR #2679, opened 2019-05-05
Add vim syntax and ftdetect files

laomaiweng:
Hi!

Here are vim syntax and ftdetect files to automatically detect and syntax-highlight firejail profiles in vim.
I built those for myself over the last few days and I feel they are functional enough for a wider audience. :-)

Please note these are my first ever written vim syntax/ftdetect files, so there's likely to be room for improvement. However, they seem to be doing alright with all the profiles I could get my hands on.

Cheers and thanks for making the awesome piece of software that firejail is. :)


Fred-Barclay:
Awesome! Merged -- thanks. :smile:

rusty-snake:
@laomaiweng as you might see, these file are now installed with `--enable-contrib`.
I'm not very familiar with vim syntax files that's why I'm asking you how we can adapt the recent dbus changes in the vim syntax.

PR #2680, opened 2019-05-06
Mumble: add new path for client data

Feandil:
Since Mumble 1.3, client data was moved from $HOME/.local/share/data/Mumble to $HOME/.local/share/Mumble.
See https://wiki.mumble.info/wiki/Client_Settings


SkewedZeppelin:
Merged, thanks!

PR #2684, opened 2019-05-08
Whitespace fix

dandelionred:
Just fixing some whitespace found with
```sh
find . -name \*.profile -exec grep -P '\s\s+' {} +
```

### DIFF
--- a/test/environment/rlimit.profile
+++ b/test/environment/rlimit.profile
@@ -1,5 +1,5 @@
-  rlimit-fsize 1024
+rlimit-fsize 1024
 rlimit-nproc 1000
-	rlimit-nofile 	500
-rlimit-sigpending	200
+rlimit-nofile 500
+rlimit-sigpending 200
 rlimit-as 123456789012
--- a/test/profiles/test.profile
+++ b/test/profiles/test.profile
@@ -1,5 +1,5 @@
-   blacklist /sbin/iptables
-blacklist		 /etc/shadow
-	blacklist		/bin/rmdir
+blacklist /sbin/iptables
+blacklist /etc/shadow
+blacklist /bin/rmdir
 blacklist ${PATH}/umount
 blacklist ${PATH}/mount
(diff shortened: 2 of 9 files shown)


test/profiles/test.profile
```diff
@@ -1,5 +1,5 @@
-   blacklist /sbin/iptables
```
dandelionred: Oops, looking into the file history I see it was intended for the sake of the parser testing.

I reverted changes to test/profiles/test.profile and test/profiles/test2.profile.

SkewedZeppelin: these are probably intended for testing

test/environment/rlimit.profile
```diff
@@ -1,5 +1,5 @@
-  rlimit-fsize 1024
```
SkewedZeppelin: this one too

dandelionred: Generic syntax checking for extra whitespaces is done here

https://github.com/netblue30/firejail/blob/f959fbb3a817dbf1452193b44efaebcc45e8cd83/test/profiles/profiles.sh#L18-L22

Hence this should only check for settings set via profile, not extra whitespace parsing

https://github.com/netblue30/firejail/blob/f959fbb3a817dbf1452193b44efaebcc45e8cd83/test/environment/environment.sh#L111-L112

Don't you think so?

rusty-snake:
Any progress here?

dandelionred:
> Any progress here?

I'm pretty sure test/environment/rlimit.profile should not include extra whitespace.

The only places to test for extra whitespace should be firejail/test/profiles/test.profile and firejail/test/profiles/test2.profile. So in case something goes wrong with extra whitespace parsing the point of failure should be syntax tests **only**, not rlimit test.

rusty-snake:
@SkewedZeppelin ?

SkewedZeppelin:
Merged, thanks.

PR #2687, opened 2019-05-10
Update keepassxc.profile

curiosity-seeker:
Adding `keepassxc-proxy` to `private-bin`as without it there is a related error message in Settings->Browser Integration (via the Keepassxc-Browser add-on).


rusty-snake:
Merged, Thanks.

PR #2688, opened 2019-05-10
nodbus enhancements

laomaiweng:
Hi,

This PR adds/changes the following behaviors of `nodbus`:

* blacklist `~/.dbus` by default
* fix the overridden value of `DBUS_SESSION_BUS_ADDRESS`

`~/.dbus` is created by `dbus-launch` and contains an environment file that can be sourced by the shell. It sets the values of `DBUS_SESSION_BUS_ADDRESS` and other DBus-related envvars, so that processes spawned by the shell can communicate with the session bus. Blacklisting it removes one way for firejailed programs to find out about the real address of the session bus (though other ways sadly still remain).

As for the overridden value of `DBUS_SESSION_BUS_ADDRESS`, it contained `DBUS_SESSION_BUS_ADDRESS=` itself at the start, which seemed like a mistake.

Hope this is helpful!


netblue30:
Quite an ugly bug, thanks!

Issue #2696, opened 2019-05-14
O_PATH undeclared (CentOS 6)


chargrnv:
OS: CentOS 6.10
Version: 0.9.56-LTS (and newer)

I believe this would be provided by the kernel, but on CentOS 6 the kernel is too old.

http://man7.org/linux/man-pages/man2/open.2.html
O_PATH (since Linux 2.6.39)

```
[root@90ea573d076b ~]# rpm -qa | grep kernel
kernel-devel-2.6.32-754.12.1.el6.x86_64
kernel-headers-2.6.32-754.12.1.el6.x86_64
```

```
fs.c: In function 'fs_private_cache':
fs.c:851:26: error: 'O_PATH' undeclared (first use in this function)
  int fd = safe_fd(cache, O_PATH|O_DIRECTORY|O_NOFOLLOW|O_CLOEXEC);
                          ^
```

0.9.54 builds OK under centos 6.

Is it possible for configure to check for the feature and have it optionally enabled/disabled?

I thought I had seen a patch posted previously, but I dont find any issues filed against this when searching for O_PATH.

Vincent43:
Centos 6.10 is near end-of-life with only crucial security updates supported. Is there a reason you didn't migrated to 7.x yet and/or you need firejail 0.9.56+ there?

chargrnv:
Moving to CentOS 7, but still have a lot of CentOS 6 and try and keep new versions of tools available.  Thought LTS was "long term support" and was trying to support older OS.

Vincent43:
> Thought LTS was "long term support" and would support older OSs.

Yes, however there are limits how old things it supports. Your kernel was released almost 10 years ago. This a time range of enterprise class support rather than most of community projects. I'm not saying support for it won't be added but it can't be taken for granted.

smitsohu:
Vanilla kernels older than 3.6 are unsupported currently, because of (from `man 2 open`)
>  O_PATH (since Linux 2.6.39)
....
The following operations can be performed on the resulting file descriptor:
close(2); fchdir(2) (since Linux 3.5); fstat(2) (since Linux 3.6).

But the patched CentOS/RedHat kernel seems to be fine, providing O_PATH is all we need to do.

Vincent43:
Fixed by https://github.com/netblue30/firejail/commit/203e34db60e9200a53708c425868a041e32eaf0d

chargrnv:
Thanks!

PR #2701, opened 2019-05-16
glibc missing O_PATH definition on CentOS 6

smitsohu:
There has been decbe96a48fc21c11349867ff2c1ada7c88a0113 and b65dfff0f3b7ff5a16771ac4eca80d0609cbbf44, but CentOS 6 is an actively maintained distribution, unlike Debian 7.

So I'd propose to bring back the fix.

 #2696


netblue30:
all set, thanks

PR #2708, opened 2019-05-18
Refactor eog and eom profiles with common redirect

glitsj16:
These profiles have mostly identical options, so it might be handy for future changes to refactor both with a common redirect file.

### DIFF
--- a/etc/eo-common.profile
+++ b/etc/eo-common.profile
@@ -38,10 +38,6 @@ seccomp
 shell none
 tracelog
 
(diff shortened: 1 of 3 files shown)


SkewedZeppelin:
I'm fine with this.
You could probably keep private-bins in each their own.

relevant: https://gitlab.gnome.org/GNOME/gnome-build-meta/issues/148

glitsj16:
@SkewedZeppelin Thanks for the info! I've returned private-bin items to their main profiles.

PR #2712, opened 2019-05-20
Add private-cwd option to control working directory within jail

apmorton:
Main motivating use case for this patch is within [compiler-explorer](https://github.com/mattgodbolt/compiler-explorer).

We want to use relative paths in profile include statements, which requires a specific working directory when launching firejail.

However, we also need to ensure the application *within* firejail runs in another working directory.

`--private-cwd` causes the jail to change to the users home directory
`--private-cwd=/path` causes the jail to change to the specified directory once inside the filesystem namespace of the jail

### DIFF
--- a/src/firejail/fs_home.c
+++ b/src/firejail/fs_home.c
@@ -370,6 +370,21 @@ void fs_check_private_dir(void) {
 	}
 }
 
+// check new private working directory (--private-cwd= option) - exit if it fails
+void fs_check_private_cwd(void) {
+	EUID_ASSERT();
+	invalid_filename(cfg.cwd, 0); // no globbing
+
+	// Expand the working directory
+	cfg.cwd = expand_macros(cfg.cwd);
+
+	// realpath/is_dir not used because path may not exist outside of jail
+	if (!cfg.cwd) {
+		fprintf(stderr, "Error: invalid private working directory\n");
+		exit(1);
+	}
+}
+
 //***********************************************************************************
 // --private-home
 //***********************************************************************************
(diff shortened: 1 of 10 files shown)


src/firejail/main.c
```diff
@@ -1773,6 +1774,14 @@ int main(int argc, char **argv) {
 			else
 				exit_err_feature("private-cache");
 		}
+		else if (strcmp(argv[i], "--private-cwd") == 0) {
+			cfg.cwd = NULL;
+			arg_private_cwd = 1;
+		}
+		else if (strncmp(argv[i], "--private-cwd=", 14) == 0) {
+			cfg.cwd = argv[i] + 14;
```
smitsohu: Can you check the argument with the `invalid_filename()` function against weird characters?

Though not strictly necessary, I also wonder if the the previous cfg.cwd should be `free`'d and the new cfg.cwd should be allocated. Just for keeping it consistent.

apmorton:
implemented `invalid_filename` checks in a similar fashion to `--private=`

src/firejail/fs_home.c
```diff
@@ -370,6 +370,23 @@ void fs_check_private_dir(void) {
...
+void fs_check_private_cwd(void) {
+	EUID_ASSERT();
+	invalid_filename(cfg.cwd, 0); // no globbing
+
+	// Expand the working directory
+	char *tmp = expand_macros(cfg.cwd);
+	cfg.cwd = realpath(tmp, NULL);
+	free(tmp);
+
+	// is_dir not checked because path may not exist outside of jail
```
smitsohu: This comment seems to indicate you want to support paths that don't exist outside the jail, but afaics it won't work because of the realpath(). Just mentioning it in case you need that functionality.

apmorton: good catch - I did not look close enough at the behavior of realpath

smitsohu:
Merged, thanks!

PR #2715, opened 2019-05-21
Add deterministic-exit-code option

apmorton:
Ensures firejail exits with the first childs exit code regardless of the termination ordering of orphaned children.

Firejail currently exits with the status code of whatever child itself exits last.
This can be non-deterministic if there are any orphaned children that outlive the process firejail launches.

Another patch motivated by [compiler-explorer](https://github.com/mattgodbolt/compiler-explorer).

We care about the exit status of the process we run within firejail.
Some compilers we run potentially spawn and orphan non-critical background tasks, telemetry agents for example.  We don't care if these auxiliary processes exit with an error.  If they do exit with an error *after* the compiler itself exits, we lose the exit status of the compiler. If the compiler exits with an error, and then the telemetry agent exits without an error, we lose the error condition from the compiler.

### DIFF
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -315,6 +315,7 @@ extern int arg_notv;	// --notv
 extern int arg_nodvd;	// --nodvd
 extern int arg_nou2f;   // --nou2f
 extern int arg_nodbus; // -nodbus
+extern int arg_deterministic_exit_code;	// always exit with first childs exit status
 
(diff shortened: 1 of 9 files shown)


chiraag-nataraj:
Looks okay to me, but I'll wait for someone else to also comment on it.

chiraag-nataraj:
Merged, thanks!

PR #2724, opened 2019-05-25
Create SECURITY.md

Fred-Barclay:
Just taking advantage of the new "security" tab on GitHub.
This lists currently supported versions and versions supported by distribution packagers.

Thoughts? I'm not sure how I feel about the red X's when the versions are still offered in distro repos especially since @reinerh is so active here; but at the same time I'm not sure we actively support those older versions.

EOL dates are when the next version was released.

Preview:
image

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,22 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported by us    | EOL  | Supported by distribution |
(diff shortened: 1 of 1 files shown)


netblue30:
Check it in, thanks!

The latest is 0.9.58.2. I have to fix the main page on wordpress (firejail.wordpress.com). Somehow I got a 0.9.58 released in January and another 0.9.58 released in February! On the download page is fine.

Issue #1078, opened 2017-01-29
Issue whitelisting or noblacklisting home dir


azimut:
I can't find a way to make firejail work with my setup. I have different users with different home directories on ex: /tree/a/b/c/ /tree/a/c/d/ , etc.

So I want to blacklist access to /tree, but whitelist access to the user home directory. There is a way to do this with the current --blacklist --noblacklist or --whitelist options? I cannot seem to find the correct ones.

I can blacklist /tree no problem, but cannot just "whitelist" the home (~/) subdir.
```
$ firejail --version
firejail version 0.9.44.8

Compile time support:
        - AppArmor support is disabled
        - AppImage support is enabled
        - bind support is enabled
        - chroot support is enabled
        - file and directory whitelisting support is enabled
        - file transfer support is enabled
        - networking support is enabled
        - overlayfs support is enabled
        - private-home support is enabled
        - seccomp-bpf support is enabled
        - user namespace support is enabled
        - X11 sandboxing support is enabled
$ rpm -qa | grep fireja
firejail-0.9.44.8-1.x86_64
```

netblue30:
Masking the user accounts is kind of tricky, because you need to know where the user accounts are located. In this moment we can handle only user accounts stored under /home. Maybe we can bring this kind of support in a future release, but for now is not supported.

azimut:
If I could --bind as non-root (with firejail suid set) I should be able to work around this. For reference "bubblewrap" allows that. But there should be a good reason to disable it here I think. Other than that my other option could be use "--chroot" with a pre-baked environment. Any guidelines written about --chroot with firejail?

netblue30:
Binding as nonroot is dangerous. For example, you can bind an new version of /etc directory with a known password in /etc/shadow.

azimut:
FWIW, I ended up having some free time to resume this. My workaround was create symlinks under /home pointing to /data/a/dir/ and add the blacklist of /data on default.profile.

Issue #1127, opened 2017-03-05
--whitelist=~/.bashrc temporarily overwrites .bashrc


Fred-Barclay:
This was brought to my attention by a chap on the Linux Mint forums. When doing something like the following, .bashrc is overwritten or overlayed inside the jail with a new, different .bashrc.:
```
firejail --whitelist=~/.bashrc
firejail --noprofile --whitelist=~/.bashrc
```
Compare the following output for `firejail --noprofile --whitelist=~/.bashrc`:
**Outside jail**
```
$ tail .bashrc
# this, if it's already enabled in /etc/bash.bashrc and /etc/profile
# sources /etc/bash.bashrc).
if ! shopt -oq posix; then
  if [ -f /usr/share/bash-completion/bash_completion ]; then
    . /usr/share/bash-completion/bash_completion
  elif [ -f /etc/bash_completion ]; then
    . /etc/bash_completion
  fi
fi
export QT_STYLE_OVERRIDE=gtk
```
**Inside jail**
```
tail .bashrc
# enable programmable completion features (you don't need to enable
# this, if it's already enabled in /etc/bash.bashrc and /etc/profile
# sources /etc/bash.bashrc).
if ! shopt -oq posix; then
  if [ -f /usr/share/bash-completion/bash_completion ]; then
    . /usr/share/bash-completion/bash_completion
  elif [ -f /etc/bash_completion ]; then
    . /etc/bash_completion
  fi
fi
```
This is on Debian Jessie.

For the bloke who originally mentioned this, he has a far more drastic change on Arch Linux, from a very long .bashrc outside of firejail to just these lines inside:
```
#
# ~/.bashrc
#

# If not running interactively, don't do anything
[[ $- != *i* ]] && return

alias ls='ls --color=auto'
PS1='[\u@\h \W]\$ '
```
Debug output is at https://gist.github.com/Fred-Barclay/fa603fb1ff6dc3084b907e7f680bf3e2

I've done a bit of digging around but I can't figure out the cause. If it were due to the `read-only ${HOME}/.bashrc` line in etc/disable-common.inc then this shouldn't occur when I use `--noprofile`, right? But it does. :confused: 

Is this expected?

netblue30:
It is copying by default /etc/skel/.bashrc, I have to disable this when .bashrc is whitelisted. Thanks!

chiraag-nataraj:
Is this still around?

Fred-Barclay:
[...]

PR #2732, opened 2019-05-30
improve/add support for arbitrary home directories

### LINKED ISSUE #2259: Whitelisting results in fs_private: Read-only file system
Using firejail 0.9.56-1 (the .deb amd64 one from SFnet).

I want to jail ffmpeg and ffprobe for the user ```peertube``` which HOME directory is ```/var/www/peertube```.
I created a ffprobe profile like that (diff from ffmpeg):
```diff
< # Firejail profile for ffmpeg
---
> # Firejail profile for ffprobe
6c6
< include /etc/firejail/ffmpeg.local
---
> include /etc/firejail/ffprobe.local
34c34
< private-bin ffmpeg
---
> private-bin ffprobe
```

And two wrappers:
```
cat /usr/local/bin/jail_ffprobe 
#!/bin/bash
/usr/bin/firejail -- /usr/bin/ffprobe "$@"
```

So I want ffprobe/ffmpeg to be able to transcode videos in ```/var/www/peertube/storage/videos``` and I've written the following profile:
```
read-write /var/www/peertube/storage/avatars
read-write /var/www/peertube/storage/logs
read-write /var/www/peertube/storage/cache
read-write /var/www/peertube/storage/cache/video-captions
read-write /var/www/peertube/storage/cache/previews
read-write /var/www/peertube/storage/torrents
read-write /var/www/peertube/storage/thumbnails
read-write /var/www/peertube/storage/videos
read-write /var/www/peertube/storage/captions
read-write /var/www/peertube/storage/previews
```

If I do ```/usr/local/bin/jail_ffprobe /var/www/peertube/storage/videos/somevideo.ext``` I get a ```No such file or directory``` (the file really exists), so I tried to whitelist that directory with:
```
whitelist /var/www/peertube/storage/videos
```

But now it results in:
```
/usr/local/bin/jail_ffprobe /var/www/peertube/storage/videos/somevideo.ext                                                
Error chown: fs_home.c:339 fs_private: Read-only file system
Error: proc 29346 cannot sync with peer: unexpected EOF
Peer 29347 unexpectedly exited with status 1
```

Whitelisting storage or the home itself results in the same error.

Any ideas how to get that working ?

smitsohu:
Relevant mostly for home directories outside /home or paths containing symbolic links. This pull request aims at removing all implicit assumptions about the location of the user home directory (which in fact can be almost anywhere in the filesystem). Instead it adds a new restriction that the home directory must be owned by the user.

#1078
Fixes #2259 (kind of)

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -264,18 +264,44 @@ static void init_cfg(int argc, char **argv) {
 		fprintf(stderr, "Error: user %s doesn't have a user directory assigned\n", cfg.username);
 		exit(1);
 	}
-	cfg.homedir = clean_pathname(pw->pw_dir);
-	assert(cfg.homedir);
-	// detect problems with user home directory
-	int fd = safe_fd(cfg.homedir, O_PATH|O_DIRECTORY|O_NOFOLLOW|O_CLOEXEC);
-	if (fd == -1) {
-		if (errno == ENOTDIR) {
-			fprintf(stderr, "Error: user directory %s is invalid (symbolic links are not allowed)\n", cfg.homedir);
-			exit(1);
+	if (checkcfg(CFG_HOMEDIR_SYMLINK)) {  // resolve symbolic links
+		cfg.homedir = realpath(pw->pw_dir, NULL);
+		if (!cfg.homedir)
+			cfg.homedir = clean_pathname(pw->pw_dir);
+		assert(cfg.homedir);
+
+		// enforce a user owned directory outside /proc and /sys
+		int fd = safe_fd(cfg.homedir, O_PATH|O_DIRECTORY|O_NOFOLLOW|O_CLOEXEC);
+		if (fd > -1) {
+			if (strncmp(cfg.homedir, "/proc/", 6) == 0 || strncmp(cfg.homedir, "/sys/", 5) == 0) {
+				fprintf(stderr, "Error: invalid user directory\n");
+				exit(1);
+			}
+			struct stat s;
+			if (fstat(fd, &s) == -1)
+				errExit("fstat");
+			if (s.st_uid != getuid()) {
+				fprintf(stderr, "Error: user directory %s is not owned by user %s\n", cfg.homedir, cfg.username);
+				exit(1);
+			}
+			close(fd);
 		}
 	}
-	else
-		close(fd);
+	else {  // reject symbolic links
+		cfg.homedir = clean_pathname(pw->pw_dir);
+		assert(cfg.homedir);
+
+		int fd = safe_fd(cfg.homedir, O_PATH|O_DIRECTORY|O_NOFOLLOW|O_CLOEXEC);
+		if (fd == -1) {
+			if (errno == ENOTDIR) {
+				fprintf(stderr, "Error: invalid user directory %s: path element is not a directory\n"
+						"Hint: following symbolic links is disabled in Firejail configuration file\n", cfg.homedir);
+				exit(1);
+			}
+		}
+		else
+			close(fd);
+	}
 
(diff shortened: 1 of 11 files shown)


smitsohu:
One observation: If the home directory is `/opt/bla/bla/bla`, and directory `/opt/bla` is whitelisted, _too_, the initialization with .bashrc and so on is lost (because a tmpfs is mounted on the top level directory, masking the initialized empty home directory).

I think though this is a different issue and related to #1127

src/firejail/main.c
```diff
@@ -256,15 +260,38 @@ static void init_cfg(int argc, char **argv) {
...
-		cfg.homedir = clean_pathname(pw->pw_dir);
-		assert(cfg.homedir);
-	}
-	else {
+	if (pw->pw_dir == NULL) {
 		fprintf(stderr, "Error: user %s doesn't have a user directory assigned\n", cfg.username);
 		exit(1);
 	}
+	// resolve symbolic links
+	cfg.homedir = realpath(pw->pw_dir, NULL);
```
smitsohu: Everything else is decoration, but this realpath() adds real attack surface. As a mitigation I enforced a user owned home directory, and disallowed /sys and /proc, as we don't want to end up with a writable tmpfs mounted in there. I'd very much appreciate thoughts if this would be considered "safe enough".

smitsohu:
I guess I'll update the pull request so the defensive way without realpath() is proposed. We can still think about it though.

smitsohu:
So, the pull request is more conservative now, but strictly excludes symbolic links in home directory paths. Everything else is allowed.

smitsohu:
What's still left now is all the interactions between the private* options and between private* and whitelisting.

If homedir is outside /home and in one of the other top level directories that support private* or whitelisting, it is easily possible to configure the sandbox in a way that inside the sandbox the home directory is eradicated from the filesystem. I'm not sure we want that.

I would propose to disallow stacking of private* options, but don't know yet what to do with the whitelisting; one possibility would be to automatically add an initialized and otherwise empty home directory, even if it was not explicitly whitelisted in that top level directory (fixing the "observation" from https://github.com/netblue30/firejail/pull/2732#issuecomment-497502556 ).

SkewedZeppelin:
Why not move this behind a config option that is disabled by default?

So if it is found insecure it wouldn't impact nearly as many users.

smitsohu:
@SkewedZeppelin Sure, we can do it this way! I'll add it to the pull request tomorrow.

smitsohu:
I think there is only a risk if the user has control over his home directory, in the sense it can be temporarily moved to somewhere else and replaced with a symbolic link... this would be a very unusual system setup though. Anyway, we are considering the scenario and symlink resolution is disabled out of the box for now (but can be enabled in firejail.config).

smitsohu:
sorry, I need to re-add the WIP badge

smitsohu:
> I think there is only a risk if the user has control over his home directory, in the sense it can be temporarily moved to somewhere else and replaced with a symbolic link

A much better solution would be to detect this situation and then decide accordingly. Such a conditional symlink resolution could be enabled by default. I'll try to it that way and will update the pr the coming week.

smitsohu:
This has escalated quite a bit; I'll close this one and break it down into smaller chunks.

smitsohu:
> A much better solution would be to detect this situation and then decide accordingly. Such a conditional symlink resolution could be enabled by default. I'll try to do it that way and will update the pr the coming week.

I have it all working now, but looking at the piles of convoluted code for handling this corner case, it doesn't seem right any more to solve this in a technical way. Maybe it's indeed better to disable symlink resolution and leave a hint in firejail.config, putting/leaving the system administrator in charge.

So we're kind of back at the start.

Issue #1569, opened 2017-09-21
Why do we have blacklist, noblacklist, and whitelist in the same profile?


Fred-Barclay:
Is there a particular reason or historical significance for including disable-programs.inc, then noblacklisting the files we need, and then whitelisting those files?
Take the firefox profile for instance. We blacklist quite a few files in ${HOME} with `include /etc/firejail/disable-programs.inc`, and then noblacklist the ones we need:
```noblacklist ~/.cache/mozilla
noblacklist ~/.config/okularpartrc
noblacklist ~/.config/okularrc
noblacklist ~/.config/qpdfview
noblacklist ~/.kde/share/apps/okular
noblacklist ~/.kde/share/config/okularpartrc
noblacklist ~/.kde/share/config/okularrc
noblacklist ~/.kde4/share/apps/okular
noblacklist ~/.kde4/share/config/okularpartrc
noblacklist ~/.kde4/share/config/okularrc
noblacklist ~/.local/share/gnome-shell/extensions
noblacklist ~/.local/share/okular
noblacklist ~/.local/share/qpdfview
noblacklist ~/.mozilla
noblacklist ~/.pki
```

We then whitelist these files:
```mkdir ~/.cache/mozilla/firefox
mkdir ~/.mozilla
mkdir ~/.pki
whitelist ${DOWNLOADS}
whitelist ~/.cache/gnome-mplayer/plugin
whitelist ~/.cache/mozilla/firefox
whitelist ~/.config/gnome-mplayer
whitelist ~/.config/okularpartrc
whitelist ~/.config/okularrc
whitelist ~/.config/pipelight-silverlight5.1
whitelist ~/.config/pipelight-widevine
whitelist ~/.config/qpdfview
whitelist ~/.kde/share/apps/okular
whitelist ~/.kde/share/config/okularpartrc
whitelist ~/.kde/share/config/okularrc
whitelist ~/.kde4/share/apps/okular
whitelist ~/.kde4/share/config/okularpartrc
whitelist ~/.kde4/share/config/okularrc
whitelist ~/.keysnail.js
whitelist ~/.lastpass
whitelist ~/.local/share/gnome-shell/extensions
whitelist ~/.local/share/okular
whitelist ~/.local/share/qpdfview
whitelist ~/.mozilla
whitelist ~/.pentadactyl
whitelist ~/.pentadactylrc
whitelist ~/.pki
whitelist ~/.vimperator
whitelist ~/.vimperatorrc
[...]

PR #2735, opened 2019-05-31
Extend profile.template with comments

jose1711:


jose1711:
The goal is to make it possible for an "average Joe" to create a new profile by following the template.

etc/templates/profile.template
```diff
@@ -1,6 +1,21 @@
...
+#
+# --- CUT HERE ---
+# This is a generic template to help you with creation of profiles
+# for new programs. PRs welcome at https://github.com/netblue30/firejail/
+#
+# Rules to follow:
+#  - make the profile as restrictive as possible while still keeping the program useful
+#    (e. g. a program that is unable to save user's work is considered a bad practice)
+#  - test profile for couple days before raising a pull request
+#  - keep the sections structure, use a single empty line as a separator
```
SkewedZeppelin: the structure is

HEADER
COMMENTS
NOBLACKLISTS
BLACKLISTS
DISABLE INCLUDES
MKDIRS
WHITELISTS
WHITELIST INCLUDES
OPTIONS (no*)
PRIVATE OPTIONS (private-*)
SPECIAL OPTIONS (mdwx, noexec, read-only)
REDIRECT INCLUDES

etc/templates/profile.template
```diff
@@ -11,6 +26,14 @@
 
 ##blacklist PATH
 
+# It is common to add files/dirs containing program-specific configuration
+# (often ${HOME}/PROGRAMNAME or ${HOME}/.config/PROGRAMNAME into disable-programs.inc
+# (included below) and then disable blacklisting below.
+# One way to retrieve the files a program uses is:
```
SkewedZeppelin: an easier method:
- disable `shell none` and `private-bin`
- launch with --private
- join the profile with bash
- use `ls -lha` to see what was created

etc/templates/profile.template
```diff
@@ -21,6 +44,13 @@
 #noblacklist /usr/local/lib/python2*
 #noblacklist /usr/local/lib/python3*
 
+# Allow perl (blacklisted by disable-interpreters.inc)
```
SkewedZeppelin: more generally, maybe we should break these out into their own so they can be included
```
include allow-python.inc
include allow-perl.inc
```
thoughts?

etc/templates/profile.template
```diff
@@ -29,6 +59,7 @@
 #include disable-programs.inc
 #include disable-xdg.inc
 
+# This section often mirrors noblacklist section above
```
SkewedZeppelin: some talk about that in #1569

etc/templates/profile.template
```diff
@@ -1,6 +1,21 @@
...
+# This is a generic template to help you with creation of profiles
+# for new programs. PRs welcome at https://github.com/netblue30/firejail/
+#
+# Rules to follow:
+#  - make the profile as restrictive as possible while still keeping the program useful
+#    (e. g. a program that is unable to save user's work is considered a bad practice)
+#  - test profile for couple days before raising a pull request
+#  - keep the sections structure, use a single empty line as a separator
+#  - entries within sections are alphabetically sorted
+#  - also update src/firecfg/firecfg.config with the name of the app (mind sorting)
```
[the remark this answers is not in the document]
jose1711: Thanks, will add a warning here

etc/templates/profile.template
jose1711: Thank you, will add

etc/templates/profile.template
jose1711: Looks neat. I will be happy to add that once we have such includes.

etc/templates/profile.template
jose1711: Will add some details.

etc/templates/profile.template
jose1711: Adjusted

etc/templates/profile.template
```diff
@@ -1,6 +1,21 @@
...
 # This file is overwritten after every install/update
+#
+# --- CUT HERE ---
+# This is a generic template to help you with creation of profiles
+# for new programs. PRs welcome at https://github.com/netblue30/firejail/
+#
+# Rules to follow:
+#  - make the profile as restrictive as possible while still keeping the program useful
+#    (e. g. a program that is unable to save user's work is considered a bad practice)
+#  - test profile for couple days before raising a pull request
```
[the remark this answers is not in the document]
jose1711: will change wording slightly, thanks

etc/templates/profile.template
```diff
@@ -11,6 +43,20 @@
 
 ##blacklist PATH
 
+# It is common to add files/dirs containing program-specific configuration
+# (often ${HOME}/PROGRAMNAME or ${HOME}/.config/PROGRAMNAME into disable-programs.inc
```
rusty-snake: `… into disable-programs.inc (keep list sorted)`

jose1711: Fixed, TY

etc/templates/profile.template
rusty-snake: @jose1711 @SkewedZeppelin `grep "^blacklist" --context=5 /etc/firejail/*.profile`
profile with `blacklist` and `noblacklist` have `blacklist` above `noblacklist`:
…
COMMENTS
BLACKLISTS
NOBLACKLISTS
DISABLE INCLUDES
…

rusty-snake: Also what means COMMENTS? the comments in the HEADER?

rusty-snake: To be complete:
PRIVATE OPTIONS (disable-mnt, private-*)
SPECIAL OPTIONS (mdwx, noexec, read-only, join-or-start)

etc/templates/profile.template
rusty-snake: I prefer `tree -a`, but that is a individual decision. Because of the most program-specific dirs are hidden:`ls -lR` -> `ls -laR` (I don't think that the `-l` must be there)

Disable `shell none` is not needed, `firejail --private --ignore=private-bin --name=<NAME> --profile=<PROFILE> <PROGRAMM>` should be work.

jose1711: `tree` is not default on my Arch Linux system, `ls` is ubiquitous. will update the rest

jose1711:
> Looks very great, only a few things left:
> 
>     * Should we remove line 4?

Do you mean the one with `--- CUT HERE ---`? actually.. I was thinking of dropping all the template comments (both block and inline ones) by means of `grep -v`, `sed` or `awk` - maybe someone has an idea how to do this nicely.

>     * Add the message of these two lines from Notes, because than we can remove Notes.
> 
> 
> ```
> * Lines with one # are often used
> * Lines with two ## are only in special situation needed
> ```

Sure, but please explain.. the uncommenting (be it single or double) is always meant to make the profile less restrictive?

rusty-snake:
1. I mean the `#\n`
2. Not sure what you mean, but as you write:
> make the profile as restrictive as possible while still keeping the program useful

Lines I commented with one # are normally used if they don't break a program (e.g. `#nodbus`) and if they make sense (e.g. `#noblacklist PYTHON`).
Lines I commented with two # are normally not used unless they are needed (e.g. `##env VAR=VALUE`, `##seccomp.drop`) or rarely make sense (e.g. `##private`, `##read-only`) or difficult to write (e.g. `##private-lib`).

jose1711:
>     1. I mean the `#\n`

removed

> Lines I commented with one # are normally used if they don't break a program (e.g. `#nodbus`) and if they make sense (e.g. `#noblacklist PYTHON`).
> Lines I commented with two # are normally not used unless they are needed (e.g. `##env VAR=VALUE`, `##seccomp.drop`) or rarely make sense (e.g. `##private`, `##read-only`) or difficult to write (e.g. `##private-lib`).

ah, I see. The confusion came from the fact that ever since I was thinking about a template I always pictured it as a profile with every possible option enabled and the user would just comment (later delete) the lines to make it work with a certain program. lines added.

> 
> EDIT: line 58 can be removed: `# - reenable 'shell none' and 'private-bin' (if applicable)`

done

rusty-snake:
Merged, Thanks.

PR #2736, opened 2019-06-01
Create allow-INTERPETER.inc

rusty-snake:
* allow-lua.inc
 * allow-perl.inc
 * allow-python2.inc
 * allow-python3.inc
 * allow-java.inc

See https://github.com/netblue30/firejail/pull/2735#discussion_r289574610

### DIFF
--- a/etc/JDownloader.profile
+++ b/etc/JDownloader.profile
@@ -8,7 +8,8 @@ include globals.local
 
 noblacklist ${HOME}/.jd
 
-# Allow access to java
+# Allow java (blacklisted by disable-interpreters.inc)
+include allow-java.inc
 
 include disable-common.inc
 include disable-devel.inc
--- a/etc/JDownloader.profile
+++ b/etc/JDownloader.profile
@@ -8,7 +8,7 @@ include globals.local
 
 noblacklist ${HOME}/.jd
 
-# Allow java (blacklisted by disable-interpreters.inc)
+# Allow java (blacklisted by disable-devel.inc)
 include allow-java.inc
 
 include disable-common.inc
--- a/etc/JDownloader.profile
+++ b/etc/JDownloader.profile
@@ -9,10 +9,6 @@ include globals.local
 noblacklist ${HOME}/.jd
 
 # Allow access to java
-noblacklist ${PATH}/java
-noblacklist /usr/lib/java
-noblacklist /etc/java
-noblacklist /usr/share/java
 
 include disable-common.inc
 include disable-devel.inc
--- a/etc/allow-java.inc
+++ b/etc/allow-java.inc
@@ -1,0 +1,4 @@
+noblacklist ${PATH}/java
+noblacklist /usr/lib/java
+noblacklist /etc/java
+noblacklist /usr/share/java
--- a/etc/allow-lua.inc
+++ b/etc/allow-lua.inc
@@ -1,0 +1,4 @@
+noblacklist ${PATH}/lua*
+noblacklist /usr/include/lua*
+noblacklist /usr/lib/lua
+noblacklist /usr/share/lua
--- a/etc/allow-perl.inc
+++ b/etc/allow-perl.inc
@@ -1,5 +1,7 @@
 noblacklist ${PATH}/cpan*
 noblacklist ${PATH}/core_perl
 noblacklist ${PATH}/perl
+noblacklist ${PATH}/site_perl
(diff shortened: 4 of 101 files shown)


SkewedZeppelin:
I think I got most of them.
I left out the template to not break #2735

rusty-snake:
HEADER
COMMENTS
BLACKLISTS
NOBLACKLISTS
**ALLOW INCLUDES**
DISABLE INCLUDES
MKDIRS
WHITELISTS
WHITELIST INCLUDES
OPTIONS (no*)
PRIVATE OPTIONS (disable-mnt, private-*)
SPECIAL OPTIONS (mdwx, noexec, read-only)
REDIRECT INCLUDES

rusty-snake:
meld, natron, qutebrowser, ranger need to be updated.

SkewedZeppelin:
That should be it.
Aside from placement.

etc/JDownloader.profile
```diff
@@ -8,7 +8,8 @@ include globals.local
 
 noblacklist ${HOME}/.jd
 
-# Allow access to java
+# Allow java (blacklisted by disable-interpreters.inc)
```
rusty-snake: java is blacklisted in disable-devel.

Vincent43:
```
HEADER
COMMENTS
BLACKLISTS
NOBLACKLISTS
ALLOW INCLUDES
```
I think NOBLACKLISTS + ALLOW INCLUDES should be before BLACKLISTS otherwise they won't work :smiley:

rusty-snake:
@Vincent43 to have the following in a profile make no sense:
```
noblacklist ${HOME}/something
blacklist ${HOME}/something
```
BLACKLIST means `blacklist` commands in the profile.

glitsj16:
To accomodate [Arch Linux's Perl policy](https://wiki.archlinux.org/index.php/Perl_Policy) I'm wondering whether we should add

```
noblacklist ${PATH}/site_perl
noblacklist ${PATH}/vendor_perl
```

to the new [allow-perl.inc](https://github.com/netblue30/firejail/blob/c62fcf782117718c1abc558e392a12a74c798335/etc/allow-perl.inc).

Fred-Barclay:
@glitsj16 I think we should

rusty-snake:
@glitsj16 feel free to commit.

Vincent43:
> to have the following in a profile make no sense:
> 
> noblacklist ${HOME}/something
> blacklist ${HOME}/something
> 

The opposite order also doesn't make sense. It doesn't make sense having noblacklist and blacklist for the same path in profile in general but noblacklist rule should precede blacklist one to be effective elsewhere and I think keeping that order here wold be consistent.

rusty-snake:
@Vincent43 to do that to have a consistent/consequent order sounds good. I use this order above because every profile with `blacklist` and `noblacklist` has this order.
`grep "^blacklist" --context=5 /etc/profile/*.profile`

OK then lets have the following:

HEADER
_COMMENTS -- don't know what this is._
**IGNORES**
NOBLACKLISTS
**ALLOW INCLUDES**
**BLACKLISTS**
DISABLE INCLUDES
MKDIRS
WHITELISTS
WHITELIST INCLUDES
OPTIONS (aa, caps, ipc-namespace, machine-id, net*, no*, protocol, seccomp, shell, tracelog)
PRIVATE OPTIONS (disable-mnt, private*)
SPECIAL OPTIONS (env, mdwx, noexec, read-only, join-or-start)
REDIRECT INCLUDES

@glitsj16 where do we `blacklist ${PATH}/site_perl` & `blacklist ${PATH}/vendor_perl` ?

glitsj16:
@rusty-snake At present we don't blacklist site_perl and vendor_perl, although I think we should. I forgot that part, nice catch. To keep this draft PR clean, would you be so kind as to add those to the current /etc/firejail/disable-interpreters.inc?

SkewedZeppelin:
@rusty-snake 
> COMMENTS -- don't know what this is.

some profiles have comments on their use
eg.
https://github.com/netblue30/firejail/blob/master/etc/Xephyr.profile
https://github.com/netblue30/firejail/blob/master/etc/xpra.profile
https://github.com/netblue30/firejail/blob/master/etc/spectre-meltdown-checker.profile

SkewedZeppelin:
Ignoring placement, is this fine to merge?

rusty-snake:
@SkewedZeppelin I think so.

PR #2741, opened 2019-06-03
Typo fix in brackets.profile

glitsj16:


rusty-snake:
@glitsj16 only as a question: why do you use a PR for changes of only one file?

With more than one file I still understand. Because if you only edit via GH you won't get more than one file in one commit. But not for a single file, because as a collaborator you can edit files directly.

glitsj16:
@rusty-snake That's just the way I like to do it. It keeps my to-do lists happy. If there's something irritating or otherwise dysfunctional about this workflow, feel free to advise. I'm always open for suggestions :-)

PR #2745, opened 2019-06-04
template profile: update private-etc templates

SkewedZeppelin:
This replaces the outdated templates from #1734
with new templates from the program used in #2093

Making as a PR, because I know we had concerns with some of these, as commented in #2093.
Since this is a template, and nothing automatic it should be OK, as they should be suggestions to the profile creator and used with discretion.

Feel free to push any changes at will.


rusty-snake:
I think we can merge this now.

PR #2746, opened 2019-06-04
firefox-common-addons.inc: + tridactyl

l29ah:


rusty-snake:
@l29ah looks fine, but we never do `blacklist ${HOME}/.local/share/tridactyl` so you can remove line 22.

Fred-Barclay:
@rusty-snake actually we should probably add `blacklist ${HOME}/.local/share/tridactyl` to disable-programs.inc and keep the noblacklist line here

rusty-snake:
@Fred-Barclay that works, too, I had decided to remove it because we don't have it with most of the others (e.g. vimperator).

rusty-snake:
@l29ah can you apply one of the advices, either the one I gave you, or rather the one @Fred-Barclay gave you.

rusty-snake:
@l29ah

rusty-snake:
Merged, Thanks.

PR #2753, opened 2019-06-07
Add davfs2 secrets file to blacklist

jose1711:
The files holds credentials to WebDAV servers in plaintext
hence it's probably a good idea to limit access to them.


rusty-snake:
@jose1711 thats good, but I think that we have some profile which need to access these files.
Or am I mistaken? If not try to find out which are this.

jose1711:
the only program i can think of is `nautilus`. but odds are that it uses its own file for auth stuff

rusty-snake:
Ok, you're right. `dnf repoquery --alldeps --whatrequires davfs2` doesn have output.

PR #2759, opened 2019-06-10
Mention macros in profile.template

jose1711:


rusty-snake:
@jose1711  Looks Great, but I'm missing `${PATH}`.

jose1711:
> @jose1711 Looks Great, but I'm missing `${PATH}`.

good point there :smile:

rusty-snake:
Merged, Thanks.

Issue #2739, opened 2019-06-02
Add automated CI checks and git hooks


jose1711:
It would be nice to have automated checks for things that tend to repeat in new profile submissions:
- header present (with description, persistent globals and local include)
- items sorted inside profile sections
- sections are divided using a single '\n' character
- `firecfg.section`, `disable-programs.inc` and similar files being sorted

rusty-snake:
I had a similar idea a few days ago and started to write it in rust, but I'm actually thinking if it's still necessary if you prominently recommend the profile.template for it (e.g. use GH pull request templates, CONTRIBUTIN.md, manpage (see my comment in the add template issue)), because it will never be possible to check it with human inteliegenz (e.g. `include disable-xdg.inc` in an image-viewer-profile but not `noblacklist ${PICTURES}`)

jose1711:
I agree for the most part - I wonder though how is this GH pull request template supposed to work. Is it difficult to implement?

rusty-snake:
Its a issue template but for pull request, see https://help.github.com/en/articles/about-issue-and-pull-request-templates.

rusty-snake:
With this script you get alphabetical sorted `private-etc` and `private-bin` lines. I will add support for automaticaly fixing profile.
```python3
#!/usr/bin/env python3
from sys import argv

with open(argv[1], "r") as profile:
    for line in profile:
        if line[:11] == "private-etc":
            print("private-etc", ",".join(sorted(line[12:-1].split(","),
                key=lambda s: s.casefold())))
        elif line[:11] == "private-bin":
            print("private-bin", ",".join(sorted(line[12:-1].split(","),
                key=lambda s: s.casefold())))
```

jose1711:
note that `[12:-1] ` means *excluding* the last character. not sure if you want that.

rusty-snake:
@jose1711 thanks. The last character is `\n`. I test this code, it works.

jose1711:
ok, but what if `private-etc` is the last line and is not terminated with `\n`? like so:
`echo -n 'private-etc foo,bar' > profile`
then I get this:
```
./fix_sorting profile                          
private-etc ba,foo
```

rusty-snake:
@jose1711 Most profiles end with a newline, but you're right that we have to consider the missing of it.
***
New version, with support for automatic fixing  several profiles.
https://gist.github.com/rusty-snake/a1010a3daf3c54e93dfe03f2f5ce3d96

Issues:
[...]

PR #2766, opened 2019-06-13
automatically fixed all private-{bin,etc} lines

rusty-snake:
Bring all private-etc and private-bin lines into an alphabetical order, using this:
https://gist.github.com/rusty-snake/a1010a3daf3c54e93dfe03f2f5ce3d96

* special characters are ignored
* the script order case insensitive

Missing (will not be implemented, only for completeness): 
 * commented lines
   * `#private-{bin,etc}`, `# private-{bin,etc}`
   * What about something like `# private-etc foo,bar -- breaks baz` and
`# private-etc must first be enabled in foobar`
 * `private-lib`
 * see also https://github.com/netblue30/firejail/issues/2739#issuecomment-499892002
 * review

### DIFF
--- a/etc/kid3.profile
+++ b/etc/kid3.profile
@@ -37,7 +37,7 @@ tracelog
 
 private-cache
 private-dev
-private-etc alternatives,drirc,fonts,kde5rc,gtk-3.0,dconf,machine-id,ca-certificates,ssl,pki,hostname,hosts,resolv.conf,pulse,,crypto-policies
+private-etc ,alternatives,ca-certificates,crypto-policies,dconf,drirc,fonts,gtk-3.0,hostname,hosts,kde5rc,machine-id,pki,pulse,resolv.conf,ssl
 private-tmp
 private-opt none
 private-srv none
--- a/etc/kid3.profile
+++ b/etc/kid3.profile
@@ -37,7 +37,7 @@ tracelog
 
 private-cache
 private-dev
-private-etc ,alternatives,ca-certificates,crypto-policies,dconf,drirc,fonts,gtk-3.0,hostname,hosts,kde5rc,machine-id,pki,pulse,resolv.conf,ssl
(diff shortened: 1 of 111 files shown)


etc/kid3.profile
```diff
@@ -37,7 +37,7 @@ tracelog
 
 private-cache
 private-dev
-private-etc alternatives,drirc,fonts,kde5rc,gtk-3.0,dconf,machine-id,ca-certificates,ssl,pki,hostname,hosts,resolv.conf,pulse,,crypto-policies
+private-etc ,alternatives,ca-certificates,crypto-policies,dconf,drirc,fonts,gtk-3.0,hostname,hosts,kde5rc,machine-id,pki,pulse,resolv.conf,ssl
```
rusty-snake: Note for me: remove the trailing comma

Vincent43:
I wonder hot to automate this and make mandatory as it's very likely that future profile updates wil break the order.

rusty-snake:
@Vincent43 I'm not sure if I understood you correctly, but that's fix every profile for now, and if in the future something like that will be enforced I can't say.

Vincent43:
Yes, this will fix current state but I doubt people would care about alphabetic order in profiles while creating future PRs so this is quite cat and mouse game. Nevertheless I have nothing against this PR.

glitsj16:
@rusty-snake Thanks for mentioning your [sort.py](https://gist.github.com/rusty-snake/a1010a3daf3c54e93dfe03f2f5ce3d96) script. I extended it so it handles private-lib too. Why not offer it up for `contrib`? We could use more of these helper scripts IMHO. Great job :+1: .

Vincent43:
I think this script may better fit for travis CI, i.e. it could fail if there is diff between its output and proposed patch in PR.

rusty-snake:
@glitsj16 I also add support for `private-lib` and improve the performance.

rusty-snake:
For @Vincent43 proposal lets move back to #2739.

PR #2771, opened 2019-06-14
mount new proc filesystem earlier

smitsohu:
a precondition for using proc magic links in the early phase of sandbox setup


netblue30:
all in, thanks!

PR #2774, opened 2019-06-15
chromium: disable nodbus

Vincent43:
Unfortunately `nodbus` prevents access to site passwords if they are stored in [GNOME Keyring](https://wiki.gnome.org/Projects/GnomeKeyring). According to [docs](https://chromium.googlesource.com/chromium/src.git/+/master/docs/linux_password_storage.md) chromium can store password in 3 different ways:

- GNOME Keyring
- KWallet 4
- plain text

As KWallet storage may be broken in a same way, using `nodbus` will force chromium to store passwords in plain text which isn't best option for security and for firejail default.

### DIFF
--- a/etc/chromium-common.profile
+++ b/etc/chromium-common.profile
@@ -29,8 +29,7 @@ include whitelist-var-common.inc
 apparmor
 caps.keep sys_chroot,sys_admin
 netfilter
-# Breaks Gnome connector - disable if you use that
-nodbus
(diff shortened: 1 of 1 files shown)


glitsj16:
Any reason why we shouldn't merge this now?

rusty-snake:
I don't see one, but :+1: for the PR.

Vincent43:
I wanted to gather some feedback before merging. As it's positive then I'm merging it.

PR #2784, opened 2019-06-18
Improve profile PRs (Related to #2739)

rusty-snake:
This PR adds my sort.py script to `contrib/` and add a GH Pull Request Template.

TODOs on sort.py:
 * [x] write a litte usage section

Feel free to improve the PR template, suggest changes, say that this is a stupid idea, …


glitsj16:
Looks great. Nitpicks:
- explain that while developing a new profile it might be wise to run sort.py on a *copy* of the file first, before doing a PR? or is this overkill?
- calling it with python3 explicitly shouldn't be needed if the shebang already takes care of this, or am I missing something? 
- for those that want to use sort.py on the whole /etc tree at once: $ find /etc -type f exec /lib64/firejail/sort.py {} +

rusty-snake:
> explain that while developing a new profile it might be wise to run sort.py on a _copy_ of the file first, before doing a PR? or is this overkill?

I leave a note that it will overwrite.

> for those that want to use sort.py on the whole /etc tree at once: $ find /etc -type f exec /lib64/firejail/sort.py {} +

You need to be root.

glitsj16:
@rusty-snake As you might have noticed, I use your script more as a regression tester after I update firejail from git master on a local copy (as my user). So I don't need root. Thanks for clearing things up :+1: .

rusty-snake:
@glitsj16 I just realized that using `find` is absolutely overkill because `sudo /lib64/firejail/sort.py /etc/firejail/*.profile` works.
***
Can anyone say if `/lib64/firejail` is on distros like debian (which not symlink `/lib64` to `/usr/lib64`)  correct, or must it `/usr/lib64/firejail`?

netblue30:
all find with me, check it in when you are done!

glitsj16:
> @glitsj16 I just realized that using find is absolutely overkill because sudo /lib64/firejail/sort.py /etc/firejail/*.profile works.

@rusty-snake I guess the rationale of me using find without specifying a name was to also catch /etc/feh-network.inc, which *.profile wouldn't. I'm fine with whatever you decide though.

> Can anyone say if /lib64/firejail is on distros like debian (which not symlink /lib64 to /usr/lib64) correct, or must it /usr/lib64/firejail?

According to packages from the [PPA maintained by @reinerh](https://launchpad.net/~deki/+archive/ubuntu/firejail) stuff from /contrib seems to go into `/usr/lib/x86_64-linux-gnu/firejail`. Maybe wait until he confirms, him being the expert on Debian/Ubuntu.

rusty-snake:
@glitsj16 you can also do `sort.py etc/*.{profile,inc,local}`.

glitsj16:
> @glitsj16 you can also do sort.py etc/*.{profile,inc,local}.

@rusty-snake I know. Maybe that's a better example, so users don't think this is only useable on *.profile files (which is what I intended in the first comment but succeeded to miscommunicate entirely)

In the mean time I can confirm `/usr/lib/x86_64-linux-gnu/firejail` it is, looking at https://packages.debian.org/experimental/amd64/firejail/filelist

rusty-snake:
Ok, then I will add `sort.py etc/*.{profile,inc,local}` to the usage in sort.py and to the PR-Template this table:

| Distro | Path |
| --- | --- |
| Arch/Fedora | `/lib64/firejail/sort.py` |
| Local git clone | `contrib/sort.py` |

If you have a better name for distro because `Local git clone` isn't a realy distro, say it.

glitsj16:
> If you have a better name for distro because Local git clone isn't a realy distro, say it.

@rusty-snake I don't. Looks ready to merge, proceed!

glitsj16:
@rusty-snake Coming along nicely :+1: . Just a question. What happened, are your thoughts on @Vincent43's comment from https://github.com/netblue30/firejail/pull/2766 that this might better fit Travis/CI? IMO these objectives might not be exclusive, but besides realizing its purpose I'm completely unfamiliar with Travis/CI. Just adding it here so it doesn't get lost in translation.

rusty-snake:
> I'm completely unfamiliar with Travis/CI

me too, but if someone wanted to do it, that'd be nice.

PR #2802, opened 2019-06-27
Streamline redirect profiles

glitsj16:
This PR does 2 things:
- use persistent spacing and wording in redirect profiles;
- avoid including `globals.local` more than once (cfr. #2006 and #2010).


etc/abrowser.profile
```diff
@@ -16,6 +16,5 @@ whitelist ${HOME}/.mozilla
 # private-etc must first be enabled in firefox-common.profile
```
rusty-snake: `include globals.local`

PR #2806, opened 2019-06-28
Add fonts to private-etc in udiskie profile

jose1711:
This prevents from seeing Fontconfig error in udiskie output.


rusty-snake:
Merged, Thanks.

PR #2825, opened 2019-07-03
Fixes #2821, riot-desktop

### LINKED ISSUE #2821: /usr/bin/riot-desktop: line 3: 8 Trace/breakpoint trap (core dumped) electron /usr/lib/riot/ "$@"
Running riot-desktop firejail makes it crash with the following message:
> /usr/bin/riot-desktop: line 3: 8 Trace/breakpoint trap (core dumped) electron /usr/lib/riot/ "$@"

Thi started happening when both electron and riot-desktop got updated 3 or so days ago.

Bandie:
Making a PR of @rusty-snake 's fix.

### DIFF
--- a/etc/riot-desktop.profile
+++ b/etc/riot-desktop.profile
@@ -7,5 +7,8 @@ include riot-desktop.local
 # added by included profile
 #include globals.local
 
+ignore seccomp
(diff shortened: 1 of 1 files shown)


rusty-snake:
:+1: 
I'm just thinking if it might be better to add it to riot-web.profile rather the riot-desktop.profile, but I don't know where the difference is (I'm not using matrix).

Bandie:
As I understand it riot-desktop needs riot-web as riot-desktop might be the electron thingy while riot-web is the real™ application. But not sure about that.

veloute:
@rusty-snake riot-desktop.profile should be fine - riot-desktop is the electron (desktop app) side of riot, which is where the problem is.

rusty-snake:
Merged, Thanks.

PR #2828, opened 2019-07-04
seccomp prevents riot-desktop from launching

veloute:


FOSSONLY:
Really, lower security to run that program? I would drop Riot completely, until they fixed their software.

rusty-snake:
@FOSSONLY #2825

PR #2832, opened 2019-07-06
Add electron4 to allow wire-desktop to launch on Arch Linux

njfox:
`electron` has recently been changed on Arch Linux and `wire-desktop` launches with `electron4` now instead of `electron`. This is the new launcher script from the Arch package:

```
#!/usr/bin/env sh

electron4 "/usr/lib/wire-desktop" "$@"
```
This PR adds `electron4` to the `private-bin` declaration to allow the program to successfully launch on Arch.


Fred-Barclay:
Merged, thanks!

PR #2835, opened 2019-07-06
Fix #2834

### LINKED ISSUE #2834: qpdfview profile broken
qpdfview profile was broken in https://github.com/netblue30/firejail/pull/1489/files#diff-e7231b67defd834b3b5ed85587ba3147R34 causing that qpdfview never open. Commenting out it line or deleting it fix the issue.

Edu4rdSHL:
Since https://github.com/netblue30/firejail/pull/1489/files#diff-e7231b67defd834b3b5ed85587ba3147R34 qpdfview profile is broken and never open the program. Reverting it fix the issue.

### DIFF
--- a/etc/qpdfview.profile
+++ b/etc/qpdfview.profile
@@ -38,5 +38,3 @@ tracelog
 private-bin qpdfview
 private-dev
 private-tmp
-
-memory-deny-write-execute


glitsj16:
Confirmed, mdwe breaks qpdfview on Arch Linux. Thanks for the fix.

@Edu4rdSHL I don't use this app myself, but a quick test seems to suggest the profile can be hardened by adding `nodbus`. Would you care to test that please?

Edu4rdSHL:
@glitsj16 tested without problems, see https://github.com/netblue30/firejail/pull/2837

glitsj16:
@Edu4rdSHL Great! Thanks for the PR, merging asap.

PR #2837, opened 2019-07-07
Harden qpdfview.profile with `nodbus`

Edu4rdSHL:


enilfodne:
Not sure if its on my machine only, but `qpdfview` does indeed work without d-bus, but only when launched directly. When launched from `Nautilus`, its crashing due to `d-bus` being unavailable.

rusty-snake:
@enilfodne has you firejailed nautilus? Is there any error message or does it just crash?

PR #2843, opened 2019-07-09
Add OWD and ARGV0 environment variables. Correctly create APPIMAGE envvar.

### LINKED ISSUE #2842: Appimage support does not send all appimage environment variables to AppRun
According to [AppImage documentation](https://docs.appimage.org/packaging-guide/environment-variables.html) `OWD` and `ARGV0` environment variables should be set for AppImage type 1 and type 2 appimages respectively.  It looks like `ARGV0` was added to the type 2 spec since #560. 

The lack of and `ARGV0` envvar is definitely affecting recent [Inkscape appimages](https://gitlab.com/inkscape/inkscape/-/jobs/247650934/artifacts/download). We should set both for AppRun.

crass:
This fixes #2842 and can be verified by testing with the Inkscape appimage linked there.  I'll merge in a few days if there are no objections.

### DIFF
--- a/src/firejail/appimage.c
+++ b/src/firejail/appimage.c
@@ -126,16 +126,25 @@ void appimage_set(const char *appimage) {
 		printf("appimage mounted on %s\n", mntdir);
 	EUID_USER();
 
+	char* abspath = realpath(appimage, NULL);
+	if (abspath == NULL)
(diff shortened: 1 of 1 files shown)


netblue30:
No objections, just put them in!

PR #2844, opened 2019-07-09
Fix issue #561. Refactor/Optimize code to get and use pid and process name.

### LINKED ISSUE #561: --trace does not work with --appimage
`--trace` does not seem to work with `--appimage`:

```
me@host:~$ firejail --trace --appimage /isodevice/Applications/Inkscape-0.91+devel.r14861.glibc2.14-x86_64.AppImage 
Reading profile /usr/local/etc/firejail/default.profile
Reading profile /usr/local/etc/firejail/disable-common.inc
Reading profile /usr/local/etc/firejail/disable-programs.inc
Reading profile /usr/local/etc/firejail/disable-passwdmgr.inc

Parent pid 5670, child pid 5673

Child process initialized
2:bash:open /dev/tty:3
2:AppRun:fopen64 inkscape.desktop:0x210c020
2:inkscape.wrappe:open /dev/tty:3
2:inkscape.wrappe:open ./bin//inkscape.wrapper:3
/tmp/firejail-mnt-UQO1O5/usr/bin/inkscape.wrapper
6:sed:fopen64 /proc/filesystems:0x15df010
6:sed:open //lib/charset.alias:-1
/tmp/firejail-mnt-UQO1O5/usr/bin/inkscape not found

```

Without trace it runs fine.

crass:
This fixes #561, which is misnamed.  This issue affects just about any non-trivial shell scripts under firejail.  Its just more obvious with appimage because the vast majority of AppRun programs are shell scripts.

The issue is that the preloaded libtrace library writes log messages to stdout and not directly to the console/tty.  So a bash construct like `FILEPATH=$(readlink -a file)`, will contain the output of `readlink` but also trace log messages for exec, stat, etc.  Thus `FILEPATH` will not just be a file path.  And the process tree then opens `FILEPATH`, which its highly likely to do, a process will try to open with a string that is not a path.

This change has log messages written to `/dev/tty`, which is the tty firejail is run with and the one you'd expect to get output on.  Also, I refactored the existing code to be a tad more efficient by setting global variables for pid and process name in the library constructors.

### DIFF
--- a/src/libtrace/libtrace.c
+++ b/src/libtrace/libtrace.c
@@ -41,54 +41,50 @@ typedef FILE *(*orig_fopen64_t)(const char *pathname, const char *mode);
 static orig_fopen64_t orig_fopen64 = NULL;
 
 //
-// pid
+// library constructor/destructor
 //
+static FILE *ftty = NULL;
 static pid_t mypid = 0;
-static inline pid_t pid(void) {
-	if (!mypid)
-		mypid = getpid();
-	return mypid;
-}
-
-//
-// process name
-//
 #define MAXNAME 16
-static char myname[MAXNAME];
-static int nameinit = 0;
-static char *name(void) {
-	if (!nameinit) {
-		// initialize the name of the process based on /proc/PID/comm
-		memset(myname, 0, MAXNAME);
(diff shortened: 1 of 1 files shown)


netblue30:
All in!

PR #2845, opened 2019-07-09
improve support for home directories outside /home

smitsohu:
whitelisting is still half broken in some cases (when home directory is in one of the other allowed top level directories), but the workarounds are getting easier :)


netblue30:
Yes, let's go for it!

PR #2851, opened 2019-07-09
Update libpostexecseccomp.c

disconnect3d:
Removes `if (size != 0)` condition, which is always true as there is a `if (size <= 0)` condition before.

Also note that if the `if (size <= 0)` condition wouldn't be there and `size` would be 0, there would have been an undefined behavior in due to division by zero in `(unsigned short) size / (unsigned short) sizeof(struct sock_filter);`.

Found with LGTM: https://lgtm.com/projects/g/netblue30/firejail/snapshot/961c4ca00425b60a7bc8543460031a8ebf3d8aa6/files/src/libpostexecseccomp/libpostexecseccomp.c#x838c24f710410160:1

### DIFF
--- a/src/libpostexecseccomp/libpostexecseccomp.c
+++ b/src/libpostexecseccomp/libpostexecseccomp.c
@@ -40,9 +40,7 @@ static void load_seccomp(void) {
 		return;
 	}
 	unsigned short entries = (unsigned short) size / (unsigned short) sizeof(struct sock_filter);
-	struct sock_filter *filter = MAP_FAILED;
-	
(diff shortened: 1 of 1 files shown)


topimiettinen:
Since `filter` is now unconditionally assigned the return value of `mmap()`, its initialization above can be removed.

disconnect3d:
@topimiettinen True, fixed :).

reinerh:
@disconnect3d Thanks!

The link doesn't work btw. ("Forbidden").

Fred-Barclay:
@netblue30 and others, since LGTM has their own CI, would it make sense to enable it for firejail? 
https://github.com/apps/lgtm-com
https://lgtm.com/#pull_requests

PR #2856, opened 2019-07-12
Make 'allow-debuggers' configurable in profiles

DropNib:
When debugging applications inside an IDE regularly, always passing the command line option `--allow-debuggers` becomes cumbersome.

This fix adds a new option `allow-debuggers` to the profile code, so if debugging capabilities are always needed, the profile of the application can simply be extended with that option.

### DIFF
--- a/src/firejail/profile.c
+++ b/src/firejail/profile.c
@@ -941,6 +941,11 @@ int profile_check_line(char *ptr, int lineno, const char *fname) {
 		return 0;
 	}
 
+	if (strcmp(ptr, "allow-debuggers") == 0) {
+		arg_allow_debuggers = 1;
+		return 0;
+	}
+
 	if (strcmp(ptr, "x11 none") == 0) {
 		arg_x11_block = 1;
 		return 0;


glitsj16:
> When debugging applications inside an IDE regularly, always passing the command line option --allow-debuggers becomes cumbersome.

You can put  `allow-debuggers` in your `globals.local` if you need/want it enabled all the time. Is that not working for you? I don't have anything against this PR, but IMHO existing functionality is already in place to cover such use case.

DropNib:
@glitsj16 I'm sorry, I probably didn't provide proper clarification. The issue is, that `allow-debuggers` was for some reason (on purpose or not, I don't know) never added as a profile option, only as a command-line option. So if I add it to `globals.local`, all I get is: `Error: line 1 in /usr/local/etc/firejail/globals.local is invalid`.

That's what my code aims to fix - adding the possibility to use allow-debuggers in profiles.

Besides that, it's not only about always using `allow-debuggers` for _every single_ application but rather about always using it for _some specific_ applications.
For instance: I always need debugging capabilities for Visual Studio Code, but for every other application (Firefox, Evolution, gedit, etc) I don't want debugging enabled. So I would add the line `allow-debuggers` to my `code.local` and be done.

glitsj16:
@DropNib Thanks for your clarification. Merged.

SkewedZeppelin:
firejail-profile man page will need to be updated

DropNib:
> firejail-profile man page will need to be updated

@SkewedZeppelin thanks for the info - opened a PR for this in #2861

PR #2861, opened 2019-07-17
document profile support for allow-debuggers in firejail-profile man page

DropNib:
Adds documentation to firejail-profile man page for the new `allow-debuggers` profile option from #2856 as discussed by @SkewedZeppelin  in https://github.com/netblue30/firejail/pull/2856#issuecomment-511560561.


rusty-snake:
Merged, Thanks.

PR #2870, opened 2019-07-25
Adding sort.py to .travis.yml

rusty-snake:


rusty-snake:
droping because trusty doesn't have `python >= 3.6`

Fred-Barclay:
@rusty-snake how about adding this to .gitlab-ci.yml instead? It should work just fine under all sections except `build_debian_package` and `build_redhat_package`

rusty-snake:
Sounfs good, If you would do that, I'm new to .travis.yml syntax right now and completely unskilled with gitlab-CI.

PR #2871, opened 2019-07-25
Create rsync.profile

rusty-snake:


etc/rsync.profile
```diff
@@ -0,0 +1,54 @@
...
+# Persistent local customizations
+include rsync.local
+# Persistent global definitions
+include globals.local
+
+# This profile is focused on using rsync as an client for downlaoding, it is not writen for using rsync
+#  - as an daemon (rsync --daemon)
+#  - creating backup of $HOME or a full system
+
+blacklist /tmp/.X11-unix
```
Vincent43: maybe `x11 none`?

rusty-snake: since `x11 none` needs `net none` to work it isn't possible here, but thanks for keeping an eye on that.

etc/rsync.profile
```diff
@@ -0,0 +1,54 @@
+# Firejail profile for rsync
+# Description: a fast, versatile, remote (and local) file-copying tool
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include rsync.local
+# Persistent global definitions
+include globals.local
+
+# This profile is focused on using rsync as an client for downlaoding, it is not writen for using rsync
```
Fred-Barclay: "downloading" :wink:

etc/rsync.profile
```diff
@@ -0,0 +1,54 @@
...
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+# Uncomment or add to rsync.local to enable extra hardening
+#whitelist ${DOWNLOADS}
```
Fred-Barclay: TBH I'm not sure why only Downloads would need to be whitelisted... but then I pretty much only use rsync for backing up my /home so perhaps I'm just not being imaginative enough :smile:

rusty-snake: The profile should be used for download (as mentioned above).  If you wan't to do backup and we include disable-*.inc there is an issue.

Fred-Barclay: @rusty-snake could you elaborate on the issue when doing backups?

rusty-snake: If you want backups from ...
 * various configurations and `include disable-programms.inc` is there
 * your documents and `include disable-xdg.inc` is there
 * your firejail profile customisations, even with `--noprofile`
...
=> If you want backups of your full ${HOME} and not only a few files you may want to omit firejail.

Fred-Barclay: @rusty-snake IMO it's probably not a great idea to include the backup-breaking options then. I don't know how common it is to use rsync directly for backups (I *assumed* it was common just because that's the way I use it!).

However, Linux Mint by default (if I understand it correctly) heavily encourages using TimeShift for backups before updates -- I believe their update manager may create the backups automatically (via timeshift) but I could be wrong there. Anyhow that's one major distro, at least, that uses timeshift heavily. For everyone not on BTRFS, then timeshift uses rsync for the backup.

TimeShift normally wouldn't include $HOME in the backup but it's easily configured to do so. Any Mint user who uses timeshift (many/most?) without btrfs (many/most?) and who also backup their $HOME with timeshift (I expect a good-sized minority) will suddenly have incomplete or failing backups.

TL,DR: If we commit an rsync profile that makes backups harder, I expect we'll see a lot of unhappy Mint users. :wink: Personally I think we shouldn't do this.

SkewedZeppelin:
The profile itself is fine for what you made it, but rsync is such a versatile tool. I don't want to see someone enabling it, and potentially losing files.

Maybe move it? like rsync-download_only.profile or something? to ensure a user has to explicitly want it

rusty-snake:
@Fred-Barclay @SkewedZeppelin rsync can do a lot so I didn't want to add it to firecfg. Besides using it as a daemon what probably the least users will do, you can use it for copying files from your own computer where firejail will be a barrier or for downloading where firejail should be no problem but IMHO rsync should be in a sandbox.

Moving it to `rsync-download_only.profile` sounds good.

PR #2879, opened 2019-07-30
qpdfview: Fix issue when opening a file from file manager

Edu4rdSHL:
I can confirm https://github.com/netblue30/firejail/pull/2837#issuecomment-511334363 when opening a file from `pcmanfm`, it doesn't open if qpdfview contains `nodbus`

### DIFF
--- a/etc/qpdfview.profile
+++ b/etc/qpdfview.profile
@@ -22,7 +22,6 @@ include whitelist-var-common.inc
 
 caps.drop all
 machine-id
-nodbus
 nodvd
 nogroups
 nonewprivs


SkewedZeppelin:
To confirm the file manager is or isn't firejailed?

Edu4rdSHL:
Is not.

SkewedZeppelin:
Does qpdfview have `DBusActivatable=true` in its .desktop?

Edu4rdSHL:
@SkewedZeppelin I'm not using Desktop files, launching it directly from terminal causes the issue.

rusty-snake:
@SkewedZeppelin Are you now fine with merging?

smitsohu:
How about leaving a comment in the profile, detailing the condition when the app is broken?

Edu4rdSHL:
I don't see that as a good option because most users just want to click over a file and open it but the profile broke that way. Or if we add a comment it should be:

```
# If you want to open files from file manager, add a qpdfview.local profile with `ignore nodbus`
```

smitsohu:
Maybe we could comment it out and leave a hint why it is disabled? It is a bit of a pitty having to remove or comment out `nodbus`, but I can reproduce the issue and I see that not being able to use a pdf viewer from a file manager is annoying.

veloute:
noticed the same issue; can confirm it's nodbus.

smitsohu:
Merged, thanks!

PR #2881, opened 2019-07-31
Add Zulip profile

flacks:


etc/zulip.profile
```diff
@@ -0,0 +1,46 @@
...
+# Description: Real-time team chat based on the email threading model
+# This file is overwritten after every install/update
+# Persistent local customizations
+include zulip.local
+# Persistent global definitions
+include globals.local
+
+ignore noexec /tmp
+
+noblacklist ${HOME}/.config/Zulip
```
rusty-snake: Add `blacklist ${HOME}/.config/Zulip` to disable-programs.inc (file is alphabetically sorted)

etc/zulip.profile
```diff
@@ -0,0 +1,46 @@
...
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
```
rusty-snake: What about `include disable-xdg.inc`?

etc/zulip.profile
```diff
@@ -0,0 +1,46 @@
...
+nodvd
+nogroups
+nonewprivs
+noroot
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
```
rusty-snake: What about `tracelog`?

etc/zulip.profile
```diff
@@ -0,0 +1,46 @@
...
+
+disable-mnt
+private-bin locale,zulip
+private-cache
+private-dev
+private-etc asound.conf,fonts,machine-id
```
rusty-snake: Network files like `ca-certificates,ssl,pki,crypto-policies...` not needed?

flacks: Zulip connects to a server just fine without those :man_shrugging:

etc/zulip.profile
flacks: Program doesn't seem to run when tracelog is added

etc/zulip.profile
flacks: Added with no problems; see corrections commit

etc/zulip.profile
flacks: Added; thank you, and sorry, it's been a while since I hacked on firejail :smile:

flacks:
Added zulip to src/firecfg/firecfg.config. See corrections commit.

etc/zulip.profile
rusty-snake: then leave it for now

flacks: as is, you mean?

rusty-snake:
Merged, Thanks.

PR #2883, opened 2019-08-01
Add Whalebird profile

flacks:
Program tested to open, connect to a TLS server, open dialog boxes (right-click) and save files using GTK's file chooser, and play sound using PulseAudio.

### DIFF
--- a/etc/whalebird.profile
+++ b/etc/whalebird.profile
@@ -1,0 +1,45 @@
+# Firejail profile for whalebird
+# Description: Electron-based Mastodon/Pleroma client
+# This file is overwritten after every install/update
+# Persistent local customizations
+include whalebird.local
+# Persistent global definitions
+include globals.local
+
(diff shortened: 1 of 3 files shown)


flacks:
Adding `tracelog` results in the program hanging / not opening.

etc/whalebird.profile
```diff
@@ -0,0 +1,45 @@
...
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+
+disable-mnt
+private-bin whalebird
+private-cache
+private-dev
+private-etc fonts,machine-id
```
SkewedZeppelin: add ca-certificates,crypto-policies,pki,ssl ?

flacks: The program doesn't seem to need them to do TLS successfully. Are they really required?

SkewedZeppelin:
Merged, thanks!

PR #2885, opened 2019-08-01
Add new Tor Browser alias

flacks:
- tor-browser in the AUR is an international package; all other
individual language variants have been removed, so, add new alias
- Add 'tor-browser' and 'mv' to private-bin in launcher profile ('mv' is
required when upgrading tor-browser versions)
- Add 'tor-browser' to firecfg.config
- Add config dir to disable-programs.inc


SkewedZeppelin:
Merged, thanks!

Issue #2106, opened 2018-09-03
OpenGL on Radeon needs /usr/lib/llvm/, which is blacklisted in disable-devel.inc


creideiki:
After switching from a GeForce to a Radeon GPU (using the open source drivers in Mesa), I no longer got hardware acceleration in Firefox.

Terminal output when starting Firefox:
```
~ $ firejail firefox
Reading profile /etc/firejail/firefox.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-common.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Warning: noroot option is not available
Parent pid 29820, child pid 29821
Blacklist violations are logged to syslog
Child process initialized in 64.10 ms
libGL error: unable to load driver: radeonsi_dri.so
libGL error: driver pointer missing
libGL error: failed to load driver: radeonsi
libGL error: unable to load driver: radeonsi_dri.so
libGL error: driver pointer missing
libGL error: failed to load driver: radeonsi
libGL error: unable to load driver: swrast_dri.so
libGL error: failed to load driver: swrast
JavaScript error: jar:file:///usr/lib64/firefox/omni.ja!/components/XULStore.js, line 65: Error: Can't 
find profile directory.
Crash Annotation GraphicsCriticalError: |[0][GFX1-]: [OPENGL] Failed to init compositor with reason: 
FEATURE_FAILURE_OPENGL_CREATE_CONTEXT (t=0.684968) [GFX1-]: [OPENGL] Failed to init compositor with 
reason: FEATURE_FAILURE_OPENGL_CREATE_CONTEXT
```
`strace`:ing the process pointed to a failure to access LLVM:
```
[pid 29900] openat(AT_FDCWD, "/usr/lib/llvm/6/lib64/libLLVMAMDGPUCodeGen.so.6", O_RDONLY|O_CLOEXEC) = 
-1 EACCES (Permission denied)
[pid 29900] openat(AT_FDCWD, "/lib64/libLLVMAMDGPUCodeGen.so.6", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No 
such file or directory)
[pid 29900] openat(AT_FDCWD, "/usr/lib64/libLLVMAMDGPUCodeGen.so.6", O_RDONLY|O_CLOEXEC) = -1 ENOENT 
(No such file or directory)
[...]

Issue #1680, opened 2017-12-16
Opengl error with wine.profile


Q3CPMA:
Hello,
after some updates (mesa-17.3 and llvm-4, I think), wine through firejail or anything GL related doesn't work anymore. Here's a simple example with and without firejail
```
> LIBGL_DEBUG=verbose firejail --profile=/etc/firejail/wine.profile glxgears
Reading profile /etc/firejail/wine.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-programs.inc
Parent pid 32211, child pid 32212
Child process initialized in 38.49 ms
libGL: Can't open configuration file /home/user/.drirc: No such file or directory.
libGL: pci id for fd 4: 1002:67df, driver radeonsi
libGL: OpenDriver: trying /usr/lib64/dri/tls/radeonsi_dri.so
libGL: OpenDriver: trying /usr/lib64/dri/radeonsi_dri.so
libGL: dlopen /usr/lib64/dri/radeonsi_dri.so failed (libLLVMX86Disassembler.so.4: cannot open shared object file: No such file or directory)
libGL error: unable to load driver: radeonsi_dri.so
libGL error: driver pointer missing
libGL error: failed to load driver: radeonsi
libGL: pci id for fd 4: 1002:67df, driver radeonsi
libGL: OpenDriver: trying /usr/lib64/dri/tls/radeonsi_dri.so
libGL: OpenDriver: trying /usr/lib64/dri/radeonsi_dri.so
libGL: dlopen /usr/lib64/dri/radeonsi_dri.so failed (libLLVMX86Disassembler.so.4: cannot open shared object file: No such file or directory)
libGL error: unable to load driver: radeonsi_dri.so
libGL error: driver pointer missing
libGL error: failed to load driver: radeonsi
libGL: OpenDriver: trying /usr/lib64/dri/tls/swrast_dri.so
libGL: OpenDriver: trying /usr/lib64/dri/swrast_dri.so
libGL: dlopen /usr/lib64/dri/swrast_dri.so failed (libLLVMX86Disassembler.so.4: cannot open shared object file: No such file or directory)
libGL error: unable to load driver: swrast_dri.so
libGL error: failed to load driver: swrast
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  154 (GLX)
  Minor opcode of failed request:  3 (X_GLXCreateContext)
[...]

PR #2886, opened 2019-08-01
noblacklist but no blacklist

rusty-snake:
Fix `noblacklist SOME` in profiles, where never is a `blacklist SOME`.

Program: https://gist.github.com/rusty-snake/9035b6d9330968322356918a7834eb86

### DIFF
--- a/etc/atom.profile
+++ b/etc/atom.profile
@@ -19,7 +19,6 @@ noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.python-history
 noblacklist ${HOME}/.python_history
 noblacklist ${HOME}/.pythonhist
-noblacklist ${HOME}/.pythonrc.py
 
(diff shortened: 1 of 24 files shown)


rusty-snake:
`blacklist /usr/lib/llvm*` in disable-devel.inc is comented (#2106) should we uncomment it or remove the noblacklists for the profiles?

smitsohu:
> blacklist /usr/lib/llvm* in disable-devel.inc is comented should we uncomment it or remove noblacklist /usr/lib/llvm* from the profiles?

Distilling previous conversations on this topic (#2106, also #1680), `blacklist /usr/lib/llvm` seems to  break things quite universally.
I'm not sure if the benefit is worth the trouble, when it's always `no3d` what users really want. Maybe this blacklist should be added to the `no3d` option?

PR #2898, opened 2019-08-11
added 'noblacklist ${PICTURES}' to mpv.profile

veloute:
${PICTURES} should be accessible to mpv.

### DIFF
--- a/etc/mpv.profile
+++ b/etc/mpv.profile
@@ -16,7 +16,7 @@ include allow-python2.inc
 include allow-python3.inc
 
 noblacklist ${MUSIC}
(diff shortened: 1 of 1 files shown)


etc/mpv.profile
```diff
@@ -16,6 +16,7 @@ include allow-python2.inc
 include allow-python3.inc
 
 noblacklist ${MUSIC}
+moblacklist ${PICTURES}
```
glitsj16: Typo: **m**oblacklist

veloute: ...how did i miss that?

PR #2900, opened 2019-08-11
Fix QOwnNotes path

fandaa:
Every time QOwnNotes was started, setup Wizard was shown. Log shown:
`Warning: cannot create xxx.config directory`


Vincent43:
Thx

PR #2907, opened 2019-08-13
Fix quiet option in archiver redirect profiles

### LINKED ISSUE #2905: Potential leakage in quiet option
Commit https://github.com/netblue30/firejail/commit/7b37c90240d760d2ce6a73e11f3e9a12edeb5955 triggered my curiosity because it introduced the `quiet` option twice in gzip archiver redirect profiles and I assumed it might break stuff. But to my surprise it didn't. After some more testing, this is what I'm observing with firejail from git master.

Placement is (still) important and might leak output:

- comment `quiet` in bunzip2.profile (included gzip.profile still has it)
```
$ time -p -f '%e' firejail /usr/bin/bunzip2 test.bz2
Reading profile /etc/firejail/bunzip2.profile
0.68
```
- uncomment `quiet` in bunzip2.profile (included gzip.profile still has it)
```
$ time -p -f '%e' firejail /usr/bin/bunzip2 test.bz2
0.52
```

This validates the double include of the `quiet` option in https://github.com/netblue30/firejail/commit/7b37c90240d760d2ce6a73e11f3e9a12edeb5955, which doesn't seem to have any/much negative impact. It also suggests we could close any potential output leakage of the `quiet` option in other (mainly archiver redirect) profiles in similar manner. Are there any objections, observations, comments on this?

glitsj16:
This fixes #2905.

### DIFF
--- a/etc/7za.profile
+++ b/etc/7za.profile
@@ -1,5 +1,6 @@
 # Firejail profile for 7za
 # This file is overwritten after every install/update
+quiet
 # Persistent local customizations
 include 7za.local
 # Persistent global definitions
--- a/etc/7zr.profile
+++ b/etc/7zr.profile
@@ -1,5 +1,6 @@
 # Firejail profile for 7zr
 # This file is overwritten after every install/update
+quiet
 # Persistent local customizations
 include 7zr.local
 # Persistent global definitions
--- a/etc/acat.profile
+++ b/etc/acat.profile
@@ -1,5 +1,6 @@
 # Firejail profile for acat
 # This file is overwritten after every install/update
+quiet
 # Persistent local customizations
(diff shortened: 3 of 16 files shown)


glitsj16:
Leaving this open for now to gather input on what is proposed here (fix #2905).

To summarize, we need to fix the `quiet` option in current archiver redirect profiles. Especially when profiles use `include foo.local` the quiet option does leak output (see examples in #2905. For profiles that don't use `include foo.local` there's currently no changes needed.

Whether or not we should always or never offer the include option for archiver redirect profiles is another matter...

glitsj16:
Going ahead and merging. We seem to have fixed the `quiet` option in all profiles now.

PR #2913, opened 2019-08-14
Prevent quiet option output leakage

glitsj16:
https://github.com/netblue30/firejail/pull/2907 fixed the `quiet` option in archiver redirect profiles, preventing potential output leakage. This PR does the same for other relevant profiles.

### DIFF
--- a/etc/Xephyr.profile
+++ b/etc/Xephyr.profile
@@ -1,6 +1,7 @@
 # Firejail profile for Xephyr
 # This file is overwritten after every install/update
 # Persistent local customizations
+quiet
 include Xephyr.local
 # Persistent global definitions
 include globals.local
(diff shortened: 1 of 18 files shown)


glitsj16:
I took the liberty of adding `quiet` to:
- all daemon/server profiles that didn't have it (dnscrypt-proxy, dnsmasq, seahorse-daemon, Xephyr, xpra and Xvfb);
- nano.
There are still quite a few discrepancies with regards to when the quiet option is deemed important enough to be added to a profile. E.g. media player 'mpv' has it, while other media players don't. I'm not sure why exactly. I left these out of the PR on purpose, as there haven't been much issue reports related to it, if any. Easyli fixed if needed/wanted in the future.

PR #2915, opened 2019-08-17
tighten private-bin and etc for torbrowser-launcher.profile

corecontingency:
Tor Browser occupies a very interesting space with firejail, as ideally we would not only want to prevent further system compromise, but also to prevent further de-anonymization (beyond your ip address, which I assume is trivially easy to get if Tor Browser is compromised, regardless of firejail).

To aid in this, Tor Browser should not be able to gain access to the name of the user, and the name of the home directory. As such,

`id` can expose user name
`pwd` can expose name of host directory
`readlink` can expose name of host directory
`realpath` can expose name of host directory

should be removed from private-bin.

Ideally, Tor Browser should not be able to access `env` (also can leak user name and name of host directory), but the program will not work if it is removed (cannot access $PATH), so that is a task for another day. However, `getconf` is not necessary, and can leak environment variables, so that has been removed from private-bin.

`hosts` and `hostname` should be removed from private-etc, as it can leak the name of the user.

This change has been tested on Fedora 30/GNOME and Arch Linux/KDE, and works fine on both, using torbrowser-launcher (from repository) and AUR version, respectively. Used firejail compiled from master for testing.

It is possible to remove far more, for instance, on my computer (Arch/KDE) I am running:
```
private-bin bash,cp,dirname,env,expr,file,grep,ln,mkdir,python*,rm,sh,tor-browser,tor-browser-en,torbrowser-launcher,xz
private-etc machine-id
```
However, I decided to err on the side of caution for general use.

### DIFF
--- a/etc/torbrowser-launcher.profile
+++ b/etc/torbrowser-launcher.profile
@@ -48,7 +48,7 @@ shell none
 #tracelog
 
 disable-mnt
-private-bin bash,cat,cp,cut,dirname,env,expr,file,getconf,gpg,grep,gxmessage,id,kdialog,ln,mkdir,mv,pwd,python*,readlink,realpath,rm,sed,sh,tail,tar,tclsh,test,tor-browser,tor-browser-en,torbrowser-launcher,update-desktop-database,xmessage,xz,zenity
+private-bin bash,cat,cp,cut,dirname,env,expr,file,gpg,grep,gxmessage,kdialog,ln,mkdir,mv,python*,rm,sed,sh,tail,tar,tclsh,test,tor-browser,tor-browser-en,torbrowser-launcher,update-desktop-database,xmessage,xz,zenity
 private-dev
-private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,hostname,hosts,ld.so.cache,machine-id,pki,pulse,resolv.conf,ssl
+private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,ld.so.cache,machine-id,pki,pulse,resolv.conf,ssl
 private-tmp


etc/torbrowser-launcher.profile
```diff
@@ -48,7 +48,7 @@ shell none
 #tracelog
 
 disable-mnt
-private-bin bash,cat,cp,cut,dirname,env,expr,file,getconf,gpg,grep,gxmessage,id,kdialog,ln,mkdir,mv,pwd,python*,readlink,realpath,rm,sed,sh,tail,tar,tclsh,test,tor-browser,tor-browser-en,torbrowser-launcher,update-desktop-database,xmessage,xz,zenity
+private-bin bash,cat,cp,cut,dirname,env,expr,file,gpg,grep,gxmessage,kdialog,ln,mkdir,mv,python*,rm,sed,sh,tail,tar,tclsh,test,tor-browser,tor-browser-en,torbrowser-launcher,update-desktop-database,xmessage,xz,zenity
 private-dev
-private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,hostname,hosts,ld.so.cache,machine-id,pki,pulse,resolv.conf,ssl
```
Vincent43: Those were added to fix otherwise broken profile: https://github.com/netblue30/firejail/commit/908736fce96da4c35697bb49d7039c1ee804668a#diff-0443272f6d05e5603c367ad95b8676ea

etc/torbrowser-launcher.profile
```diff
@@ -48,7 +48,7 @@ shell none
 #tracelog
 
 disable-mnt
-private-bin bash,cat,cp,cut,dirname,env,expr,file,getconf,gpg,grep,gxmessage,id,kdialog,ln,mkdir,mv,pwd,python*,readlink,realpath,rm,sed,sh,tail,tar,tclsh,test,tor-browser,tor-browser-en,torbrowser-launcher,update-desktop-database,xmessage,xz,zenity
+private-bin bash,cat,cp,cut,dirname,env,expr,file,gpg,grep,gxmessage,kdialog,ln,mkdir,mv,python*,rm,sed,sh,tail,tar,tclsh,test,tor-browser,tor-browser-en,torbrowser-launcher,update-desktop-database,xmessage,xz,zenity
```
Vincent43: Some of those were added by @pizzadude in initial profile: https://github.com/netblue30/firejail/commit/255b087f869bd8ecb8135c3452a84b262fdfb7ef#diff-0443272f6d05e5603c367ad95b8676ea

Some were added by @SkewedZeppelin : https://github.com/netblue30/firejail/commit/78bb84ddf277dab653a08f97303894e35433402f#diff-0443272f6d05e5603c367ad95b8676eaR35

@SkewedZeppelin does your approval means they are no longer needed?

netblue30:
Works fine for me on Debian 9 and 10 on the archive distributed by Tor. Thanks for the patch!

Vincent43:
@netblue30 could you look at my comments?

netblue30:
Sorry, I've missed the discussion! Add back whatever is needed!

Vincent43:
@netblue30  Possibly all of those are needed so it may need to be fully reverted.

rusty-snake:
FYI: Removing `id` show error messages in the terminal, but nothing is broken.

SkewedZeppelin:
I approved it because I quickly tested it working.

id is only used to check if it is running as root and display a warning
getconf is only used to check if you have the right build for your architecture
readlink/realpath are used to get correct path if launch via symlink and should be kept
pwd is only needed if $PWD isn't set

hosts/hostsname is probably safe to remove

rusty-snake:
I prever to have a more liberal profile which doesn't break anything and can be hardened from the user.

rusty-snake:
`update-desktop-database` can also be removed since it is only needed for `--register-app`.

netblue30:
Question: why do we need python in /bin? This python thing is more dangerous than all the other put together.

rusty-snake:
The torbrowser-launcher is written in python.

corecontingency:
> quickly reviewing what is used by the actual start-tor-browser script included with the bundle:
> 
>     id is only used to check if it is running as root and display a warning
>     getconf is only used to check if you have the right build for your architecture
>     readlink/realpath are used to get correct path if launch via symlink and should be kept
>     pwd is only needed if $PWD isn't set
> 
> hosts/hostname is probably safe to remove

Yeah, this makes sense. Sorry, I didn't actually check how my proposed changes would affect the actual start-tor-browser script! My bad. Just was doing some trial and error to see what stuff would break the program outright.

Vincent43:
> hosts/hostname is probably safe to remove

@SkewedZeppelin  Then it will bypass hosts that were supposed to be blocked on system.

SkewedZeppelin:
@Vincent43 tor already ignores /etc/hosts iirc

Vincent43:
@SkewedZeppelin I couldn't find definitive answer but it seems[ tor uses /etc/hosts for something](https://trac.torproject.org/projects/tor/wiki/doc/TorInChroot#Configurationfiles).

Issue #2809, opened 2019-06-29
firejail hanging when using strace, but ptrace is not allowed


reinerh:
A Debian user [reported the issue](https://bugs.debian.org/931051), that firejail is just hanging, when he was using `strace ls` with a profile that had the ptrace syscall dropped, even though `--allow-debuggers` has been specified.

> /usr/bin/firejail --allow-debuggers --profile=firefox strace /bin/ls

So there seem to be two issues:
 * why does firejail just hang, instead of terminating when the program tries to ptrace something
 * should `--allow-debuggers` specified on the command line not override the dropped syscalls from the profile? According to the manpage this should be the case.

smitsohu:
`firejail --allow-debuggers --seccomp` -> works
`firejail --allow-debuggers --seccomp.drop=ptrace` -> is broken

It seems that strace pauses soon after starting up, before any `ptrace` syscalls are issued:
```
set_robust_list(0x762e4e55aa20, 24) = 0
pause(
```

rusty-snake:
> `allow-debuggers`: Allow tools such as strace and gdb inside the sandbox by  whitelisting system calls ptrace and process_vm_readv.

> `seccomp`: Enable seccomp filter and blacklist the syscalls in the default list (@default).

#### whats wrong?

 - `seccomp` doesn't use thes default list
 - `allow-debuggers` doesn't whitelist syscalls

#### whats realy happening?

`seccomp` blacklist the syscalls in `@default-nodebuggers` unless `allow-debuggers` is given, the it use `@default`

=> `allow-debuggers` switch the list used by `seccomp`

https://github.com/netblue30/firejail/blob/8dfe3eb5656e17af51d700038cb2bc29a9a53a0c/src/fseccomp/seccomp.c#L29-L32

(TODO: update manpage)

#### whats the issue with `seccomp.drop`?

> `seccomp.drop`: Enable seccomp filter, and blacklist the syscalls or the syscall groups specified by the command.

`seccomp.drop` use always the given list of syscalls/groups and no group by default, so `allow-debuggers` can't swtich this. Here `allow-debuggers` would have to work like `noblacklist` in profiles, but this is not implemented.

https://github.com/netblue30/firejail/blob/8dfe3eb5656e17af51d700038cb2bc29a9a53a0c/src/fseccomp/seccomp.c#L72

PR #2919, opened 2019-08-21
Profiles: add I2P

corecontingency:
Added an I2P profile.

This was tested with Arch/KDE, installing I2P using the official distro-independent java installer and  Xubuntu, installing I2P using the official PPA. 

This does not support the AUR packages for I2P, as they install it really funky and require root privileges to run I2P. Didn't see the point of supporting that when there is an official distro-independent java installer, that only requires normal user privileges to run I2P.

Was not able to create a private-bin for I2P, unfortunately. I'm sure it is possible, but it uses a java wrapper shell script to start I2P, and I thought I copied all the necessary binaries to get it running, but no luck. Are there some debugging tools for firejail that simplifies the process of creating a private-bin?

By default, after starting up I2P opens your default browser to the I2P router console at http://127.0.0.1:7657/home. I couldn't get this to work, so currently it just shows some error messages that it can't open the browser. Luckily, this can be turned off in the I2P settings. I don't think it is that much of an inconvenience to manually open your browser and go to http://127.0.0.1:7657/home instead.

### DIFF
--- a/etc/i2prouter.profile
+++ b/etc/i2prouter.profile
@@ -54,7 +54,7 @@ notv
 nou2f
 novideo
 protocol unix,inet,inet6
-seccomp.drop @default-nodebuggers
+seccomp
 shell none
 
 disable-mnt
--- a/etc/i2prouter.profile
+++ b/etc/i2prouter.profile
@@ -1,0 +1,64 @@
+# Firejail profile for I2P
+# Description: A distributed anonymous network
+# This file is overwritten after every install/update
+# Persistent local customizations
+include i2prouter.local
+# Persistent global definitions
+include globals.local
+
+# Notice: default browser will not be able to automatically open, due to sandbox.
+# Auto-opening default browser can be disabled in the I2P router console.
+# This profile will not currently work with any Arch User Repository i2p packages,
+# use the distro-independent official java installer instead
+
+# Only needed if i2prouter binary is not in home directory, ubuntu official ppa package does this
+ignore noexec ${HOME}
+
+noblacklist ${HOME}/.config/i2p
+noblacklist ${HOME}/.i2p
+noblacklist ${HOME}/.local/share/i2p
+noblacklist ${HOME}/i2p
+# Only needed if wrapper is placed in /usr/sbin/, ubuntu official ppa package does this
+noblacklist /usr/sbin
+
+# Allow java (blacklisted by disable-devel.inc)
+include allow-java.inc
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+whitelist ${HOME}/.config/I2P
+whitelist ${HOME}/.i2p
+whitelist ${HOME}/.local/share/I2P
+whitelist ${HOME}/i2p
+# Only needed if wrapper is placed in /usr/sbin/, ubuntu official ppa package does this
+whitelist /usr/sbin/wrapper*
+
+# May break I2P if wrapper is placed in the home directory
+# If using ubuntu official ppa, this should be fine to uncomment, as it puts wrapper in /usr/sbin/
+#apparmor
+caps.drop all
+ipc-namespace
+machine-id
+netfilter
+no3d
+nodvd
+nogroups
+nonewprivs
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp.drop @default-nodebuggers
+shell none
+
+disable-mnt
+private-cache
+private-dev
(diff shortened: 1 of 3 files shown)


etc/i2prouter.profile
```diff
@@ -0,0 +1,64 @@
...
+noblacklist ${HOME}/.config/i2p
+noblacklist ${HOME}/.i2p
+noblacklist ${HOME}/.local/share/i2p
+noblacklist ${HOME}/i2p
+# Only needed if wrapper is placed in /usr/sbin/, ubuntu official ppa package does this
+noblacklist /usr/sbin
+
+# Allow java (blacklisted by disable-devel.inc)
+include allow-java.inc
+include disable-common.inc
```
rusty-snake: empty line
```
include allow-java.inc

include disable-common.inc
```

etc/i2prouter.profile
```diff
@@ -0,0 +1,64 @@
...
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+whitelist ${HOME}/.config/I2P
+whitelist ${HOME}/.i2p
+whitelist ${HOME}/.local/share/I2P
```
rusty-snake: I2P vs. i2p

etc/i2prouter.profile
```diff
@@ -0,0 +1,64 @@
...
+include allow-java.inc
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+whitelist ${HOME}/.config/I2P
```
rusty-snake: mkdir is needed
```
mkdir ${HOME}/.config/I2P
mkdir ${HOME}/.i2p
mkdir ${HOME}/.local/share/I2P
mkdir ${HOME}/i2p
```

etc/i2prouter.profile
```diff
@@ -0,0 +1,64 @@
...
+include i2prouter.local
+# Persistent global definitions
+include globals.local
+
+# Notice: default browser will not be able to automatically open, due to sandbox.
+# Auto-opening default browser can be disabled in the I2P router console.
+# This profile will not currently work with any Arch User Repository i2p packages,
+# use the distro-independent official java installer instead
+
+# Only needed if i2prouter binary is not in home directory, ubuntu official ppa package does this
```
rusty-snake: > Only needed if i2prouter binary is not in home directory 

?? rather ` Only needed if i2prouter binary is in home directory`?

etc/i2prouter.profile
```diff
@@ -0,0 +1,64 @@
...
+no3d
+nodvd
+nogroups
+nonewprivs
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp.drop @default-nodebuggers
```
rusty-snake:
```suggestion
seccomp
```

etc/i2prouter.profile
```diff
@@ -0,0 +1,64 @@
...
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+whitelist ${HOME}/.config/I2P
+whitelist ${HOME}/.i2p
+whitelist ${HOME}/.local/share/I2P
+whitelist ${HOME}/i2p
+# Only needed if wrapper is placed in /usr/sbin/, ubuntu official ppa package does this
+whitelist /usr/sbin/wrapper*
```
rusty-snake: `include whitelist-common.inc`

rusty-snake:
> Are there some debugging tools for firejail that simplifies the process of creating a private-bin?

`firejail --build PROGRAM`
`firejail --trace [OPTIONS] PROGRAM | grep bin`
`strrace PROGRAM | grep -E 'open.+bin'` (or `| less`)

corecontingency:
Thanks! Tried using firejail --build earlier, but didn't get much out of it. Hopefully the other commands will help. Might give private-bin another shot in a couple days, we will see. (I'll open another pull request if I do figure something out.)

Yeah, good catch on that ignore home exec though, got the ubuntu ppa and the java installer mixed up.

SkewedZeppelin:
@corecontingency 
> is it manual? I couldn't find where it is in the repository

its [src/firecfg/firecfg.config](https://github.com/netblue30/firejail/blob/master/src/firecfg/firecfg.config)

rusty-snake:
Merged, Thanks.

rusty-snake:
@corecontingency FYI why to use `seccomp` instead of `seccomp.drop @default-nodebuggers`: https://github.com/netblue30/firejail/issues/2809#issuecomment-523876836

PR #2926, opened 2019-08-25
Allow exceptions to seccomp lists

### LINKED ISSUE #1366: seccomp.keep and seccomp
The [seccomp guide](https://firejail.wordpress.com/documentation-2/seccomp-guide/) is very helpful when creating a list of whitelisted syscalls for a specific program. However, I have a comprehension question regarding `seccomp.keep`:

`seccomp` blacklists the syscalls in the [default list](https://github.com/netblue30/firejail/blob/master/src/fseccomp/seccomp.c).  But when creating a list of whitelisted syscalls it is necessary to add syscalls to `seccomp.keep` which are **not** explicitly blacklisted by `seccomp`. But why?

Wouldn't it be possible to change the behavior of `seccomp.keep` accordingly? The list(s) of whitelisted syscalls wouldn't be that mile-long anymore, and I think that it would be easier to create new profiles.

But perhaps I'm simply missing a decisive argument.

topimiettinen:
Prefix `!` can be used to make exceptions to system call blacklists and
whitelists used by `seccomp`, `seccomp.drop` and `seccomp.keep`.

Closes #1366


topimiettinen:
! is a bit unfriendly with shells.

topimiettinen:
Is `!` the way to go, or how about `~`? Though shells can also expand that if someone has a username which matches a syscall. `-` should be safe (it's not confused with options as there's always prefix like "seccomp=".

rusty-snake:
Or a bit simelar to regexp's `[a-z]` `[^a-z]` `^` can be used.

netblue30:
! is as good as any other character. Nothing seems safe from the shell these days. We already had this problem with numeric syscalls (--seccomp=$83) that need to be escaped in the shell like this:

`````firejail --seccomp=\$83

In case you are in a script, you need to escape it one more time:

`````firejail --seccomp=\\\$83 ...

topimiettinen:
This should be enhanced to allow making exceptions using system call groups: `!@chown`.

rusty-snake:
You often need to only allow a few syscalls, full groups are IMHO unlikely, however. How would (easy example) `seccomp.drop @default-nodebugges,!@ipc` be handel `process_vm_readv`?

PR #2927, opened 2019-08-25
Use new seccomp syntax

rusty-snake:
Use the new seccomp-syntax from #2926

### DIFF
--- a/etc/kmail.profile
+++ b/etc/kmail.profile
@@ -51,7 +51,7 @@ nou2f
 novideo
 protocol unix,inet,inet6,netlink
 # we need to allow chroot, io_getevents, ioprio_set, io_setup, io_submit system calls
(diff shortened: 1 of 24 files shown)


rusty-snake:
I've writen a little, ugly script that helps migrating from `seccomp.drop` to the new exceptions syntax from #2926: https://gitlab.com/snippets/1888887.

etc/kmail.profile
```diff
@@ -51,7 +51,7 @@ nou2f
 novideo
 protocol unix,inet,inet6,netlink
 # we need to allow chroot, io_getevents, ioprio_set, io_setup, io_submit system calls
-seccomp.drop @clock,@cpu-emulation,@debug,@module,@obsolete,@raw-io,@reboot,@resources,@swap,acct,add_key,bpf,fanotify_init,io_cancel,io_destroy,kcmp,keyctl,mount,name_to_handle_at,nfsservctl,ni_syscall,open_by_handle_at,personality,pivot_root,process_vm_readv,ptrace,remap_file_pages,request_key,setdomainname,sethostname,syslog,umount,umount2,userfaultfd,vhangup,vmsplice
+seccomp !chroot,!io_getevents,!io_submit,!io_submit,!ioprio_set
```
smitsohu: you have !io_submit twice, one of which should be !io_setup

rusty-snake: Thx.

PR #2928, opened 2019-08-25
Add further seccomp groups

topimiettinen:
Get further seccomp group definitions from systemd.

### DIFF
--- a/src/fseccomp/syscall.c
+++ b/src/fseccomp/syscall.c
@@ -50,6 +50,99 @@ static const SyscallEntry syslist[] = {
 }; // end of syslist
 
 static const SyscallGroupList sysgroups[] = {
+	{ .name = "@aio", .list =
+#ifdef SYS_io_cancel
+	  "io_cancel,"
+#endif
+#ifdef SYS_io_destroy
+	  "io_destroy,"
+#endif
+#ifdef SYS_io_getevents
+	  "io_getevents,"
+#endif
+#ifdef SYS_io_pgetevents
+	  "io_pgetevents,"
+#endif
+#ifdef SYS_io_setup
+	  "io_setup,"
+#endif
+#ifdef SYS_io_submit
+	  "io_submit"
+#endif
+	},
+	{ .name = "@basic-io", .list =
+#ifdef SYS__llseek
+	  "_llseek,"
+#endif
+#ifdef SYS_close
+	  "close,"
+#endif
+#ifdef SYS_dup
+	  "dup,"
+#endif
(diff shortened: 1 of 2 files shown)


src/fseccomp/syscall.c
```diff
@@ -190,6 +283,382 @@ static const SyscallGroupList sysgroups[] = {
...
+#ifdef SYS_select
+	  "select"
+#endif
+	},
+	{ .name = "@ipc", .list =
+#ifdef SYS_ipc
+	  "ipc,"
+#endif
+#ifdef SYS_memfd_create
+	  "memfd_create,"
```
rusty-snake: also blocked from `memory-deny-write-execute`

rusty-snake:
IDK if it has any effect, but the last syscall in some groups has a trailing comma and in some there is no.

src/fseccomp/syscall.c
topimiettinen: Yes, but I don't think that's so much of a problem, especially with the exceptions in place.

topimiettinen:
@rusty-snake Thanks for the review. I only found (and fixed) one instance of trailing comma, did you find more?

rusty-snake:
I found this while generating new lists for syscalls.txt, `@default` has it also (because mincore is commented). I know that this is not from this PR.

rusty-snake:
@topimiettinen Beside that I'm asking me for what `@system-service` is good and what it means. I'm wondering why it include `@chown` and `@resources` twices.
```
chown<-privileged
privileged,resources<-default
default<-system-services
```

topimiettinen:
@rusty-snake `@system-service` can be used as a default whitelist for systemd services with typical system calls allowed. For firejail, which has strong focus for user apps instead of system services, it's probably not so interesting unless someone would like to use firejail instead of systemd for a service, or otherwise use `@system-service` as a starting point for constructing filters.

rusty-snake:
From the point of view it makes sense, using for allow and not to deny. Thx.

But what is with the double includes?

topimiettinen:
Including `@chown` in `@privileged` makes sense because these system calls are indeed privileged. But indeed `@default` and `@system-service` could be optimized by dropping the redundant groups. 

Systemd's `@default` is used differently to Firejail's `@default`, that's why they do not match and why there could be these accidental overlaps.

rusty-snake:
If `@system-service` is primary for whitelisting, and `@default` primary for blacklisting, it's a bit problematic to include `@default` in `@system-service`, because `@default` includes also `@obsolete`.

PR #2935, opened 2019-08-30
Fix profile builder

crass:
@netblue30 
Per @smitsohu's suggestion, I've fixed the `--build` option to work with fix for `--trace`.  This actually was a little more involved because the builder parses the trace output to build an rough profile.  Since `--trace` outputs to console now and not `stdout`, the output would be harder to capture.  So I've allowed `--trace` to accept an optional parameter which is a path to the trace log file.  Now the builder logs to a logfile and parses that log file. This is actually better than before, where program, firejail debug, and trace output were all combined together.

One part I'm slightly concerned with is that `--output` has code that has a bit of validation on the given path and then pipes `ftee` to capture the output.  Is the usage of `ftee` just to make sure all the output is capture from the beginning? It the purpose of the path validation just for better error messaging? Why are hardlinks not allowed? The `git log` provides no clues because this code predates the import into this git repo.  I've some some tests to check that symbolic links to files the user doesn't have permission to access are not allowed to be accessed and that the given path is created and opened for writing as an unprivileged user (when run as an unprivileged user).  I'm wondering if I'm missing a security issue?

### DIFF
--- a/src/firejail/fs_trace.c
+++ b/src/firejail/fs_trace.c
@@ -45,14 +45,16 @@ void fs_trace_preload(void) {
 		if (arg_debug)
 			printf("Creating an empty trace log file: %s\n", arg_tracefile);
 		// create a bind mounted trace logfile that the sandbox can see
+		EUID_USER();
 		FILE *fp = fopen(arg_tracefile, "w");
 		if (!fp)
 			errExit("fopen");
 		SET_PERMS_STREAM(fp, firejail_uid, firejail_gid, S_IRUSR | S_IWRITE | S_IRGRP | S_IROTH);
 		fclose(fp);
+		EUID_ROOT();
 		fp = fopen(RUN_TRACE_FILE, "w");
 		if (!fp)
-			errExit("fopen");
+			errExit("fopen " RUN_TRACE_FILE);
 		fclose(fp);
 		fs_logger2("touch ", arg_tracefile);
 		if (mount(arg_tracefile, RUN_TRACE_FILE, NULL, MS_BIND|MS_REC, NULL) < 0)
--- a/src/firejail/fs_trace.c
+++ b/src/firejail/fs_trace.c
@@ -41,6 +41,25 @@ void fs_trace_preload(void) {
 		fclose(fp);
 		fs_logger("touch /etc/ld.so.preload");
 	}
+	if (arg_tracefile) {
+		if (arg_debug)
+			printf("Creating an empty trace log file: %s\n", arg_tracefile);
+		// create a bind mounted trace logfile that the sandbox can see
+		FILE *fp = fopen(arg_tracefile, "w");
+		if (!fp)
+			errExit("fopen");
+		SET_PERMS_STREAM(fp, firejail_uid, firejail_gid, S_IRUSR | S_IWRITE | S_IRGRP | S_IROTH);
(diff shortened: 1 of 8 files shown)


src/firejail/fs_trace.c
```diff
@@ -41,6 +41,27 @@ void fs_trace_preload(void) {
...
+			errExit("fopen");
+		SET_PERMS_STREAM(fp, firejail_uid, firejail_gid, S_IRUSR | S_IWRITE | S_IRGRP | S_IROTH);
+		fclose(fp);
+		EUID_ROOT();
+		fp = fopen(RUN_TRACE_FILE, "w");
+		if (!fp)
+			errExit("fopen " RUN_TRACE_FILE);
+		fclose(fp);
+		fs_logger2("touch ", arg_tracefile);
+		if (mount(arg_tracefile, RUN_TRACE_FILE, NULL, MS_BIND|MS_REC, NULL) < 0)
```
smitsohu: There is a tocttou between the fopen and the actual mount. An evil user might succeed in replacing arg_tracefile just in time with a symbolic link (it would allow him to access files in all directories, no matter if he has search permission on the path or not, because we mount as root and mount follows symbolic links). 

In order to prevent this kind of symlink attacks, one trick is to open the resolved path as user with safe_fd(), run necessary or good-to-have checks on the file descriptor, and then mount using a symbolic link in /proc/self/fd. We are doing something like that already in a number of places, for example check out the x11_xorg() function.

src/firejail/main.c
```diff
@@ -1296,6 +1297,10 @@ int main(int argc, char **argv) {
 		}
 		else if (strcmp(argv[i], "--trace") == 0)
 			arg_trace = 1;
+		else if (strncmp(argv[i], "--trace=", 8) == 0) {
+			arg_trace = 1;
+			arg_tracefile = argv[i] + 8;
```
smitsohu: consider checking for invalid characters

src/libtrace/libtrace.c
```diff
@@ -65,14 +67,24 @@ void init(void) {
...
+			logfile = "/dev/tty";
 #else
-	ftty = stderr;
+			logfile = "/proc/self/fd/2";
 #endif
-	tprintf(ftty, "=== tracelib init() === \n");
+	}
+	
+	// logfile
+	ftty = orig_fopen(logfile, "a");
```
smitsohu: what do we do if ftty is NULL

src/firejail/fs_trace.c
```diff
@@ -41,6 +41,27 @@ void fs_trace_preload(void) {
 		fclose(fp);
 		fs_logger("touch /etc/ld.so.preload");
 	}
+	if (arg_tracefile) {
```
smitsohu: should we break this out into an own function? fs_trace_preload() is called up to three times, whereas this code only needs to run once.

smitsohu:
Of course with this fix it is much better than before.

I'm going to add some adjustments. Apologies for my pedantry ;)

crass:
@smitsohu thanks for your comments. I was planning on making changes based on them, but got busy.  And then my github account got hijacked.  I've just regained control of it several days ago.  It looks like you already fixed the security issue you mentioned, and better than I would've. Thanks!

Issue #2936, opened 2019-08-30
Firefox and Thunderbird profiles broken


curiosity-seeker:
Sorry - I haven't had the time to investigate the details. 

But commenting `seccomp.drop` in `firefox-common.profile` fixed the problem. So it seems that it is caused by the recent seccomp commits.

veloute:
can confirm the recent sccomp commits are causing issues with a lot of other profiles than just firefox and thunderbird.

curiosity-seeker:
Yes, indeed: kate, quiterss, ...

corecontingency:
It is weird, because at least for me it is intermittent. When I first boot up, it doesn't work, however, tor-browser profile works just fine, and then once that is opened, the firefox profile works fine too, even if tor-browser is closed first!

Teamspeak is broken anytime you try to change the options though...

Both problems are fixed by commenting out the seccomp lines.

corecontingency:
Looks like the problem is @resources, if I replace @resources with what is [supposed to be equal to](https://github.com/netblue30/firejail/blob/master/etc/templates/syscalls.txt) (mbind,migrate_pages,move_pages,set_mempolicy), then both firefox and teamspeak work.

curiosity-seeker:
> Looks like the problem is @resources, if I replace @resources with what is [supposed to be equal to](https://github.com/netblue30/firejail/blob/master/etc/templates/syscalls.txt) (mbind,migrate_pages,move_pages,set_mempolicy), then both firefox and teamspeak work.

I don't think that this explains the problem completely. The 2 other examples mentioned earlier (quiterss and kate) only contain the `seccomp` line in their profiles.

corecontingency:
Sorry, don't think I was being clear. Writing `seccomp` in a profile is exactly the same as writing

    seccomp.drop @cpu-emulation,@debug,@obsolete,@privileged,@resources,add_key,fanotify_init,io_cancel,io_destroy,io_getevents,io_setup,io_submit,ioprio_set,kcmp,keyctl,name_to_handle_at,ni_syscall,open_by_handle_at,remap_file_pages,request_key,syslog,umount,userfaultfd,vmsplice

in a profile. So I found that `@resources` is the problem. Writing `mbind,migrate_pages,move_pages,set_mempolicy` is supposed to be the same as writing `@resources`, but it seems that something in the most recent commits messed this up, and replacing `@resources` with the constituent parts fixes the problem.

So replace `seccomp` with 

[...]

PR #2937, opened 2019-08-31
Fix seccomp for firefox, kate, and others

corecontingency:
See discussion at [issues/2936.](https://github.com/netblue30/firejail/issues/2936)

Tested on my computer running Arch/KDE with wayland. No errors regarding `seccomp` present in `sudo journalctl -f` while running each program.

Fixes firefox, thunderbird, kate, teamspeak, quiterss, transmission, ark, and okular.

All of these would not work, just hang without opening, before this change to seccomp.

### DIFF
--- a/etc/feedreader.profile
+++ b/etc/feedreader.profile
@@ -36,7 +36,7 @@ notv
 nou2f
 novideo
 protocol unix,inet,inet6
-seccomp
+seccomp !sched_setscheduler,!sched_setaffinity
 shell none
 
 disable-mnt
--- a/etc/feedreader.profile
+++ b/etc/feedreader.profile
@@ -36,7 +36,7 @@ notv
 nou2f
 novideo
 protocol unix,inet,inet6
-seccomp !sched_setscheduler,!sched_setaffinity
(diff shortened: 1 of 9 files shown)


Vincent43:
I think amount of breaks is too big and it's just tip of the iceberg as nobody tested most of the apps yet. If the new default seccomp filter needs so much exceptions then it's not suitable default and should be reverted.

corecontingency:
At minimum `!sched_setscheduler,!sched_setaffinity` should be removed from `@default`, as that seems to be necessary for most programs to both actually work, and run with no errors to `journalctl`, respectively.

I am going to be leaving this pull request up until `seccomp` is fixed, for reference, in case someone is using a version compiled from github and wants to get some programs working again.

etc/feedreader.profile
```diff
@@ -36,7 +36,7 @@ notv
 nou2f
 novideo
 protocol unix,inet,inet6
-seccomp
+seccomp !sched_setscheduler,!sched_setaffinity
```
rusty-snake: Update: `fchown` is also in journal.

PR #2939, opened 2019-09-03
Revert changes in #2928 to seccomp group @default

topimiettinen:


topimiettinen:
I did not revert changes to `@default` caused by additions to `@privileged` (chown etc.), I think those should be OK.

corecontingency:
> I did not revert changes to @default caused by additions to @privileged (chown etc.), I think those should be OK.

I agree. At least on my system, some of the additions to `@privileged` only broke firefox, but seemed to not cause problems elsewhere, as opposed to the additions to `@resources`. 

I think that if it is only a program or two that breaks, then the syscall should still be in the default, i.e. `chroot`. It's easy now with the new `!` syntax to remove a call for a program that needs it.

I'll retest firefox again after this is merged, to see what things break the program, now that my journal is actually logging properly now. BTW, thank you rusty-snake for your help regarding that. (I disabled logging with a kernel parameter couple of months ago when getting plymouth to work, and forgot about it)

Vincent43:
I disagree. `chroot` is bad example because it broke much more than one or two apps since electron started requiring it. Moreover firejail release cycle takes usually couple of months and in that time users either have broken apps or are forced to use firejail git snapshot which made situation even worse recently. In this circumstances whack a mole approach isn't appropriate and we should make sure 99% apps will safely work with default seccomp filter. We have already two apps which needs `chown` (firefox, feedreader) and it's hard to believe those are only ones among hundreds which we support.

topimiettinen:
Perhaps there should be another default group besides `@default` for apps wanting to execute setuid helpers (`@default-suid`). Or the reverse: `@default-nosuid`. It could be part of bigger knob, like `--disable-suid` which could enable also other restrictions.

rusty-snake:
What about a new tight `@default` and using it for all programs where we (or anywhere else) has tested that it works. For untested profiles we use `seccomp.old` (must be implemented) or `seccomp.drop @old-default`.

That will enable stricter `seccomp` and keeping old profiles working (after one round sed).

topimiettinen:
For maximum compatibility there could be `@default-0.9.60`, `@default-0.9.61`, `@default-latest`. But maybe that's too complex. There will always be some instability because of new system calls added to kernel and libraries switching to use those, so it's not a good idea to guarantee that the lists will not change over time.

rusty-snake:
@topimiettinen chown must also be removed.
@corecontingency here is an example from reality why we best don't change anything when we don't know it works.

I just tried to socialize with gnome-builder again, but all my changes to the code were not saved (and yes I saved). But with `seccomp !chown,!fchown` they were saved. Now consider this in reality: A user install firejail and run firecfg, then he starts gnome-builder and wirtes a lot of code, close gnome-builder and lost work from a few hours.

topimiettinen:
@rusty-snake I see. With this version I think I reverted all changes to `@default` (except adding `idle` syscall to `@obsolete`). Introducing new groups can be another PR.

rusty-snake:
I am now fine with this PR, it can be merged and #2936, #2937 can be closed (but waiting for further feedback). Then we can start implementing one of my ideas (or another), or discuss future changes to seccomp in a new issue.

PR #2960, opened 2019-09-13
Update SkypeForLinux profile for latest version

### LINKED ISSUE #2933: skypeforlinux 8.51.0.86 now requires SYS_ADMIN, SYS_CHROOT capabilities
The newest stable Skype version 8.51.0.86 stopped working in `firejail`.
I got latest profile version from `master`.
It seems that they completely moved to Chome Sandbox, which requires broad set of permissions.

### TL;DR

By experimenting for a while I managed to run Skype with these minimal options:
```sh
firejail --ignore=nonewprivs --ignore=noroot --ignore=protocol --ignore=seccomp --ignore=caps.drop --caps.keep=sys_admin,sys_chroot skypeforlinux
```

### CAP_SYS_CHROOT
The error with `CAP_SYS_CHROOT` disabled:
```
Cannot chroot into /proc/ directory: Operation not permitted
```

### CAP_SYS_ADMIN
The error with `CAP_SYS_ADMIN` disabled:
```
```
The same error is also caused by applying `nonewprivs`, `protocol` and `seccomp` settings. Any Seccomp filter also breaks the app.

### `noroot`

Since the binary `/usr/share/skypeforlinux/chrome-sandbox` is now having setuid bit installed, `noroot` needs to be ignored as well.

vutny:
Fixes #2933

### DIFF
--- a/etc/skypeforlinux.profile
+++ b/etc/skypeforlinux.profile
@@ -16,16 +16,13 @@ include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
+include disable-xdg.inc
 
-caps.drop all
+caps.keep sys_admin,sys_chroot
 netfilter
 nodvd
 nogroups
-nonewprivs
-noroot
 notv
-protocol unix,inet,inet6,netlink
-seccomp !chroot
 shell none
 
 disable-mnt


rusty-snake:
Merged, Thanks.

PR #2962, opened 2019-09-14
"Net None" Option Breaks Functionality

circuitshaker:
In my opinion "Netfilter" option is enough, but "net none" option breaks functionality of marketplace.


Fred-Barclay:
I agree, vs code should be able to install extensions and so on by default.
Merged - thanks!

PR #2964, opened 2019-09-15
Deleted Clamav.profile cause it breaks AV completely

circuitshaker:
This profile breaks ClamAv completely which i can't understand why is there a profile for an AV in the beginning. AV's needs permissions for working stable, if you have problems with AV's because of their *potential* attack surface you shouldn't use it at all.


chiraag-nataraj:
Please open an issue describing the issues you have with the profile. If at possible, we should fix the issues rather than delete the profile entirely. Thanks!

Vincent43:
> if you have problems with AV's because of their potential attack surface you shouldn't use it at all.

I don't follow that logic. The only job of AV is to deal with untrusted input. It's common attack vector to leverage that fact and attack AV itself to gain its system-wide privs. Ideally every AV would sandbox themselves but using 3rd party sandbox from firejail is another option.

chiraag-nataraj:
@circuitshaker I'm going to go ahead and close this pull request. Please open an issue on the bug tracker and we'll take a look at what's going on. Thanks!

PR #2965, opened 2019-09-16
Removed disable-interpreters.inc from w3m.profile

AN3223:
Including disable-interpreters.inc consistently prevents the help screen from loading within w3m.


AN3223:
Okay I included allow-perl.inc and that seems to work fine.

rusty-snake:
Merged, Thanks.

Issue #2860, opened 2019-07-16
seccomp causes steam (and other) games to freeze


LamaRaz:
Nvidia 418.56 Ubuntu 18.04 up-to-date w/ Gnome. Firejail built today from master.

So to track down some issues I ran steam with noprofile and when I added seccomp Proton/Wine games would freeze. Usually 1-2 seconds of intro video (or something else) would play with sound before freezing completely. The freeze is not global so alt tabbing works, mouse is not released though. Ctrl C on the firejail steam terminal window works.

I settled on

--disable-mnt --net=eth0 --defaultgw=192.168.0.1 --dns=1.1.1.1 --private=~/jailed --noroot --nonewprivs --nogroups --noprofile steam

And steam and its proton games run with no issues.

The same is true of Lutris (python wine manager), seccomp causes games to not even start and game launchers to start slowly - the wine virtual desktop starts. In this particular case I also noted that Lutris itself was unable to kill the wine processes it started (it has an option to kill all wine processes), probably related. Running the above options with Lutris solved all problems.

rusty-snake:
@LamaRaz can you try if `firejail --ignore=seccomp steam` also works.

LamaRaz:
only these:
`audit[5477]: SECCOMP auid=1000 uid=1000 gid=1000 ses=2 pid=5477 comm="steam" exe="/home/*/.steam/ubuntu12_32/steam" sig=31 arch=40000003 syscall=26 compat=1 ip=* code=0x0

kernel: audit: type=1326 audit(1563350012.331:81): auid=1000 uid=1000 gid=1000 ses=2 pid=5477 comm="steam" exe="/home/*/.steam/ubuntu12_32/steam" sig=31 arch=40000003 syscall=26 compat=1 ip=* code=0x0

audit[5798]: SECCOMP auid=1000 uid=1000 gid=1000 ses=2 pid=5798 comm="wineserver" exe=2F686F6D652F*******************************************************************************5736572766572 sig=31 arch=c000003e syscall=101 compat=0 ip=* code=0x0

kernel: audit: type=1326 audit(1563350036.867:82): auid=1000 uid=1000 gid=1000 ses=2 pid=5798 comm="wineserver" exe=2F686F6D652F***********************************************************************766572 sig=31 arch=c000003e syscall=101 compat=0 ip=* code=0x0`

rusty-snake:
@LamaRaz can you try `firejail --allow-debuggers steam`

[...]

PR #2971, opened 2019-09-19
Add allow-debuggers to steam.profile

corecontingency:
With the default steam profile, steam runs fine, but several games (all running via proton) do not open or open to a black screen. Checking `journalctl`, each segfaulted with firejail reporting they tried to access syscall 101, which is `ptrace`.

Games: Twilight Struggle, ZeroK, Supreme Commander 2

Adding `seccomp !ptrace` or `allow-debuggers` seemed to fix the problem, and now the games run fine.

I added `allow-debuggers` to the profile, as I think that was the preferred method of allowing `ptrace`, but I can change it to `seccomp !ptrace` instead.

Just checked now, and seems like other people are having this problem: [Issue 2860](https://github.com/netblue30/firejail/issues/2860)

### DIFF
--- a/etc/steam.profile
+++ b/etc/steam.profile
@@ -38,6 +38,8 @@ include disable-programs.inc
 
 include whitelist-var-common.inc
 
+# allow-debuggers needed for running some games with proton
+allow-debuggers
 caps.drop all
 #ipc-namespace
 netfilter


SkewedZeppelin:
fwiw the flatpak does this

rusty-snake:
Merged, Thanks.

PR #2978, opened 2019-09-26
KeePassXC: Added a warning regarding tray icon

tihadot:
nodbus also breaks the (minimize to-) system tray feature of KeePassXC. Users should be aware of that.


rusty-snake:
Merged, Thanks.

PR #2981, opened 2019-09-28
Update evince

### LINKED ISSUE #2980: Evince crashes when 2-page side-by-side is chosen
Nope, not a duplicate of #2621 :smiley:. But same way to crash evince with the difference that now `ignore private-lib` does the trick.

rusty-snake:
closes #2980


rusty-snake:
Merged from commandline.

PR #2982, opened 2019-09-29
Move chroot entirely from path based to file descriptor based mounts

smitsohu:
chroot directory validation was based on a file descriptor logic already, move now on top of that to file descriptor based mounts and resolv.conf copying.

Also collects all code in a new chroot module.


smitsohu:
added a commit to base all checks and all mounts on same file descriptor

netblue30:
Cool!

PR #2972, opened 2019-09-21
whitelist-usr-share-common.inc

rusty-snake:
Any suggestions,  testing, ... welcome :wink: .

Related issue:  #2505

PR #2992, opened 2019-10-05
Wusc fixes

glitsj16:
Some fixes for #2972. @rusty-snake Apologies for not responding earlier to your excellent work with wusc, been a bit hectic on the personal front...

### DIFF
--- a/etc/eom.profile
+++ b/etc/eom.profile
@@ -8,6 +8,9 @@ include globals.local
 
 noblacklist ${HOME}/.config/mate/eom
 
+whitelist /usr/share/eom
+include whitelist-usr-share-common.inc
+
 # private-bin, private-etc and private-lib break 'Open With' / 'Open in file manager'
 # comment those if you need that functionality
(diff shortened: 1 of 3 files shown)


rusty-snake:
Thanks. Maybe we can now move wusc to eo-common.profile since now eog and eom have it.

rusty-snake:
@glitsj16 long time ago that I was using Arch. Is `/usr/share/ca-certificates` only used by Mozilla software or also by other e.g. wget,mpv,gnome-maps ?

glitsj16:
@rusty-snake Good idea to move wusc to eo-common. Regarding `/usr/share/ca-certificates` I can confirm nothing besides Mozilla software uses that, at least not on Arch and Debian/Ubuntu, so no further action needs to be taken for that. During the coming week I'll have some time to test adding wusc to some more profiles. Nice addition :+1:.

netblue30:
Sorry, wrong commit number (2992), it was supposed to be 2922.

PR #3002, opened 2019-10-15
Profiles: add signal-cli profile

tihadot:
signal-cli: https://github.com/AsamK/signal-cli

Tested on Arch (including AUR package) using command line and dbus interface. I've opted for commenting private-etc out to ensure compatibility with all supported Java configurations. I'm still on the fence about that though, since it should work with private-etc in the majority of environments.


tihadot:
(please excuse the messy commit history... I'm still learning the finer details of squashing/rebasing)

etc/signal-cli.profile
```diff
@@ -0,0 +1,49 @@
...
+# Description: signal-cli provides a commandline and dbus interface for signalapp/libsignal-service-java
+# This file is overwritten after every install/update
+# Persistent local customizations
+include signal-cli.local
+# Persistent global definitions
+include globals.local
+
+blacklist /tmp/.X11-unix
+
+noblacklist ${HOME}/.local/share/signal-cli
```
rusty-snake: Add `blacklist ${HOME}/.local/share/signal-cli` to `disable-programs.inc`.

etc/signal-cli.profile
```diff
@@ -0,0 +1,49 @@
...
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
```
rusty-snake: What about `tracelog`?

etc/signal-cli.profile
```diff
@@ -0,0 +1,49 @@
...
+include disable-xdg.inc
+
+mkdir ${HOME}/.local/share/signal-cli
+whitelist ${HOME}/.local/share/signal-cli
+include whitelist-common.inc
+include whitelist-var-common.inc
+
+caps.drop all
+netfilter
+no3d
```
rusty-snake: `nodbus`?

etc/signal-cli.profile
tihadot: Done

etc/signal-cli.profile
tihadot: You're right, missed that one

etc/signal-cli.profile
tihadot: That would break a main feature, the daemon mode via dbus interface

tihadot:
Thank you for your feedback!

etc/signal-cli.profile
rusty-snake: I over read your 
> and dbus interface

 :upside_down_face:

rusty-snake:
Note about `private-etc` if `private-etc java*` would be supported it would work.

rusty-snake:
Merged, Thanks.

PR #3005, opened 2019-10-18
Add wusc to more profiles

glitsj16:
Leaving this open for a while to give reviewers a chance to look at it. Nothing surprising, just expanding the range of profiles that can benefit from wusc.

### DIFF
--- a/etc/QMediathekView.profile
+++ b/etc/QMediathekView.profile
@@ -27,9 +27,6 @@ include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-xdg.inc
 
-whitelist /usr/share/ca-certificates
-whitelist /usr/share/qt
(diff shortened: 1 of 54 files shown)


etc/QMediathekView.profile
```diff
@@ -27,6 +27,11 @@ include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-xdg.inc
 
+whitelist /usr/share/ca-certificates
+whitelist /usr/share/qt
+whitelist /usr/share/qt4
```
rusty-snake: Can be remove. Is in wusc now.

glitsj16:
@rusty-snake I agree, adding ca-certs to wusc makes it a bit more robust/less likely to break. I'll make the needed changes. Thanks for your input.

PR #3010, opened 2019-10-25
Update main.c

xeniter:
Eight byte stack alignment is a requirement of the Procedure Call Standard for the ARM Architecture [AAPCS]. This specifies that functions must maintain an eight-byte aligned stack address (for example: 0x00, 0x08, 0x10, 0x18, 0x20) on all external interfaces

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -55,7 +55,7 @@ uid_t firejail_uid = 0;
 gid_t firejail_gid = 0;
 
 #define STACK_SIZE (1024 * 1024)
-static char child_stack[STACK_SIZE];		// space for child's stack
+static char child_stack[STACK_SIZE] __attribute__((aligned(8)));		// space for child's stack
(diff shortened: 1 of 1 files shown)


xeniter:
without this fix, linux killed firejail cause of unaligment exception trap, cause the stack pointer was not 8 byte aligned:

root@myboard:~# firejail

Parent pid 25756, child pid 257[ 1479.741141] Alignment trap: not handling instruction f9440a1f at [<b6ea4890>]
57
[ 1479.750378] Unhandled fault: alignment exception (0x801) at 0x00579b4c
[ 1479.757192] pgd = 56ee8009
[ 1479.759936] [00579b4c] *pgd=43171835, *pte=4214c75f, *ppte=4214cc7f
Error: proc 25756 cannot sync with peer: unexpected EOF
Peer 25757 unexpectedly killed (Bus error)

netblue30:
Thanks for the fix!

PR #3011, opened 2019-10-25
Add Comment Mentioning that `nodbus` Breaks Native Notifications

julianschacher:
After installing a notification daemon yesterday, I realized that `nodus` in the `firefox-common.profile` also breaks native notifications.
So I extended the comment to mention that, so that users have an easier time discovering this issue in the future.


rusty-snake:
Merged, Thanks.

PR #3019, opened 2019-10-30
Slack profile: use temporary cache

### LINKED ISSUE #3015: Slack Desktop 4.1.1 cannot open external links
Hi, 

I know this is a common issue for many sandboxed applications.
I've got that with Slack, which was unable to open any external URLs. Also it was impossible to sign-in into new workspace when running in the jail.

I had slow disk, so I tried to take advantage of having `~/.cache` directory on temporary file system to reduce disk writes.
And profile option `private-cache` comes in handy.
After adding it to `slack.profile` I've got external links being opened in default desktop browser, to my surprise. I haven't tested the sign-in process.

Not sure how those things are connected, though.
I think we could add it as the default into the profile, because it does no harm and even improves isolation a little bit. The only drawback I observed is slightly increased start-up time for Slack. But that should not be an issue on relatively modern hardware.

vutny:
Fixes #3015

### DIFF
--- a/etc/slack.profile
+++ b/etc/slack.profile
@@ -33,6 +33,7 @@ shell none
 
 disable-mnt
 private-bin locale,slack
+private-cache
 private-dev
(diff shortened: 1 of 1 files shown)


rusty-snake:
Merged, Thanks.

Issue #3030, opened 2019-11-07
Dia not working


matu3ba:
Reading profile /usr/local/etc/firejail/dia.profile
Reading profile /usr/local/etc/firejail/disable-common.inc
Reading profile /usr/local/etc/firejail/disable-devel.inc
Reading profile /usr/local/etc/firejail/disable-exec.inc
Reading profile /usr/local/etc/firejail/disable-interpreters.inc
Reading profile /usr/local/etc/firejail/disable-passwdmgr.inc
Reading profile /usr/local/etc/firejail/disable-programs.inc
Reading profile /usr/local/etc/firejail/disable-xdg.inc
Parent pid 13747, child pid 13748
Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Child process initialized in 326.87 ms
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
ImportError: No module named site

Parent is shutting down, bye...

matu3ba:
firejail --audit --debug
Autoselecting /bin/bash as shell
Building quoted command line: 'dia' 
Command name #dia#
Found dia.profile profile in /usr/local/etc/firejail directory
Reading profile /usr/local/etc/firejail/dia.profile
Found disable-common.inc profile in /usr/local/etc/firejail directory
Reading profile /usr/local/etc/firejail/disable-common.inc
Found disable-devel.inc profile in /usr/local/etc/firejail directory
Reading profile /usr/local/etc/firejail/disable-devel.inc
Found disable-exec.inc profile in /usr/local/etc/firejail directory
Reading profile /usr/local/etc/firejail/disable-exec.inc
Found disable-interpreters.inc profile in /usr/local/etc/firejail directory
Reading profile /usr/local/etc/firejail/disable-interpreters.inc
Found disable-passwdmgr.inc profile in /usr/local/etc/firejail directory
Reading profile /usr/local/etc/firejail/disable-passwdmgr.inc
Found disable-programs.inc profile in /usr/local/etc/firejail directory
Reading profile /usr/local/etc/firejail/disable-programs.inc
Found disable-xdg.inc profile in /usr/local/etc/firejail directory
Reading profile /usr/local/etc/firejail/disable-xdg.inc
DISPLAY=:0.0 parsed as 0
Parent pid 13631, child pid 13632
Initializing child process
PID namespace installed
[...]

PR #3032, opened 2019-11-08
dia apparently wants access to python and crashes without

matu3ba:
Even though it is written in C, maybe another fix would be better


matu3ba:
fixes #3030

netblue30:
I see it also on Debian stable, thanks for the patch!

rusty-snake:
Why disabling `disable-interpreters.inc`, is `include allow-pythonX.inc` not enough?

netblue30:
You are right, I put a new fix in.

PR #3037, opened 2019-11-11
Resolve #3029: drop outdated Skype profile

### LINKED ISSUE #3029: Drop legacy Skype profile
The file `etc/skype.profile` is outdated and not suitable for running current versions of Skype For Linux proprietary software.
Since the package provided by Microsoft sets the binary name to `skypeforlinux`, I think it is sufficient to keep only `skypeforlinux.profile`, which is up to date.

vutny:
This is no longer applicable.


vutny:
Thanks for your guidance, @rusty-snake .
I have removed the profile from `firecfg.config` file. And, indeed, `~/.Skype` directory is obsoleted and no longer in use.

netblue30:
all in, thanks!

PR #3044, opened 2019-11-13
RFC: profiles: allow nc in ssh profile by default

reinerh:
nc is currently disallowed in the ssh profile.
I think it could be allowed by default, as it can not only be used for tor, but has more general uses (see ProxyCommand in ssh_config(5)). And I think it's also not harmful to allow it.

Not blacklisting nc would allow a smoother out-of-the-box experience for tor users that use nc in ProxyCommand.

But I'm interested in other opinions if it's a sensible change, or if the profile should be kept more closed.


netblue30:
Let's pull it in. If anybody has a problem with it, we can remove it later.

PR #3061, opened 2019-11-25
Add new profile: gist

glitsj16:
https://github.com/defunkt/gist


rusty-snake:
@glitsj16 makes it sense to add `blacklist /tmp/.X11-unix`?

glitsj16:
@rusty-snake It makes sense yes, thanks for the hint :+1:

PR #3065, opened 2019-11-26
Minor profile tweaks.

the-antz:
thunderbird-wayland profile did not include thunderbird-wayland.local


smitsohu:
Merged, thanks!

PR #3086, opened 2019-12-17
remove read access from fs log file

smitsohu:
Removing read access from the fs log file should make it slightly more difficult to gather information about the sandbox

### DIFF
--- a/src/firejail/fs_logger.c
+++ b/src/firejail/fs_logger.c
@@ -97,7 +97,7 @@ void fs_logger_print(void) {
 		perror("fopen");
 		return;
 	}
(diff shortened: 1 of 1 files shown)


netblue30:
We are using it for firejail --fs.print=... and it goes into firetools to print the filesystem. So I have to reject it, at least for now.

smitsohu:
@netblue30 I thought it was ok because `--fs.print` runs as root, but I see that it can fail when Firejail itself runs in a restricted environment like inside a container.

However, could you confirm the sandbox pid leak there is necessary? I might be wrong, but it is my impression this information is not consumed currently (except maybe during debugging).

Issue #2944, opened 2019-09-06
Firejail breaks Brave browser default sandboxing


corecontingency:
Running brave with firejail has this error:

    gzip: /proc/config.gz: Permission denied
    User namespaces are not detected as enabled on your system, Brave will run with the sandbox disabled

Brave opens fine, and runs fine, but the sandbox does not work. Image is below.

![brave_no_sandbox_a](https://user-images.githubusercontent.com/54042889/64463847-60521f80-d0ba-11e9-8372-c5be600fd008.png)

It seems like [brave checks](https://github.com/brave/brave-browser/issues/3420) if the kernel has user namespaces enabled by checking `/proc/config.gz`. Unfortunately, this file cannot be read from the firejail sandbox, so brave assumes user namespaces are disabled (even when they are actually enabled), so it runs with the `--no-sandbox` flag.

I believe this issue can be fixed if `/proc/config.gz` is made readable. Unfortunately, even if using a completely blank profile, this file still cannot be read. I have tried passing 

    noblacklist /proc/config.gz
    read-only /proc/config.gz
    whitelist /proc/config.gz

but the first two do nothing, and the last says it is an invalid whitelist, and firejail won't run.

How do I make this file accessible from the sandbox?

rusty-snake:
> Unfortunately, even if using a completely blank profile

A blank profile (aka `--noprofile`) can't help here because the blacklist is hardcoded:
https://github.com/netblue30/firejail/blob/2e64e781bb351448edb6e6f31f7c9085ca580382/src/firejail/fs.c#L636

rusty-snake:
@corecontingency 
still an issue?

corecontingency:
Yes. Just replicated this exact same bug in a Manjaro VM, using the firejail-git and brave-bin packages from the AUR. 

I think that this might be fixed if Brave was able to read `/proc/config.gz` from the sandbox, but I don't know how hard removing the hardcoded blacklist would be. What is the reasoning behind blacklisting `/proc/config.gz`? Could allowing a program to read this file lead to a sandbox escape?

I am not sure how much effort should be put into this, considering chromium already has a sandbox already, and the issues that firejail is having with electron based apps.

glitsj16:
[...]

PR #3087, opened 2019-12-17
Fix Brave's native sandbox

### LINKED ISSUE #2914: Command "firejail --seccomp skypeforlinux" used to work until skype's rpm update to 8.51.0.72-1.x86_64.rpm
Hi, I noticed that I can no longer run skypeforlinux in firejail after update to 8.51.0.72-1.x86_64.rpm
It works properly on 8.50.0.38-1.x86_64.rpm
https://repo.skype.com/rpm/stable/

glitsj16:
**IMPORTANT**: leave this unmerged until collaborators have had ample opportunity to review.

This is a tentative fix for #2944. The patch enables using `noblacklist /proc/config.gz` in `/etc/firejail/brave.profile`. Momentarily this is only needed for that one profile, so I'm not entirely convinced this is the best way to fix #2914. Thoughts?

### DIFF
--- a/etc/brave.profile
+++ b/etc/brave.profile
@@ -20,5 +20,8 @@ whitelist ${HOME}/.config/brave
 whitelist ${HOME}/.config/BraveSoftware
 whitelist ${HOME}/.gnupg
 
+# Brave sandbox needs read access to /proc/config.gz
+noblacklist /proc/config.gz
+
 # Redirect
 include chromium-common.profile
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -452,3 +452,6 @@ blacklist ${HOME}/Mail
 blacklist ${HOME}/mail
 blacklist ${HOME}/postponed
 blacklist ${HOME}/sent
+
+# kernel configuration
+blacklist /proc/config.gz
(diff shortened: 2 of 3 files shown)


Vincent43:
Honestly I don't believe `/proc/config.gz` is so sensitive that it requires special treatment. I would propose dropping it from `fs.c` and adding it to `disable-common.inc` instead.

rusty-snake:
`firejail cat /proc/cmdline` :heavy_check_mark: 
`firejail --blacklist=/proc/cmdline cat /proc/cmdline` :x:

glitsj16:
@Vincent43 I've updated the PR, implementing your suggestions.

glitsj16:
For me this looks fine now. It would be appreciated if any of the reviewers could merge this after checking the updated commits. TIA.

Vincent43:
@glitsj16 looks fine to me however the commits may be cleaned up to not introduce changes that are later reverted. Alternatively the whole thing may be squashed.

glitsj16:
@netblue30 Do we need to backport this to the **release-0.9.62** branch?

rusty-snake:
@glitsj16 since there are some more commits with (important) fixes I opened an issue to discuss this: #3089

PR #3091, opened 2019-12-22
Rework thunderbird.profile

rusty-snake:
- ~remove `include firefox-common.profile`; make standalone profile~
- enable wusc; **need some test users**
- add disable-passwdmgr.inc and disable-xdg.inc, novideo.
- reorder
- improve some other deficit
---
## Needs Testing - Test welcome

**Tested:**

Arch:
   - [ ] thunderbird
   - [ ] thunderbird + enigmail

Debian:
   - [ ] thunderbird
   - [ ] thunderbird + enigmail

Fedora:
   - [x] thunderbird
   - [ ] thunderbird + enigmail

### DIFF
--- a/etc/thunderbird.profile
+++ b/etc/thunderbird.profile
@@ -16,12 +16,13 @@ writable-run-user
 # TODO: we have whitelist-var-common.inc sure that here is no whitelsit requiered
 # writable-var
 
-# Uncomment the next 4 lines or put them in your thunderbird.local to
-# allow Firefox to load your profile when clicking a link in an email
+# Allow Firefox
 #noblacklist ${HOME}/.cache/mozilla
 #noblacklist ${HOME}/.mozilla
 #whitelist ${HOME}/.cache/mozilla/firefox
 #whitelist ${HOME}/.mozilla
+#seccomp !chroot
+#ignore tracelog
 
 # TODO: allow chromium
 
@@ -36,6 +37,7 @@ include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
+include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-xdg.inc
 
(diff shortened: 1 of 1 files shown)


Vincent43:
As TB is strongly based on firefox esr then why shouldn't it use `firefox-common.profile`? Similarly all chromium based apps include `chromium-common.profile` instead of setting same things themselves.

rusty-snake:
There are also differences (xdg-open, enigmail, gnupg, ...). The reason why I decided to remove the include was that nobody has an eye on TB while editing firefox-common.

Vincent43:
The xdg-open/private-tmp issue occurs in firefox as well. The differences like enigmail/gnupg are handled in thunderbird profile right now and more things could be handled there as well. I think splitting profiles is unnecessary duplication at this point as most things will be the same.

rusty-snake:
Ok, will revert this.

Fred-Barclay:
@rusty-snake do you still need testers? I can check with Arch some time over the next few days (maybe Debian too)

rusty-snake:
@Fred-Barclay yes. Currently it is only tested that thunderbird start and the calendar works under Fedora. I does not use enigmail and are not familiarly with it. My worry is that wusc breaks enigmail.

Fred-Barclay:
@rusty-snake I don't really use enigmail either but it seems to work with my limited testing on Arch (prompts me for a key, tries to autoencrypt, etc)...

Everything else seems good. The first attempt it didn't want to save a draft message but I think that may have been a TB bug; it worked on subsequent starts. Did you notice anything like this?

I need to set up a new key anyways so if you're in no rush to merge I can try again in a few days more exhaustively.

Cheers!
Fred

rusty-snake:
I will merge this on Sunday if no one catch me.

smitsohu:
If we put disable-passwdmgr.inc and disable-xdg.inc in the thunderbird.profile, we might as well migrate everything and put them right into firefox-common.profile.

It would need some additional noblacklists in firefox-common-addons.inc to not break the password managers. Does that sound reasonable/are there objections?

Apologies for being late to the party

Issue #2139, opened 2018-10-05
Allow for --join to wait until sandbox is ready instead of dying when its not.


crass:
My specific use case is that I have a script which first creates an X11 sandbox with WM.  Then it tries to execute a process in that sandbox via `--join`.  However, the sandbox isn't fully setup at this point, so firejail exits with the error `Error: no valid sandbox`.  It would be nice if `--join` would wait until the sandbox is setup.

smitsohu:
It's taken a while, but yes! I think we can do it, possibly the timeout can be configurable in /etc/firejail/firejail.config

PR #3093, opened 2019-12-23
add join timeout and make it configurable

smitsohu:
per this patch, firejail tries for 5 seconds to join a sandbox, and only then exits with an error. I'm not sure if the defaults are chosen wisely, you may want to adjust both the default timeout and the frequency of the checks (every 100 msec currently)

see #2139

### DIFF
--- a/src/firejail/join.c
+++ b/src/firejail/join.c
@@ -255,9 +255,10 @@ static void extract_umask(pid_t pid) {
 	fclose(fp);
 }
 
-// return 1 if the sandbox identified by pid is not fully set up yet or if
-// it is no firejail sandbox at all, return 0 if the sandbox is complete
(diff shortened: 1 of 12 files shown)


src/firejail/join.c
```diff
@@ -255,32 +255,114 @@ static void extract_umask(pid_t pid) {
 	fclose(fp);
 }
 
+// return 0 if the sandbox identified by pid is not fully set up yet or if
```
topimiettinen: Could be nicer with 'bool'.

topimiettinen:
Looks good.

smitsohu:
@topimiettinen I will update is_ready_for_join() so it returns a boolean, thanks for the review

netblue30:
All fine with me, so check it in when you are done.

Issue #3095, opened 2019-12-23
Firecfg CLI archivers - missing and/or broken on Arch makepkg


glitsj16:
I noticed that [firecfg](https://github.com/netblue30/firejail/blob/master/src/firecfg/firecfg.config) enables **some** of the available command-line archivers, but not others (see compiled list below). Is there any (official) `policy` to decide which archivers are supported in `firecfg` and which not? If this is simply due to omissions, we need to add the missing ones IMHO.

List of currently missing archiver (redirect) profiles in firecfg:

7z, 7za, 7zr, bsdcat, bsdcpio, bsdtar, cpio, gtar, gunzip, gzexe, gzip, lbunzip2, lbzcat, lbzip2, lzcat, lzcmp, lzdiff, lzegrep, lzfgrep, lzgrep, lzip, lzless, lzma, lzmainfo, lzmore, p7zip, tar, uncompress, unlzma, unrar, unxz, unzip, xz, xzcat, xzcmp, xzdiff, xzegrep, xzfgrep, xzgrep, xzless, xzmore, zcat, zcmp, zdiff, zegrep, zfgrep, zforce, zgrep, zless, zmore, znew

rusty-snake:
Tools like *grep, *diff, *less, *more, .. should be added IMHO (after removing whitelist-var). Tools like xz, lzma, cpio maybe conflict with some packet-managers.

glitsj16:
> Tools like *grep, *diff, *less, *more, .. should be added IMHO (after removing whitelist-var).

I just checked. None of the CLI archivers includes `whitelist-var-common.inc`, so that shouldn't pose any problems.

rusty-snake:
@glitsj16 you right, but maybe some `noblacklist` and `writable-var-log` (and `writable-var`?).

> **--writable-var-log**
Use the real /var/log directory, not a clone. By default, a tmpfs  is
mounted  on  top  of /var/log directory, and a skeleton filesystem is
created based on the original /var/log.

Vincent43:
Yeah, archivers are tricky due to package managers issues. `bzip2` already causes trouble for me in Arch.

glitsj16:
[...]

PR #3097, opened 2019-12-26
Drop CLI archivers from firecfg

glitsj16:
More research/testing is needed to make CLI-based archivers work with makepkg on Arch (based distributions). See ongoing discussion in #3095.


netblue30:
I'll grab it also for 0.9.62 branch, thanks!

Issue #3026, opened 2019-11-04
Configuring network interface with DHCP


kris7t:
I am using the network isolation feature of firejail with a network bridge managed by libvirt (`virbr0`), so I can set up networking for sandboxes similarly to virtual machines. While firejail can auto-assign an IPv4 address by ARP scanning for a free address, there is no such feature for IPv6. Moreover, I would like to avoid clashes between sandboxed and (powered-off) virtual machines, so I would like to let the DHCP server (dnsmasq) managed by libvirt to handle address allocation of my sandboxes. (While I could assign an address manually to each and every sandbox, that would be much more error prone. I prefer to do manual assignments in one place in the libvirt config by MAC address.)

I set up DHCP for IPv4 (from a /24 block) and stateful DHCP6 for IPv6 (from a /112 prefix) in libvirt. I was experimenting with running `dhcpcd` in a `--noprofile` sandbox (in a real setup, an `sbox_run` call during sandbox initialization would probably replace `sudo` and `--noprofile`, but I didn't get that far) as follows:

```
$ firejail --net=virbr0 --ip=none --noprofile --debug bash
Autoselecting /usr/bin/zsh as shell
Building quoted command line: 'bash' 
Command name #bash#
get interface virbr0 configuration
MTU of virbr0 is 1500.
Bridge device virbr0 at 192.168.122.1/24
DISPLAY=:0 parsed as 0
Parent pid 667124, child pid 667125
Initializing child process
sbox run: /usr/lib/firejail/fnet create veth veth667126eth0 eth0 virbr0 667125 (null) 
Set caps filter 3000
Host network configured
PID namespace installed
Mounting tmpfs on /run/firejail/mnt directory
Creating empty /run/firejail/mnt/seccomp directory
Creating empty /run/firejail/mnt/seccomp/seccomp.protocol file
Creating empty /run/firejail/mnt/seccomp/seccomp.postexec file
sbox run: /run/firejail/lib/fnet ifup lo (null) 
Set caps filter 3000
sbox run: /run/firejail/lib/fnet ifup eth0 (null) 
Set caps filter 3000
Warning: cannot configure default route
Network namespace enabled

sbox run: /run/firejail/lib/fnet printif (null) 
Set caps filter 3000
Interface        MAC                IP               Mask             Status
lo                                  127.0.0.1        255.0.0.0        UP    
Default gateway configuration failed

Basic read-only filesystem:
Mounting read-only /etc
Mounting noexec /etc
[...]

PR #3102, opened 2019-12-30
DHCP client support

kris7t:
As per my issue in #3026, I was trying to use DHCP to configure network interfaces in firejail. This patch implements integration with the ISC dhclient for both IPv4 and IPv6.

I left the commits as-is so that they can be reviewed individually, but I can squash them if that is preferred.

* Added the options `--ip=dhcp` and `--ip6=dhcp` for IPv4 and IPv6 configuration by DHCP.
* `--ip=dhcp` is handled similarly to `--ip=none`. In particular, firejail does not do any network configuration on its own, not even setting up routing.
* dhclient is invoked in forking mode. Its main process exits when it successfully acquired a lease, while the daemon keeps running in the background. This ensures that firejail only proceeds when the network is working inside the container (running dhclient as a foreground process and getting notified about leases would be a bit trickier).
* I added a capability filter for `CAP_NET_BIND_SERVICE` so that dhclient can bind to low ports.
* PID files and lease files are written to `/run/firejail/mnt/dhclient`. The PID file is read (hopefully without race conditions) to find the PID of the dhclient daemon process. The sandbox may terminate if only the dhlient daemons are running.
* No effort is made to properly terminate (`-x`) dhlient or release the DHCP lease (`-r`), as neither is required by the DHCP protocol. The dclient processes just die when the sandbox terminates. It would be possible to release the lease properly (this is allegedly required by some ISPs, but not in the more common situation when the sandboxes are connected to a local virtual bridged network, such as libvirt), but would require keeping some privileges until container termination (either in the firejail main process, or an auxiliary process just for communicating with dhclient).
* Especially when used solely for DHCPv6, dhclient may fail to bind to the network interface if it has no link-local IPv6 address (or the LL address is only tentative). When both DHCPv4 and DHCPv6 is in use, the delay caused by waiting for the DHCPv4 lease is virtually always enough for IPv6 LL address to become active. I added a subcommand to `fnet` to wait for IPv6 LL addresses. Unfortunately, I had to use the rather complex rtnetlink interface, because this is the only way to access the tenative flag of an address.

### DIFF
--- a/src/firejail/dhcp.c
+++ b/src/firejail/dhcp.c
@@ -117,6 +117,21 @@ static void dhcp_start_dhclient(const Dhclient *client) {
   *(client->pid) = dhcp_read_pidfile(client);
 }
 
+static void dhcp_waitll(const char *ifname) {
+  sbox_run(SBOX_ROOT | SBOX_CAPS_NETWORK | SBOX_SECCOMP, 3, PATH_FNET, "waitll", ifname);
+}
+
+static void dhcp_waitll_all() {
+  if (cfg.bridge0.arg_ip6_dhcp)
+    dhcp_waitll(cfg.bridge0.devsandbox);
+  if (cfg.bridge1.arg_ip6_dhcp)
+    dhcp_waitll(cfg.bridge1.devsandbox);
+  if (cfg.bridge2.arg_ip6_dhcp)
+    dhcp_waitll(cfg.bridge2.devsandbox);
+  if (cfg.bridge3.arg_ip6_dhcp)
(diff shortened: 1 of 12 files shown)


netblue30:
All merged, thanks!

rusty-snake:
We need to update the manpage (and --help).

kris7t:
I will probably not have enough time to work on the documentation this weekend (qual exams are coming up), but will get to it some time next week.

glitsj16:
@kris7t Would you have a look at https://github.com/netblue30/firejail/issues/3174? Travis CI is broken due to this.

Issue #2302, opened 2018-12-18
Qutebrowser 1.5.2 (latest) not working with seccomp.drop name_to_handle_at 


D-Nice:
After having updated all packages on my distro (void linux), I noticed qutebrowser would start, seemingly hang for 10 seconds, and then close. 

Removing name_to_handle_at from the default profile, under seccomp.drop got qutebrowser working for me again without issues. Please let me know if there's any further information needed, to properly diagnose this issue.

Versions:
```
firejail version 0.9.56
```
```
qutebrowser v1.5.2
Git commit: 
Backend: QtWebEngine (Chromium 65.0.3325.230)

CPython: 3.6.7
Qt: 5.11.3 (compiled 5.11.2)
PyQt: 5.11.3

sip: 4.19.13
colorama: no
pypeg2: 2.15
jinja2: 2.10
pygments: 2.2.0
yaml: 3.13
cssutils: no
attr: 18.2.0
PyQt5.QtWebEngineWidgets: yes
PyQt5.QtWebKitWidgets: no
pdf.js: no
sqlite: 3.26.0
QtNetwork SSL: LibreSSL 2.8.3

Style: QFusionStyle
Platform: Linux-4.19.10_1-x86_64-with-glibc2.3.4, 64bit
Linux distribution: void (void)
Frozen: False
Imported from /usr/lib/python3.6/site-packages/qutebrowser
Using Python from /usr/bin/python3
Qt library executable path: /usr/lib/qt5/libexec, data path: /usr/share/qt5

Paths:
cache: /home/user/.cache/qutebrowser
config: /home/user/.config/qutebrowser
data: /home/user/.local/share/qutebrowser
runtime: /tmp/runtime-user/qutebrowser
system data: /usr/share/qutebrowser
```

D-Nice:
a quick addendum also is that init_module and delete_module were identified as missing and bad by the firejail audit, and I've added those in, without any noticeable issues so far.

rusty-snake:
@D-Nice to confirm:
works:
```
[...]

PR #3111, opened 2020-01-03
Get rid of #2302

rusty-snake:
@D-Nice @f4814 @q3cpma can someone confirm that this profile works.


q3cpma:
Works for me, I use this too on my local profile.

PR #3114, opened 2020-01-03
Allow Tor Browser to run /usr/bin/id

creideiki:
The `start-tor-browser` script tries to run `/usr/bin/id` to check that
it isn't root before starting the browser. See
https://gitweb.torproject.org/builders/tor-browser-build.git/tree/projects/tor-browser/RelativeLink/start-tor-browser?id=41fd236bbb7d3d75a27473f927be31f7dd8fdc99#n94

If `id` is not in the `private-bin` directory, the test still works by
accident, but prints these error messages:
```
  ./Browser/start-tor-browser: line 94: id: command not found
  ./Browser/start-tor-browser: line 94: [: : integer expression expected
```
Add `id` to the `private-bin` directory to make it run as intended.

### DIFF
--- a/etc/torbrowser-launcher.profile
+++ b/etc/torbrowser-launcher.profile
@@ -48,7 +48,7 @@ shell none
 #tracelog
 
 disable-mnt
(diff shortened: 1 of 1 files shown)


creideiki:
As long as someone has thought about it, I don't particularly care either way. I was mostly concerned with the script changing and suddenly no longer able to be run under Firejail.

Vincent43:
I'm in favor of adding `id`.

corecontingency:
Probably makes sense to add `id` back in the interests of providing a "future-proofed" generic profile for everyone, although I might personally be sticking to my custom hardened profile! :smile:

rusty-snake:
Merged, Thanks.

PR #3115, opened 2020-01-03
Add barrier profile

adrianlshaw:
Barrier is a keyboard and mouse sharing application that can be used across multiple networked computers. Here is a profile for it. 

I tried to make a whitelist for it but it was taking a huge amount of time. So here it is as a blacklist. It needs dbus, netlink, and doesn't work with a private /etc or private binary.

### DIFF
--- a/etc/barrier.profile
+++ b/etc/barrier.profile
@@ -1,5 +1,5 @@
 # Firejail profile for barrier
-# Keyboard and mouse sharing application
+# Description: Keyboard and mouse sharing application
 # This file is overwritten after every install/update
 
 # Persistent local customizations
--- a/etc/barrier.profile
+++ b/etc/barrier.profile
@@ -19,7 +19,6 @@ include disable-xdg.inc
 
 include whitelist-var-common.inc
 
-
 caps.drop all
 machine-id
 netfilter
--- a/etc/barrier.profile
+++ b/etc/barrier.profile
@@ -1,0 +1,41 @@
+# Firejail profile for barrier
+# Keyboard and mouse sharing application
+# This file is overwritten after every install/update
+
+# Persistent local customizations
+include barrier.local
+
+# Persistent global definitions
+include globals.local 
+
+noblacklist ${HOME}/.config/Debauchee/Barrier.conf
+noblacklist ${HOME}/.local/share/barrier/SSL/Barrier.pem
+noblacklist /etc/xdg/Debauchee/Barrier.conf
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+caps.drop all
+netfilter
+no3d
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6,netlink
+seccomp
+shell none
+
+private-dev
+private-cache
+private-tmp
+memory-deny-write-execute
--- a/etc/barrier.profile
+++ b/etc/barrier.profile
@@ -35,6 +35,7 @@ protocol unix,inet,inet6,netlink
 seccomp
 shell none
 
+disable-mnt
 private-dev
 private-cache
 private-tmp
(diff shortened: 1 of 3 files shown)


etc/barrier.profile
```diff
@@ -0,0 +1,42 @@
+# Firejail profile for barrier
+# Description: Keyboard and mouse sharing application
+# This file is overwritten after every install/update
+
```
rusty-snake:
```suggestion

```

etc/barrier.profile
```diff
@@ -0,0 +1,42 @@
...
+# Persistent local customizations
+include barrier.local
+
```
rusty-snake:
```suggestion

```

etc/barrier.profile
```diff
@@ -0,0 +1,42 @@
...
+# Persistent global definitions
+include globals.local 
+
+noblacklist ${HOME}/.config/Debauchee/Barrier.conf
+noblacklist ${HOME}/.local/share/barrier/SSL/Barrier.pem
```
rusty-snake: Add blacklist entrys  to disable-programs.inc.

etc/barrier.profile
```diff
@@ -0,0 +1,42 @@
...
+noblacklist /etc/xdg/Debauchee/Barrier.conf
```
rusty-snake:
```suggestion

```

Never blacklisted.

etc/barrier.profile
rusty-snake:
```suggestion
noblacklist ${HOME}/.local/share/barrier
```

etc/barrier.profile
```diff
@@ -0,0 +1,42 @@
...
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6,netlink
+seccomp
+shell none
+
```
rusty-snake: What about `tracelog`?

etc/barrier.profile
```diff
@@ -0,0 +1,42 @@
...
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+caps.drop all
+netfilter
```
rusty-snake: What about `machine-id`?

etc/barrier.profile
```diff
@@ -0,0 +1,42 @@
...
+noblacklist /etc/xdg/Debauchee/Barrier.conf
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
```
rusty-snake: `include whitelis-var-common.inc`?

etc/barrier.profile
```diff
@@ -0,0 +1,42 @@
...
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6,netlink
+seccomp
+shell none
+
+private-dev
```
rusty-snake: `disable-mnt`?

adrianlshaw:
Thanks for the review Rusty!

etc/barrier.profile
```diff
@@ -0,0 +1,46 @@
...
+shell none
+tracelog
+
+disable-mnt
+private-dev
+private-cache
+private-tmp
+
+memory-deny-write-execute
+noexec ${HOME}
```
SkewedZeppelin: redundant with disable-exec.inc

etc/barrier.profile
```diff
@@ -0,0 +1,47 @@
...
+# Description: Keyboard and mouse sharing application
+# This file is overwritten after every install/update
+# Persistent local customizations
+include barrier.local
+# Persistent global definitions
+include globals.local 
+
+noblacklist ${HOME}/.config/Debauchee/Barrier.conf
+noblacklist ${HOME}/.local/share/barrier
+noblacklist ${PATH}/openssl
```
SkewedZeppelin: most programs shouldn't need the openssl program, only the library.
the program is commonly used by linux ransomware

adrianlshaw: /usr/bin/openssl appeared as a blacklist violation in syslog. I think the program uses the CLI app to create certs during setup rather than through a library.

adrianlshaw:
Thanks for the review @SkewedZeppelin, I've made the changes.

rusty-snake:
Merged, Thanks.

PR #3123, opened 2020-01-05
Update i2prouter profile, and remove from firecfg

corecontingency:
- Added some networking options to private-etc copied from firefox-common, as I2P wasn't able to reseed, leading to it being unable to bootstrap into the I2P network.
- Removed from firecfg, as it only works if using the ubuntu PPA, but not if using the distro-independent java installer, or the AUR file. I2P also is manually started on the command line in any case, there is no desktop file, so it doesn't really make sense to have a firecfg entry anyways.
- Edited the comments to make them more clear.
- As of now, I2P only officially supports using openJDK 8 and 9, but in the interests of "future-proofing" I added some extra java versions.

### DIFF
--- a/etc/i2prouter.profile
+++ b/etc/i2prouter.profile
@@ -6,19 +6,19 @@ include i2prouter.local
 # Persistent global definitions
 include globals.local
 
-# Notice: default browser will not be able to automatically open, due to sandbox.
+# Notice: default browser will most likely not be able to automatically open, due to sandbox.
 # Auto-opening default browser can be disabled in the I2P router console.
-# This profile will not currently work with any Arch User Repository i2p packages,
-# use the distro-independent official java installer instead
+# This profile will not currently work with any Arch User Repository I2P packages,
+# use the distro-independent official I2P java installer instead
 
-# Only needed if i2prouter binary is in home directory, java installer does this
+# Only needed if i2prouter binary is in home directory, official I2P java installer does this
 ignore noexec ${HOME}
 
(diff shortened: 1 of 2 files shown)


etc/i2prouter.profile
```diff
@@ -67,5 +67,5 @@ shell none
 disable-mnt
 private-cache
 private-dev
-private-etc alternatives,ca-certificates,crypto-policies,i2p,java-8-openjdk,pki,ssl
+private-etc alternatives,ca-certificates,crypto-policies,dconf,group,hostname,hosts,i2p,java-8-openjdk,java-9-openjdk,java-10-openjdk,java-11-openjdk,java-12-openjdk,java-13-openjdk,java-openjdk,ld.so.cache,localtime,machine-id,nsswitch.conf,passwd,pki,resolv.conf,ssl
```
glitsj16: Perhaps we can get away with `java-*-openjdk` instead?

corecontingency: I tried that at first, but I don't know if private-etc accepts regular expressions. I keep getting:
```
Error: "java-*-openjdk" is an invalid filename
Error: proc 10805 cannot sync with peer: unexpected EOF
Peer 10806 unexpectedly exited with status 1
```
So I just decided to manually enter the versions.

glitsj16:
@corecontingency Apologies for the late review, it slipped under my radar. Looks fine, let's merge this and deal with potential fall-out later (if any). Thanks!

PR #3128, opened 2020-01-06
Allow sound for hexchat

haarp:
Hexchat can play sounds for notifications or other events. It makes sense to allow it to play them.

Cheers!


Vincent43:
Thx

PR #3134, opened 2020-01-08
cmus: allow access to resolv.conf

nblock:
DNS resolution is useful in case streams are referenced by DNS.


Fred-Barclay:
Merged, thanks!

PR #3149, opened 2020-01-15
Blacklisting openrc paths by defaults

GovanifY:
This PR blacklists openrc path that could be used to identify running services out of the sandbox, which could be used for information gathering.


GovanifY:
Tor does have its own sandbox by default(enabled by Sandbox 1 in /etc/torrc), which is actually fairly solid.
When it comes to other services I would be quite surprised if that would break them as runlevels only define the services set to launch as startup and init.d the actual init scripts, both of which should not be needed under normal use.

netblue30:
Thanks @Govanif, let's merge it and we'll fix it later if anything is breaking.

Issue #3147, opened 2020-01-15
ffmpeg.profile needs ld.so.cache on arch linux


jose1711:
Arch Linux needs `ld.so.cache` added to `private-etc` in `ffmpeg.profile`. Otherwise it gives:
```
$ ffmpeg
/usr/bin/ffmpeg: error while loading shared libraries: libmfx.so.1: cannot open shared object file: No such file or directory
```

rusty-snake:
Fixed, Thanks.

rusty-snake:
@glitsj16 I pinned this issue for now, so people see it (hopefully).

Vincent43:
The problem should be fixed in Arch Linux soon without need for changes in firejail: https://bugs.archlinux.org/task/65238 although the workarounds would be still useful to keep for similar cases in future.

florianbegusch:
> Arch Linux needs `ld.so.cache` added to `private-etc` in `ffmpeg.profile`. Otherwise it gives:
> 
> ```
> $ ffmpeg
> /usr/bin/ffmpeg: error while loading shared libraries: libmfx.so.1: cannot open shared object file: No such file or directory
> ```

For some reason the package changed the location to [/usr/lib](https://git.archlinux.org/svntogit/community.git/commit/trunk?h=packages/intel-media-sdk&id=79ee9971dc419ba02a249ee763866ad741538127) again.  
ffmpeg runs fine without ```ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload``` again. Probably any other program depending on it would too.

Issue #3076, opened 2019-12-09
[Feature] Profile Updates


FOSSONLY:
I would like to suggest a manual or automatic update for profiles. It is noted that there are often problems with the profiles. Especially since not all users use the same versions of different programs, and the profiles per version of Firejail are usually adapted to certain program versions. By automatically updating the profiles directly via git, any adjustments to the profiles could be immediately received by the users. This could prevent typical problems with program updates. What do you think?

rusty-snake:
:+1: a way to get profile fixes and enhancements to debian stable for example would be great.

FOSSONLY:
Yeah. I mean, theoretically, anyone could do that quickly by hand. Just install git and download the profiles to "/etc/firejail". The idea was to simplify it and make it easier for beginners.

rusty-snake:
> theoretically, anyone could do that quickly by hand.

You missed one point, new/changed options does not work with older firejail binaries.

Example: Debian stable has firejail 0.9.58
- mdwe breaks all most every GNOME-APP since firejail 0.9.60
- allow-debuggers works only with firejail master in profiles
- the seccomp exception syntax
- ...

Vincent43:
Yes, there is no guarantee that newer profiles will work with older firejail so shipping those separately isn't possible atm.

glitsj16:
I'm not sure if all distros offer a firejail-git package like the one from [AUR](https://aur.archlinux.org/packages/firejail-git/), but we could ask packagers to promote that if they do. I like the idea of a 'rolling-release-kind' firejail, it would offer some interesting opportunities (besides unburdening collaborators a bit regarding issues management/small profile fixes). E.g. https://github.com/netblue30/firejail/pull/3150.

rusty-snake:
If I look now at the relnotes, IMHO it is not possible to use master profile with a stable firejail after some scripts.

matu3ba:
@FOSSONLY So your proposal wants to fetch one of the  releases and adjust the paths for each distributions, because the distribution package is out of date?
Thats the job of distribution packagers.

Sadly there dont exist programs or scripts that fetch on what distribution you are and adjust the installation paths/rules for the distro accordingly, because the paths are tracked globally different on every distribution with `packet managers`.

Maybe you can be more specific what you want to do. Doing things for only a subset of the distributions belongs to anothet project.

rusty-snake:
Every distro I know uses `/etc/firejail`, that's not an issue.

PR #3150, opened 2020-01-16
Fix missing lib libmfx.so.1 for ffmpeg

florianbegusch:
To fix (on Arch):
```code
$ ffmpeg
/usr/bin/ffmpeg: error while loading shared libraries: libmfx.so.1: cannot open shared object file: No such file or directory
```

### DIFF
--- a/etc/ffmpeg.profile
+++ b/etc/ffmpeg.profile
@@ -48,6 +48,7 @@ private-bin ffmpeg
 private-cache
 private-dev
 private-etc alternatives,ca-certificates,crypto-policies,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,nsswitch.conf,pkcs11,pki,resolv.conf,ssl
+private-lib /opt/intel/mediasdk/lib/libmfx.so.1
 private-tmp
 
 # memory-deny-write-execute - it breaks old versions of ffmpeg


glitsj16:
Cannot reproduce this on Arch. The ffmpeg 1:4.2.2-1 package has a [new dependency on intel-media-sdk](https://git.archlinux.org/svntogit/packages.git/commit/trunk?h=packages/ffmpeg&id=f4949b65ed63375e4dca8d21c194a851b895a169), which does have that libmfx.so.1 file. Is your Arch fully updated?

florianbegusch:
It is fully updated and the file is present:
```code
$ pacman -Qs ffmpeg
local/ffmpeg 1:4.2.2-1
    Complete solution to record, convert and stream audio and video
local/gst-libav 1.16.2-1
    GStreamer open-source multimedia framework FFmpeg plugin
[ ~ (master u=) 0 ]
$ pacman -Qs intel-media-sdk
local/intel-media-sdk 19.4.0-1
    API to access hardware-accelerated video decode, encode and filtering on Intel
    platforms with integrated graphics


$ ls -alh /opt/intel/mediasdk/lib/libmfx.so.1
lrwxrwxrwx 1 root root 14 Jan  3 20:08 /opt/intel/mediasdk/lib/libmfx.so.1 -> libmfx.so.1.31
```

I filed this pull request because firejail did not provide the library to "ffmpeg".  
As far as I'm concerned firejail does not look into "/opt" to resolve libs. The "intel-media-sdk" package places the file into "/opt":
```code
$ pkgfile -l intel-media-sdk  | grep libmfx.so.1
community/intel-media-sdk       /opt/intel/mediasdk/lib/libmfx.so.1
community/intel-media-sdk       /opt/intel/mediasdk/lib/libmfx.so.1.31
```

glitsj16:
@florianbegusch Thanks for your detailed response. I see now what explains my different observation. I run firejail from git, which saw [this recent commit](https://github.com/netblue30/firejail/commit/31772d81f534d1537736dbec02098a80544182d9). Can you check if the changes made there fix this for you?

In any case, IMHO it wouldn't be advised to add private-lib with only libmfx.so.1 in it to the ffmpeg profile. We leave private-lib out of the picture because it pulls in a rather unpredictable set of dependencies with files in /usr/lib that would break it for a lot of users.

glitsj16:
@netblue30 Might this warrant a fresh release of 0.9.62 to unbreak ffmpeg (and avoid users reporting issues cfr. #3147)?

rusty-snake:
#3076

florianbegusch:
> @florianbegusch Thanks for your detailed response. I see now what explains my different observation. I run firejail from git, which saw [this recent commit](https://github.com/netblue30/firejail/commit/31772d81f534d1537736dbec02098a80544182d9). Can you check if the changes made there fix this for you?
> 
> In any case, IMHO it wouldn't be advised to add private-lib with only libmfx.so.1 in it to the ffmpeg profile. We leave private-lib out of the picture because it pulls in a rather unpredictable set of dependencies with files in /usr/lib on different Linux distributions that would break it for a lot of users.

It does fix it. Thank you for the quick response!

PR #3151, opened 2020-01-16
Fix missing lib libmfx.so.1 (standardnotes-desktop)

florianbegusch:
To fix:
```shell
$ firejail --profile=/etc/firejail/standardnotes-desktop.profile /usr/bin/standardnotes-desktop
Reading profile /etc/firejail/standardnotes-desktop.profile
Reading profile /etc/firejail/standardnotes-desktop.local
Parent pid 292648, child pid 292675
Warning: skipping alternatives for private /etc
Warning: skipping crypto-policies for private /etc
Warning: skipping pki for private /etc
Private /etc installed in 47.93 ms
Post-exec seccomp protector enabled
Child process initialized in 146.13 ms
/usr/bin/electron: error while loading shared libraries: libmfx.so.1: cannot open shared object file: No such file or directory

```

### DIFF
--- a/etc/standardnotes-desktop.profile
+++ b/etc/standardnotes-desktop.profile
@@ -39,5 +39,5 @@ seccomp !chroot
 disable-mnt
 private-dev
 private-tmp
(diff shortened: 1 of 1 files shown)


glitsj16:
Thanks. Merged!

PR #3152, opened 2020-01-16
Add profile for offical Linux Teams application

Karneades:
Based on the existing teams-for-linux profile here's the profile which matches with Microsoft's official Teams application for Linux.

Currently I had to disable private-bin and private-etc because of immediate crashes after start but it's worth having it in the repo nevertheless so others can contribute to that profile to enhance it if someone come up with the fix for private-bin and private-etc.

### DIFF
--- a/etc/teams.profile
+++ b/etc/teams.profile
@@ -18,6 +18,9 @@ include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 
+mkdir ${HOME}/.config/teams
+mkdir ${HOME}/.config/Microsoft
+
 whitelist ${HOME}/.config/teams
 whitelist ${HOME}/.config/Microsoft
 whitelist ${DOWNLOADS}
--- a/etc/teams.profile
+++ b/etc/teams.profile
@@ -4,6 +4,8 @@
 # Persistent local customizations
 include teams.local
 # Persistent global definitions
+# added by included profile
+#include globals.local
 
 # TODO add these 2 paths as blacklist to disable-programs.inc
 noblacklist ${HOME}/.config/teams
--- a/etc/teams.profile
+++ b/etc/teams.profile
@@ -7,7 +7,6 @@ include teams.local
 # added by included profile
 #include globals.local
 
-# TODO add these 2 paths as blacklist to disable-programs.inc
 noblacklist ${HOME}/.config/teams
 noblacklist ${HOME}/.config/Microsoft
 
@@ -21,13 +20,10 @@ whitelist ${HOME}/.config/teams
 whitelist ${HOME}/.config/Microsoft
 include whitelist-common.inc
 include whitelist-var-common.inc
-# TODO Check if can we enable wusc - if so, do we need to whitelist something in /usr/share/?
-#include whitelist-usr-share-common.inc
 
 nou2f
 shell none
-# TODO Check if can we enable tracelog?
-#tracelog
+tracelog
 
 disable-mnt
 private-cache
--- a/etc/teams.profile
+++ b/etc/teams.profile
@@ -1,48 +1,37 @@
-# Firejail profile for offical Teams application
-#
-# Description: Teams is an application for Microsoft's team collaboration and chat program
-# URL:         https://teams.microsoft.com/downloads
-#
+# Firejail profile for teams
+# Description: Official Microsoft Teams client for Linux using Electron.
 # This file is overwritten after every install/update
 # Persistent local customizations
 include teams.local
 # Persistent global definitions
-include globals.local
 
+# TODO add these 2 paths as blacklist to disable-programs.inc
 noblacklist ${HOME}/.config/teams
+noblacklist ${HOME}/.config/Microsoft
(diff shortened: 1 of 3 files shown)


etc/teams.profile
```diff
@@ -0,0 +1,45 @@
...
+
+noblacklist ${HOME}/.config/teams
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+
```
rusty-snake:
```suggestion

mkdir ${HOME}/.config/teams
mkdir ${HOME}/.config/Microsof
```

Karneades: Add both mkdir as suggested. thanks!

rusty-snake:
Is the offical Teams also electron-based? If so this should be an redirect profile. If not we can harden it.

Karneades:
Yes, I just used the existing teams-for-linux as basis which was also already an electron app. Will look now into https://github.com/netblue30/firejail/blob/master/etc/electron.profile and use that as include/redirect as others do using

```
# Redirect
include electron.profile
```

rusty-snake:
@Karneades I mean redirecting to teams-for-linux.profile.

rusty-snake:
https://github.com/netblue30/firejail/blob/master/etc/templates/redirect_alias-profile.template

Karneades:
Got that with the redirect, but I would not use teams-for-linux as basis rather using the redirect to electron which is used in multiple profiles. teams-for-linux should instead also use an redirect to electron.profile. The settings in teams-for-linux do not work with teams (e.g. private-bin issue).

rusty-snake:
```
# Firejail profile for Mictosoft Teams
# Description: DESCRIPTION
include teams.local
# added by included profile
#include globals.local

ignore private-bin
ignore private-etc

noblacklist ${HOME}/.config/teams

mkdir ${HOME}/.config/teams
mkdir ${HOME}/.config/Microsoft
whitelist ${HOME}/.config/teams
whitelist ${HOME}/.config/Microsoft

# Redirect
include teams-for-linux.profile
```

Karneades:
Hope my questions are ok...

why not using the basic electron.profile?

the teams-for-linux is a different application and also the directories are named differently.

rusty-snake:
different but similar.

If you strongly want to use electron.profile do it.

glitsj16:
@Karneades @rusty-snake Can we work on this some more? I have refactored teams-for-linux, whalebird and wire-desktop as electron redirect profiles. To speed up this PR for merging, may I suggest that @Karneades takes a look at a [temporary teams.profile](https://gist.github.com/glitsj16/e16988634e58a3a0bf9a854cec976500), tries to check/answer the questions in that file marked with `# ***` and integrates it here? I do realize this request diverges from the 'regular' PR-routine (if there is such a thing), but I think it's worthwhile to have a nice, manageable profile for this application (which is the sole goal of this request).

Karneades:
Thanks @glitsj16 Will check that next week, test it and commit asap.

Karneades:
The blacklist entries are missing to for teams-for-linux in disable-programs.inc, will add those soon too.

etc/teams.profile
```diff
@@ -0,0 +1,37 @@
+# Firejail profile for teams
+# Description: Official Microsoft Teams client for Linux using Electron.
+# This file is overwritten after every install/update
+# Persistent local customizations
+include teams.local
+# Persistent global definitions
```
rusty-snake:
```suggestion
# Persistent global definitions
# added by included profile
#include globals.local
```

etc/teams.profile
```diff
@@ -0,0 +1,37 @@
...
+
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+
+mkdir ${HOME}/.config/teams
+mkdir ${HOME}/.config/Microsoft
+whitelist ${HOME}/.config/teams
+whitelist ${HOME}/.config/Microsoft
+whitelist ${DOWNLOADS}
```
rusty-snake:
```suggestion

```

Karneades: Download is used for file sharing through Teams. Should that essential feature be just in local profile overwrites?

rusty-snake: `whitelist ${DOWNLOADS}` is in electron.profile

Karneades: Sorry, my bad, remove that line from the profile asap

Karneades:
I cleanup the profile after testing and enabled tracelog which logs blacklist access to syslog now. I had to use `ignore apparmor` in my local config - don't know if it's an issue on my side or with (apparmor && teams) in general.

rusty-snake:
> don't know if it's an issue on my side or with (apparmor && teams) in general.

better save then sorry, lets comment it for now.

Karneades:
@rusty-snake Added a comment in Teams profile regarding apparmor. Please let me know if anything's missing.

rusty-snake:
Merged, Thanks.

Issue #3153, opened 2020-01-16
Whois not working


aminvakil:
```
firejail --profile=/etc/firejail/whois.profile --private-bin=strace --allow-debuggers strace -e trace=open,openat whois 185.121.130.130
openat(AT_FDCWD, "/etc/ld.so.preload", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/run/firejail/lib/libtracelog.so", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/tls/haswell/x86_64/libidn2.so.0", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/tls/haswell/libidn2.so.0", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/tls/x86_64/libidn2.so.0", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/tls/libidn2.so.0", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/haswell/x86_64/libidn2.so.0", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/haswell/libidn2.so.0", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/x86_64/libidn2.so.0", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib/libidn2.so.0", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/usr/lib/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/usr/lib/libdl.so.2", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/usr/lib/libunistring.so.2", O_RDONLY|O_CLOEXEC) = 3
[...]

PR #3154, opened 2020-01-16
Update whois.profile

aminvakil:
More details in https://github.com/netblue30/firejail/issues/3153

### DIFF
--- a/etc/whois.profile
+++ b/etc/whois.profile
@@ -45,8 +45,8 @@ private
 private-bin bash,sh,whois
 private-cache
 private-dev
-private-etc alternatives,hosts,jwhois.conf,services,whois.conf
-private-lib
+private-etc alternatives,hosts,jwhois.conf,services,whois.conf,resolv.conf
+private-lib gconv
 private-tmp
 
 memory-deny-write-execute
--- a/etc/whois.profile
+++ b/etc/whois.profile
@@ -45,7 +45,7 @@ private
 private-bin bash,sh,whois
 private-cache
 private-dev
-private-etc alternatives,hosts,jwhois.conf,services,whois.conf,resolv.conf
+private-etc alternatives,hosts,jwhois.conf,resolv.conf,services,whois.conf
 private-lib gconv
 private-tmp
 


etc/whois.profile
```diff
@@ -45,8 +45,8 @@ private
 private-bin bash,sh,whois
 private-cache
 private-dev
-private-etc alternatives,hosts,jwhois.conf,services,whois.conf
-private-lib
+private-etc alternatives,hosts,jwhois.conf,services,whois.conf,resolv.conf
```
rusty-snake:
```suggestion
private-etc alternatives,hosts,jwhois.conf,resolv.conf,services,whois.conf
```

aminvakil: What's the difference?

glitsj16:
I **wouldn't merge this yet**. IMHO the OP might be using firecfg, in which case the strace examples would need to refer to '/usr/bin/whois' instead of 'whois'. I'm asking some more info on the situation in #3153.

etc/whois.profile
glitsj16: I'm trying to reproduce this. Please see #3153 for details.

etc/whois.profile
```diff
@@ -45,8 +45,8 @@ private
...
+private-lib gconv
```
glitsj16: I'm trying to reproduce this. Plaese see #3153 for details.

etc/whois.profile
Vincent43: alphabetic sorting

glitsj16:
Can someone confirm whois is working as expected as-is, so without the changes proposed in this PR? It is for me on both Arch and Ubuntu, but after prolonged discussion in #3153 it would be nice to make sure before closing this due to 'not a bug'.

etc/whois.profile
netblue30: I get it on Debian stable:

$ firejail whois google.com
getaddrinfo(whois.verisign-grs.com): Name or service not known

netblue30:
OK, I'm getting lost in this resolve-conversation thing github put in. So, I would merge @aminvakil, and change:

private-etc alternatives,hosts,jwhois.conf,services,whois.conf,resolv.conf
private-lib gconv

This is working fine on Debian stable.

PR #3155, opened 2020-01-16
profiles: whitelist transmission-daemon config directory

reinerh:
Reported at: https://bugs.debian.org/948993


glitsj16:
Looks fine. I have had this in my transmission-daemon.local for a while now without issues. Additionally we could mkdir/whitelist ${HOME}/.cache/transmission-daemon too.

reinerh:
It looks like it does not need a cache directory. At least during my short test it didn't create one (while unjailed).

reinerh:
Thanks for reviewing :)

Issue #3001, opened 2019-10-15
firejail --get and --put fail on filenames containing brackets


JKjakay:
Hi, I am using firejail to sandbox firefox, and I love it! However, I have noticed a problem.

I'm running firefox with very restricted settings (--seccomp --private). So, in order to be able to retrieve downloaded files I use --get and --put to access the file within the sandbox. This works ok, except if the filename contains round brackets like "(" or ")", and maybe other special characters. I get an "invalid filename" error. 

This happens with a filename like this:

foo(321).jpg

Escaping the illegal character wiith \ does not work. Using quotation marks also does not work:  --get "/home/user/Downloads/foo(321).jpg"

Im using bash on Ubuntu 19.04. Regards JK

JKjakay:
Anybody got any ideas? I guess brackets may not work because there is no kind of filename expansion active with --put and --get. Probably for security, I think. But it still seems like a bug to me.

smitsohu:
I doubt you'll find a workaround, it is by design. Firejail rejects filenames if they contain certain characters.

JKjakay:
Thank you for clarifying this. It should be mentioned somewhere in the documentation (which is, sadly, not good). If you can use a filename as an argument, you would expect it to behave according to Posix standards.

kodo-pp:
I agree with JKjakay with regard to this issue: the restrictions on the filenames don't seem to fit right with POSIX standards and, without being mentioned in the documentation (at least I couldn't find anything about them in the manual page `man firejail`), are quite surprising and counter-intuitive.

From the security viewpoint, I don't see any benefit of deliberately restricting the set of characters used in filenames in the context of `--get`, `--put` and `--ls` operations (which are supposed to work with **any** filenames). Firejail just has to access the specified file, which should not treat characters like brackets or exclamation sign in any special way. (I hope Firejail doesn't use any unquoted shell magic for these operations, does it?)

In fact, brackets are quite common in the filenames, especially of downloaded files, which often have names like `some important file (2).zip`. Managing such files is a typical use case (I guess, at least what `--get`/`--put`/`--ls` are designed for), so I see disallowing the work with some commonly used characters in filenames is a misfeature which harms the usability.

Also, I'm a bit surprised that there's been no action taken about this issue since November. Some discussion took place, but the issue wasn't resolved at all (it wasn't rejected or fixed in any way), so I wonder whether we can actually do something about it.

smitsohu:
@netblue30 ?

[...]

PR #3156, opened 2020-01-17
print rejected character in invalid filenames

smitsohu:
provide feedback about why a filename was rejected; see issue #3001

### DIFF
--- a/src/firejail/macros.c
+++ b/src/firejail/macros.c
@@ -258,6 +258,28 @@ char *expand_macros(const char *path) {
 	return rv;
 }
 
+// replace control characters with a '?'
+static char *fix_control_chars(const char *fname) {
+	assert(fname);
+
+	size_t len = strlen(fname);
+	char *rv = malloc(len + 1);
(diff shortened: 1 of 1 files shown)


smitsohu:
Actually this does not work well if the filename contains a control character. I would like to add a few lines to filter control characters as well (commit follows)

smitsohu:
There needs to be a better error message if filename contains a control character. Just printing the filename does not work, as 1) these characters are not printable, and 2) the printed filename is sometimes hard to decipher garbage. So I'd propose to replace them with a `?`.

netblue30:
Sure, let's go for '?', there is no better way. () are quite common in downloads.

smitsohu:
Ok, I can remove () from the set.

rusty-snake:
@smitsohu https://gitlab.com/Firejail/firejail_ci/-/jobs/422887566#L174

smitsohu:
@rusty-snake Thanks for the pointer

PR #3161, opened 2020-01-17
blacklist ${RUNUSER}/wayland-* in every profile with blacklist /tmp/.X11-unix or x11 none

rusty-snake:
This PR adds `blacklist ${RUNUSER}/wayland-*` (the default location for wayland-sockets) to every profile with `blacklist /tmp/.X11-unix` or `x11 none`


glitsj16:
@rusty-snake I think devilspie{,2} also have to be included in this. Nice job!

Issue #3164, opened 2020-01-18
end of python2 


rusty-snake:
> The Python 2 language, i.e. Python 2.7.x, was officially discontinued on January 1, 2020 (first planned for 2015) after which security patches and other improvements will not be released for it. With Python 2's end-of-life, only Python 3.5.x and later are supported.

_Source: https://en.wikipedia.org/wiki/Python2_

Python2 is getting a security risk. Till it's completely dead we should remove `include allow-python2.inc` from every profile where it's not requiered. We could also add `disable-python2.inc` for profiles where we can't add `disable-interpreters.inc`.

List of profile with `include allow-python2.inc`:

 - [ ] [anki.profile](https://github.com/netblue30/firejail/blob/master/etc/anki.profile)
 - [ ] [arm.profile](https://github.com/netblue30/firejail/blob/master/etc/arm.profile)
 - [ ] [autokey-common.profile](https://github.com/netblue30/firejail/blob/master/etc/autokey-common.profile)
 - [ ] [bleachbit.profile](https://github.com/netblue30/firejail/blob/master/etc/bleachbit.profile)
 - [ ] [blender.profile](https://github.com/netblue30/firejail/blob/master/etc/blender.profile)
 - [ ] [caja.profile](https://github.com/netblue30/firejail/blob/master/etc/caja.profile)
 - [ ] [cameramonitor.profile](https://github.com/netblue30/firejail/blob/master/etc/cameramonitor.profile)
 - [ ] [catfish.profile](https://github.com/netblue30/firejail/blob/master/etc/catfish.profile)
 - [ ] [celluloid.profile](https://github.com/netblue30/firejail/blob/master/etc/celluloid.profile)
 - [ ] [cherrytree.profile](https://github.com/netblue30/firejail/blob/master/etc/cherrytree.profile)
 - [ ] [ddgtk.profile](https://github.com/netblue30/firejail/blob/master/etc/ddgtk.profile)
 - [ ] [deluge.profile](https://github.com/netblue30/firejail/blob/master/etc/deluge.profile)
 - [ ] [d-feet.profile](https://github.com/netblue30/firejail/blob/master/etc/d-feet.profile)
 - [ ] [dia.profile](https://github.com/netblue30/firejail/blob/master/etc/dia.profile)
 - [ ] [display.profile](https://github.com/netblue30/firejail/blob/master/etc/display.profile)
[...]

PR #3167, opened 2020-01-19
clarify dropping python2 support in meld.profile

glitsj16:
This is just a suggestion relating to the discussion on dropping support for python2 in https://github.com/netblue30/firejail/issues/3164.

### DIFF
--- a/etc/meld.profile
+++ b/etc/meld.profile
@@ -23,7 +23,7 @@ noblacklist ${HOME}/.subversion
 
 # Allow python (blacklisted by disable-interpreters.inc)
 # Python 2 is EOL (see #3164). Uncomment the next line (or put it into your meld.local) if you understand the risks but want python 2 support for older meld versions.
-include allow-python3.inc
+#include allow-python3.inc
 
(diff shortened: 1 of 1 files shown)


etc/meld.profile
```diff
@@ -22,6 +22,7 @@ noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.subversion
 
 # Allow python (blacklisted by disable-interpreters.inc)
+# Python 2 is EOL (see #3164). Uncomment the next line (or put it into your meld.local) if you understand the risks but want python 2 support for older meld versions.
```
rusty-snake:
```suggestion
# Python 2 is EOL (see #3164). Uncomment the next line (or put it into your meld.local) if you understand the risks but want python 2 support for older meld versions.
#include allow-python2.inc
```

glitsj16: Hehe, my subconscious python2 love showed itself... Nice catch, and more importantly, nice job proposing this in #3164 in the first place :+1:

PR #3177, opened 2020-01-22
Allow request_key syscall for udiskie

### LINKED ISSUE #3175: udiskie fails to open drive with seccomp blocking request_key
udiskie fails to automatically open my drive when running under firejail.

When running udiskie without firejail, I get the notification that the drive has been added, then a popup to enter the drive password (the drive is LUKS encrypted).

When running under firejail, the notification that the drive has been added appears, but no prompt for my password appears. Unfortunately `udiskie -v` doesn't show any errors.

Adding `ignore seccomp` to `udiskie.local` fixes the issue.

Debugging the blocked syscalls it appears that `request_key` is being blocked and causing the hang.

carloabelli:
Fixes #3175.

### DIFF
--- a/etc/udiskie.profile
+++ b/etc/udiskie.profile
@@ -31,7 +31,7 @@ notv
 nou2f
 novideo
 protocol unix
-seccomp
+seccomp !request_key
 shell none
 tracelog
 


rusty-snake:
Merged, Thanks.

PR #3178, opened 2020-01-22
Allow mbind syscall for GIMP

### LINKED ISSUE #2681: Gimp Fails to Run with seccomp
Gimp 2.10.10 fails to load (hangs without any output).

Firejail version 0.9.58.2 on Arch Linux. Runs fine with `--noprofile`. Disabling `seccomp` in `gimp.profile` fixes the issue.

carloabelli:
Fixes #2681.

### DIFF
--- a/etc/gimp.profile
+++ b/etc/gimp.profile
@@ -45,7 +45,7 @@ nosound
 notv
 nou2f
 protocol unix
-seccomp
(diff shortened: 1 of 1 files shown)


rusty-snake:
Merged, Thanks.

PR #3183, opened 2020-01-24
fix test catchsignal-master.sh

glitsj16:
While moving copyright to 2020 I noticed that _catchsignal2.sh_ was never being run. Being totally unfamiliar with the test codebase, I could be wrong here. I'd appreciate a review before breaking stuff.

### DIFF
--- a/test/utils/catchsignal-master.sh
+++ b/test/utils/catchsignal-master.sh
@@ -4,4 +4,4 @@
 # License GPL v2
 
 ./catchsignal.sh &
-./catchsignal.sh &
+./catchsignal2.sh &


reinerh:
I'm not completely sure, but it might be intentional.

catchsignal.sh is currently executed from shutdown2.exp.
catchsignal2.sh is executed from shutdown3.exp.
catchsignal-master.sh is executed from shutdown4.exp.

It could be that catchsignal-master.sh was intended to run two times the catchsignal.sh script.
But it could also be that it should run catchsignal.sh and catchsignal2.sh like you proposed.

ping @netblue30

glitsj16:
@reinerh Thanks for your response, that would make sense.

netblue30:
I'll do a run through all the tests

rusty-snake:
Any progress here?

glitsj16:
I'm closing this. If it indeed was a problem it should have shown itself by now.

PR #3186, opened 2020-01-25
blacklist gjs in disable-interpreters

rusty-snake:
TODO: reverse dependencies check on more distros (actually only Fedora).

### DIFF
--- a/etc/allow-gjs.inc
+++ b/etc/allow-gjs.inc
@@ -1,0 +1,6 @@
+noblacklist ${PATH}/gjs
+noblacklist ${PATH}/gjs-console
+noblacklist /usr/lib/gjs
+noblacklist /usr/lib64/gjs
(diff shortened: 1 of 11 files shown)


glitsj16:
@rusty-snake On Arch we would need to blacklist /usr/lib/libgjs.* too. I suppose allow-gjs.inc will be added later?

rusty-snake:
Nope, I only commited change files and forget to commit new files, good catch.

glitsj16:
@rusty-snake Looks fine. Unless you want to add anything, we can merge. On a side-note. I just realized we don't have local persistent customizations for allow-*.inc and implemented it via https://github.com/netblue30/firejail/pull/3187. What do you think? If that is merged we'd need a smallish change to allow-gjs.inc, but we can take care of that later.

Issue #3174, opened 2020-01-21
Travis CI fail: error: comparison between signed and unsigned integer expressions


Vincent43:
Since https://github.com/netblue30/firejail/commit/80b50300b8349a5b47a9ba05cd2a8f9a4f1c17c5 the [Travis CI](https://travis-ci.org/netblue30/firejail/builds/639986140) show following errors:
```
sandbox.c: In function ‘monitor_application’:

sandbox.c:340:15: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]

       if (pid == dhclient4_pid || pid == dhclient6_pid)

               ^

sandbox.c:340:39: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]

       if (pid == dhclient4_pid || pid == dhclient6_pid)

                                       ^

cc1: all warnings being treated as errors

make[1]: *** [sandbox.o] Error 1

make[1]: Leaving directory `/home/travis/build/netblue30/firejail/src/firejail'

make: *** [src/firejail] Error 2

The command "( ./configure --enable-fatal-warnings --prefix=/usr && make && sudo make install && make test-travis )" exited with 2.
```
It doesn't look related to aforementioned commit though.

rusty-snake:
https://travis-ci.org/netblue30/firejail/builds shows me https://github.com/netblue30/firejail/commit/94d018cec4daa4d409877640277156d2038628ed as first broken commit.

kris7t:
Build is fixed by 93cecca281c42fe7b96605b612f3c206d528cae2

PR #3190, opened 2020-01-27
DHCP client code quality fixes

kris7t:
I finally had some time to fix up my implementation for DHCP client support. Sorry for the delay!
* Unfortunately, I used spaces instead of tabs for my code in #3102  (I messed up something in Emacs, I think), which resulted in inconsistent indentation at least on GIthub. My first commit replaces the spaces with tabs, however, it does mess up `git blame` output a bit, so it is debatable whether we should merge it.
* I fixed the unsigned comparison warning in #3174 that made the build fail due to `-Werror`.
* Updated documentation (usage, firejail(1) and firejail-profile(5) manpages) with the new options.
By the way, there seem to be some differences between default CFLAGS produces by `./configure` on my machine and on Travis. ~Compiling with `-Wall -Wextra -Werror` seems prudent, but does not happen for me locally.~ Thanks for @Vincent43 for pointing out the `./configure --enable-fatal-warnings` option! On the other hand, on Travis, C99 support is not enabled. Travis also uses older netlink headers, I think, for example, it does not define `IFA_FLAGS`.

### DIFF
--- a/src/firejail/dhcp.c
+++ b/src/firejail/dhcp.c
@@ -29,130 +29,133 @@ pid_t dhclient4_pid = 0;
 pid_t dhclient6_pid = 0;
 
 typedef struct {
-  char *version_arg;
-  char *pid_file;
-  char *leases_file;
-  uint8_t generate_duid;
-  char *duid_leases_file;
-  pid_t *pid;
-  ptrdiff_t arg_offset;
+	char *version_arg;
+	char *pid_file;
+	char *leases_file;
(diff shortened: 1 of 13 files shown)


Vincent43:
> By the way, there seem to be some differences between default CFLAGS produces by ./configure on my machine and on Travis. Compiling with -Wall -Wextra -Werror seems prudent, but does not happen for me locally.

Travis uses `./configure --enable-fatal-warnings`.

Issue #3192, opened 2020-01-28
zathura fails without /etc/ld.so.cache


s-hamann:
I tried opening a PDF file in zathura. Zathura started, but could not open the document. I got the following error message on the console:
```
error: Could not load plugin '/usr/lib64/zathura/libpdf-poppler.so' (libstdc++.so.6: cannot open shared object file: No such file or directory).
error: Unknown file type: 'application/pdf'
```

I was able to resolve this by adding `ld.so.cache` to the `private-tmp` directive.
Could this please be added to the default profile?

rusty-snake:
https://github.com/netblue30/firejail/commit/22f74522156f67f26d6be5c8943e905a75a7f1ee

glitsj16:
@s-hamann After seeing this I had a first-time look at our zathura profile. It could be considerably hardened IMHO. I installed zathura on Arch and did some testing with the profile changes in #3193. Would you be willing to test that profile please? You can leave your comments, remarks etcetera in the PR.

Issue #3121, opened 2020-01-04
evince.profile issues (Gentoo)


reagentoo:
@glitsj16 Hi.
I've tested again the profile after your commit (https://github.com/netblue30/firejail/pull/3120/commits/bd7394f6e5a23c6bf68b37b74c5f786ee16b4712). And as I said in our previous discussion (https://github.com/netblue30/firejail/issues/3118) `libgcc_s.so.*,libstdc++.so.*` does not solve.
I can view PDFs only if `,gcc` used in `private-lib`. But It's not properly way to fix I think (not removes warnings).

Also may be it's necessary to remove `nodbus` due to line in log:
```
Warning: An abstract unix socket for session D-BUS might still be available. Use --net or remove unix from --protocol set.
```

glitsj16:
> Also may be it's necessary to remove nodbus due to line in log:
> 
> Warning: An abstract unix socket for session D-BUS might still be available. Use --net or remove unix from --protocol set.
> 

@reagentoo Hi again. Let's tackle this one first. As the warning and the comment inside /etc/firejail/evince.profile states `net none` only has confirmed issues on Ubuntu with AppAmor. On Gentoo you can add `net none` to your evince.local. That should take care of the D-BUS warning. Firejail also has the `--audit` option to check/confirm that:
`
$ firejail --audit /usr/bin/evince
`

> I can view PDFs only if `,gcc` used in private-lib. But It's not properly way to fix I think (not removes warnings).

If adding `gcc` is the only way to get evince working on Gentoo, I'll happily add it. IMHO it shouldn't impact the strength of the profile. Besides that it doesn't remove the warnings, does evince work for you with all its functionalities if you add gcc to private-lib?

rusty-snake:
IIRC you can do `private-lib gcc/something/libstdc++.so` which will add `/usr/lib/libstdc++.so` in the sandbox.

reagentoo:
@glitsj16 
> As the warning and the comment inside /etc/firejail/evince.profile states net none only has confirmed issues on Ubuntu with AppAmor.

Actually I have both enabled sandboxes AppArmor and Firejail on Gentoo.

> On Gentoo you can add net none to your evince.local. That should take care of the D-BUS warning.

I've gotten log without DBus warning after adding this.

> Besides that it doesn't remove the warnings, does evince work for you with all its functionalities if you add gcc to private-lib?

Right.

glitsj16:
[...]

Issue #3118, opened 2020-01-03
Can't load libstdc++.so.6 due to private-etc (Gentoo)


reagentoo:
`ERROR: ffmpeg: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory`

Affected applications: celluloid, evince, youtube-dl

My temporary fix is to remove next lines:
celluloid.profile:
```
private-etc alternatives,ca-certificates,crypto-policies,dconf,drirc,fonts,gtk-3.0,hosts,libva.conf,localtime,machine-id,pkcs11,pki,resolv.conf,selinux,ssl,xdg
```
evince.profile:
```
private-etc alternatives,fonts,group,machine-id,passwd
private-lib evince,gconv,gdk-pixbuf-2.*,gio,gvfs/libgvfscommon.so,libdjvulibre.so.*,libgconf-2.so.*,libgraphite2.so.*,libpoppler-glib.so.*,librsvg-2.so.*,libspectre.so.*
```
youtube-dl.profile:
```
private-etc alternatives,ca-certificates,crypto-policies,hostname,hosts,mime.types,pki,resolv.conf,ssl,youtube-dl.conf
```

reagentoo:
Adding `ld.so.cache` to `private-etc` solve this. But except for evince:
```
Warning fldd: cannot find libstdc++.so.6, skipping...
Warning fldd: cannot find libgcc_s.so.1, skipping...
```

glitsj16:
> Adding `ld.so.cache` to private-etc solve this. But except for "evince":
Warning fldd: cannot find libstdc++.so.6, skipping...
Warning fldd: cannot find libgcc_s.so.1, skipping...

@reagentoo Hi, thanks for reporting. The evince profile also has `private-lib`. Can you try if that works when adding `libgcc_s.so.*,libstdc++.so.*` to it?

reagentoo:
@glitsj16 Hi.
Adding `libgcc_s.so.*,libstdc++.so.*` to `private-lib` does not helps.
I've tried to add `,gcc` instead of them (This dir contains this libs). So I can see pdf images in the application after that, but I still see this warnings that I showed above.

glitsj16:
[...]

PR #3193, opened 2020-01-28
updates for zathura.profile

glitsj16:
After seeing #3192 it seemed the zathura profile could be hardened considerably. I've tested this PR on Arch Linux and will be asking OP of #3192 to do the same. Additional input is always welcome as this was the first time I used zathura so there might be omissions.

### DIFF
--- a/etc/zathura.profile
+++ b/etc/zathura.profile
@@ -20,11 +20,15 @@ include disable-xdg.inc
 
 mkdir ${HOME}/.config/zathura
 mkdir ${HOME}/.local/share/zathura
+whitelist /usr/share/zathura
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
 
 caps.drop all
+ipc-namespace
 machine-id
-# net none
-# nodbus
+net none
+nodbus
 nodvd
 nogroups
 nonewprivs
@@ -42,6 +46,7 @@ private-bin zathura
 private-cache
 private-dev
 private-etc alternatives,fonts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,machine-id
+private-lib libarchive.so.*,libdjvulibre.so.*,libgirara-gtk*,libpoppler-glib.so.*,libspectre.so.*,zathura
 private-tmp
 
 read-only ${HOME}
--- a/etc/zathura.profile
+++ b/etc/zathura.profile
@@ -18,6 +18,9 @@ include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-xdg.inc
 
+mkdir ${HOME}/.config/zathura
+mkdir ${HOME}/.local/share/zathura
+
 caps.drop all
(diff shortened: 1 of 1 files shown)


s-hamann:
Does not work for me. I get the following output:
```
error: Could not load plugin '/usr/lib64/zathura/libpdf-poppler.so' (libstdc++.so.6: cannot open shared object file: No such file or directory).
error: Unable to initialize database. Bookmarks won't be available.
error: Unknown file type: 'application/pdf'
```

The first error is caused by the `private-lib` directive, but I couldn't figure out what to add to get the files into the sandbox.
The third error is caused by the first one.

The second error is unrelated to `private-lib`, as it is printed even after removing the line from the profile. It does not prevent Zathura from showing the PDF.
I never consciously used the database and do not know how it works or what it needs.

s-hamann:
Adding `libstdc++.so.*` does not help for me. I think my issue is that `/usr/lib64/zathura` is empty in the sandbox.

glitsj16:
@s-hamann Thanks for testing, this really helps to improve the profile.

> error: Could not load plugin '/usr/lib64/zathura/libpdf-poppler.so' (libstdc++.so.6: cannot open shared object file: No such file or directory).

I pushed a commit that should fix this. Had to add libstdc++.so.* to private-lib. That should take care of the first and third error IMHO. About the second error. There are a lot of profiles that show errors like this. As long as the application works as expected we prefer to have a tighter profile instead of an error-free terminal output. So I'll proceed to merge this and keep an eye on zathura here too. Thanks again for testing, very much appreciated.

Regards

glitsj16:
> Adding libstdc++.so.* does not help for me. I think my issue is that /usr/lib64/zathura is empty in the sandbox.

@s-hamann I've disabled private-lib for now. It wasn't in the profile before so we still have made some progress. I'll have another closer look at zathura later to see if we can fix private-lib. Thanks for the speedy reply.

s-hamann:
I applied the solution from #3121 to my Zathura profile and it now works, even with `private-lib`.
The full line is now:
```
private-lib libarchive.so.*,libdjvulibre.so.*,libgirara-gtk*,libpoppler-glib.so.*,libspectre.so.*,libstdc++.so.*,zathura,gcc/*/*/libstdc++.so.*,gcc/*/*/libgcc_s.so.*
```

Note: This is on Gentoo as well. And since #3192 is the equivalent of #3118 for Zathura, I suspect that more profiles might have similar issues.

glitsj16:
@s-hamann Thanks for taking the time to report this Gentoo issue. I'll re-enable private-lib in zathura.profile shortly. Also, after a preliminary search through private-lib in all our profiles, we can update geekbench and unf accordingly for Gentoo. Those at least have references to either libgcc_s.so.* or libstdc++.so.* so it's easy to fix them as well. If you stumble on others that need this for Gentoo, please give us a ping!

PR #3194, opened 2020-01-28
profiles: whitelist /usr/share/doc

reinerh:
It contains documentation and example files for installed programs and libraries.

I'm not sure if it should be allowed here globally, or in individual profiles, where it makes sense.
(For example in browsers, which are often used for reading documentation from there.)

Reported at: https://bugs.debian.org/950007

### DIFF
--- a/etc/devhelp.profile
+++ b/etc/devhelp.profile
@@ -16,6 +16,8 @@ include disable-programs.inc
 include disable-xdg.inc
 
 whitelist /usr/share/devhelp
+whitelist /usr/share/doc
+whitelist /usr/share/gtk-doc/html
 include whitelist-common.inc
 include whitelist-usr-share-common.inc
 
--- a/etc/evince.profile
+++ b/etc/evince.profile
@@ -17,6 +17,7 @@ include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-xdg.inc
 
+whitelist /usr/share/doc
 whitelist /usr/share/evince
 whitelist /usr/share/poppler
 whitelist /usr/share/tracker
--- a/etc/firefox.profile
+++ b/etc/firefox.profile
@@ -14,6 +14,8 @@ mkdir ${HOME}/.mozilla
 whitelist ${HOME}/.cache/mozilla/firefox
 whitelist ${HOME}/.mozilla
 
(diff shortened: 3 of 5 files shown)


glitsj16:
Did a quick count. Currently we have over 120 profiles using wusc, of which firefox is the only web browser. If we decide to add /usr/share/doc to wusc only for web browsers (which indeed makes more sense) we need to move it from firefox.profile to firefox-common.profile to cover redirects. And do something similar in chromium-common.profile if that works with wusc obviously.

reinerh:
Just noticed for example evince also uses wusc. We should allow PDF viewers as well.

rusty-snake:
> we need to move it from firefox.profile to firefox-common.profile to cover redirects.

No, we don't want to enable wusc for all this profiles. Except if you test wusc for all profiles.

```
$ grep "include firefox-common.profile" /etc/firejail/*
/etc/firejail/abrowser.profile:include firefox-common.profile
/etc/firejail/basilisk.profile:include firefox-common.profile
/etc/firejail/cliqz.profile:include firefox-common.profile
/etc/firejail/cyberfox.profile:include firefox-common.profile
/etc/firejail/firefox.profile:include firefox-common.profile # has wusc
/etc/firejail/icecat.profile:include firefox-common.profile
/etc/firejail/palemoon.profile:include firefox-common.profile
/etc/firejail/thunderbird.profile:include firefox-common.profile # has wusc
/etc/firejail/waterfox.profile:include firefox-common.profile
```

glitsj16:
> No, we don't want to enable wusc for all this profiles. Except if you test wusc for all profiles.

@rusty-snake You're correct, didn't think it through.

reinerh:
I moved the whitelist now to some profiles where I think it makes sense.

glitsj16:
> I moved the whitelist now to some profiles where I think it makes sense.

To me this looks ready to merge :+1:.

glitsj16:
> devhelp (IDK were the gnome-docs are stored)

I see files under `/usr/share/gtk-doc/html` on Arch and Ubuntu 16.04 LTS. This path might be whitelisted both in devhelp and firefox (if we want both these applications to read them).

reinerh:
done

PR #3208, opened 2020-02-07
Fixes for fix_private-bin.py

KOLANICH:
Added typing.
Replaced os.path with pathlib.
Made it work.
Fixes in .gitignore.


glitsj16:
Let's merge. Thanks!

PR #3239, opened 2020-02-23
Harden dhcp by checking for /sbin/dhclient

kris7t:
In order to avoid picking up some random executable from `$PATH` and potentially running it root privileges, `dhcp.c` should check for `/sbin/dhclient` or `/usr/sbin/dhclient` directly. This might cause problems on distributions where `dhclient` is located elsewhere, but as far as I know, `/sbin` is the default location (and even Arch, which uses only `/usr/bin`, symlinks `/sbin` to `/usr/bin`). This is the same approach that `netfilter.c` takes when running `iptables-restore`.

I formulated the check for root privileges according to https://github.com/netblue30/firejail/blob/master/src/firejail/x11.c#L1103, which allows the binary to either be owned by root, or the root group. A more paranoid version would use `open`, `fstat` and `fexecve` to make sure ownership of the binary does not change between the `stat` and `exec` calls. On the other hand, manipulation of a binary already owned by root means the system is already compromised, so there's probably not much point to be careful about TOCTOU here. `netfiler.c` even skips the check altogether, and relies on `/sbin` being only writable by root.

Hopefully, we don't blindly execute a binary with root privileges anywhere else in the codebase.

_A more meta question: what is the policy with running external executable from firejail with root privilages (even if capability and seccomp sandboxed)?_ The potential for introducing setuid-related vulnerabilities is high. If we want to integrate additional external tools (like `xdg-dbus-proxy`), we should try to keep it reasonably safe.

### DIFF
--- a/src/firejail/dhcp.c
+++ b/src/firejail/dhcp.c
@@ -18,6 +18,7 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 #include "firejail.h"
+#include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <errno.h>
@@ -56,9 +57,9 @@ static const Dhclient dhclient6 = {
 	.arg_offset = offsetof(Bridge, arg_ip6_dhcp)
 };
 
-static void dhcp_run_dhclient(const Dhclient *client) {
+static void dhcp_run_dhclient(char *dhclient_path, const Dhclient *client) {
 	char *argv[256] = {
-		"dhclient",
+		dhclient_path,
 		client->version_arg,
 		"-pf", client->pid_file,
 		"-lf", client->leases_file,
@@ -115,8 +116,8 @@ static pid_t dhcp_read_pidfile(const Dhclient *client) {
 	return found;
 }
 
-static void dhcp_start_dhclient(const Dhclient *client) {
-	dhcp_run_dhclient(client);
+static void dhcp_start_dhclient(char *dhclient_path, const Dhclient *client) {
+	dhcp_run_dhclient(dhclient_path, client);
 	*(client->pid) = dhcp_read_pidfile(client);
 }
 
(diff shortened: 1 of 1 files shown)


src/firejail/dhcp.c
```diff
@@ -100,7 +101,7 @@ static pid_t dhcp_read_pidfile(const Dhclient *client) {
 			long pid;
 			if (fscanf(pidfile, "%ld", &pid) == 1) {
 				char *pidname = pid_proc_comm((pid_t) pid);
-				if (pidname && strcmp(pidname, "dhclient") == 0)
+				if (pidname && strcmp(pidname, dhclient_path))
```
smitsohu: you don't want equal strings any more?

src/firejail/dhcp.c
```diff
@@ -139,18 +140,32 @@ void dhcp_start(void) {
 	if (!any_dhcp())
 		return;
 
+	char *dhclient_path = "/sbin/dhclient";
+	struct stat s;
+	if (stat(dhclient_path, &s) == -1) {
+		dhclient_path = "/usr/sbin/dhclient";
+		if (stat(dhclient_path, &s) == -1) {
+			fprintf(stderr, "Error: dhclient was not found.\n");
+			exit(-1);
```
smitsohu: Can you return `1`? Just to keep it consistent.

smitsohu:
IMHO sbox is not really in the position to figure out what to execute. Maybe we should replace the `execvp` with `execv`?

src/firejail/dhcp.c
kris7t: Whoops, good catch!

kris7t: I don't know why I was fumbling around with this. `pid_proc_comm` returns `"dhclient"` even if it was started as `/sbin/dhclient`, so the previous check was actually correct.

kris7t:
Yeah, `execv` sounds much safer. We could even move the uid check to `sbox`, since it would receive the absolute path of the executable.

src/firejail/dhcp.c
kris7t: Done.

smitsohu:
> Hopefully, we don't blindly execute a binary with root privileges anywhere else in the codebase.

Quickly checked and there seem to be no other instances. Also means that replacing `execvp` with `execv` is indeed a trivial thing.

> We could even move the uid check to sbox, since it would receive the absolute path of the executable.

:+1:

smitsohu:
Merged, thanks!

smitsohu:
I'm going to remove the execvp in a separate patch.

kris7t:
@smitsohu  Oh, I am also trying to put together a patch for -- with some extra checking. Playing around with some symlinks, I think the paranoia with `fecexve` might be warranted, although I am still figuring out how to use it properly.

smitsohu:
@kris7t Go ahead! I did not start yet :)

PR #3241, opened 2020-02-23
Harden sbox_run by using fexecve instead of execvp

kris7t:
Based on the discussion in #3239, this patch aims to make `sbox_run` less likely to execute malicious binaries as root.

* The `execvp` call is replaces with `fexecve`, which does not perform PATH resolution. Because the environment was already cleared, this already wasn't a serious problem. But now, by passing absolute paths only to sbox_run, we get a bit more control.
* We check the binary to be owned by root (either the user or the group) and not be world-writable. As we are running system utilities, not having these would indicate a serious and dangerous misconfiguration, or some kind of an attack.
* The use of `open` with `O_PATH` and `fexecve` prevents TOCTOU errors with the permission check. This prevents a situation when the containing folder is somehow owned by a normal user (maybe by abusing some other setuid functionality to mount a tmpfs over it), and the binary we wish to run is replaced by a malicious one after we check its permissions.

### DIFF
--- a/src/firejail/dhcp.c
+++ b/src/firejail/dhcp.c
@@ -143,10 +143,6 @@ void dhcp_start(void) {
 			exit(1);
 		}
 	}
-	if (s.st_uid != 0 && s.st_gid != 0) {
-		fprintf(stderr, "Error: invalid dhclient executable\n");
-		exit(1);
-	}
 
 	EUID_ROOT();
 	if (mkdir(RUN_DHCLIENT_DIR, 0700))
--- a/src/firejail/dhcp.c
+++ b/src/firejail/dhcp.c
@@ -88,9 +88,6 @@ static void dhcp_run_dhclient(char *dhclient_path, const Dhclient *client) {
 static pid_t dhcp_read_pidfile(const Dhclient *client) {
 	// We have to run dhclient as a forking daemon (not pass the -d option),
 	// because we want to be notified of a successful DHCP lease by the parent process exit.
-	// However, try to be extra paranoid with race conditions,
-	// because dhclient only writes the daemon pid into the pidfile
-	// after its parent process has exited.
 	int tries = 0;
 	pid_t found = 0;
 	while (found == 0 && tries < 10) {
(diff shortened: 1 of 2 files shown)


smitsohu:
I found something weird. On recent kernels, `fexecve` won't work without access to `/dev/fd`. https://lore.kernel.org/patchwork/patch/517783 might provide some insights. 

In Firejail this can be provoked for example with `firejail --noprofile --debug --seccomp --whitelist=/dev/null`

It means that on a current kernels plain `execve` (with `/proc/self/fd/<fd>` as first argument) is actually the friendlier interface, as it only needs a mounted procfs and doesn't require us to make any assumptions about `/dev`.

kris7t:
That's quite interesting. For me,
```
firejail --noprofile --debug --seccomp --whitelist=/dev/null --net=virbr0 --ip=dhcp --ip6=dhcp --shell=/usr/bin/bash
```
(which executes plenty of programs with `sbox_run`, inclduing `fnet` and `dhclient`) with my patch works fine on Linux 5.5.5.

How does the command fail for your?

I assumed that `fexecve` would use `exeveat` to execute the program without looking at any files, purely by looking at the file descriptor table. Sure, the program gets `/proc/self/fd/3` as its `argv[0]`, so it cannot e.g. just exec `argv[0]` to restart itself. But that would fail anyways, since the fd passed to `fexecve` is normally `O_CLOEXEC`.

I guess more cautious approach would be to check whether `/proc` is indeed a procfs (and not some evildoer playing us with mountpoints) and try to `execve` `/proc/self/fd/<fd>`, or if `/proc` is not mounted, `fexecve` the fd instead.

smitsohu:
> How does the command fail for you?

```
sbox run: /usr/lib/firejail/fsec-print /run/firejail/mnt/seccomp/seccomp.32 
Dropping all capabilities
Drop privileges: pid 2, uid 1000, gid 1000, nogroups 1
No supplementary groups
fexecve: No such file or directory
Error: failed to run /usr/lib/firejail/fsec-print
Error: proc 5634 cannot sync with peer: unexpected EOF
Peer 5635 unexpectedly exited with status 1
```
That's what I get.

> Sure, the program gets /dev/fd/3 as its argv[0], so it cannot e.g. just exec argv[0] to restart itself. But that would fail anyways, since the fd passed to fexecve is normally O_CLOEXEC.

Right! Maybe the real reason is somewhere else, let me look again...

kris7t:
> ENOENT The close-on-exec flag is set on fd, and fd refers to a script.  See BUGS.

Maybe we should check whether we are trying to run a script. Still no clue why it doesn't fail for me.

smitsohu:
@kris7t 
> Maybe we should check whether we are trying to run a script, and print an appropriate error message.

On the other hand we control these commands, and if we want to execute a script, we can always just call the interpreter and pass the script as an argument. IMHO a comment in the source should be sufficient. If anything.

> ENOENT The close-on-exec flag is set on fd, and fd refers to a script. See BUGS.

Removing the O_CLOEXEC flag unfortunately doesn't fix it for me, and fsec-print is a compiled binary. I'll try to figure out what is going on, but it will take a bit of time (two, three weeks I guess).

PR #3242, opened 2020-02-24
integrate AppArmor with join options

smitsohu:


kris7t:
Looks good to me.

smitsohu:
@kris7t Sure, will do.

I was thinking to merge this next week, if there is no protest.

PR #3243, opened 2020-02-24
add lua support for mpv

glitsj16:
On Arch mpv needs access to lua:

```
$ pacman -Q mpv
mpv 1:0.32.0-1
$ firejail /usr/bin/mpv ~/Downloads/video.mp4
/usr/bin/mpv: error while loading shared libraries: liblua5.2.so.5.2: cannot open shared object file: Permission denied
```

On a side-note, Python support (enabled in the mpv profile) doesn't seem to be needed, at least not on Arch. Need to check on a Ubuntu machine to confirm.

### DIFF
--- a/etc/allow-lua.inc
+++ b/etc/allow-lua.inc
@@ -4,5 +4,6 @@ include allow-lua.local
 
 noblacklist ${PATH}/lua*
 noblacklist /usr/include/lua*
+noblacklist /usr/lib/liblua*
 noblacklist /usr/lib/lua
 noblacklist /usr/share/lua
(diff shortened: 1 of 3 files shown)


etc/allow-lua.inc
```diff
@@ -4,5 +4,6 @@ include allow-lua.local
 
 noblacklist ${PATH}/lua*
 noblacklist /usr/include/lua*
+noblacklist /usr/lib/liblua*
```
rusty-snake: Where is the blacklist? `grep "/usr/lib/liblua" etc/*.inc`

glitsj16: @rusty-snake I had that in my disable-interpreters.local and forgot about it. Thanks for the catch!

PR #3246, opened 2020-02-24
additional Lua blacklisting

glitsj16:
Follow-up for #3243.

### DIFF
--- a/etc/disable-interpreters.inc
+++ b/etc/disable-interpreters.inc
@@ -12,10 +12,12 @@ blacklist /usr/lib64/libgjs*
 
 # Lua
 blacklist ${PATH}/lua*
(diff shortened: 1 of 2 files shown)


etc/disable-interpreters.inc
```diff
@@ -12,10 +12,12 @@ blacklist /usr/lib64/libgjs*
 
 # Lua
 blacklist ${PATH}/lua*
+blacklist /usr/include/lauxlib.h
```
rusty-snake: `/etc/firejail/disable-devel.inc:blacklist /usr/include`

glitsj16: Should be fixed now, thanks.

PR #3251, opened 2020-02-27
add xournal.profile

eighthave:
xournal is a funny document editor that can do overlay edits on PDFs.  It is meant as a diary/journal note taking app.  It turns out it is a great app for overlaying images of signatures onto PDF contracts.  Since it means editing PDFs from outside sources, it should be firejailed.  Also, it needs very little access to work.

I'm new to this, so I didn't attempt _private-lib_ at all.  This should probably have a custom _private-lib_ entry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,4 +151,4 @@ We also keep a list of profile fixes for previous released versions in [etc-fixe
 
 ### New profiles:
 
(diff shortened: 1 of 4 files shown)


rusty-snake:
Merged, Thanks.

rusty-snake:
Merged, Thanks.

Issue #3250, opened 2020-02-27
conky needs lua


curiosityseeker:
Since a few days I've noticed that conky didn't start any more here on Arch Linux. Starting it in the console produced the error:

`/bin/conky: error while loading shared libraries: liblua.so.5.3: cannot open shared object file: Permission denied`

Adding `allow-lua.inc` to the profile solved the problem. (It just comes to my mind that I didn't check if `noblacklist /usr/lib/liblua*` would be sufficient ...)

Vincent43:
Adding `allow-lua.inc` to the profile should be  ok. Feel free to open PR.

rusty-snake:
> Adding `allow-lua.inc` to the profile solved the problem. (It just comes to my mind that I didn't check if `noblacklist /usr/lib/liblua*` would be sufficient ...)

I'm very sure that it solve the issue. Has you added it above the disable-* block?

curiosityseeker:
> I'm very sure that it solve the issue. 

Yes, it does - I just tried it.

> Has you added it above the disable-* block?

I added it to my conky.profile in {HOME}/.config/firejail above `include /etc/firejail/conky.profile`

PR #3255, opened 2020-02-28
conky needs lua

curiosityseeker:
See issue #3250

### DIFF
--- a/etc/conky.profile
+++ b/etc/conky.profile
@@ -8,6 +8,9 @@ include globals.local
 
 noblacklist ${PICTURES}
 
+# Allow lua (blacklisted by disable-interpreters.inc)
+include allow-lua.inc
+
 include disable-common.inc
 include disable-devel.inc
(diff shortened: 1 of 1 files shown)


etc/conky.profile
```diff
@@ -8,6 +8,8 @@ include globals.local
 
 noblacklist ${PICTURES}
 
+noblacklist /usr/lib/liblua*
```
reinerh: you could include allow-lua.inc.

Vincent43: Yes, even if not all lines from `allow-lua.inc` are needed using includes is cleaner for maintenance than  custom `noblacklist` rules in profile.

curiosityseeker: Well, I always try to keep the attack surface as small as possible. But if you prefer to add the inc file I'll change that.

rusty-snake:
Merged, Thanks.

PR #3256, opened 2020-02-28
support netcat variants

glitsj16:
Currently we only use ${PATH}/nc and ${PATH}/ncat to refer to `netcat`. Due to packaging philosophy this binary can be available under several other disguises. E.g. on Debian/Ubuntu it also exists as ${PATH}/nc.openbsd and ${PATH}/nc.traditional. Let's play safe with such a tool and add support for these known variants. Not sure if we would prefer to have it sandboxed via firecfg by default, so I left that aspect out of this PR.

I'll keep this open for a while to give people a chance to suggest other variants if they exist.

### DIFF
--- a/etc/disable-common.inc
+++ b/etc/disable-common.inc
@@ -391,6 +391,8 @@ blacklist ${PATH}/ksu
 blacklist ${PATH}/mount
 blacklist ${PATH}/mount.ecryptfs_private
 blacklist ${PATH}/nc
+blacklist ${PATH}/nc.openbsd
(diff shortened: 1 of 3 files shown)


reinerh:
I'm wondering if this is necessary. I thought firejail was resolving symlinks, so it would automatically blacklist e.g. nc.openbsd if nc is a symlink to it.

glitsj16:
@reinerh I stand corrected, firejail indeed does that!

Issue #3184, opened 2020-01-25
access to the system DBus


rusty-snake:
Thanks @jonleivent  for bringing #3078 up.

**Current situation:**
Profiles with `nodbus` have no access to `/run/dbus/system_bus_socket`, profiles without `nodbus` have access. The must profiles without `nodbus` didn't have it because they need access to the _session-bus_.

**Proposed change (soft):**
Adding a `nodbus.system` option. If only this option is specified, the program can access the session-bus but not the system-bus.

**Proposed change (hard):**

Firejail blocks system-bus access by default, unless an allow-system-dbus-access option is specified.

Vincent43:
I wonder if we could bring in [xdg-dbus-proxy](https://github.com/flatpak/xdg-dbus-proxy) integration for fine grained dbus access control instead if all or nothing.

rusty-snake:
Sure we should. However this is a lot more work and blocking the system-bus works for almost all programs and prevents them from doing "privileged" actions.

jonleivent:
Would it be possible to enclose a session bus within a jail with other applications, so that they can use it without interfering with applications and session busses outside the jail?

jonleivent:
I did some experimenting, and it is possible with some additional firejail protections on /run and setting XDG_RUNTIME_DIR to launch dbus-launch --exit-with-session icewm-session into a x11=xephyr firejail container.  Some testing shows that the session dbus inside the jail works.  I'm going to do more testing.  Anyone seen this configuration before?

rusty-snake:
untested: `firejail --noprofile --nodbus --private dbus-run-session APP`

ned64:
I have experimented with `dbus-run-session` but needed to add a new `/etc` and `/tmp` for that to work and isolate properly. Perhaps this could be integrated as `--new-dbus-session`?

jonleivent:
@ned64 
What in particular did you need to do to /etc and /tmp?

kris7t:
@Vincent43 Provided we had a convenient way to run `xdg-dbus-proxy`, what kind of policies would we want to enforce? I imagine usual suspects would be allowing notifications, and perhaps MPRIS or idle inhibition, depending on the app.

I am tempted to implement `xdg-dbus-proxy` support. It seems to have a straightforward format for filters (`\0`-separated options read from a given fd), but we will probably need something a bit more composable for profiles.

rusty-snake:
[...]

PR #3265, opened 2020-03-02
Fine-grained DBus sandboxing

kris7t:
This PR adds the `dbus-user` and `dbus-system` options to individually control access to the session and system DBus buses, as per #3184. Access policies to the buses can be `allow`, which completely allows the bus, `filter`, which runs [`xdg-dbus-proxy`](https://github.com/flatpak/xdg-dbus-proxy), and `none`, which disables access. The `nodbus` options, which is equivalent to `dbus-user none` `dbus-system none` is kept for compatibility.

Filter rules for `xdg-dbus-proxy` can be specified with the `dbus-user.talk`, `dbus-user.own`, `dbus-system.talk`, and `dbus-system.own` options. While `xdg-dbus-proxy` implements finer-grained rules, these four should hopefully be enough to convert filter rules from Flatpak manifests.

On the implementation side:
- The new filter rules are added to the profile similarly to `blacklist` and `whitelist` commands.
- `xdg-dbus-proxy` runs outside the sandbox namespace (to maintain access to the original DBus sockets), but contained by a modified version of `sbox_run`. It is linked to the parent firejail process by a pipe fd, closing which triggers its exit, so it (hopefully) quits when the parent firejail process exits for any reason.
- Filtered DBus sockets are in `/run/firejail/dbus` and are bind-mounted to their usual locations inside the sandbox. Even the system DBus socket is owned the user running the sandbox, not root, because if we `chown` it to root, xdg-dbus-proxy cannot clean them up when it exits (an the parent firejail process might not be around to clean the up instead). So inside the sandbox, the system DBus socket is owned by a normal user. This does not cause any problems when connecting to the bus, but might be used to detect the sandbox.

This PR does not contain any profile changes, so all profiles remain to constrain DBus access in a coarse-grained way. Hopefully, in the future, profiles can be updated to take advantage of DBus filtering, but that might require a hard dependency on xdg-dbus-proxy, or some auto-detection mechanism.

### DIFF
--- a/src/firejail/dbus.c
+++ b/src/firejail/dbus.c
@@ -33,10 +33,10 @@
 
 #define DBUS_SOCKET_PATH_PREFIX "unix:path="
 #define DBUS_USER_SOCKET_FORMAT "/run/user/%d/bus"
-#define DBUS_USER_SOCKET_PATH_FORMAT DBUS_SOCKET_PATH_PREFIX DBUS_USER_SOCKET_FORMAT
+#define DBUS_USER_SOCKET_FORMAT2 "/run/user/%d/dbus/user_bus_socket"
 #define DBUS_SYSTEM_SOCKET "/run/dbus/system_bus_socket"
-#define DBUS_SYSTEM_SOCKET_PATH DBUS_SOCKET_PATH_PREFIX DBUS_SYSTEM_SOCKET
 #define DBUS_SESSION_BUS_ADDRESS_ENV "DBUS_SESSION_BUS_ADDRESS"
+#define DBUS_SYSTEM_BUS_ADDRESS_ENV "DBUS_SYSTEM_BUS_ADDRESS"
 #define DBUS_USER_DIR_FORMAT RUN_FIREJAIL_DBUS_DIR "/%d"
 #define DBUS_USER_PROXY_SOCKET_FORMAT DBUS_USER_DIR_FORMAT "/%d-user"
 #define DBUS_SYSTEM_PROXY_SOCKET_FORMAT DBUS_USER_DIR_FORMAT "/%d-system"
@@ -170,6 +170,36 @@ static void dbus_create_user_dir(void) {
 	free(path);
 }
 
+static char *find_user_socket_by_format(char *format) {
+	char *dbus_user_socket;
+	if (asprintf(&dbus_user_socket, format, (int) getuid()) == -1)
+		errExit("asprintf");
+	struct stat s;
+	if (stat(dbus_user_socket, &s) == -1) {
+		if (errno == ENOENT)
+			goto fail;
+		return NULL;
+		errExit("stat");
+	}
+	if (!S_ISSOCK(s.st_mode))
+		goto fail;
+	return dbus_user_socket;
+ fail:
+	free(dbus_user_socket);
+	return NULL;
+}
+
+static char *find_user_socket(void) {
+	char *socket1 = find_user_socket_by_format(DBUS_USER_SOCKET_FORMAT);
+	if (socket1 != NULL)
+		return socket1;
+	char *socket2 = find_user_socket_by_format(DBUS_USER_SOCKET_FORMAT2);
+	if (socket2 != NULL)
+		return socket2;
+	fprintf(stderr, "DBus user socket was not found.\n");
+	exit(1);
+}
+
 void dbus_proxy_start(void) {
 	dbus_create_user_dir();
 
@@ -203,26 +233,35 @@ void dbus_proxy_start(void) {
 			errExit("close");
 
 		if (arg_dbus_user == DBUS_POLICY_FILTER) {
-			char *dbus_user_path_env = getenv(DBUS_SESSION_BUS_ADDRESS_ENV);
-			if (dbus_user_path_env == NULL) {
-				write_arg(args_pipe[1], DBUS_USER_SOCKET_PATH_FORMAT, getuid());
+			char *user_env = getenv(DBUS_SESSION_BUS_ADDRESS_ENV);
+			if (user_env == NULL) {
+				char *dbus_user_socket = find_user_socket();
+				write_arg(args_pipe[1], DBUS_SOCKET_PATH_PREFIX "%s",
+						  dbus_user_socket);
+				free(dbus_user_socket);
 			} else {
-				write_arg(args_pipe[1], dbus_user_path_env);
+				write_arg(args_pipe[1], "%s", user_env);
 			}
 			if (asprintf(&dbus_user_proxy_socket, DBUS_USER_PROXY_SOCKET_FORMAT,
 						 (int) getuid(), (int) getpid()) == -1)
 				errExit("asprintf");
-			write_arg(args_pipe[1], dbus_user_proxy_socket);
+			write_arg(args_pipe[1], "%s", dbus_user_proxy_socket);
 			write_arg(args_pipe[1], "--filter");
 			write_profile(args_pipe[1], "dbus-user.");
 		}
 
 		if (arg_dbus_system == DBUS_POLICY_FILTER) {
-			write_arg(args_pipe[1], DBUS_SYSTEM_SOCKET_PATH);
+			char *system_env = getenv(DBUS_SYSTEM_BUS_ADDRESS_ENV);
+			if (system_env == NULL) {
+				write_arg(args_pipe[1],
+						  DBUS_SOCKET_PATH_PREFIX DBUS_SYSTEM_SOCKET);
+			} else {
+				write_arg(args_pipe[1], "%s", system_env);
+			}
 			if (asprintf(&dbus_system_proxy_socket, DBUS_SYSTEM_PROXY_SOCKET_FORMAT,
 						 (int) getuid(), (int) getpid()) == -1)
 				errExit("asprintf");
-			write_arg(args_pipe[1], dbus_system_proxy_socket);
+			write_arg(args_pipe[1], "%s", dbus_system_proxy_socket);
 			write_arg(args_pipe[1], "--filter");
 			write_profile(args_pipe[1], "dbus-system.");
 		}
@@ -294,6 +333,16 @@ static void socket_overlay(char *socket_path, char *proxy_path) {
 	close(fd);
 }
 
+static char *get_socket_env(const char *name) {
+	char *value = getenv(name);
+	if (value == NULL)
+		return NULL;
+	if (strncmp(value, DBUS_SOCKET_PATH_PREFIX,
+				strlen(DBUS_SOCKET_PATH_PREFIX)) == 0)
+		return value + strlen(DBUS_SOCKET_PATH_PREFIX);
+	return NULL;
+}
+
 static void disable_socket_dir(void) {
 	struct stat s;
 	if (stat(RUN_FIREJAIL_DBUS_DIR, &s) == 0)
@@ -314,36 +363,41 @@ void dbus_apply_policy(void) {
 		return;
 	}
 
-	char *dbus_new_user_socket_path;
-	if (asprintf(&dbus_new_user_socket_path, DBUS_USER_SOCKET_PATH_FORMAT, getuid()) == -1)
-		errExit("asprintf");
-	char *dbus_new_user_socket = dbus_new_user_socket_path + strlen(DBUS_SOCKET_PATH_PREFIX);
-	char *dbus_user_path_env = getenv(DBUS_SESSION_BUS_ADDRESS_ENV);
-	char *dbus_orig_user_socket_path;
-	if (dbus_user_path_env != NULL
-		&& strncmp(DBUS_SOCKET_PATH_PREFIX, dbus_user_path_env, strlen(DBUS_SOCKET_PATH_PREFIX)) == 0) {
-		dbus_orig_user_socket_path = dbus_user_path_env;
-	} else {
-		dbus_orig_user_socket_path = dbus_new_user_socket_path;
-	}
-	char *dbus_orig_user_socket = dbus_orig_user_socket_path + strlen(DBUS_SOCKET_PATH_PREFIX);
+	create_empty_dir_as_root(RUN_DBUS_DIR, 0755);
+	create_empty_file_as_root(RUN_DBUS_USER_SOCKET, 0700);
+	create_empty_file_as_root(RUN_DBUS_SYSTEM_SOCKET, 0700);
 
 	if (arg_dbus_user != DBUS_POLICY_ALLOW) {
 		if (arg_dbus_user == DBUS_POLICY_FILTER) {
 			assert(dbus_user_proxy_socket != NULL);
-			socket_overlay(dbus_new_user_socket, dbus_user_proxy_socket);
+			socket_overlay(RUN_DBUS_USER_SOCKET, dbus_user_proxy_socket);
 			free(dbus_user_proxy_socket);
-		} else { // arg_dbus_user == DBUS_POLICY_BLOCK
-			disable_file_or_dir(dbus_new_user_socket);
 		}
 
-		if (strcmp(dbus_orig_user_socket, dbus_new_user_socket) != 0)
-			disable_file_or_dir(dbus_orig_user_socket);
+		char *dbus_user_socket;
+		if (asprintf(&dbus_user_socket, DBUS_USER_SOCKET_FORMAT,
+					 (int) getuid()) == -1)
+			errExit("asprintf");
+		disable_file_or_dir(dbus_user_socket);
 
-		// set a new environment variable:
-		// DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/<UID>/bus
-		if (setenv(DBUS_SESSION_BUS_ADDRESS_ENV, dbus_new_user_socket_path, 1) == -1) {
-			fprintf(stderr, "Error: cannot modify " DBUS_SESSION_BUS_ADDRESS_ENV " required by --dbus-user\n");
+		char *dbus_user_socket2;
+		if (asprintf(&dbus_user_socket2, DBUS_USER_SOCKET_FORMAT2,
+					 (int) getuid()) == -1)
+			errExit("asprintf");
+		disable_file_or_dir(dbus_user_socket2);
+
+		char *user_env = get_socket_env(DBUS_SESSION_BUS_ADDRESS_ENV);
+		if (user_env != NULL && strcmp(user_env, dbus_user_socket) != 0 &&
+			strcmp(user_env, dbus_user_socket2) != 0)
+			disable_file_or_dir(user_env);
+
+		free(dbus_user_socket);
+		free(dbus_user_socket2);
+
+		if (setenv(DBUS_SESSION_BUS_ADDRESS_ENV,
+				   DBUS_SOCKET_PATH_PREFIX RUN_DBUS_USER_SOCKET, 1) == -1) {
+			fprintf(stderr, "Error: cannot modify " DBUS_SESSION_BUS_ADDRESS_ENV
+							" required by --dbus-user\n");
 			exit(1);
 		}
 
@@ -355,14 +409,25 @@ void dbus_apply_policy(void) {
 		free(path);
 	}
 
-	free(dbus_new_user_socket_path);
+	if (arg_dbus_system != DBUS_POLICY_ALLOW) {
+		if (arg_dbus_system == DBUS_POLICY_FILTER) {
+			assert(dbus_system_proxy_socket != NULL);
+			socket_overlay(RUN_DBUS_SYSTEM_SOCKET, dbus_system_proxy_socket);
+			free(dbus_system_proxy_socket);
+		}
 
-	if (arg_dbus_system == DBUS_POLICY_FILTER) {
-		assert(dbus_system_proxy_socket != NULL);
-		socket_overlay(DBUS_SYSTEM_SOCKET, dbus_system_proxy_socket);
-		free(dbus_system_proxy_socket);
-	} else if (arg_dbus_system == DBUS_POLICY_BLOCK) {
 		disable_file_or_dir(DBUS_SYSTEM_SOCKET);
+
+		char *system_env = get_socket_env(DBUS_SYSTEM_BUS_ADDRESS_ENV);
+		if (system_env != NULL && strcmp(system_env, DBUS_SYSTEM_SOCKET) != 0)
+			disable_file_or_dir(system_env);
+
+		if (setenv(DBUS_SYSTEM_BUS_ADDRESS_ENV,
+				   DBUS_SOCKET_PATH_PREFIX RUN_DBUS_SYSTEM_SOCKET, 1) == -1) {
+			fprintf(stderr, "Error: cannot modify " DBUS_SYSTEM_BUS_ADDRESS_ENV
+							" required by --dbus-system\n");
+			exit(1);
+		}
 	}
 
 	// Only disable access to /run/firejail/dbus here, when the sockets have been bind-mounted.
--- a/src/firejail/dbus.c
+++ b/src/firejail/dbus.c
@@ -22,9 +22,14 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <errno.h>
 #include <unistd.h>
 #include <fcntl.h>
 #include <string.h>
+
+#ifndef O_PATH
+#define O_PATH 010000000
+#endif
 
 #define DBUS_SOCKET_PATH_PREFIX "unix:path="
 #define DBUS_USER_SOCKET_FORMAT "/run/user/%d/bus"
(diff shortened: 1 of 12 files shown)


kris7t:
I have been using this for a couple of days with local profiles and it seems to work fine. For example,
```
dbus-user filter
dbus-user.own org.mpris.MediaPlayer2.spotify
dbus-system none
```
Lets Spotify display notifications and be controller over MPRIS.

However, the situation is not as rosy with e.g. Electron based apps. They need a rather permissive
```
dbus-user.own org.kde.*
```
because they need to own names in the form `org.kde.StatusNotifierItem-40-819` (with the two numbers variable) to display tray icons -- tray icons and Chromium are quite wonky, anyways, since it writes they tray icons directly into `/tmp` (and thus cannot be run with `private-tmp`).

I also tried to use desktop portals with
```
dbus-user.talk org.freedesktop.portal.*
```
and `GTK_USE_PORTAL=1`, but had no success yet. Perhaps I am missing some components on the DE side.

Vincent43:
> However, the situation is not as rosy with e.g. Electron based apps. They need a rather permissive
> 
> dbus-user.own org.kde.*
> 
> because they need to own names in the form org.kde.StatusNotifierItem-40-819 (with the two numbers variable) to display tray icons -- tray icons and Chromium are quite wonky, anyways, since it writes they tray icons directly into /tmp (and thus cannot be run with private-tmp).

Hm, I knew about those tray icon issues for qt apps but electron apps (which are more gtk than qt) in flatpak are able to show tray icon with just` --talk-name=org.kde.StatusNotifierWatcher`

The /tmp issue may be fixed with `TMPDIR` env variable. Perhaps pointing it to `/run/user/xxxx` aka `$RUNUSER` would work most widely as it's shared with host most of the time.

> I also tried to use desktop portals with
> 
> dbus-user.talk org.freedesktop.portal.*
> 
> and GTK_USE_PORTAL=1, but had no success yet. Perhaps I am missing some components on the DE side.

Installing xdg-desktop-portal & xdg-desktop-portal-{gtk,kde} and enabling systemd user services (if they aren't already):
```
xdg-desktop-portal.service
xdg-document-portal.service
xdg-permission-store.service
```
should be sufficient. Did you tested apps that actually support portals? AFAIK chromium/electron doesn't. Note that for Qt apps you need to use `SNAP=1` instead of `GTK_USE_PORTAL=1`

kris7t:
> Hm, I knew about those tray icon issues for qt apps but electron apps (which are more gtk than qt) in flatpak are able to show tray icon with just` --talk-name=org.kde.StatusNotifierWatcher`
I tried that, but failed to display an icon & complained about not owning the well-known name for the icon. Maybe because I am trying to run Wavebox 10, which is not "really" an Electron app (more like a forked Chromium, with some Electron bits).

> The /tmp issue may be fixed with `TMPDIR` env variable. Perhaps pointing it to `/run/user/xxxx` aka `$RUNUSER` would work most widely as it's shared with host most of the time.
Thanks for these pointers!
 
> Did you tested apps that actually support portals? AFAIK chromium/electron doesn't.
Oh that's a pity.

kris7t:
> manpage: example filter

Thanks, good catch!

> For what is dbus-* allow? Not using dbus-* at all will also allow D-Bus access.

It seemed to be the logical thing to do. On the other hand, maybe only `--dbus-*=filter` and `--dbus-*=none` should be supported (and `DBUS_POLICY_ALLOW` should only be the default value for the `arg_dbus_*` variables in the program, not something user selectable).

> just noting: nodbus vs. dbus-* allow: the later will win

One way to fix this would be to force the policies to the always _strengthened_ by subsequent profile entries, i.e., `--nodbus --dbus-user=filter` should throw an error. Any thoughts on whether this behavior would be preferable?

> future of HAS_NODBUS?

Logically, a sandbox with blocked DBus `HAS_NODBUS`, but maybe there should be conditionals for other settings, too.

rusty-snake:
More conditions? Sure. I mean what will/should be the behaviour for e.g.
```
dbus-user allow
dbus-system none
?HAS_NODBUS: blacklist /etc/fstab
```

kris7t:
That's a good question. Searching through the firejail repo in order to find uses of ?HAS_NODBUS that are hopefully similar to uses in other profiles in the wild, I could basically find two:
* In https://github.com/netblue30/firejail/blob/df1c73a00f68b3ee2503b75d3220e65f99a7f760/etc/disable-common.inc, the ksocket is blacklisted when DBus is blacklisted to tighten the sandbox. In this case, we should probably apply the ?HAS_NODBUS condition if any kind of DBus filtering is in effect (since we don't filter the ksocket).
* In https://github.com/netblue30/firejail/blob/32c3669115a7168e5a7fa13347bd6f8daf838be0/etc/chromium-common.profile, ?HAS_NODBUS is used to fix a regression when DBus is not available. If the apply ?HAS_NODBUS even if there is some DBus access, at worst we apply an unnecessary workaround, but do not degrade functionality (compared to the case when DBus was not available at all).

To sum up, setting ?HAS_NODBUS if any DBus filtering or blocking is in effect is safe for profiles which use it for optionally tightening the sandbox, while does not degrade profiles that use it to apply workarounds. Profiles of the latter kind could be eventually updated to use fine-grained filters to avoid the regressions in the first place.

smitsohu:
Naive question: Would it be possible to run the proxy early as a child of the sandbox process?

EDIT: There is actually not much of a difference, but running it inside all of the sandbox namespaces maybe could add a bit extra peace of mind.

smitsohu:
> Naive question: Would it be possible to run the proxy early as a child of the sandbox process?

Or maybe that's not a good idea. If it runs in the same pid namespace, it can be vulnerable to attacks with ptrace, which the current design avoids.

src/firejail/preproc.c
```diff
@@ -58,6 +58,10 @@ void preproc_build_firejail_dir(void) {
 		create_empty_dir_as_root(RUN_FIREJAIL_X11_DIR, 0755);
 	}
 
+	if (stat(RUN_FIREJAIL_DBUS_DIR, &s)) {
+		create_empty_dir_as_root(RUN_FIREJAIL_DBUS_DIR, 01777);
```
smitsohu: I wonder if we can somehow avoid the world-writable directory. Maybe per user subdirectories?
Might be also good to mount `RUN_FIREJAIL_DBUS_DIR` noexec.

src/man/firejail.txt
```diff
@@ -325,6 +325,112 @@ $ firejail \-\-list
...
+.br
+
+.br
+The \fBfilter\fR policy enables the system DBus filter. This option requires
+installing the xdg-dbus-proxy utility. Permissions for well-known can be
+specified with the --dbus-system.talk and --dbus-system.own options.
+.br
+
+.br
+The \fBnone\fR policy disables access to the sesson DBus.
```
rusty-snake:
```suggestion
The \fBnone\fR policy disables access to the session DBus.
```

Same for dbus-user.

src/firejail/dbus.c
```diff
@@ -18,41 +18,310 @@
...
+
+int dbus_check_name(const char *name) {
+	unsigned long length = strlen(name);
+	if (length == 0 || length > DBUS_MAX_NAME_LENGTH)
+		return 0;
+	const char *p = name;
+	int segments = 1;
+	int in_segment = 0;
+	while (*p) {
+		int alpha = (*p >= 'a' && *p <= 'z') || (*p >= 'A' && *p <= 'Z');
```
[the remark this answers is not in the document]
kris7t: Sorry, I dont' really understand what you are saying.

Originally, I was using `isalpha`, but realized that it takes locale into account (so potentially allows some non-ascii characters). However, https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus says that
> A bus name is of type STRING, meaning that it must be valid UTF-8. However, [...] Each element must only contain the ASCII characters "[A-Z][a-z][0-9]_-", with "-" discouraged in new bus names.

So bus names are explicitly UTF-8, but restricted to the subset of UTF-8 that coincides with ASCII alphanumericals and punctuation. In that case, the the lower and upper case letters are contiguous. Of course, it is theoretically possible that some locale is set where characters are not contiguous, but I can't think of any what would be usable as locale under Linux (stuff like UTF-8, ISO 8859 and Shift JIS are ASCII compatible regarding alphanumerical character). An admittedly superficial search confirms that this compatibility is pretty much required for running programs: https://unix.stackexchange.com/questions/102040/how-to-add-ebcdic-support-on-debian-ubuntu

Nevertheless, I really don't like this conditional, but couldn't find a more POSIXy way without having to deal with locales.

src/firejail/dbus.c
```diff
@@ -18,41 +18,310 @@
...
+	char *dbus_new_user_socket = dbus_new_user_socket_path + strlen(DBUS_SOCKET_PATH_PREFIX);
+	char *dbus_user_path_env = getenv(DBUS_SESSION_BUS_ADDRESS_ENV);
+	char *dbus_orig_user_socket_path;
+	if (dbus_user_path_env != NULL
+		&& strncmp(DBUS_SOCKET_PATH_PREFIX, dbus_user_path_env, strlen(DBUS_SOCKET_PATH_PREFIX)) == 0) {
+		dbus_orig_user_socket_path = dbus_user_path_env;
+	} else {
+		dbus_orig_user_socket_path = dbus_new_user_socket_path;
 	}
+	char *dbus_orig_user_socket = dbus_user_path_env + strlen(DBUS_SOCKET_PATH_PREFIX);
```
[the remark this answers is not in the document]
kris7t: Thanks! Yeah, that should be `dbus_new_user_socket_path`. Fixing it.

src/firejail/dbus.c
```diff
@@ -18,41 +18,310 @@
...
+		dbus_user_proxy_socket = NULL;
+	}
+	if (dbus_system_proxy_socket != NULL) {
+		free(dbus_system_proxy_socket);
+		dbus_system_proxy_socket = NULL;
+	}
+}
+
+static void socket_overlay(char *socket_path, char *proxy_path) {
+	if (mount(proxy_path, socket_path, NULL, MS_BIND, NULL) == -1)
```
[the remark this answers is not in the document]
kris7t: Sure! The solution from `fs_x11` will work fine. It's a pity there is no `fmount` system call, though...

src/firejail/main.c
```diff
@@ -2812,6 +2885,9 @@ printf("**********************************\n");
 	// end of signal-safe code
 	//*****************************
 
+	// stop dbus proxy (if any)
+	dbus_proxy_stop();
```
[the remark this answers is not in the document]
kris7t: Yeah, this belongs to `myexit`. Moving it.

src/firejail/dbus.c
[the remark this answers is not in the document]
smitsohu: > it is theoretically possible that some locale is set where characters are not contiguous, but I can't think of any what would be usable as locale under Linux

Me neither. I was indeed thinking of `isalpha`, which would work in combination with the C locale that Firejail uses. But in the glibc implementation of the C locale the alphabetic characters are a contiguous stretch anyway. Feel free to ignore the comment.

src/firejail/preproc.c
kris7t: Mounting a tmpfs noexec on `RUN_FIREJAIL_DBUS_DIR` and creating per-uid directories will work fine, I think.

kris7t: However, there is a potential race condition if we mount `RUN_FIREJAIL_DBUS_DIR` in `preproc_build_firejail_dir`: a concurrent firejail process might have created and mounted the directory. `create_empty_dir_as_root` already handles this race by not failing on `EEXIST`, but we might potentially mount tmps twice.

smitsohu: Why the tmpfs? I thought we can just remount it as is.

My impression is that we don't need to worry too much about the tiny race window. In the worst case there are one (or few) extra entries in the mount table, there won't be any other real world effect.

kris7t: Ah, indeed, a simple remount works too.

kris7t: But `RUN_FIREJAIL_DBUS_DIR` is not a mountpoint, only `RUN_FIREJAIL_BASEDIR` is, isn't it?

smitsohu: > But RUN_FIREJAIL_DBUS_DIR is not a mountpoint

Right, you need to make it a mount point first (bind-mount it onto itself) before you can do the remount.

kris7t: Oh, I see what you mean now.

src/man/firejail.txt
kris7t: Thanks, fixed these.

kris7t:
I revised the options `--dbus-user` and `--dbus-system`:
* Not there is not `allow` policy, only `filter` and `block`.
* The policy cannot be relaxed, only tightened. If we set `--nodbus`, then `--dbus-user=filter` will fail. Nevertheless, a `.local` profile can override the built-in `nodbus` command by e.g. `ignore nodbus` and `dbus-user filter`, `dbus-system block`.
* Adding a filter rule requires the `filter` or `block` policy. The `block` policy in conjunction with filter rules will produce a warning.

I am still not entirely satisfied by this design: to relax a profile from filtering to DBus allowed in a `.local` profile, we not only need to `ignore dbus-user filter`, but also ignore all filter rules. I don't know whether that's a good thing. On the one hand, it's pretty convenient. On the other hand, this forces the user to review any changes to filter rules in the main profile (and decide whether leaving DBus completely open is still warranted).

src/firejail/dbus.c
```diff
@@ -18,41 +18,355 @@
...
+	if (fstat(fd, &s) == -1)
+		errExit("fstat");
+	if (!S_ISSOCK(s.st_mode)) {
+		errno = ENOTSOCK;
+		errExit("mounting DBus proxy socket");
+	}
+	char *proxy_fd_path;
+	if (asprintf(&proxy_fd_path, "/proc/self/fd/%d", fd) == -1)
+		errExit("asprintf");
+	if (mount(proxy_path, socket_path, NULL, MS_BIND | MS_REC, NULL) == -1)
```
smitsohu: Would it be possible to instead set the `DBUS_SESSION_BUS_ADDRESS` environment variable to the proxy socket and blacklist the original session bus socket (or somehow avoid this mount on the session bus socket in another way)?

My corner case goes like this: Because some time passes before the mount is called, a user could replace `socket_path` with a link and mount `proxy_path` wherever he wants, including to places without search or write permission. Then a second race starts to control xdg-dbus-proxy by means of ptrace and friends. This possibly leads to issues if a tool with file capabilities or the setuid bit set has a socket path hard coded that it reasonably expects to be root controlled.

kris7t: Clearly, I wasn't at my best when writing this, since I mean to mount `proxy_fd_path` to `socket_path`, of course, not `proxy_path`. But your point still stands eminently: the user can mount `proxy_`(`fd_`)`path` wherever she wishes, if `socket_path` is a symbolic link. This isn't even that racy, because the proxy will only try to connect to `socket_path` one a connection to the proxy socket is made -- that's plenty of time to replace the link with a real socket. Even without ptrace, this could be used to effectively take over the system DBus (at least within the sandbox).

Updating `DBUS_SESSION_BUS_ADDRESS` (and hoping no program has hardcoded `/run/user/<uid>/bus`) is probably the best idea.

smitsohu: > By the way, don't we have the same problem with the X11 socket?

I think it's a little bit different, because the mount target is a file we create ourselves in an empty tmpfs, and this tmpfs doesn't exist in the parent mount namespace. The implicit assumption is that no unprivileged process has access to the new sandbox mount namespace at this timepoint. If this assumption turns out to be wrong, the X11 socket whitelisting has exactly the same problem, yes.

As an additional hardening measure we could consider removing the `fchown` call:
https://github.com/netblue30/firejail/blob/c96efe7d0af04414ab3c9cd1256afd6ee678ff95/src/firejail/x11.c#L1308
This would mean the mount target is always owned by root, and in combination with the sticky bit on `/tmp/.X11-unix` that no unprivileged user can play with it.

smitsohu: > By the way, don't we have the same problem with the X11 socket?

I'm going to prepare a patch to remove that attack surface

kris7t: Sorry for not getting back to this earlier, COVID-19 lockdown had added some extra tasks to my to-do list. Anyways, I realized that we _must_ bind-mount the proxy socket somewhere, so that we can bind-mount any empty directory over `/run/firejail/dbus/`. Otherwise, sandboxed apps will still be able to access the proxy sockets of other apps, which may have more permissive filters. So I guess one way to handle this is:
1. Bind-mount an empty file over the original `DBUS_SESSION_BUS_ADDRESS`
2. Bind-mount the session bus proxy socket to somewhere in `/run/firejail/mnt/` and set `DBUS_SESSION_BUS_ADDRESS accordingly.
3. Bind-mound the system bus socket to its usual place. This is safe, because `/run/dbus/system_bus_socket` is owned by root.
4. Bind-mount an empty directory over `/run/firejail/dbus/`.

Out of these, step 2 may still be problematic and suffer from the same issue as the X11 socket. However, the target of the bind-mount at step 2 may be owned by root, because only the proxy socket itself has to be owned by the user (to allow `xdg-dbus-proxy` clean up after itself when exiting). The bind-mount will be cleaned up without intervention from a user process when the mount namespace is destroyed.

smitsohu: I think that sounds good.

abranson:
Thanks for this work - this is exactly what I've been looking for! One note -  some systemd based systems use a different path for the user dbus socket: /run/user/%d/dbus/user_bus_socket. These should have the correct path set in the DBUS_SESSION_BUS_ADDRESS env var. Perhaps this should be checked first, then if it's not set fallback to /run/user/%d/bus.

kris7t:
@rusty-snake @glitsj16 What do you think about the interface for profiles? With the changes (https://github.com/netblue30/firejail/pull/3265#issuecomment-596095652) it is hopefully a bit harder to misuse, and compatibility with `nodbus` profiles should still be preserved.

src/firejail/usage.c
```diff
@@ -53,6 +53,12 @@ static char *usage_str =
 #endif
 	"    --cpu=cpu-number,cpu-number - set cpu affinity.\n"
 	"    --cpu.print=name|pid - print the cpus in use.\n"
+	"    --dbus-system=allow|filter|none - set system DBus access policy.\n"
```
rusty-snake:
```suggestion
	"    --dbus-system=filter|none - set system DBus access policy.\n"
```

src/firejail/usage.c
```diff
@@ -53,6 +53,12 @@ static char *usage_str =
...
+	"    --dbus-system.own=name - allow ownership of name on the system DBus.\n"
+	"    --dbus-system.talk-name - allow talking to name on the system DBus.\n"
+	"    --dbus-user=allow|filter|none - set session DBus access policy.\n"
```
rusty-snake:
```suggestion
	"    --dbus-user=filter|none - set session DBus access policy.\n"
```

kris7t:
@rusty-snake Currently, firejail will `exit(1)` if `/usr/bin/xdg-dbus-proxy` is not found. Alternatively, we could do one of the following:

1. We could fall back to `--*-dbus=none` in that case, which is always safe (but could nevertheless break applications, if we update profiles without `nodbus` to e.g., `dbus-user.talk`).
2. We could fall back to ignoring `--*-dbus`, which is not safe, but won't break any profile that currently does not have `nodbus`.
3. In addition to 1 or 2, we could add a conditonal (e.g., `HAS_DBUS_FILTER` or `NO_HAS_DBUS_FILTER`) to control fallback behavior per profile. For example, if we chose blocking DBus access as the fallback behavior, we can add
```
?HAS_DBUS_FILTER: dbus-user filter
?HAS_DBUS_FILTER: dbus-user.talk some.name.*
```
to profiles that currently does not have `nodbus`.

However, 3 hints at the suboptimality of the current filter rule design: if e.g.). `dbus-user.talk` causes an error unless `dbus-user filter` is set, it is very inconvenient to conditionally enable filters (we must repeat `?HAS_DBUS_FILTER:`). Maybe we should downgrade that error to a warning.

rusty-snake:
I would prefer 2 with a warning and dependencies for the deb/rpm packages.

kris7t:
(By fallback behavior, I was talking about how to handle `--dbus-*=filter`. Of course, profiles without any DBus commands would not change their behavior unless we add the DBus filter command later. But it is prudent to think about how users without `xdg-dbus-proxy` would be affected by such profile changes.)

rusty-snake:
[firejail_dbus_permissions_from_flathub.tar.xz__for_github__.zip](https://github.com/netblue30/firejail/files/4403404/firejail_dbus_permissions_from_flathub.tar.xz__for_github__.zip)

kris7t:
@rusty-snake I updated the behavior to option 2:  filters get downgraded to allowing everything (with a warning) if `xdg-dbus-proxy` is not installed. A warning is emitted if there are ignored filter rules (either because we allow or block everything, or because we got downgraded because there is no proxy installed).

glitsj16:
@kris7t Just a FYI, I'm following the work you have been doing here without explicitly participating in the discussion. As @rusty-snake is on to it, we should be fine to adjust profiles when this gets merged. Very much appreciated!

smitsohu:
Apologies for coming back, just one more idea: Would it make sense to run dbus_proxy_start with dropped privileges and then start the proxy in its own Firejail sandbox?

kris7t:
@smitsohu Yeah, I was thinking about that, since the proxy only needs access to `/run/firejail/dbus/<uid>/`, the original DBus sockets, and not much else. So sandboxing it would definitely be neat. Maybe we could even have a `xdg-dbus-proxy.profile` and just exec `/proc/self/exe` with some special arguments (adding the original sockets to `--whitelist`, and setting `--net=none` if neither of the original sockets is an abstract socket).

One compilcation is the handling of the file descriptors, though. Perhaps we could add a general-purpose `--keep-fd=` flag to pass open file descriptors to the sandbox (close them in the main firejail process, but keep them open in the forked namespace).

On the other hand, `xdg-dbus-proxy` is running within a seccomp sandbox already (via `sbox_exec`). I don't know whether having a dedicated namespace for the proxy improves security at all, since if you can take over the proxy in any way, you are already able to send any DBus message to session bus you want. That's plenty of tools to escape, e.g., in Gnome, you can spawn a new Terminal with any command, or run any javascript code in the Shell.

smitsohu:
@kris7t 

> I don't know whether having a dedicated namespace for the proxy improves security at all, since if you can take over the proxy in any way, you are already able to send any DBus message to session bus you want. That's plenty of tools to escape, e.g., in Gnome, you can spawn a new Terminal with any command, or run any javascript code in the Shell.

You're right, I didn't think of that. The benefits of a full sandbox would be limited.

rusty-snake:
@kris7t I'm with @glitsj16 once this is in master the profile can be modified.

As a follow-up: xdg-dbus-proxys --log argument. It logs all the dbus-traffic to stdout (needs to be redirected). Either as --dbus-log (just a simple (maybe raw) log) or better as `--dbus-tracelog` to match any traffic against the policy in the profile. This could be very help-full while writing profiles or in issue about broken profiles because e.g. gtk only throws a `Failed to register: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: org.freedesktop.DBus.Error.ServiceUnknown` and does not tell which path it can't register.

rusty-snake:
@kris7t can you rebase.

kris7t:
@rusty-snake done.

src/man/firejail-profile.txt
```diff
@@ -522,12 +570,6 @@ Disable 3D hardware acceleration.
 Disable automatic ~/.config/pulse init, for complex setups such as remote
 pulse servers or non-standard socket paths.
 .TP
-\fBnodbus
-Disable D-Bus access. Only the regular UNIX socket is handled by
-this command. To disable the abstract socket, you would need to
-request a new network namespace using the net command. Another
-option is to remove unix from protocol set.
-.TP
 \fBnodvd
```
Fred-Barclay, line 573: `nodbus` would still work and be in the existing profiles, right? If so I'd prefer to leave these lines in as documentation, but mark as deprecated... i.e.
`\fBnodbus (deprecated)`
...

Then once we phase `nodbus` out off **all** profiles, let's remove.


EDIT: or do like the main firejail.txt documentation, line 1280, and specify what nodbus is equivalent to.

rusty-snake: it was moved: see line 498

rusty-snake: > Then once we phase nodbus out off all profiles, let's remove.

I would keep it for one release and add a deprecated warning. D-Bus can be used to escape the sandbox in several ways and users might have it in their .locals or custom profiles.

kris7t: I marked the `--nodbus` option and the `nodbus` command as deprecated in the manpages.

kris7t:
Big thanks to everyone for the reviews and comments!

As a next step, I'll work on adding DBus logging (https://github.com/netblue30/firejail/pull/3265#issuecomment-609477194) to help profile creation.

PR #3268, opened 2020-03-06
remount hardening: move to file descriptor based mounts

smitsohu:
Harden `noexec`, `read-only`, `read-write` options by moving entirely to file descriptor based mounts.

Along the way fix long-time broken `writable-var-log` option, which didn't work any more because of an ownership check.


startx2017:
Sorry for the review thing! Looks fine, let's merge it.

PR #3276, opened 2020-03-13
seccomp: allow defining separate filters for 32-bit arch

topimiettinen:
System calls (names and numbers) are not exactly the same for 32 bit
and 64 bit architectures. Let's allow defining separate filters for
32-bit arch using seccomp.32, seccomp.32.drop, seccomp.32.keep. This
is useful for mixed 64/32 bit application environments like Steam and
Wine.

Lists of supported system calls are also updated.


src/firejail/main.c
```diff
@@ -1275,6 +1288,18 @@ int main(int argc, char **argv) {
 			else
 				exit_err_feature("seccomp");
 		}
+		else if (strncmp(argv[i], "--seccomp.32.drop=", 15) == 0) {
+			if (checkcfg(CFG_SECCOMP)) {
+				if (arg_seccomp32) {
+					fprintf(stderr, "Error: seccomp,32 already enabled\n");
```
topimiettinen: seccomp.32

src/firejail/sandbox.c
```diff
@@ -1113,9 +1113,14 @@ int sandbox(void* sandbox_arg) {
 	// if a keep list is available, disregard the drop list
 	if (arg_seccomp == 1) {
 		if (cfg.seccomp_list_keep)
-			seccomp_filter_keep();
+			seccomp_filter_keep(true);
 		else
-			seccomp_filter_drop();
+			seccomp_filter_drop(true);
+
```
topimiettinen: should check arg_seccomp32 instead

src/include/syscall_x86_64.h
```diff
@@ -0,0 +1,347 @@
+{ "_sysctl", 156 },
```
topimiettinen: forgot to remove old system call file

FOSSONLY:
All this work just for these terrible proprietary programs? Steam is also available as a flatpak if you really need it. You should never install proprietary software natively. I also don't think it should be the job of firejail or similar to address the years of intentional flaws in proprietary software, while Valve still rests on absurd and tired arguments to effectively avoid 64 bit Steam. And why on earth is Wine relevant? In my opinion a completely superfluous software. And all this just to run disgusting software that by its very nature wants to harmonize exclusively with Windows, and sometimes even deliberately sabotages it from working via Wine. Nothing against the free developers and their social commitment, but after decades it should finally be clear that you can expect nothing from the antisocial proprietary world, no matter how often you give it a helping hand.

topimiettinen:
Generating incorrect seccomp filters is clearly a bug. This also could be extended to handle x64 (x86_64 but in 32 bit address space) though it seems to be dead.

I think it's great to have Steam on Linux as I don't have to run Windows anymore. Many of my Steam games are native 64 bit Linux apps, with very good performance. The rest are emulated using Proton library, which I think is derived from Wine. Nowadays I play the very few games which only run on Windows very rarely. It's also more fun using Windows as I can quickly reinstall it without caring too much anymore (because there isn't anything valuable) to fix for example broken Windows Update.

The reverse is also possible, using Windows Subsystem for Linux (esp. v2) to have almost Linux-like experience on Windows, but hardening the Linux system with tools like Firejail, systemd, TOMOYO, SELinux etc. gives me much more secure feeling.

topimiettinen:
I haven't found a good idea to solve the libpostexecseccomp (also libtrace) issue. Perhaps the preload libraries should be disabled when 32 bit filters are active or with a new command line option. Also libpostexecseccomp seems to be activated even when not needed (only needed to block execve() and a few others).

topimiettinen:
In this version Firejail exits if `seccomp.32` etc are used together with trace or postexecseccomp. Now libpostexecseccomp should also be triggered only if system calls in `@default-keep` are blocked.

topimiettinen:
Perhaps there should be only a warning for the libraries, exiting seems too restrictive.

topimiettinen:
Also mdwx and protocol filters need 32 bit versions.

PR #3286, opened 2020-03-19
Whitelist runuser common

rusty-snake:
introduce whitelist-runuser-common.inc
60f822e

 * If an applications does not need a whitelist it can/should be
   nowhitelisted. Example:

     nowhitelist ${RUNUSER}/pulse

 * ${RUNUSER}/bus is inaccessible with nodbus regardless of the
   whitelist. (as it should)

 * strange wayland setups with an second wayland-compostior need to
   whitelist ${RUNUSER}/wayland-1, ${RUNUSER}/wayland-2 and so on.

 * some display-manager store there Xauthority file in ${RUNUSER}.
   test results with fedora 31:
   - ssdm: ~/.Xauthority is used
   - lightdm: /run/lightdm/USER/Xauthority
   - gdm: /run/user/UID/gdm/Xauthority

 * IMPORTANT: ATM we can only enable this for non-graphical and GTK3
   programs because mutter (GNOMEs window-manger) stores the Xauthority
   file for Xwayland under /run/user/UID/.mutter-Xwaylandauth.XXXXXX
   where XXXXXX is random. Until we have whitelist globbing we can't
   whitelist this file. QT/KDE and other toolkits without full wayland
   support won't be able to start.

 wru update 1
286ae11

- add wru to more profiles.
- blacklist ${RUNUSER} works for the most cli programs too.

~todo: rebase and apply to gnome-screenshot~ done

### DIFF
--- a/RELNOTES
+++ b/RELNOTES
@@ -3,6 +3,7 @@ firejail (0.9.63) baseline; urgency=low
   * DHCP client support
   * SELinux labeling support
   * 32-bit seccomp filter
+  * restrict ${RUNUSER} in serveral profiles
   * new condition: HAS_NOSOUND
   * new profiles: gfeeds, firefox-x11, tvbrowser, rtv, clipgrab, muraster
   * new profiles: gnome-passwordsafe, bibtex, gummi, latex, mupdf-x11-curl
--- a/etc/0ad.profile
+++ b/etc/0ad.profile
@@ -26,7 +26,6 @@ whitelist ${HOME}/.local/share/0ad
 whitelist /usr/share/0ad
 include whitelist-common.inc
 include whitelist-usr-share-common.inc
-#X11: include whitelist-runuser-common.inc
 include whitelist-var-common.inc
 
 caps.drop all
--- a/etc/0ad.profile
+++ b/etc/0ad.profile
@@ -26,6 +26,7 @@ whitelist ${HOME}/.local/share/0ad
 whitelist /usr/share/0ad
 include whitelist-common.inc
 include whitelist-usr-share-common.inc
+include whitelist-runuser-common.inc
 include whitelist-var-common.inc
 
 caps.drop all
--- a/etc/0ad.profile
+++ b/etc/0ad.profile
@@ -26,7 +26,7 @@ whitelist ${HOME}/.local/share/0ad
 whitelist /usr/share/0ad
 include whitelist-common.inc
 include whitelist-usr-share-common.inc
-include whitelist-runuser-common.inc
+#X11: include whitelist-runuser-common.inc
 include whitelist-var-common.inc
 
 caps.drop all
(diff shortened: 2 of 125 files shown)


topimiettinen:
Looks good. Just a suggestion, feel free to ignore: I'd sort include lines alphabetically, so `include whitelist-runuser-common.inc` should be before `include whitelist-usr-share-common.inc`.

kris7t:
> ATM we can only enable this for non-graphical and GTK3
programs because mutter (GNOMEs window-manger) stores the Xauthority
file for Xwayland under /run/user/UID/.mutter-Xwaylandauth.XXXXXX
where XXXXXX is random.

Does this mean that we can also enable this for Qt programs, provided they run under either KDE, sway, or use [QGnomePatform](https://github.com/FedoraQt/QGnomePlatform) under GNOME? AFAIK the main reason why Qt does not use Wayland everywhere is the lack of client-side decoration support in some compositors, but otherwise, it should work fine.

rusty-snake:
yes, with kwin_wayland, sway, mutter (x11) or QT_QPA_PLATFORM=wayland (only qt) there are no issues related to wayland/X/XWayland to expect. I'm using `include whitelist-runuser-common.inc` in my `globals.local` for a few days now and it works fine. Only for firefox and keepassxc I need to add `include ignore-wruc.inc` because the keepassxc-browser addon does not work with wruc. Fedora: 31 GNOME: 3.34. IDK why but also XWayland programs like firefox, freeoffice, wps, onlyoffice, torbrowser works with this.

kris7t:
I think, at least on the Firefox side, it would be enough to whitelist `/run/user/<uid>/kpxc_server`. On the other hand, that may be insufficient if keepassxc starts after Firefox.

rusty-snake:
> On the other hand, that may be insufficient if keepassxc starts after Firefox.

You got it. Unfortunately, this socket path is hardcoded and cannot be easily changed to XDG_RUNTIME_DIR/KeePassXC/browser or similar.

glitsj16:
> IMPORTANT: ATM we can only enable this for non-graphical and GTK3
> programs because mutter (GNOMEs window-manger) stores the Xauthority
> file for Xwayland under /run/user/UID/.mutter-Xwaylandauth.XXXXXX
> where XXXXXX is random. Until we have whitelist globbing we can't
> whitelist this file. QT/KDE and other toolkits without full wayland
> support won't be able to start.

Now https://github.com/netblue30/firejail/commit/601df2fbb9cdfedc6ab71cbe47d275e39c935dca brings in whitelist globbing, do you have any suggestions as to how we can harden wruc?

rusty-snake:
Adding `whitelist ${RUNUSER}/.mutter-Xwaylandauth.*` to wruc allows us to add wruc to all kinds of program.

We could add `whitelist ${RUNUSER}/wayland-?`, but I don't think that this is necessary.

Issue #992, opened 2016-12-18
SDL error when trying to run certain games in steam


craftyguy:
I am unable to launch certain games (e.g. Mount and Blade: Warband) when using the steam profile provided with firejail, or even when using --noprofile option. I am using firejail version 0.9.45. OS is Arch Linux with kernel 4.9.

Output in terminal when running `firejail --noprofile steam`:

```
CHIDDeviceListSDL(): Couldn't load libSDL2-2.0.so.0, not enumerating devices
```

Here's the output when the game is launched:

```
Installing breakpad exception handler for appid(steam)/version(1481930308)
Game update: AppID 48700 "Mount & Blade: Warband", ProcID 211, IP 0.0.0.0:0
ERROR: ld.so: object '/home/clayton/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
>>> Adding process 211 for game ID 48700
ERROR: ld.so: object '/home/clayton/.local/share/Steam/ubuntu12_64/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS64): ignored.
/home/clayton/.local/share/Steam/SteamApps/common/MountBlade Warband/mb_warband_linux: /usr/lib32/libcurl-gnutls.so.4: no version information available (required by /home/clayton/.local/share/Steam/SteamApps/common/MountBlade Warband/mb_warband_linux)
ERROR: ld.so: object '/home/clayton/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
>>> Adding process 212 for game ID 48700
>>> Adding process 214 for game ID 48700
>>> Adding process 215 for game ID 48700
Game removed: AppID 48700 "Mount & Blade: Warband", ProcID 211

```

In kernel log when the game is launched:

```
[ 4279.759294] SDLHotplugALSA[14952]: segfault at 0 ip 00000000f75d1e88 sp 00000000ed2ff240 error 4 in libSDL2-2.0.so.0[f753d000+e4000]
```

craftyguy:
[...]

PR #3292, opened 2020-03-22
steam.profile: correctly blacklist unneeded directories in user's home

davidebeatrici:
`noblacklist` directives prevent following ones from blacklisting the specified directory/file.

The profile currently has a `noblacklist` directive for each directory used by Steam and/or its games, which is fine.

However, there are no directives blacklisting the user's home, thus all directories and files inside it are accessible by Steam.

This pull request fixes the issue by adding `whitelist` directives, which automatically blacklist the parent directory (in this case the user's home).

`mkdir` and `mkfile` directives are added so that the directories/files are created if they don't exist.

Thanks to @SkewedZeppelin for suggesting to keep `noblacklist` and use `mkdir` and `mkfile`.

### DIFF
--- a/etc/steam.profile
+++ b/etc/steam.profile
@@ -36,6 +36,34 @@ include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 
+mkdir ${HOME}/.killingfloor
+mkdir ${HOME}/.local/share/3909/PapersPlease
+mkdir ${HOME}/.local/share/aspyr-media
+mkdir ${HOME}/.local/share/cdprojektred
+mkdir ${HOME}/.local/share/feral-interactive
+mkdir ${HOME}/.local/share/Steam
+mkdir ${HOME}/.local/share/SuperHexagon
+mkdir ${HOME}/.local/share/Terraria
+mkdir ${HOME}/.local/share/vpltd
+mkdir ${HOME}/.local/share/vulkan
+mkdir ${HOME}/.steam
+mkfile ${HOME}/.steampath
+mkfile ${HOME}/.steampid
+whitelist ${HOME}/.killingfloor
+whitelist ${HOME}/.local/share/3909/PapersPlease
+whitelist ${HOME}/.local/share/aspyr-media
+whitelist ${HOME}/.local/share/cdprojektred
+whitelist ${HOME}/.local/share/feral-interactive
+whitelist ${HOME}/.local/share/Steam
+whitelist ${HOME}/.local/share/SuperHexagon
+whitelist ${HOME}/.local/share/Terraria
+whitelist ${HOME}/.local/share/vpltd
+whitelist ${HOME}/.local/share/vulkan
+whitelist ${HOME}/.steam
+whitelist ${HOME}/.steampath
+whitelist ${HOME}/.steampid
+whitelist ${HOME}/.steampid
+include whitelist-common.inc
 include whitelist-var-common.inc
 
 caps.drop all


SkewedZeppelin:
This was discussed a while ago iirc.
I only added those after @valoq added them in 6bf6d5ed51fa6f4247f10e83e85ca0b69650e389
which specifically were from my original steam.profile from my (awful) generator program in 

> * 8b66325 - (Tue Oct 4 07:48:12 2016) Steam profile updates - Tad

Merged as is would likely cause players to end up with lost game saves.

More games would need to be tested and have their paths added.

I have no interest in updating steam.profile as I've been using the flatpak version of Steam for a while now.

For that you can use
`~/.local/share/flatpak/overrides/com.valvesoftware.Steam`
```
[Context]
filesystems=!xdg-documents;!xdg-pictures;!xdg-music;
devices=dri;!all;
shared=!network;
features=!bluetooth;!canbus;
```
when you want to give network (to download games)
`flatpak run --share=network com.valvesoftware.Steam`

davidebeatrici:
Thank you very much for the detailed explanation!

> Edit:
as is is actually completely broken.
you need to leave those
add the mkdir/mkfiles
add the whitelist
and also whitelist-common.inc's

Before creating the pull request I tested the profile with the changes applied and it seems to work without any issues.

Why should `noblacklist` and `mkdir` be required? To create directories if they don't exist yet?

rusty-snake:
> Why should `noblacklist` and `mkdir` be required? To create directories if they don't exist yet?

noblacklist: the paths are still blacklisted even if they are whitelisted. test: `firejail --noprofile '--blacklist=${HOME}/Downloads' '--whitelist=${HOME}/Downloads' ls Downloads`

mkdir: non existing paths can't be whitelisted

davidebeatrici:
Done!

etc/steam.profile
```diff
@@ -7,18 +7,45 @@ include steam.local
 include globals.local
 
 # Blacklist the entire home directory, except for the specified files/directories
```
rusty-snake: remove

etc/steam.profile
```diff
@@ -19,6 +20,34 @@ noblacklist ${HOME}/.local/share/vulkan
...
+mkdir ${HOME}/.local/share/vpltd
+whitelist ${HOME}/.local/share/vpltd
+mkdir ${HOME}/.local/share/vulkan
+whitelist ${HOME}/.local/share/vulkan
+mkdir ${HOME}/.steam
+whitelist ${HOME}/.steam
+mkfile ${HOME}/.steampath
+whitelist ${HOME}/.steampath
+mkfile ${HOME}/.steampid
+whitelist ${HOME}/.steampid
```
rusty-snake: group and move down. example:

```
mkdir ${HOME}/.killingfloor
mkdir ${HOME}/.local/share/3909/PapersPlease
whitelist ${HOME}/.killingfloor
whitelist ${HOME}/.local/share/3909/PapersPlease
include whitelist-common.inc
```

davidebeatrici:
> I'm not a steam user, but if I get @SkewedZeppelin there are still games in the world with other paths. 

I assume you're referring to `${HOME}/My Games`.
I believe the path is wrong, because on Windows it's `%UserProfile\Documents\My Games`.

Also, I'm not aware of any games that save into `~/Documents/My Games`.
Windows games are run through Proton, they have their own "virtual" home directory under `~/.steam`.

SkewedZeppelin:
Once sorted in place the profile is OK for what it is.
My issue is that is with the small number of paths.
I don't want a user having their saves deleted.

https://github.com/netblue30/firejail/issues/992#issuecomment-277497630
Mentions more paths:
```
~/.paradoxinteractive
~/.mbwarband
~/.local/share/Paradox Interactive
~/.config/unity3d/Feral Interactive
```
More native games will need to be run through and see if they store outside of the Steam userdata folder. Non-native games run via Proton can all be assumed to only be in their compatdata folder.

davidebeatrici:
Fixed, sorry about that.

I agree that we should investigate whether there are missing paths.

PR #3293, opened 2020-03-23
Update wire-desktop.profile

0x7969:
At least on ArchLinux, wire-desktop is now using electron6.

### DIFF
--- a/etc/wire-desktop.profile
+++ b/etc/wire-desktop.profile
@@ -26,7 +26,7 @@ nou2f
 shell none
 
 disable-mnt
-private-bin bash,electron,electron4,env,sh,wire-desktop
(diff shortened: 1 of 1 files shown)


etc/wire-desktop.profile
```diff
@@ -26,7 +26,7 @@ nou2f
 shell none
 
 disable-mnt
-private-bin bash,electron,electron4,env,sh,wire-desktop
+private-bin bash,electron,electron6,env,sh,wire-desktop
```
rusty-snake:
```suggestion
private-bin bash,electron,electron4,electron6,env,sh,wire-desktop
```

Just for the compatible, does not harm.

rusty-snake:
Merged, Thanks.

PR #3300, opened 2020-03-27
Added compatibility with BetterDiscord

Atrate:
BetterDiscord is a nice utility that allows direct modifications to Discord if the user wishes so. To make it work, dirs in .config and .local/share need to be whitelisted.

Signed-off-by: Atrate <Atrate@protonmail.com>

### DIFF
--- a/etc/discord-common.profile
+++ b/etc/discord-common.profile
@@ -15,6 +15,8 @@ include disable-passwdmgr.inc
 include disable-programs.inc
 
 whitelist ${DOWNLOADS}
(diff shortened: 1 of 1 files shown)


glitsj16:
If the same 2 lines need to be added for BetterDiscord support, it would make more sense to add those to discord-common.profile IMO.

Atrate:
Force push. Ready for review.

glitsj16:
Thanks @Atrate!

Issue #3288, opened 2020-03-21
MPV fails with gpu/vdpau options (ADM gpu with mesa drivers)


gui-don:
Hi!

Here is the options I use:
```
hwdec=vdpau-copy
vo=gpu
profile=gpu-hq
gpu-api=vulkan
gpu-context=x11vk
```
`firejail mpv "my_video.mp4"` output:
 ```
[vo/gpu/vulkan/libplacebo] Failed creating instance: VK_ERROR_EXTENSION_NOT_PRESENT
[vo/gpu/vulkan/libplacebo] Failed initializing vulkan instance
[vo/gpu] Failed initializing any suitable GPU context!
Error opening/initializing the selected video_out (--vo) device.
```
`mpv "my_video.mp4"` runs fine.

Notes:
- Removing the options makes `mpv` run fine within firejail too.
- Using openGL instead of vulkan *does not work* either, `mpv` runs indefinitly.

Thanks for your help!

EDIT: Vulkan gpu API works with @rusty-snake fix.
Still stuck without vulkan: see my other comment

**Comment by rusty-snake** [comment]
Duplicate of #3107? Try this fix: https://github.com/netblue30/firejail/commit/08061fb590acda23d9f8fff4eebdd1366caf65fa

**Comment by gui-don** [comment]
Thanks @rusty-snake;

So that solves the vulkan run. However, when using these options:
```
hwdec=vdpau-copy
vo=gpu
profile=gpu-hq
```
`mpv` runs indefinitly without video nor sound. I tried to strace without firejail to find the accessed files and did some adjustements but no luck so far.

gui-don:
Alright, another quick research and test and I have to ignore `seccomp`, then mpv launches.

Thanks!

rusty-snake:
mesa, amd, seccomp? #3219, #3267.

Try `seccomp !kcmp`

PR #3301, opened 2020-03-27
Changeable seccomp error action

topimiettinen:
Let user specify the action when seccomp filters trigger:
- 'kill' (default): kill the process as before
- errno name like EPERM/ENOSYS: return errno and let the process continue.

Not killing the process weakens Firejail slightly when trying to
contain intrusion, but it may also allow tighter filters if the
alternative is to always allow a system call. It's still possible to
use errno return per syscall.

### DIFF
--- a/src/firejail/usage.c
+++ b/src/firejail/usage.c
@@ -209,6 +209,7 @@ static char *usage_str =
 	"    --seccomp.print=name|pid - print the seccomp filter for the sandbox\n"
 	"\tidentified by name or PID.\n"
 	"    --seccomp.32[.drop,.keep][=syscall] - like above but for 32 bit architecture.\n"
+	"    --seccomp-error-action=errno|kill - change error code or kill process.\n"
 #endif
 	"    --shell=none - run the program directly without a user shell.\n"
 	"    --shell=program - set default user shell.\n"
--- a/src/man/firejail-profile.txt
+++ b/src/man/firejail-profile.txt
@@ -411,6 +411,9 @@ Enable seccomp filter and whitelist the system calls in the list.
 \fBseccomp.32.keep syscall,syscall,syscall
 Enable seccomp filter and whitelist the system calls in the list for 32 bit system calls on a 64 bit architecture system.
 .TP
+\fBseccomp-error-action kill | ERRNO
+Return a different error instead of EPERM to the process or kill it when an attempt is made to call a blocked system call.
+.TP
 \fBx11
 Enable X11 sandboxing.
 .TP
--- a/RELNOTES
+++ b/RELNOTES
@@ -1,5 +1,10 @@
 firejail (0.9.63) baseline; urgency=low
   * work in progress
+  * The blocking action of seccomp filters has been changed from
+    killing the process to returning EPERM to the caller. To get the
+    previous behaviour, use --seccomp-error-action=kill or
+    syscall:kill syntax when constructing filters, or override in
+    /etc/firejail/firejail.config file.
   * DHCP client support
   * SELinux labeling support
   * 32-bit seccomp filter
--- a/etc/firejail.config
+++ b/etc/firejail.config
@@ -146,3 +146,6 @@
 
 # Xvfb command extra parameters.  None by default; this is an example.
 # xvfb-extra-params -pixdepths 8 24 32
+
+# Seccomp error action, kill or errno (EPERM, ENOSYS etc)
+# seccomp-error-action EPERM
--- a/src/firejail/checkcfg.c
+++ b/src/firejail/checkcfg.c
@@ -18,6 +18,8 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 #include "firejail.h"
+#include "../include/seccomp.h"
+#include "../include/syscall.h"
 #include <sys/stat.h>
 #include <linux/loop.h>
 
@@ -32,6 +34,7 @@ char *xvfb_screen = "800x600x24";
 char *xvfb_extra_params = "";
 char *netfilter_default = NULL;
 unsigned long join_timeout = 5000000; // microseconds
+char *config_seccomp_error_action_str = "EPERM";
 
 int checkcfg(int val) {
 	assert(val < CFG_MAX);
@@ -51,6 +54,7 @@ int checkcfg(int val) {
 		cfg_val[CFG_DISABLE_MNT] = 0;
 		cfg_val[CFG_ARP_PROBES] = DEFAULT_ARP_PROBES;
 		cfg_val[CFG_XPRA_ATTACH] = 0;
+		cfg_val[CFG_SECCOMP_ERROR_ACTION] = -1;
 
 		// open configuration file
 		const char *fname = SYSCONFDIR "/firejail.config";
@@ -218,6 +222,24 @@ int checkcfg(int val) {
 			// timeout for join option
 			else if (strncmp(ptr, "join-timeout ", 13) == 0)
 				join_timeout = strtoul(ptr + 13, NULL, 10) * 1000000; // seconds to microseconds
+
+			// seccomp error action
+			else if (strncmp(ptr, "seccomp-error-action ", 21) == 0) {
+#ifdef HAVE_SECCOMP
+				if (strcmp(ptr + 21, "kill") == 0)
+					cfg_val[CFG_SECCOMP_ERROR_ACTION] = SECCOMP_RET_KILL;
+				else {
+					cfg_val[CFG_SECCOMP_ERROR_ACTION] = errno_find_name(ptr + 21);
+					if (cfg_val[CFG_SECCOMP_ERROR_ACTION] == -1)
+						errExit("seccomp-error-action: unknown errno");
+				}
+				config_seccomp_error_action_str = strdup(ptr + 21);
+				if (!config_seccomp_error_action_str)
+					errExit("strdup");
+#else
+				warning_feature_disabled("seccomp");
+#endif
+			}
 
 			else
 				goto errout;
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -186,6 +186,7 @@ typedef struct config_t {
 	char *seccomp_list_drop, *seccomp_list_drop32;	// seccomp drop list
 	char *seccomp_list_keep, *seccomp_list_keep32;	// seccomp keep list
 	char *protocol;			// protocol list
+	char *seccomp_error_action;			// error action: kill or errno
 
 	// rlimits
 	long long unsigned rlimit_cpu;
@@ -572,6 +573,7 @@ int seccomp_install_filters(void);
 int seccomp_load(const char *fname);
 int seccomp_filter_drop(bool native);
 int seccomp_filter_keep(bool native);
+int seccomp_filter_mdwx(bool native);
 void seccomp_print_filter(pid_t pid);
 
 // caps.c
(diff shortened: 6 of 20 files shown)


topimiettinen:
I based this on 32-bit filter commit to avoid rebasing.

Vincent43:
Why keep kill as default? I think we want default to work with as many apps as possible otherwise cases like mesa depending on new blocked syscall will keep breaking apps.

topimiettinen:
For compatibility, then each app profile can opt-in to enable it or keep the default if the profile just works.

topimiettinen:
There should be a way to override this per syscall, like `--seccomp-error-action=EPERM --seccomp.drop=kcmp:kill`.

Vincent43:
How moving to EPERM/ENOSYS can decrease compatibility? I think it can only increase it. The point is not having to _whack a mole_ with fixing particular profiles one by one which often takes weeks or months until fixes land in distros. Apps and their ecosystem constantly change and we don't control what syscall get called. Default _kill_ results in profiles that just worked yesterday are broken tomorrow. Kill should be user explicit override, not default.

topimiettinen:
With compatibility, I mean that the behaviour does not change until each profile is changed. Right now, all profiles should work with no change. Someone who cares for a specific app should check if the new option makes sense (probably does), enable it but preferably also check if the filters can be tightened since the app may respond intelligently to getting ENOSYS/EPERM from system call filters which previously were not there because of kill action. I don't like the idea of enabling this globally for all apps, then in the worst case nobody would make the checks for tightening the filters.

Vincent43:
As I said the point was not having to change each of hundreds of profiles individually. I guess mesa kcmp stuff breaks many, many apps right now, who will be fixing them? The reality is people are [disabling seccomp completely](https://github.com/netblue30/firejail/issues/3288#issuecomment-602064408) because broken security is worse than no security.

> I don't like the idea of enabling this globally for all apps, then in the worst case nobody would make the checks for tightening the filters.

With _default kill_ nobody would tightening the filters neither because it's impossible to do. So in worst case you end up with same security (but working apps), in best case security will improve as filters will be tightened. ENOSYS/EPERM is really no brainier here, everyone else learned that already and it's time for firejail to fix itself too.

topimiettinen:
How would it be impossible to tighten the filters? Just add `seccomp-error-action EPERM´ to the profile.

Vincent43:
Yes, it would be possible this way, yet it needs additional work about which you said nobody will do. I don't understand why we can't fix everything here.

topimiettinen:
Updated with `syscall:kill` syntax to override the default.

We don't know how much there's willingness do the changes and the checks. Perhaps other developers should also comment which would be their preference?

glitsj16:
I don't consider myself a developer technically, my programming skills are way to limited for that alas. As a contributor I probably have above-average time to spend on firejail (profile maintenance) due to all kinds of personal circumstances. So I for one would be willing to make an extra effort to keep our profiles in good working condition, if that's what it takes. Please note that I don't fully understand _all_ the implications of what this PR is about, but I wanted to chime in to express a growing concern that firejail profiles in general are getting harder to maintain. The technically very reasonable MR about [fine-grained D-Bus control](https://github.com/netblue30/firejail/pull/3265) is a case in point and IMO also relevant in this context. Not that I don't appreciate all the work being done here, quite the opposite. Profile complexity will keep rising, there's no way around that. But it wouldn't hurt to have a discussion on what we as a community can do to keep it manageable. _Bottomline: I tend to prefer implementations that don't touch (individual) profiles, or just as little as possible otherwise._

topimiettinen:
All right, I changed the default action to EPERM.

topimiettinen:
Some thoughts about the filters. Let's say we have three sets of system calls for an application X:
- A: always need to be allowed
- B: not really needed, but the app or the libraries can try to use the system call
- C: never needed

Previously the kill error action has forced the filters to allow both A and B because denying B would kill the application. With default seccomp-error-action of EPERM it's possible to deny B.

This could be improved so that while set B returns EPERM as the new default, it's easy to use kill method for set C (or parts of it like the most suspicious system calls). The syntax could be something like `seccomp.drop=@obsolete:kill`.

FOSSONLY:
It should not be possible to change this, and certainly not reduce security globally. And it's not the profiles that are broken, but the programs that are the problem if they go against security. Finally, "kill (process tree)" is the Linux standard for Seccomp, and for very good reasons. Compatibility is also clearly secondary to security. Not the security mechanisms have to become more compatible, but the programs have to obey the rules. It can't make sense to give in all the time, just so that certain programs that don't like to be restricted work properly.

Vincent43:
> It should not be possible to change this, and certainly not reduce security globally

This change will only increase security globally as more syscalls can be blocked now. Default seccomp policy could be also changed from blacklist to whitelist as new syscalls will be gracefully denied but this needs some more work.

> And it's not the profiles that are broken, but the programs that are the problem if they go against security

Firejail can't fix any programs, it can only refuse to support it however if program profile exist in firejail repo then it should be supported.

> Finally, "kill (process tree)" is the Linux standard for Seccomp

It's not. Neither of snap, flatpak, systemd, or docker does it. [Here's](https://lwn.net/Articles/796253/) comment from Lennart Poettering which among other things admits that killing process with seccomp was a mistake.

> Compatibility is also clearly secondary to security.

Yes, broken apps are secure. Also useless 😄 .

> Not the security mechanisms have to become more compatible, but the programs have to obey the rules.

How would you force them _to obey the rules_? Magic? Programs do what was written in their code.

> It can't make sense to give in all the time, just so that certain programs that don't like to be restricted work properly.

When app crash on startup because it gets killed by seccomp policy then what you would do? Stick to kill and allow problematic syscall or keep blocking that syscall and use EPERM? What do you think is more secure?

matu3ba:
> Default seccomp policy could be also changed from blacklist to whitelist as new syscalls will be gracefully denied but this needs some more work.
Could you explain a possible strategy on changing to whitelists? Removing code to simplify is always a good thing, but the problem is the compatibility (which needs alot testing under different distros).
To me it looks like some way to to keep 2 parallel installations to test the distro update by users would be needed. Then the program maintainers or firejail devs could be notified to fix the problem.
Do you know any simple enough tools for that with large enough compatibility?

> Firejail can't fix any programs, it can only refuse to support it however if program profile exist in firejail repo then it should be supported.
You definition of support is vaguely. Maybe the definition of support regarding security should be specified somewhere like in the README. If you say that is the problem, then keeping track of severity would be the first step. Do you have numbers/a list? Do they use common functionality?
> > Finally, "kill (process tree)" is the Linux standard for Seccomp
> 
> It's not. Neither of snap, flatpak, systemd, or docker does it. [Here's](https://lwn.net/Articles/796253/) comment from Lennart Poettering which among other things admits that killing process with seccomp was a mistake.
True, but the questions is about the project goal(intention) and default behavior and how simple/complex the solution would be.

> > Compatibility is also clearly secondary to security.
> 
> Yes, broken apps are secure. Also useless smile .
It depends on what the project goal is and if you want to take that compromise. However this should be specified.

> When app crash on startup because it gets killed by seccomp policy then what you would do? Stick to kill and allow problematic syscall or keep blocking that syscall and use EPERM? What do you think is more secure?
Please be constructive and suggest an alternative behavior with its complexity/simplicity tradeoffs from a coding standpoint.

Sorry for being picky your comment in special, but probably you could at best weight the tradeoffs of the argument against another at best.

Vincent43:
> Could you explain a possible strategy on changing to whitelists?

Take a current blacklist and reverse it - instead of listing blocked syscalls, list allowed syscalls which are ones not in blacklist. The advantage of this is that any new syscalls added won't be allowed automatically. This is what docker does.

> You definition of support is vaguely. Maybe the definition of support regarding security should be specified somewhere like in the README. If you say that is the problem, then keeping track of severity would be the first step. Do you have numbers/a list? Do they use common functionality?

I think it's safe to assume that if firejail ships profile for specific app then that app should work and not be killed by seccomp policy. There are a lot of issues reported regarding seccomp being too aggressive, latest one is `kcmp` syscall used by recent mesa which affects many apps.

> True, but the questions is about the project goal(intention) and default behavior and how simple/complex the solution would be.

Firejail has same goal as all examples I mentioned - enhance security. I believe we don't want to argue here if making app unworkable enhances its security. The complexity doesn't change.

> Please be constructive and suggest an alternative behavior with its complexity/simplicity tradeoffs from a coding standpoint.

I literally provided two practical alternatives: one when seccomp violation kills offending processes (current behavior) and one when seccomp policy just blocks specific syscall (what this PR does) which user can switch back.

topimiettinen:
If someone does not like the new default, it's still possible to get the old behaviour with `--seccomp-error-action=kill`. This can also be done per profile basis or locally. Though when the filters get tightened so that previously allowed system calls return EPERM, this will not be so simple, also the filters need to be tuned so the processes are not killed.

Maybe some filters could be annotated with tags or something that specify whether it's OK to use either kill or error, or error/allow only (`syscall:?ERROR_ALLOW_NOKILL:EPERM` or something crazy like that). That would not help the goal to keep the job of profile maintainers relatively sane.

Vincent43:
@topimiettinen could you rebase this to resolve conflicts?

@netblue30 @Fred-Barclay @smitsohu @reinerh could any (or all) of you review this if there are any objections?

src/firejail/usage.c
```diff
@@ -208,6 +208,7 @@ static char *usage_str =
 	"\twhitelist the syscalls specified by the command.\n"
 	"    --seccomp.print=name|pid - print the seccomp filter for the sandbox\n"
 	"\tidentified by name or PID.\n"
+	"    --seccomp-error-action=errorno|kill - change error code or kill process.\n"
```
reinerh, line 211: There are a few places where "errorno" is used.
I think it's clearer if "errno" would be used instead.

src/man/firejail-profile.txt
```diff
@@ -411,6 +411,9 @@ Enable seccomp filter and whitelist the system calls in the list.
 \fBseccomp.32.keep syscall,syscall,syscall
 Enable seccomp filter and whitelist the system calls in the list for 32 bit system calls on a 64 bit architecture system.
 .TP
+\fBseccomp-error-action kill | ERRNO
+Return a different error instead EPERM to the process or kill it when an attempt is made to call a blocked system call.
```
reinerh, line 415: "instead of"

reinerh:
I don't have objections. I think it could help with maintainability of profiles.
But I would have an additional request.
Could you make the default action also configurable via /etc/firejail/firejail.config?
This would make it much easier for users to switch back to the old behaviour, if they prefer to kill the processes instead.

(It would also be nice to have tests for the new behavior. And I would also expect that some older tests are now failing?)

rusty-snake:
> Could you make the default action also configurable via /etc/firejail/firejail.config?
This would make it much easier for users to switch back to the old behaviour, if they prefer to kill the processes instead.

IMHO globals.local with exceptions in <profile>.local would be easier. However a `seccomp-force-kill yes` (like `force-nonewprivs yes`) could be interesting.

> (It would also be nice to have tests for the new behavior. And I would also expect that some older tests are now failing?)

`firejail --audit`

Fred-Barclay:
I like it! I'm with @reinerh, it would be great to configure with /etc/firejail/firejail.config (I slightly prefer this to globals.inc but either are good).

But this is needed for sure. :+1:

src/firejail/usage.c
topimiettinen: OK. Though "errorno" was used elsewhere, it's slightly less technical than Unix term "errno".

src/man/firejail-profile.txt
topimiettinen: OK

topimiettinen:
Added also /etc/firejail/firejail.config option.

PR #3310, opened 2020-03-31
Preserve CFLAGS given to configure in common.mk.in

Liorst4:
Wanted to add some CFLAGS during ./configure, but it didn't work.

```bash
./configure CFLAGS='-Wall'
```

Added this line to make it work.


netblue30:
all set, thanks!

PR #3319, opened 2020-04-05
Simple sanity checks for arguments and environment

topimiettinen:
Restrict number of program arguments and their length as well as
number of environment variables and their length.


netblue30:
merged!

Issue #2685, opened 2019-05-09
Firefox 66: can't save downloaded files due to unwriteable /tmp/user/$UID


chrsmrtnx:
Running Debian buster with Firefox 66 and firejail 0.9.58.2, I found that with firejail's default Firefox profile, I couldn't download files. When clicking on a link to a file in Firefox, I'd get an error that the file "could not be saved, because you cannot change the contents of that folder. Change the folder properties and try again, or try saving in a different location".

To solve the problem, I had to comment out "private-tmp" in firefox-common.profile or, alternatively, add the line "whitelist /tmp/user/1000" (1000 being the relevant userid).

I suspect that the problem is that Firefox wants to write the downloaded file to /tmp/user/$UID before it moves it to the final download folder, typically under $HOME. But it can't because /tmp/user/$UID is read-only when "private-tmp" is set, unless you explicitly whitelist /tmp/user/1000 or whatever. (If I run "firejail --profile=firefox /bin/bash", I can't write to /tmp/user/1000: "touch /tmp/user/1000/foo" fails, whereas of course it works fine outside of firejail.)

Thanks.

Vincent43:
Do you use [pam-tmpdir](https://packages.debian.org/sid/libpam-tmpdir)?

chiraag-nataraj:
Btw, one way to solve this is to right-click and hit "Save As...", which (I think) _should_ work? I actually opened a related bug [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1497405).

chrsmrtnx:
> Do you use [pam-tmpdir](https://packages.debian.org/sid/libpam-tmpdir)?

Yes, I do have libpam-tmpdir installed.

chrsmrtnx:
> Btw, one way to solve this is to right-click and hit "Save As...", which (I think) _should_ work? I actually opened a related bug [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1497405).

I tested this, and it does work, oddly. But then clicking on a link to a file still fails as described before. Odd.

rusty-snake:
@chrsmrtnx you can make this as an default. about:preferences -> always ask when downloading files (not sure about the correct english name) (`browser.download.useDownloadDir = false`)

chrsmrtnx:
> @chrsmrtnx you can make this as an default. about:preferences -> always ask when downloading files (not sure about the correct english name) (`browser.download.useDownloadDir = false`)

[...]

PR #3322, opened 2020-04-06
Filter environment variables

topimiettinen:
Save all environment variables for later use in sandboxes, clear
environment and re-apply only whitelisted variables for the main
firejail process. All variables will be reapplied for the sandboxes by
env_apply().

### DIFF
--- a/src/firejail/env.c
+++ b/src/firejail/env.c
@@ -25,8 +25,8 @@
 
 typedef struct env_t {
 	struct env_t *next;
-	char *name;
-	char *value;
+	const char *name;
+	const char *value;
 	ENV_OP op;
 } Env;
 static Env *envlist = NULL;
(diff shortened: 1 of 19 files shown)


topimiettinen:
The commit eliminates the difference between start_application(0, fp) and start_application(1, fp) as now variables are applied unconditionally (because some of them don't exist anymore after clearenv()). To be honest, I'm not sure if there is now a change in behavior in some cases.

topimiettinen:
Actually I'm pretty sure that all variables changed with setenv() in main Firejail process may be overwritten with the original values. So those setenv()s have to be changed to env_store()s.

src/firejail/env.c
```diff
@@ -225,3 +226,46 @@ void env_apply(void) {
...
+	"FIREJAIL_FILE_COPY_LIMIT",
+	"FIREJAIL_PROMPT",
+	"FIREJAIL_QUIET",
+	"FIREJAIL_X11",
+	"LANG",
+	"LANGUAGE",
+	"LC_MESSAGES",
+	"PATH",
+	"SHELL",
+	"TMP",
```
Vincent43, line 241: Why `TMP` is whitelisted?

smitsohu: We used `TMP` to fix #2685

smitsohu:
> To be honest, I'm not sure if there is now a change in behavior in some cases.

A while back I had what I think is the same problem with sanitizing the umask. So if you like you can  look out for `umask(orig_umask)` calls.

topimiettinen:
In this version, the variables which Firejail uses internally are also moved to env storage for consistency. Thus the whitelist only applies to variables used by external libraries (libc/pthreads, libapparmor) and then it can be much shorter: only LANG, LANGUAGE, LC_MESSAGES.

Perhaps the sandboxed helper apps (fseccomp & friends) should also be run with a short list of variables: LANG etc. and internal variables starting with FIREJAIL_, otherwise they might need similar sanitizing (or perhaps they also need some sanitizing in any case to be sure). Only the final user app should get all variables restored.

topimiettinen:
Latest build failure happens because tests expect that using `--env=FIREJAIL_TEST_ARGUMENTS=yes` (or exporting the environment variable FIREJAIL_TEST_ARGUMENTS from shell), the variable is passed also to the helper apps. Maybe this should be instead flagged with some other means, for example adding arguments to `faudit` like `faudit --test-arguments`. It would be simpler to just let the the variable pass.

msva:
@topimiettinen ping?

Any progress here? ;)

We're all struggling (from env length checks) without your work ;)

topimiettinen:
Sorry, I actually made a version which lets FIREJAIL_TEST_ARGUMENTS pass, but forgot to follow up. Thanks for the reminder. Let's see what the CIs think of the rebased version.

topimiettinen:
Adding also PATH let the CI pass!

msva:
@netblue30 ping? :D

netblue30:
all merged!

PR #3326, opened 2020-04-07
dbus filter profiles (1)

rusty-snake:
Opening as PR so we can disscuss about them. The most are just extracted from flathub, see https://github.com/netblue30/firejail/pull/3265#issuecomment-606030794.

**EDIT:** everyone fell free to push more profiles here.

### DIFF
--- a/etc/templates/profile.template
+++ b/etc/templates/profile.template
@@ -33,6 +33,7 @@
 #   WHITELIST INCLUDES
 #   OPTIONS (caps*, net*, no*, protocol, seccomp*, shell none, tracelog)
 #   PRIVATE OPTIONS (disable-mnt, private-*, writable-*)
+#   DBUS FILTER
 #   SPECIAL OPTIONS (mdwx, noexec, read-only, join-or-start)
 #   REDIRECT INCLUDES
 #
@@ -136,6 +137,10 @@ include globals.local
 #net none
 #netfilter
 #no3d
+<<<<<<< HEAD
+=======
+##nodbus (deprecated, use 'dbus-user none' and 'dbus-system none')
+>>>>>>> dbus filter (1)
 #nodvd
 #nogroups
 #nonewprivs
@@ -185,7 +190,18 @@ include globals.local
 ##writable-var
 ##writable-var-log
 
-#dbus-user none
+# Since 0.9.63 also a more granular regulation of dbus is supported.
+# To get the dbus-addresses to which an application needs access to you can look
+# at flatpak if the application is also distriputed with flatpak:
+#    flatpak remote-info --show-metadata flathub <APP-ID>
+# Notes:
+#  - flatpak implicitly allows an app to own <APP-ID> on the session bus
+#  - In order to make dconf work (if it is used by the app) you need to allow
+#    'ca.desrt.dconf' even if it is not allowed by flatpak.
+#dbus-user filter
+#dbus-user.own com.github.netblue30.firejail
+#dbus-user.talk ca.desrt.dconf
+#dbus-user.talk org.freedesktop.Notifications
 #dbus-system none
 
 ##env VAR=VALUE
--- a/etc/templates/profile.template
+++ b/etc/templates/profile.template
@@ -188,13 +188,15 @@ include globals.local
 ##writable-var-log
 
 # Since 0.9.63 also a more granular regulation of dbus is supported.
-# To get the dbus-addresses to which an application needs access to you can look
-# at flatpak if the application is also distriputed with flatpak:
+# To get the dbus-addresses to which an application needs access to.
+# You can look at flatpak if the application is also distriputed via flatpak:
 #    flatpak remote-info --show-metadata flathub <APP-ID>
 # Notes:
 #  - flatpak implicitly allows an app to own <APP-ID> on the session bus
 #  - In order to make dconf work (if it is used by the app) you need to allow
 #    'ca.desrt.dconf' even if it is not allowed by flatpak.
+# Notes and Policiy about addresses can be found at
+# <https://github.com/netblue30/firejail/wiki/Restrict-D-Bus>
 #dbus-user filter
 #dbus-user.own com.github.netblue30.firejail
 #dbus-user.talk ca.desrt.dconf
--- a/etc/templates/profile.template
+++ b/etc/templates/profile.template
@@ -137,10 +137,7 @@ include globals.local
 #net none
 #netfilter
 #no3d
-<<<<<<< HEAD
-=======
-##nodbus (deprecated, use 'dbus-user none' and 'dbus-system none')
->>>>>>> dbus filter (1)
+##nodbus (deprecated, use 'dbus-user none' and 'dbus-system none', see below)
 #nodvd
 #nogroups
 #nonewprivs
--- a/etc/profile-a-l/celluloid.profile
+++ b/etc/profile-a-l/celluloid.profile
@@ -49,8 +49,6 @@ private-tmp
 dbus-user filter
 dbus-user.own io.github.celluloid_player.Celluloid
 dbus-user.talk org.gnome.SettingsDaemon.MediaKeys
-dbus-user.talk org.gtk.vfs.*
-dbus-user.talk org.gtk.vfs
 dbus-system none
 
 read-only ${HOME}
--- a/etc/profile-a-l/celluloid.profile
+++ b/etc/profile-a-l/celluloid.profile
@@ -46,9 +46,12 @@ private-etc alternatives,ca-certificates,crypto-policies,dconf,drirc,fonts,gtk-3
 private-dev
 private-tmp
 
-# uses dconf, MPRIS
-# dbus-user none
-# dbus-system none
+dbus-user filter
+dbus-user.own io.github.celluloid_player.Celluloid
+dbus-user.talk org.gnome.SettingsDaemon.MediaKeys
+dbus-user.talk org.gtk.vfs.*
+dbus-user.talk org.gtk.vfs
+dbus-system none
 
 read-only ${HOME}
 read-write ${HOME}/.config/celluloid
--- a/etc/profile-a-l/com.github.dahenson.agenda.profile
+++ b/etc/profile-a-l/com.github.dahenson.agenda.profile
@@ -54,6 +54,11 @@ private-dev
 private-etc dconf,fonts,gtk-3.0
 private-tmp
 
+dbus-user filter
+dbus.own com.github.dahenson.agenda
+dbus.talk ca.desrt.dconf
+dbus-system none
+
 read-only ${HOME}
 read-write ${HOME}/.cache/agenda
 read-write ${HOME}/.config/agenda
--- a/etc/profile-a-l/dconf-editor.profile
+++ b/etc/profile-a-l/dconf-editor.profile
@@ -44,3 +44,9 @@ private-dev
 private-etc alternatives,dconf,fonts,gtk-3.0,machine-id
 private-lib
 private-tmp
+
+dbus-user filter
+dbus-user.own ca.desrt.dconf-editor
+dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.gnome.SettingsDaemon.Color
+dbus-system none
--- a/etc/profile-a-l/dconf-editor.profile
+++ b/etc/profile-a-l/dconf-editor.profile
@@ -48,5 +48,4 @@ private-tmp
 dbus-user filter
 dbus-user.own ca.desrt.dconf-editor
 dbus-user.talk ca.desrt.dconf
-dbus-user.talk org.gnome.SettingsDaemon.Color
 dbus-system none
--- a/etc/profile-a-l/eog.profile
+++ b/etc/profile-a-l/eog.profile
@@ -15,5 +15,10 @@ whitelist /usr/share/eog
 # or put 'ignore private-bin', 'ignore private-etc' and 'ignore private-lib' in your eog.local
 private-bin eog
 
+dbus-user filter
+dbus-user.own org.gnome.Eog
+dbus-user.talk ca.desrt.dconf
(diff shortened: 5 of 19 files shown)


rusty-snake:
BTW: `firejail --noprofile --dbus-user=filter --audit | grep D-Bus` `MAYBE: D-Bus socket /run/firejail/mnt/dbus/user is available`.

etc/ghostwriter.profile
```diff
@@ -48,3 +48,8 @@ private-dev
 # passwd,login.defs,firejail are a temporary workaround for #2877 and can be removed once it is fixed
 private-etc alternatives,ca-certificates,crypto-policies,dbus-1,dconf,firejail,fonts,gconf,groups,gtk-2.0,gtk-3.0,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,login.defs,machine-id,mime.types,nsswitch.conf,pango,passwd,pki,protocols,resolv.conf,rpc,services,ssl,texlive,Trolltech.conf,X11,xdg
 private-tmp
+
+dbus-user filter
+dbus-user.own io.github.wereturtle.ghostwriter
+dbus-user.talk com.canonical.AppMenu.Registrar
```
kris7t, line 54: I think this would be needed for any Gnome application for top menu support, as well as for Qt applications with top menu support on X11. Top menu support on Wayland seems much more complicated, however (I was musing on that earlier: https://www.reddit.com/r/swaywm/comments/fn1uf9/global_app_menus/ , but the protocols are a mess of DBus and custom Wayland extensions).

kris7t:
> BTW: `firejail --noprofile --dbus-user=filter --audit | grep D-Bus` `MAYBE: D-Bus socket /run/firejail/mnt/dbus/user is available`.

Well, that socket is the proxy socket, so it should be present. But yeah, I should fix the audit so it does not emit a warning, it makes no sense to warn for a properly running filter.

etc/templates/profile.template
```diff
@@ -186,6 +187,20 @@ include globals.local
...
+# Since 0.9.63 also a more granular regulation of dbus is supported.
+# To get the dbus-addresses to which an application needs access to you can look
+# at flatpak if the application is also distriputed with flatpak:
+#    flatpak remote-info --show-metadata flathub <APP-ID>
+# Notes:
+#  - flatpak implicitly allows an app to own <APP-ID> on the session bus
+#  - In order to make dconf work (if it is used by the app) you need to allow
+#    'ca.desrt.dconf' even if it is not allowed by flatpak.
+#dbus-user filter
+#dbus-user.own com.github.netblue30.firejail
```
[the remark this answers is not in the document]
rusty-snake: Thats just an example, it would not pass any review. However we can change it to `com.example.appname`.

etc/templates/profile.template
```diff
@@ -186,6 +187,20 @@ include globals.local
...
+#dbus-user.talk ca.desrt.dconf
```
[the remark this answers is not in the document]
rusty-snake: > I wonder what's the flatpak solution for this vulnerability.

https://docs.flatpak.org/en/latest/sandbox-permissions.html#dconf-access, a lot of flatpaks still use direct access. gnome-builder suggests it.

etc/firefox.profile
```diff
@@ -28,5 +28,11 @@ include whitelist-usr-share-common.inc
 # private-etc must first be enabled in firefox-common.profile
 #private-etc firefox
 
+dbus-user filter
+dbus-user.own org.mozilla.firefox.*
```
kris7t, line 32: Starting with Firefox 75, there should probably also be a `dbus-user.own org.mpris.MediaPlayer2.firefox.*` for hardware media key support.

etc/templates/profile.template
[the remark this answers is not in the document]
rusty-snake: > A malicious application could, for example, change stuff under `org.gnome.system.proxy` to MITM traffic.

Or even worser: `/org/gnome/terminal/legacy/profiles:/<UUID>/custom-command`

etc/celluloid.profile
```diff
@@ -47,5 +46,12 @@ private-etc alternatives,ca-certificates,crypto-policies,dconf,drirc,fonts,gtk-3
 private-dev
 private-tmp
 
+dbus-user filter
+dbus-user.own io.github.celluloid_player.Celluloid
+dbus-user.talk org.gnome.SettingsDaemon.MediaKeys
+dbus-user.talk org.gtk.vfs.*
+dbus-user.talk org.gtk.vfs
```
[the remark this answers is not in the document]
rusty-snake: _Resolved other `org.gtk.vfs` commend in order to have one thread._

Your right, the most programs don't need it.

TODO: can it be used to bypass `disable-mnt`/`blacklist`/`whitelist`.

etc/templates/profile.template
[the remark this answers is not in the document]
rusty-snake: I only added it there as an example since it is common used.

etc/firefox.profile
rusty-snake: I wanted to start minimal, there a lot more path for firefox + extensions. The policy used by the inoffical firefox-flatpak:
```
dbus-user filter
dbus-user.own org.mozilla.Firefox
dbus-user.own org.mozilla.firefox.*
dbus-user.talk org.freedesktop.FileManager1
dbus-user.talk org.gnome.SessionManager
dbus-system filter
dbus-system.talk org.freedesktop.NetworkManager
```

etc/ghostwriter.profile
rusty-snake: Maybe we should just drop `com.canonical.AppMenu.Registrar` from profiles and add a FAQ point to use globals.local if wanted.

etc/firefox.profile
kris7t: Whoah, `org.freedesktop.NetworkManager` sounds quite evil in this context (MITM via adding a VPN, or leaking IP and MAC addresses of interfaces otherwise isolated with `--net=`.

Fred-Barclay:
@rusty-snake I was going to replace all `nodbus` in profiles with
```
dbus-system none
dbus-user none
```
as a general starting point

Ok to push here or should I open a new PR? It's probably lots of files so didn't want to clutter up this PR.

rusty-snake:
@Fred-Barclay if you push upstream, I can rebase.

Nitpick: I used `dbus-user` before `dbus-system`.

Fred-Barclay:
@rusty-snake I'm usually bad at alphabetising... in this case though for consistency's sake I'd guess `dbus-system` before `dbus-user`. What do you think?

rusty-snake:
The must profiles will disable dbus or look like the example below. For my eyes it was nicer to have a `dbus-system none` at the bottom instead of the top.

```
dbus-user filter
dbus-user.own com.foo.bar
dbus-user.talk ca.desrt.dconf
dbus-user.talk org.freedesktop.smile
dbus-system none
```

The addresses (grouped by own and talk) should be alphabetical sure.

Fred-Barclay:
@rusty-snake done, see 3848b98961614e1776b29ecfb76ef4c750b6b25f

There are a few small changes left, like how to handle `?HAS_NODBUS`, but that should affect >5 files.

etc/firefox.profile
```diff
@@ -28,5 +28,11 @@ include whitelist-usr-share-common.inc
...
+dbus-user.talk org.freedesktop.Notifications
+dbus-system none
+ignore nodbus
```
glitsj16, line 35: I get why we need to have `ignore nodbus` here. Does that mean we can remove it from firefox-common-addons.inc?

glitsj16:
Suppose a profile has `dbus-user.talk org.freedesktop.Notifications` and a user wants to override that, and only that. Is `ignore` still the correct/preferred way to do this with regards to the DBus filters?

etc/firefox.profile
rusty-snake: Now needs to be `ignore dbus-user none`.

firefox-common.profile is used by firefox, thunderbird, waterfox, ... and disables D-Bus access. Of course we can't add the own rule to firefox-common. Because some program throw errors messages if D-Bus isn't available, but hardfail on D-Bus error due to work policy, I added this for now only to firefox.

etc/templates/profile.template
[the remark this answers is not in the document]
abranson: A sandboxed dconf similar to flatpak's would be a logical next step after DBus

rusty-snake:
I created a short list about some of the D-Bus addresses.
```
Legend:
# |0| => harmless
# |!| => Danger sandbox escape possible
# |!!| => Very dangerous
# |$| => sensitive

# |0| MediaKeys
org.gnome.SettingsDaemon.MediaKeys
# |!| Dconf
ca.desrt.dconf
# |0| NightMode/Screen temperaur
org.gnome.SettingsDaemon.Color
# |!!| Native notifications
org.freedesktop.Notifications
# |$| Keyring
org.freedesktop.secrets
# |$|
org.gnome.OnlineAccounts
# |0|
org.mpris.MediaPlayer2.APP
# |!!| Filesystem access (blacklist/whitelist esacape)
org.freedesktop.FileManager1
# |$| Unlock screen
org.freedesktop.ScreenSaver
# |$| reboot(?)
org.gnome.SessionManager
# |!|
org.freedesktop.NetworkManager
# |$|
org.gnome.evolution.dataserver.*
# |$|
org.freedesktop.GeoClue2
# |!|
org.gnome.Shell
# |!!|
org.gnome.Shell.*
# |!|
org.freedesktop.login1
# |!|
org.gtk.vfs.*


Things that `org.freedesktop.Notifications` can do:

# Screenshots

firejail --noprofile --dbus-user=filter --dbus-user.talk=org.freedesktop.Notifications dbus-send --session --print-reply --dest=org.freedesktop.Notifications /org/gnome/Shell/Screenshot org.gnome.Shell.Screenshot.Screenshot boolean:false boolean:false string:screenshot.png

# Unlock your screen

firejail --noprofile --dbus-user=filter --dbus-user.talk=org.freedesktop.Notifications dbus-send --session --print-reply --dest=org.freedesktop.Notifications /org/gnome/ScreenSaver org.gnome.ScreenSaver.SetActive boolean:false

# Install gnome-shell-extensions

# /org/gnome/Shell org.gnome.Shell **Eval** any code (gjs)

# /org/gnome/Mutter/ScreenCast org.gnome.Mutter.ScreenCast

Record your screen

# /org/gnome/keyring/Prompter org.gnome.keyring.internal.Prompter

IDK what it can, but keyring sounds not good.

# /org/gnome/SessionManager/EndSessionDialog org.gnome.SessionManager.EndSessionDialog

Logout; Poweroff?; Reboot?

# /org/gnome/Mutter/RemoteDesktop org.gnome.Mutter.RemoteDesktop

forward desktop

# /org/gnome/Mutter/DisplayConfig

turnoff the monitor

# force you to hard-poweroff
```

Fred-Barclay:
@rusty-snake Awesome, care to toss it in the wiki?

rusty-snake:
:+1: Definitely something for the wiki. with more comments/descriptions and better formating.

---

What policy should be apply for `org.freedesktop.Notifications`?

glitsj16:
> What policy should be apply for org.freedesktop.Notifications?

That's one of the reasons I wanted to add an example to the firejail-profile man page. I admit not spending a whole lot of time on it and decided to add an example on how to override a given setting. But for this I would have liked a `dbus-system.talk org.freedesktop.Notifications none` option (or something similar). Probably not easily - if at all - implementable, but I for one will be scanning future profiles that allow it and do whatever I can to disable. Just a personal opinion...

rusty-snake:
flatpak like: `dbus-user.no-talk`

I would say making it opt-in for whitelisting profiles.

kris7t:
something like `dbus-user.no-talk` is potentially tricky (originally, I considered it, but decided not to implement for now) due to the presence of `.*` patterns in talk rules. Basically, what should happen if we have (fictitious example, obvious not a good idea to apply these exact rules):

```
dbus-user.talk org.freedesktop.*
dbus-user.no-talk org.freedesktop.Notifications
```

The challenge here is that, as far as I could find, there is no way to pass _negative_ rules to `xdg-dbus-proxy`. We could resort to an approach similar to filesystem whitelisting, i.e., allow all names below `org.freedesktop` _current present_ on the bus except `Notifications`, but that is complicated (we have to connect to the bus and list all names) and wrong (new names may appear all the time). Alternatively, we could just throw an error if a `no-talk` rule is superseded by a `talk` rule (i.e., a `no-talk` with no corresponding `talk` rule or a `talk` rule which would whitelist that exact pattern of bus name(s) is ok, but anything more general than that is an error). That's pretty much what an `ignore` command does, tho.

I should really look into how Flatpak handles this.

rusty-snake:
I would say `dbus-user.no-talk` can act like (no)blacklist, path/address of the no command *must* match the path/address of the blacklist/talk command.

glitsj16:
@kris7t Very informative ^. I might have missed it, but have there been suggestions on how to deal with `org.freedesktop.portal.*` (see [here](https://github.com/flatpak/xdg-desktop-portal)) yet?

rusty-snake:
https://github.com/netblue30/firejail/wiki/Restrict-D-Bus -- [disscussion](https://github.com/netblue30/firejail/issues/3349)

kris7t:
@glitsj16 Unfortunately, I don't know much about xdg-portals. I think Gnome has some kind of settings dialog for at least flatpak app permissions (filesystem access, webcam, microphone, etc.). I don't know whether that's just black/whitelisting the D-Bus names of portals, or something implemented inside the portals themselves.

rusty-snake:
From my side we could merge.

glitsj16:
> From my side we could merge.

@rusty-snake LGTM. Great work and a big thank you to all involved!

PR #3331, opened 2020-04-08
Don't change default xpra window title

alnionet:
Currently firejail adds the --title option to the xpra attach command. There is no way to customize this title and it's really annoying in case of browsers where you would expect to have the page title in the window title.


dmfreemon:
I use multiple home directories, and this PR would eliminate any indication of which jail which browser is running in (I often have multiple browsers running in different jails).  I personally don't find much value in the default titlebar of "web page title, the word firefox, and the hostname", but maybe others do.  Perhaps there could be an option or something to switch between these behaviors.

alnionet:
Maybe adding a suffix to the existing title like xpra does is a good idea.

crgithubmoth:
Please by default add to the existing xpra title, but even better to have an option to use a custom string, or by string+ to prefix or +string for a suffix.
I need especially to be able to add the user to the xpra window.

startx2017:
> I personally don't find much value in the default titlebar of "web page title, the word firefox, and the hostname",

That's a valid point, better to have the name of the sandbox there.

PR #3334, opened 2020-04-09
Request behavior change description in bug reports

matu3ba:
Do not miss firejail --noprofile PROGRAM as default bug reporter input.


rusty-snake:
Merged, Thanks.

PR #3337, opened 2020-04-09
Build improvements

topimiettinen:
Sometimes concurrent build could fail if the filter apps were not
made before attempting to make the filters.


netblue30:
Merged, thanks.

Issue #3333, opened 2020-04-09
Creating temporary filesystem from shell fails


matu3ba:
**Describe the bug**
Running `firejail --private firefox` or simply `firejail firefox` ie fails with
```
Error: execute permission denied for /usr/local/bin/firefox
Error: no suitable firefox executable found
```

**To Reproduce**
1. Run in shell `firejail firefox`
2. Get output:
Reading profile /usr/local/etc/firejail/firefox.profile
Reading profile /usr/local/etc/firejail/whitelist-usr-share-common.inc
Reading profile /usr/local/etc/firejail/firefox-common.profile
Reading profile /usr/local/etc/firejail/disable-common.inc
Reading profile /usr/local/etc/firejail/disable-devel.inc
Reading profile /usr/local/etc/firejail/disable-exec.inc
Reading profile /usr/local/etc/firejail/disable-interpreters.inc
Reading profile /usr/local/etc/firejail/disable-programs.inc
Reading profile /usr/local/etc/firejail/whitelist-common.inc
Reading profile /usr/local/etc/firejail/whitelist-var-common.inc
Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
Parent pid 63204, child pid 63205
Warning: An abstract unix socket for session D-BUS might still be available. Use --net or remove unix from --protocol set.
Warning: cleaning all supplementary groups
Warning: cleaning all supplementary groups
Warning: cleaning all supplementary groups
Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Warning: not remounting /run/user/1000/gvfs
Warning: cleaning all supplementary groups
Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
Child process initialized in 118.04 ms
Error: execute permission denied for /usr/local/bin/firefox
Error: no suitable firefox executable found

Parent is shutting down, bye...

**Expected behavior**
Firefox should start with a temporary  filesystems.

**Desktop (please complete the following information):**
 - Linux distribution and version
Manjaro Linux 19.0.2
 - Firejail version (output of `firejail --version`)
firejail version 0.9.63

Compile time support:
	- AppArmor support is disabled
	- AppImage support is enabled
	- chroot support is enabled
	- file and directory whitelisting support is enabled
	- file transfer support is enabled
	- firetunnel support is enabled
	- networking support is enabled
	- overlayfs support is enabled
	- private-home support is enabled
	- seccomp-bpf support is enabled
[...]

PR #3339, opened 2020-04-09
early decision in bug report if using git version

matu3ba:


matu3ba:
Questions like [this](https://github.com/netblue30/firejail/issues/3333#issuecomment-611334207) could be prevented by this.

Fred-Barclay:
Merged, thanks!

PR #3359, opened 2020-04-15
32bit ARM syscall table

abranson:
Submitting a syscall table as requested during build. It's a bit manual - the header was a bit too complex for the given awk so I cleaned it up manually. There were a few of these undefined for eabi, but I thought it didn't hurt to keep them in.


netblue30:
merged, thanks!

PR #3362, opened 2020-04-17
Profile for jitsi-meet-desktop 

Kishore96in:
Profile for the Jitsi (https://jitsi.org/jitsi-meet/) desktop app. Jitsi works with this profile on my machine.

### DIFF
--- a/etc/jitsi-meet-desktop.profile
+++ b/etc/jitsi-meet-desktop.profile
@@ -1,5 +1,5 @@
 # Firejail profile for jitsi-meet-desktop
-# Description: Build cross platform desktop apps with web technologies
+# Description: Jitsi Meet desktop application powered by Electron
 # This file is overwritten after every install/update
 # Persistent local customizations
 include jitsi-meet-desktop.local
--- a/etc/jitsi-meet-desktop.profile
+++ b/etc/jitsi-meet-desktop.profile
@@ -18,7 +18,9 @@ include disable-xdg.inc
 nowhitelist ${DOWNLOADS}
 
 mkdir ${HOME}/.config/Jitsi Meet
+
 whitelist ${HOME}/.config/Jitsi Meet
+
 include whitelist-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-runuser-common.inc
--- a/etc/jitsi-meet-desktop.profile
+++ b/etc/jitsi-meet-desktop.profile
@@ -30,5 +30,6 @@ protocol unix,inet,inet6,netlink
 seccomp !chroot
 
 disable-mnt
+private-bin jitsi-meet-desktop,bash
 private-dev
 private-tmp
--- a/etc/jitsi-meet-desktop.profile
+++ b/etc/jitsi-meet-desktop.profile
@@ -1,0 +1,30 @@
+# Firejail profile for jitsi-meet-desktop
+# Description: Build cross platform desktop apps with web technologies
+# This file is overwritten after every install/update
+# Persistent local customizations
+include jitsi-meet-desktop.local
+# Persistent global definitions
+include jitsi-meet-desktop.local
+
+noblacklist ${HOME}/.config/Jitsi Meet
+
+include disable-common.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+
+whitelist ${DOWNLOADS}
+mkdir ${HOME}/.config/Jitsi Meet
+whitelist ${HOME}/.config/Jitsi Meet
+
+apparmor
+caps.drop all
+netfilter
+nodvd
+nogroups
+nonewprivs
+noroot
+notv
+protocol unix,inet,inet6,netlink
+seccomp !chroot
+
+private-tmp
--- a/etc/jitsi-meet-desktop.profile
+++ b/etc/jitsi-meet-desktop.profile
@@ -6,10 +6,13 @@ include jitsi-meet-desktop.local
 # Persistent global definitions
 include globals.local
 
+ignore noexec /tmp
+
 noblacklist ${HOME}/.config/Jitsi Meet
 
 include disable-common.inc
 include disable-devel.inc
+include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
--- a/etc/jitsi-meet-desktop.profile
+++ b/etc/jitsi-meet-desktop.profile
@@ -10,12 +10,9 @@ ignore noexec /tmp
 
 noblacklist ${HOME}/.config/Jitsi Meet
 
-include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
-include disable-passwdmgr.inc
-include disable-programs.inc
 include disable-xdg.inc
 
 mkdir ${HOME}/.config/Jitsi Meet
@@ -25,15 +22,6 @@ include whitelist-usr-share-common.inc
 include whitelist-runuser-common.inc
 include whitelist-var-common.inc
 
-apparmor
-caps.drop all
-netfilter
-nodvd
-nogroups
-nonewprivs
-noroot
-notv
-protocol unix,inet,inet6,netlink
 seccomp !chroot
 
 disable-mnt
@@ -41,3 +29,6 @@ private-bin jitsi-meet-desktop,bash
 private-cache
 private-dev
 private-tmp
+
+# Redirect
+include electron.profile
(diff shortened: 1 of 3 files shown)


etc/jitsi-meet-desktop.profile
```diff
@@ -0,0 +1,30 @@
+# Firejail profile for jitsi-meet-desktop
+# Description: Jitsi Meet desktop application powered by Electron
+# This file is overwritten after every install/update
+# Persistent local customizations
+include jitsi-meet-desktop.local
+# Persistent global definitions
+include jitsi-meet-desktop.local
```
rusty-snake, line 7:
```suggestion
include globals.local
```

etc/jitsi-meet-desktop.profile
```diff
@@ -0,0 +1,30 @@
...
+
+noblacklist ${HOME}/.config/Jitsi Meet
```
rusty-snake, line 9: Add `blacklist ${HOME}/.config/Jitsi Meet` to disable-programs.inc

etc/jitsi-meet-desktop.profile
```diff
@@ -0,0 +1,30 @@
...
+
+include disable-common.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
```
rusty-snake, line 13: What about:
 - disable-devel
 - disable-exec
 - disable-interpreters
 - disable-xdg

etc/jitsi-meet-desktop.profile
```diff
@@ -0,0 +1,30 @@
...
+
+whitelist ${DOWNLOADS}
+mkdir ${HOME}/.config/Jitsi Meet
+whitelist ${HOME}/.config/Jitsi Meet
```
rusty-snake, line 17: Add `include whitelist-common.inc`

what about whitelist-var-common, whitelist-usr-share-common, whitelist-runuser-common?

etc/jitsi-meet-desktop.profile
```diff
@@ -0,0 +1,30 @@
...
+netfilter
+nodvd
+nogroups
+nonewprivs
+noroot
+notv
+protocol unix,inet,inet6,netlink
+seccomp !chroot
+
+private-tmp
```
rusty-snake, line 30: What about 
- `disable-mnt`
- `private-cache`
- `private-bin`
- `private-etc`
- `private-dev` (if it not break cam/micro)

etc/jitsi-meet-desktop.profile
Kishore96in: It doesn't launch with disable-exec. I've added the others

Kishore96in:
I didn't use the electron profile because the application doesn't launch with ```seccomp``` (we need to use ```seccomp !chroot```).

etc/jitsi-meet-desktop.profile
Kishore96in: Jitsi Meet doesn't seem to use the ```.cache``` directory, so I have not used ```private-cache```. There is a cache directory in ```.config/Jitsi Meet```, which is what the application seems to use.

Kishore96in: For ```private-etc```, I still haven't been able to figure out which files in ```/etc/``` it expects to be able to access. I have only been able to get it to show a blank window till now. I'll add it the option when I figure that out.

etc/jitsi-meet-desktop.profile
Kishore96in: I do not understand the need for this. It seems to run fine without including all these whitelist rules. Wouldn't this just unnecessarily give it the ability to edit the whitelisted files?

etc/jitsi-meet-desktop.profile
Kishore96in: Done.

etc/jitsi-meet-desktop.profile
rusty-snake: Try `disable-exec.inc` and `ignore noexec ${HOME}` (maybe also `ignore noexec /tmp`).

The `ignore` commands should be put between `include globals.local` and `noblacklist` (with blank lines).

etc/jitsi-meet-desktop.profile
rusty-snake: > give it the ability to edit the whitelisted files?

Some of the paths there are read-only (in whitelist-common or disable-common).

> It seems to run fine without including all these whitelist rules. 

- `whitelist-common.inc`: for you it works, but it may need some file for there on other setup. this is for easier maintenance 
- wvc, wusc, wruc: add these increases the security. Example wvc: w/o it all files under /var are visible. with wvc only the few whitelisted paths are visible.

etc/jitsi-meet-desktop.profile
rusty-snake: You can use `firejail --build path/to/jitsi-meet-desktop`.

> Jitsi Meet doesn't seem to use the .cache directory, so I have not used private-cache.

`private-cache` means also hide the the caches of other apps. (This is also done by whitelist, but if a user deactivates whitelist the cache is still protected)

rusty-snake:
You can redirect to electron.profile and set the seccomp exception. Like riot-desktop.

Kishore96in:
The electron profile doesn't seem to contain a lot of the extra options which you mentioned (```whitelist-*.inc``` and ```private-*```). Should I leave them out, or include them in this profile before redirecting to the electron one?

rusty-snake:
The last two lines of jitsi-meet-desktop.profile should look like this.
```
# Redirect
include electron.profile
```

Then you can remove every thing from jitis-meet-desktop.profile already set in electron.profile.

Kishore96in:
I noticed the ```electron.profile``` whitelists the Downloads directory, while this program doesn't have any file sharing / downloading functionality AFAIK. Should I add a ```nowhitelist```, or leave it as it is?

etc/jitsi-meet-desktop.profile
Kishore96in: The manpage says under ```---build``` that Chromium-based browsers won't work, which I guess explains why the program doesn't launch with ```firejail --build```.

Any ideas on how one would figure out which filesin ```/etc``` it would need?

rusty-snake: `--trace` will also break, you could try it with the templates.

Try this (still need ordering), of it does not work leave it out.
`private-etc alternatives,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,mime.types,xdg,passwd,group,ca-certificates,ssl,pki,crypto-policies,nsswitch.conf,resolv.conf,hosts,host.conf,hostname,protocols,services,rpc,alsa,asound.conf,pulse,machine-id,fonts,pango,X11,drirc,glvnd,bumblebee,nvidia`

Kishore96in: The one you suggested works. Thanks a lot.

rusty-snake:
Sounds not bad, removing unused permissions.

---

```
private-bin bash,jitsi-meet-desktop
private-etc alsa,alternatives,asound.conf,bumblebee,ca-certificates,crypto-policies,drirc,fonts,glvnd,group,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,machine-id,mime.types,nsswitch.conf,nvidia,pango,passwd,pki,protocols,pulse,resolv.conf,rpc,services,ssl,X11,xdg
```

etc/jitsi-meet-desktop.profile
```diff
@@ -0,0 +1,36 @@
...
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/Jitsi Meet
+whitelist ${HOME}/.config/Jitsi Meet
+include whitelist-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-runuser-common.inc
+include whitelist-var-common.inc
+
+nowhitelist ${DOWNLOADS}
```
Kishore96in, line 25: I couldn't find any ```nowhitelist``` command in ```profile.template```. Is this the right place to do this, or should I put it with the ```whitelist``` commands?

rusty-snake: I recently added it just before the whitelist-block.

Kishore96in: Moved. I've also put in some newlines to separate sections.

rusty-snake:
Merged, Thanks.

Issue #3360, opened 2020-04-15
Cannot start Libreoffice with the latest Firejail


youknow16:

matu3ba:
Looks like this may be [unrelated](https://ask.libreoffice.org/en/question/136435/lo-wont-start/). 
Libreoffice profiles may just [break](https://bugs.documentfoundation.org/show_bug.cgi?id=98027).

Please try resetting your profile and report.
Meta: Are you running libreoffice as superuser?

If the problem persists:
It is weird that the program fails to read from that path, because libreoffice contains `noblacklist ${HOME}/.config/libreoffice`.
What is your output of `firejail --debug libreoffice` ?

rusty-snake:
> Meta: Are you running libreoffice as superuser?

https://github.com/netblue30/firejail/wiki/Frequently-Asked-Questions#ive-noticed-the-title-bar-in-firefox-shows-as-superuser-is-this-normal

> It is weird that the program fails to read from that path, because the libreoffice profile contains noblacklist ${HOME}/.config/libreoffice.

> Describe clear and concise what changed calling firejail --noprofile PROGRAM in a shell.
The exact same thing happens with firejail --noprofile libreoffice.
The only way to run libreoffice is by directly running /usr/bin/libreoffice

youknow10:
> What is your output of `firejail --debug libreoffice` ?
Here is the result of that command:  [output.txt](https://github.com/netblue30/firejail/files/4499931/output.txt)

matu3ba:
```
Searching $PATH for libreoffice
trying #/home/ThisUser/anaconda3/bin/libreoffice#
trying #/home/ThisUser/anaconda3/condabin/libreoffice#
trying #/usr/local/sbin/libreoffice#
trying #/usr/local/bin/libreoffice#
trying #/usr/sbin/libreoffice#
trying #/usr/bin/libreoffice#
Installing /run/firejail/mnt/seccomp/seccomp seccomp filter
Installing /run/firejail/mnt/seccomp/seccomp.32 seccomp filter
Installing /run/firejail/mnt/seccomp/seccomp.protocol seccomp filter
monitoring pid 10

I/O warning : failed to load external entity "/home/ThisUser/.config/libreoffice/4/user/config/javasettings_Linux_X86_64.xml"
javaldx failed!
Warning: failed to read path from javaldx

[...]

Issue #3361, opened 2020-04-16
jdownloader profile doesn't work


Micha-Btz:
Hey,

I don't now since when, the jdownloader profile doesn't work for me. I have tried to recover it, but without luck.

```
~/.config/firejail$ cat JDownloader2.profile 
# Firejail profile for JDownloader
# This file is overwritten after every install/update
# Persistent local customizations
include JDownloader2.local
include /etc/firejail/firefox.profile
# Persistent global definitions
 include globals.local
  
 noblacklist ${HOME}/.jd
 noblacklist ${HOME}/media/jd2
 noblacklist ${HOME}/.scripte/
 noblacklist ${HOME}/.install4j
 
 noblacklist /usr/lib/jvm/
 
 mkdir ${HOME}/media/jd2
 whitelist ${HOME}/media/jd2
 mkdir ${HOME}/.scripte/
 noblacklist ${HOME}/.scripte/
 
 noblacklist ${HOME}/.mozilla
 
 # Allow java (blacklisted by disable-devel.inc)
 include allow-java.inc
 
 include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-xdg.inc
 
 mkdir ${HOME}/.jd
 whitelist ${HOME}/.jd
 whitelist ${DOWNLOADS}
 include whitelist-common.inc
 include whitelist-var-common.inc
 include whitelist-usr-share-common.inc 
 
 
 caps.drop all
 netfilter
 no3d
 nodbus
 nodvd
 nogroups
 nonewprivs
 noroot
 nosound
 notv
 nou2f
 novideo
 protocol unix,inet,inet6
 seccomp
 #shell none
 tracelog
 
 disable-mnt
 private-cache
 private-dev
 private-tmp
```

I use some different paths, this are normally located in the .local file, but for test I have put it together.
In the Jdownloader2 start script I have set INSTALL4J_JAVA_HOME_OVERRIDE=/usr/lib/jvm/java-11-openjdk-amd64/ which worked without firejail without problems.

Here is the debug output:

 
 ```
seccomp filter configured
 Mounting read-only /run/firejail/mnt/seccomp
 Dropping all capabilities
 noroot user namespace installed
 Dropping all capabilities
 NO_NEW_PRIVS set
 Drop privileges: pid 1, uid 1000, gid 1000, nogroups 1
 No supplementary groups
[...]

PR #3370, opened 2020-04-20
github issue template ask unusual setups,debug warnings/errors

matu3ba:
attempt to simplify support for cases like #3312 and #3360

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,27 +6,30 @@ labels: ''
 assignees: ''
 
 ---
-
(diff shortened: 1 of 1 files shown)


matu3ba:
i dont know how to test visuals however  
tracking issue for improving `--debug` would be also helpful  
discussion:
lately we had weird java/dbus issues hard to debug in #3360 and #3361  
do you have ideas on that?

case #3312 could be used as improvement of `--debug` to emit default program usage, but I fear every distro cooks its own meal

glitsj16:
@matu3ba Could you resolve the conflict please. We can merge it when that's done. Thanks!

PR #3373, opened 2020-04-21
update --build

### LINKED ISSUE #2150: --build should not assume that firejail's default profile dir is /etc/firejail
The option `--build` is hardcoded to build profiles assuming the firejail profile dir is `/etc/local`, but it may not be if firejail was installed in a non-default location.  It should also use the `${CFG}` macro in the output.

rusty-snake:
The profile generated by --build are quite outdated. There are still a
lot of things left to do.

 - fix #2150 (whitelist-common.inc is still opened from /etc/firejail)
 - include wusc and wvc (todo: remove whitelists in wusc/wvc from the
   generated profile.)
 - fix parsing wc / use ${HOME} macro instead of ~
 - update profile headers
 - include all disable includes (mustly commented) in the output
 - reorder the filesystem section

### DIFF
--- a/src/fbuilder/build_fs.c
+++ b/src/fbuilder/build_fs.c
@@ -165,10 +165,12 @@ void build_var(const char *fname, FILE *fp) {
 
 	process_files(fname, "/var", var_callback);
 
-	if (var_out == NULL)
+	if (var_out == NULL) {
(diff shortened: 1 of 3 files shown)


netblue30:
let's try it!

PR #3382, opened 2020-04-25
firecfg: Only use fix_desktop_files automatically when run through sudo

backspac:
It also fixes this error when running `firecfg` as root:
```
Error: this option is not supported for root user; please run as a regular user.
```

### DIFF
--- a/src/firecfg/main.c
+++ b/src/firecfg/main.c
@@ -370,7 +370,6 @@ int main(int argc, char **argv) {
 			return 0;
 		}
 		else if (strcmp(argv[i], "--fix") == 0) {
-			// fix .desktop files in ~/.local/share/applications directory
 			fix_desktop_files(home);
 			return 0;
 		}
@@ -486,6 +485,10 @@ int main(int argc, char **argv) {
 
 	if (arg_debug)
 		printf("%s %d %d %d %d\n", user, getuid(), getgid(), geteuid(), getegid());
+
+	// if runs as regular user, fix .desktop files in ~/.local/share/applications directory
+	if (getuid() != 0)
+		fix_desktop_files(home);
 
 	return 0;
 }
--- a/src/firecfg/main.c
+++ b/src/firecfg/main.c
@@ -370,6 +370,7 @@ int main(int argc, char **argv) {
 			return 0;
 		}
 		else if (strcmp(argv[i], "--fix") == 0) {
+			// fix .desktop files in ~/.local/share/applications directory
 			fix_desktop_files(home);
 			return 0;
 		}
@@ -485,9 +486,6 @@ int main(int argc, char **argv) {
 
 	if (arg_debug)
 		printf("%s %d %d %d %d\n", user, getuid(), getgid(), geteuid(), getegid());
-
-	// fix .desktop files in ~/.local/share/applications directory
-	fix_desktop_files(home);
 
 	return 0;
 }


rusty-snake:
Why? A loot user will end with unfixed .desktop files.

IMHO it would be better to make it conditional either add `--nofix` or better:
```
if ROOT AND SUDO {
  fix_desktop_files()
}
```

backspac:
Right. I see what you mean.
My main issue was that I was trying to setup a [Pacman hook](https://wiki.archlinux.org/index.php/Firejail#Using_Firejail_by_default) and it ran with errors.
In these conditions, I still have to manually run `firecfg --fix` as my regular user whenever I install a new desktop application.
Why doesn't `firecfg` handle `/usr/local/share/applications`? I don't see why it overwrites executables system-wide but desktop files user-wide.

glitsj16:
>  My main issue was that I was trying to setup a Pacman hook and it ran with errors.

What went wrong exactly?

>  Why doesn't firecfg handle /usr/local/share/applications? I don't see why it overwrites executables system-wide but desktop files user-wide.

I think the choice for ${HOME}/.local/share/applications is to ensure desktop integration works (more) reliable due to it having **precedence** over /usr/local/share/applications. E.g. a non-firecfg'ed _firefox.desktop_ under ${HOME} would be executed rather than what firecfg would drop under /usr/local. That would onlt create confusion and/or a false sense of security IMO.

backspac:
> What went wrong exactly?

When I run `pacman -Syu`, I get this "error":
```
(#/#) Configure symlinks in /usr/local/bin based on firecfg.config...
error: command failed to execute correctly
```
I say "error" because everything went fine in the end. It's just reported as such. It's mainly due to `firecfg` returning `1`.

@rusty-snake's fix is a good solution to this issue IMHO.

> That would only create confusion and/or a false sense of security IMO.

That makes sense, but don't you think it would be even better to make the package manager take care of everything, like putting desktop files in `/usr/local/share/applications`, and then giving the user the possibility to run `firecfg --fix` locally to go beyond security-wise.
In this way, it wouldn't require a manual action from the user every time they install a new application.

rusty-snake:
@backspac do you want do add this conditional fix?

backspac:
There you go.
Should I make a new issue for the handling of `/usr/local/share/applications`?

rusty-snake:
> Should I make a new issue for the handling of /usr/local/share/applications?

You can, but ATM there is one who fix firecfg bugs. That's why I started firecfg.py, I plan to suport also /etc/xdg/autostart.

src/firecfg/main.c
```diff
@@ -487,5 +486,9 @@ int main(int argc, char **argv) {
 	if (arg_debug)
 		printf("%s %d %d %d %d\n", user, getuid(), getgid(), geteuid(), getegid());
 
+	// if runs as regular user, fix .desktop files in ~/.local/share/applications directory
+	if (getuid() != 0)
```
rusty-snake, line 490: If we check `geteuid`, it would continue to work with sudo.

backspac: I don't see why, since `firecfg` doesn't have the `setuid` bit.

Also, `sudo` sets both `uid` and `euid` to the same value.
You can check it by running `sudo firecfg --debug`.

rusty-snake: Your right so far, but sudo also sets the `SUDO_USER` env-var, which is checked by `get_user` 

https://github.com/netblue30/firejail/blob/2359fa901d6ac4a266ca2a1c0ddf31939a3a9863/src/firecfg/main.c#L286-L287

which is used here
https://github.com/netblue30/firejail/blob/2359fa901d6ac4a266ca2a1c0ddf31939a3a9863/src/firecfg/main.c#L323-L328

and then here

https://github.com/netblue30/firejail/blob/2359fa901d6ac4a266ca2a1c0ddf31939a3a9863/src/firecfg/main.c#L476-L484

backspac: Agreed. But I don't see what it has to do with `geteuid()`.

`sudo firecfg` is still working as expected on my system.

rusty-snake: If you start it with sudo, the real user-id is always 0, right? And `getuid` checks the real-user-id (https://linux.die.net/man/2/getuid). Therefore fix_desktop_files is never called. When checking the effective-uid it would get called.

what am I missing?

backspac: `sudo` sets both `uid` and `euid` to the same value. So, if you start it with `sudo`, both the real user ID and the effective user ID will be set to `0`.

I think you are missing the permissions drop that happens just before:
https://github.com/netblue30/firejail/blob/2359fa901d6ac4a266ca2a1c0ddf31939a3a9863/src/firecfg/main.c#L482-L483

After these lines, both `uid` and `euid` are set back to the original user ID.

rusty-snake: https://linux.die.net/man/2/setuid
If I had read not only the first, but also the second sentence…

rusty-snake:
Merged, Thanks.

PR #3386, opened 2020-04-26
Increase MAX_ENVS to 256

### LINKED ISSUE #3350: Question: Environment variables
Currently the number of environment variables is restricted to 100 in `firejail.h`.
```
#define MAX_ENVS 100	// some sane maximum number of environment variables
```
For tmux or screen (or weird shell quirks) these however blow up to in my case 171. When I call from within tmux git push/pull, this fails with `fprintf(stderr, "Error: too many environment variables\n");`. Any standard shell uses ~80 environment variables (`printenv | wl -l`).  
What is the advisory for such programs? Do there exist techniques to drop the variables for the program call, so it has no access to the unnecessary environment information or is this still wild west-like?

topimiettinen:
Some applications like Byobu, tmux and screen like to use environment
and then 100 environment variables may be too few.

Closes #3350

### DIFF
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -649,7 +649,7 @@ int check_namespace_virt(void);
 int check_kernel_procs(void);
 void run_no_sandbox(int argc, char **argv);
 
-#define MAX_ENVS 100			// some sane maximum number of environment variables
+#define MAX_ENVS 256			// some sane maximum number of environment variables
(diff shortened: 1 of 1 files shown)


glitsj16:
@topimiettinen Looks fine and should avoid issues being reported (frequently) on it in the future. It was a learning experience to have it capped at 100 originally; great incentive to cleanup my env var organization :+1:

brotherzhao2019:
Can I increase MAX_ENVS to 2048?

PR #3390, opened 2020-05-01
Disable browser drm by default.

Liorst4:
Browser DRM was not disabled by default, even though etc/firejail/firejail.config stated that it was.


glitsj16:
Nice catch! If you checked that the `?BROWSER_ALLOW_DRM:` conditional still works as expected we can merge this.

Liorst4:
Checked by running bash in firefox profile.

firefoxs profile uses `?BROWSER_ALLOW_DRM` to deiced whether `${HOME}` should be non executable.

Created an executable shell script in /home/$USER/Downloads.

When browser-allow is unchanged in `firejail.config`, I got `permission denied`.

When browser-allow was set to yes, The script ran.

When browser-allow was set to no, I got `permission denied`.

rusty-snake:
Merged, Thanks.

PR #3397, opened 2020-05-03
Follow-up for #3326

glitsj16:
Purely cosmetic stuff, no functional changes.

### DIFF
--- a/etc/profile-a-l/chromium-common.profile
+++ b/etc/profile-a-l/chromium-common.profile
@@ -41,7 +41,7 @@ disable-mnt
 
 # prevents access to passwords saved in GNOME Keyring and KWallet, also breaks Gnome connector
 # dbus-user none
-# dbus-system none
+dbus-system none
 
 # the file dialog needs to work without d-bus
 ?HAS_NODBUS: env NO_CHROME_KDE_FILE_DIALOG=1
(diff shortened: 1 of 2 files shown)


etc/profile-a-l/chromium-common.profile
```diff
@@ -40,5 +39,9 @@ disable-mnt
 ?BROWSER_DISABLE_U2F: private-dev
 # private-tmp - problems with multiple browser sessions
 
+# prevents access to passwords saved in GNOME Keyring and KWallet, also breaks Gnome connector
+# dbus-user none
+# dbus-system none
```
rusty-snake, line 44: Can we block the system-bus?

glitsj16: @rusty-snake Not sure, I'm actually not that familiar with chrome-based browsers. So if you say the system bus can be blocked without issues, I'm ok with that. Thanks for the input, I didn't want to commit this without seeking some advice.

rusty-snake: Password are per user, therefore on the session-bus. 

In general: I block the system-bus in my `pre-globals.local` for around two month now, and only allowed it for d-feet and searhorse. IHMO we can add `dbus-system none` to the most profiles.

rusty-snake: But I'm not sure if it maybe still breaks something in chromium.

glitsj16: @rusty-snake That makes sense, thanks for your input, as always. I've blocked the system bus, we can change this later if we get issue reports.

Issue #3402, opened 2020-05-04
implement xdg-dbus-proxy --log / log denied D-Bus access tries


rusty-snake:
Implement xdg-dbus-proxy --log to log blocked D-Bus calls. Without this it is difficult debug things like #3400 or #3399.

_Alternative we could log all calls for writing for writing profiles or simply forward the --log output (stdout of xdg-dbus-proxy) to a file (e.g. /tmp/pid_of_sandbox-dbus) and parse it with a python/bash/... program in contrib._

see also https://github.com/netblue30/firejail/pull/3265#issuecomment-609477194

cc @kris7t

kris7t:
Sorry for not getting back to this earlier. I think capturing `xdg-dbus-proxy` output is the way forward, then post-processing it with a script (either to create a more human-readable log, or to generate profile line directly). The format of the logs looks way too annoying to process in a C program.

A complication is that `xdg-dbus-proxy` doesn't log separately for the session and system buses. Sometimes, interactive with the session and system buses is easily to filter out, but some names may be present on both buses. So to avoid confusion, I am now experimenting with single `--dbus-user.log` and `--dbus-session.log` options, as well as a `dbus-log=` option to direct output to a log file. A post-processing Python script may attempt to ingest a single log, connecting to the session and system buses to discern where each names belong (this can be somewhat of an approximation, if some names have disappeared since generating the log file), or take two logs (but then we have to run the application with logging two separate times).

In principle, we could even generate very fine-grained policies with `--call` and `--broadcast` filters that take bus name, interface member, object path triples. I added `--dbus-{user,system}.{call,broadcast}=` options to firejail to this end. However, these profiles can be unwieldy to write (hence the need for automatic generation from logs).

rusty-snake:
Thanks for the explanations.

> I think capturing xdg-dbus-proxy output is the way forward, then post-processing it with a script (either to create a more human-readable log, or to generate profile line directly).

Going this way, would allow different scripts for different things (generate profile, warn on violation, ...)

> Another annoyance is that while filter rules are bound to well-known names, but communication on the bus might use unique names. See e.g.,

From that what I saw (with --log and `dbus-monitor`) the output still contains the addresses (at least one time).

PR #3406, opened 2020-05-07
DBus filtering enhancements

kris7t:
This PR exposes additional functionalities from `xdg-dbus-proxy`:
* `--dbus-user.see` and `--dbus-system.see` allow setting the SEE policy for DBus names.
* `--dbus-{user,system}.{call,broadcast}` allows setting per interface and per object path policies. For example, we can set
  
  ```
  dbus-user.call org.freedesktop.Notifications=org.freedesktop.Notifications.*@/org/freedesktop/Notifications
  dbus-user.broadcast org.freedesktop.Notifications=org.freedesktop.Notifications.*@/org/freedesktop/Notifications
  ```

  to safely allow notifications without exposing additional objects (e.g. `/org/gnome/Shell`).
* `--dbus-{user,system}.log` to turn on DBus proxy logging (#3402). Unfortunately, this is not as useful as it may seem, because without `--filer`, `xdg-dbus-proxy --log` will not actually log anything. So we can use the options for debugging applications that fail due to DBus filters, but not for building DBus profiles from scratch.

In addition, this PR updates `faudit` so it won't report MAYBE when DBus filters are in effect (https://github.com/netblue30/firejail/pull/3326#issuecomment-610338629).

A next step would be to either find a better DBus logging option (maybe with `dbus-monitor`), or stay with this one, and create some contrib scripts to generate filter rules from (rejected) DBus communications.


kris7t:
1. Well, it means the app can find out that the bus name is being held by someone, and retrieve the objects from that bus name by introspection. So the app will know that it shouldn't take ownership of the name. However, if it tries to talk to objects exposed by the bus name, it will still fail, and may crash the application if it is not prepared for D-Bus errors.
2. Yeah, it works to some extends, but not as useful as I'd hoped (no logging unless filtering is active).
3. `call` and `broadcast` imply `see` (and potentially relax it). So if you have `dbus-user.call org.freedesktop.Notifications=org.freedesktop.Notifications.*@/org/freedesktop/Notifications`, that means the application can see the `org.freedesktop.Notifications` name, but only method calls of the interface `org.freedesktop.Notifications` to the object `/org/freedesktop/Notifications` will be successful. So it is pointless to combine this with `talk`, which would relax the policy even further (any method call succeeds and any signal can be received). `call`/`broadcast` should mostly be used in pairs, e.g.,  if you have `call` on `org.freedesktop.Notifications`, but not broadcast, you won't be called back if the user click the notification. So we could have something like `dbus-notifications.inc` with:

```
dbus-user.call org.freedesktop.Notifications=org.freedesktop.Notifications.*@/org/freedesktop/Notification
dbus-user.broadcast org.freedesktop.Notifications=org.freedesktop.Notifications.*@/org/freedesktop/Notification
```

and then

```
include dbus-notifications.inc
```

in any profile that needs access to the notification daemon.

kris7t:
Sorry for not checking back earlier, I was unfortunately very busy with projects at uni.

Does this look good to merge?

rusty-snake:
From the (low-level) profile interface standpoint, yes.

rusty-snake:
@kris7t xdg-dbus-proxy is started with arg0=3 which looks really strength in system-monitoing tools.


Screenshot_2020-06-02_00-17-31-fs8

glitsj16:
Confirming @rusty-snake's observation. E.g. when running dconf-editor the xdg-dbus-proxy process shows up with process names `0` in gnome-system-monitor. Not sure if there is anything that can be done about it, but if that's the case we can add it to the wiki page to avoid confused users filing issues on it.

kris7t:
@rusty-snake @glitsj16 That might be a side effect of using `fexecve` to start `/usr/bin/xdg-dbus-proxy` and gnome-system-monitor getting a bit confused. `argv[0]` should certainly be `"/usr/bin/xdg-dbus-proxy"`, but this is a bit weird, because we are starting it from and fd and not a path (#3241). At least for me, the correct `/usr/bin/xdg-dbus-proxy` name shows up when I view the process tree in htop. So this might be a bug in xdg-system-monitor.

glitsj16:
@kris7t Thanks for explaining, very informative. Using htop does indeed show the correct name. Personally I don't think it's a high-priority issue, but I'll try to file a bug report somewhere nonetheless. Take care!

rusty-snake:
@kris7t :+1: Just a question for my understanding, why we can not simply execute `fexecve(3, ["xdg-dbus-proxy", "--filter", ...], ...)`?

kris7t:
@rusty-snake That's exactly what we are doing: https://github.com/netblue30/firejail/blob/d1dd36355aa7cd62b374aefa47822ebf5bb0ecac/src/firejail/sbox.c#L233

Here `arg` is the `argv` passed to the binary we want to exec. First we `fopen` `arg[0]`, `fstat` it to check permissions, and if there is no tampering, we `fexecve` the opened fd, and pass the full `arg` (including `arg[0]`, which should be `"/usr/bin/xdg-dbus-proxy"` in our case now), as well as the sanitized environment.

rusty-snake:
> At least for me, the correct /usr/bin/xdg-dbus-proxy name shows up when I view the process tree in htop. So this behavior might be a bug in xdg-system-monitor.

Hmm, `ps -eH` shows me `3`, `ps -eFH` `/usr/bin/xdg-dbus-proxy --fd=9 --args=10`.

PR #3409, opened 2020-05-10
Add several games to steam and disable-programs

corecontingency:
Add Faster Than Light, Into the Breach, Paradox Interactive, and mbwarband
to disable-programs.inc.

Also, add Faster Than Light and Into the Breach into steam.profile. This
fixes saved games being lost when steam is closed, and also lets Steam
cloud sync work properly.

Lastly, remove a duplicate whitelist ${HOME}/.steampid from
steam.profile.


rusty-snake:
Merged, Thanks.

PR #3410, opened 2020-05-11
add new profile: plv

glitsj16:
Add profile for [pacmanlogviewer](https://www.archlinux.org/packages/community/x86_64/pacmanlogviewer/).

### DIFF
--- a/etc/profile-m-z/plv.profile
+++ b/etc/profile-m-z/plv.profile
@@ -56,3 +56,4 @@ dbus-system none
 
 #memory-deny-write-execute - breaks opening file-chooser
 read-only ${HOME}
+read-write ${HOME}/.config/PacmanLogViewer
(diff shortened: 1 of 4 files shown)


etc/profile-m-z/plv.profile
```diff
@@ -0,0 +1,58 @@
...
+private-etc alternatives,fonts
+private-opt none
+private-tmp
+writable-var-log
+
+dbus-user none
+dbus-system none
+
+#memory-deny-write-execute - breaks opening file-chooser
+read-only ${HOME}
```
rusty-snake, line 58: `read-write ${HOME}/.config/PacmanLogViewer`?

glitsj16: Nice catch, forgot about the write access to its config. Thanks @rusty-snake!

PR #3411, opened 2020-05-11
disable-shell.inc

rusty-snake:


rusty-snake:
Initially I wanted to add it to disable-interpreters.inc, but to many program require allow-shell.inc. I think it is better to add it as an own disable-*.inc

rusty-snake:
> Should we document somewhere that including disable-shell.inc will break join?

:+1:, we should also note that `private-bin` w/o $SHELL breaks join too.

etc/profile-a-l/aria2c.profile
```diff
@@ -19,6 +19,7 @@ include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
+include disable-shell.inc
```
glitsj16, line 22: Aria2 supports `on-download-complete=foo` and we don't know what a user might put in a shell script. Perhaps we can add a comment, but the safest would be to not include disable-shell.inc here I guess. What do you think?

etc/profile-a-l/cower.profile
```diff
@@ -16,6 +16,7 @@ include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
+include disable-shell.inc
```
glitsj16, line 19: More a FYI, this package has been removed from AUR a while back, due to incompatibilities with pacman. So perhaps we can drop this profile. But the include would be fine regardless.

Fred-Barclay:
@rusty-snake how does this compare to `shell none`?

rusty-snake:
`disable-shell`: `blacklist`s shell inside the sandbox.
`shell none`: make firejail searching for the program in $PATH and `exec*` it direct by full path instead of `/bin/bash -c firefox`.

firejail will fail to start a program in a sandbox for profiles with disable-shell but w/o `shell none`.

PR #3418, opened 2020-05-15
Allow google-chrome access to the custom flags files in ~/.config.

mace015:
This PR modifies the google-chrome.profile to allow chrome access to its custom flag files in ~/.config.

### DIFF
--- a/etc/profile-a-l/google-chrome-beta.profile
+++ b/etc/profile-a-l/google-chrome-beta.profile
@@ -8,10 +8,16 @@ include globals.local
 noblacklist ${HOME}/.cache/google-chrome-beta
 noblacklist ${HOME}/.config/google-chrome-beta
 
+noblacklist ${HOME}/.config/chrome-beta-flags.conf
+noblacklist ${HOME}/.config/chrome-beta-flags.config
+
 mkdir ${HOME}/.cache/google-chrome-beta
 mkdir ${HOME}/.config/google-chrome-beta
 whitelist ${HOME}/.cache/google-chrome-beta
 whitelist ${HOME}/.config/google-chrome-beta
+
+whitelist ${HOME}/.config/chrome-beta-flags.conf
+whitelist ${HOME}/.config/chrome-beta-flags.config
 
 # Redirect
 include chromium-common.profile
--- a/etc/profile-a-l/google-chrome-unstable.profile
+++ b/etc/profile-a-l/google-chrome-unstable.profile
@@ -8,10 +8,16 @@ include globals.local
 noblacklist ${HOME}/.cache/google-chrome-unstable
 noblacklist ${HOME}/.config/google-chrome-unstable
 
+noblacklist ${HOME}/.config/chrome-unstable-flags.conf
+noblacklist ${HOME}/.config/chrome-unstable-flags.config
+
 mkdir ${HOME}/.cache/google-chrome-unstable
 mkdir ${HOME}/.config/google-chrome-unstable
 whitelist ${HOME}/.cache/google-chrome-unstable
 whitelist ${HOME}/.config/google-chrome-unstable
+
+whitelist ${HOME}/.config/chrome-unstable-flags.conf
+whitelist ${HOME}/.config/chrome-unstable-flags.config
 
 # Redirect
 include chromium-common.profile
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -161,6 +161,12 @@ blacklist ${HOME}/.config/cantata
 blacklist ${HOME}/.config/catfish
 blacklist ${HOME}/.config/celluloid
 blacklist ${HOME}/.config/cherrytree
+blacklist ${HOME}/.config/chrome-flags.conf
+blacklist ${HOME}/.config/chrome-flags.config
+blacklist ${HOME}/.config/chrome-beta-flags.conf
+blacklist ${HOME}/.config/chrome-beta-flags.config
+blacklist ${HOME}/.config/chrome-unstable-flags.conf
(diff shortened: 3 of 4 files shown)


glitsj16:
Nice catch! It would be nice if you could keep the PR conform our current google-chrome profiles. I'll try to explain what I mean below.

In disable-programs.inc add:
blacklist ${HOME}/.config/chrome-flags.conf
blacklist ${HOME}/.config/chrome-flags.config

Use the noblacklist/mkfile option in google-chrome.profile above the whitelisting:
noblacklist ${HOME}/.config/chrome-flags.conf
noblacklist ${HOME}/.config/chrome-flags.config

mkfile ${HOME}/.config/chrome-flags.conf
mkfile ${HOME}/.config/chrome-flags.config

Also, this needs to be done for google-chrome-beta.profile and google-chrome-unstable.profile too. This all means a bit of extra work, sorry about that :-)

rusty-snake:
> mkfile ${HOME}/.config/chrome-flags.conf
mkfile ${HOME}/.config/chrome-flags.config

I'm not sure, but if they are created by the user and not by chrome, we don't need mkfile.

glitsj16:
> I'm not sure, but if they are created by the user and not by chrome, we don't need mkfile.

Not sure if google-chrome creates these. Never used google-chrome. But @mace015 should be able to figure that out I guess, and follow your advise accordingly.

mace015:
Neither google-chrome nor chromium create these files, they are not needed for both to function either, should I include that anyway? 

The chromium profile also doesn't implement `mkfile` for the config files.

glitsj16:
> Neither google-chrome nor chromium create these files, they are not needed for both to function either, should I include that anyway?

@mace015 Thanks for confirming that. We do want to support these custom flags if they exist. So the only thing you can drop are the `mkfile` lines. Apologies for any confusion caused by my oversight in this context. As soon as you can do the work, we can merge.

mace015:
I've updated the PR, thanks for helping me out with this! If there is anything I should change please let me know!

etc/profile-a-l/google-chrome-beta.profile
```diff
@@ -8,10 +8,16 @@ include globals.local
 noblacklist ${HOME}/.cache/google-chrome-beta
 noblacklist ${HOME}/.config/google-chrome-beta
 
+noblacklist ${HOME}/.config/chrome-beta-flags.conf
```
glitsj16, line 11: We should keep the file names of the flags configuration cfr. what you used in google-chrome.profile. I'm pretty sure the beta version will otherwise not be able to find those (even when they exist). So just drop the '-beta' part. Same goes for the whitelists and the unstable profile. I'll mark those as well, just so you know where to edit.

etc/profile-a-l/google-chrome-beta.profile
```diff
@@ -8,10 +8,16 @@ include globals.local
...
+noblacklist ${HOME}/.config/chrome-beta-flags.config
+
 mkdir ${HOME}/.cache/google-chrome-beta
 mkdir ${HOME}/.config/google-chrome-beta
 whitelist ${HOME}/.cache/google-chrome-beta
 whitelist ${HOME}/.config/google-chrome-beta
 
+whitelist ${HOME}/.config/chrome-beta-flags.conf
```
glitsj16, line 19: Same remarks as above.

etc/profile-a-l/google-chrome-unstable.profile
```diff
@@ -8,10 +8,16 @@ include globals.local
 noblacklist ${HOME}/.cache/google-chrome-unstable
 noblacklist ${HOME}/.config/google-chrome-unstable
 
+noblacklist ${HOME}/.config/chrome-unstable-flags.conf
```
glitsj16, line 11: Same remarks as above.

etc/profile-a-l/google-chrome-unstable.profile
```diff
@@ -8,10 +8,16 @@ include globals.local
...
+noblacklist ${HOME}/.config/chrome-unstable-flags.config
+
 mkdir ${HOME}/.cache/google-chrome-unstable
 mkdir ${HOME}/.config/google-chrome-unstable
 whitelist ${HOME}/.cache/google-chrome-unstable
 whitelist ${HOME}/.config/google-chrome-unstable
 
+whitelist ${HOME}/.config/chrome-unstable-flags.conf
```
glitsj16, line 19: Same remarks as above.

etc/profile-a-l/google-chrome-beta.profile
mace015: I've tested the profiles as they are now and google-chrome-beta does seem to look for `chrome-beta-flags.conf / chrome-beta-flags.config`.

etc/profile-a-l/google-chrome-beta.profile
mace015: I've tested the profiles as they are now and google-chrome-beta does seem to look for `chrome-beta-flags.conf / chrome-beta-flags.config`.

etc/profile-a-l/google-chrome-unstable.profile
mace015: I've tested the profiles as they are now and google-chrome-unstable does seem to look for `chrome-unstable -flags.conf / chrome-unstable -flags.config`.

etc/profile-a-l/google-chrome-unstable.profile
mace015: I've tested the profiles as they are now and google-chrome-unstable does seem to look for `chrome-unstable -flags.conf / chrome-unstable -flags.config`.

PR #3430, opened 2020-05-22
Add Ubuntu's renamed version of dino

karoshi42:
Ubuntu ships the dino instant messenger as ``dino-im``, so I added a profile for it, which adds the name to the private-bin parameter and then redirects to the dino profile. Also it's added to firecfg.

### DIFF
--- a/etc/profile-a-l/dino-im.profile
+++ b/etc/profile-a-l/dino-im.profile
@@ -1,0 +1,14 @@
+# Firejail profile for dino-im
+# Description: Modern XMPP Chat Client using GTK+/Vala, Ubuntu specific bin name
+# This file is overwritten after every install/update
+##quiet
+# Persistent local customizations
+include dino-im.local
+# Persistent global definitions
+include globals.local
+
+# Add Ubuntu specific binary name
+private-bin dino-im
+
+# Redirect
+include dino.profile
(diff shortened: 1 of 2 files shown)


rusty-snake:
Merged, Thanks.

etc/profile-a-l/dino-im.profile
```diff
@@ -0,0 +1,14 @@
+# Firejail profile for dino-im
+# Description: Modern XMPP Chat Client using GTK+/Vala, Ubuntu specific bin name
+# This file is overwritten after every install/update
+##quiet
```
glitsj16, line 4: I think we can drop this line.

etc/profile-a-l/dino-im.profile
```diff
@@ -0,0 +1,14 @@
...
+# Persistent local customizations
+include dino-im.local
+# Persistent global definitions
+include globals.local
```
glitsj16, line 8: We should avoid including globals.local twice, so it needs:
# added by included profile
#include globals.local

etc/profile-a-l/dino-im.profile
karoshi42: You mean the ``##quiet``? I was confused about it, because it is in the template, but not in other profiles I looked at.

karoshi42:
I changed the two lines, should I create a new pull request for it?

rusty-snake:
yes

PR #3433, opened 2020-05-23
Update dino-im.profile

karoshi42:
Comment out ``include globals.local`` so it's not included twice. Also remove the ``##quiet`` line.

### DIFF
--- a/etc/profile-a-l/dino-im.profile
+++ b/etc/profile-a-l/dino-im.profile
@@ -4,6 +4,7 @@
 # Persistent local customizations
 include dino-im.local
 # Persistent global definitions
+# added by included profile
 #include globals.local
 
 # Add Ubuntu specific binary name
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/dino-im.profile
```diff
@@ -1,11 +1,10 @@
 # Firejail profile for dino-im
 # Description: Modern XMPP Chat Client using GTK+/Vala, Ubuntu specific bin name
 # This file is overwritten after every install/update
-##quiet
 # Persistent local customizations
 include dino-im.local
 # Persistent global definitions
-include globals.local
+#include globals.local
```
glitsj16, line 7: Add the below line just above '#include globals.local' please,  to make it consistent with our other profiles
`# added by included profile`

karoshi42: Oh sorry, missed that, added it now.

PR #3437, opened 2020-05-27
new profile: mocp

glitsj16:


etc/profile-m-z/mocp.profile
```diff
@@ -0,0 +1,52 @@
...
+private-cache
+private-dev
+private-etc alternatives,asound.conf,ca-certificates,crypto-policies,group,machine-id,pki,pulse,resolv.conf,ssl
+private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+read-only ${HOME}
```
rusty-snake, line 52:
```suggestion
read-only ${HOME}
read-write ${HOME}/.moc
```

???

glitsj16: Nice catch, I always seem to forget that. Thanks for the review!

M1ndo:
@glitsj16 there's an issue somewhere in the mocp profile config , when detaching from moc server (Using q) the window get stuck but the server still runs (like in the screenshot below) however if you sent a kill signal with (ctrl+c) the server shutdown and the music stops .
2021-02-25-122230_1078x367_scrot

glitsj16:
@M1ndo Thanks for bringing this to our attention. For the moment though I'm having trouble reproducing it. But that's entirely due to my **very** hacked mocp setup. For starters I use a custom built moc PKGBUILD on Arch Linux instead of the official one from the AL repo's. I use Shift+Q to quit. To top it all off I decided to sandbox mocp with a whitelist profile. Together these changes work very nicely for me. I'll temporary revert of all my fiddlings and re-test the original profile without local overrides as soon as possible.

On a sidenote, there's [this FAQ item](https://moc.daper.net/node/345) upstream that might be helpful. Do you happen to have 'ShowTime = yes' and/or increased 'TagsCacheSize' in your ${HOME}/.config/moc/config?

glitsj16:
@M1ndo Okay, here's some more info I've been able to gather. What you're seeing is indeed the behaviour I can reproduce with the Arch Linux repo package. Don't know what mocp is actually doing under the hood, but even with --noprofile detaching with 'q' does not stop the server. That's probably why I bound `/usr/bin/mocp --exit` to Shift+Q in the first place. Been too long ago to remember exactly. You could try this and if it proves to be a workable solution I'll add a note to the mocp.profile to that effect.

For this hack to work you need to use the `full` path to the `unsandboxed` mocp executable when binding the key you'd like to use. I'll need more time to investigate this and try to find a proper fix.

HTH

M1ndo:
> TagsCacheSize
Thanks replying back, In my config i set ShowTime To "no" And TagsCacheSize is unset however that's not the issue here .When detaching from mocp i don't wanna kill the moc server with (shift-q), i keep it running in the background while streaming music using (q) and that's where the window doesn't close (like in the screenshot i attached previously).
Using the full path to the ```unsandboxed``` mocp this doesn't happen at all .

glitsj16:
I'm getting a 404 on that screenshot. But I think I understand now, 'mocp --exit' won't do in that cae. Alas, I can't provide a quick fix. What terminal app are you running mocp in? Is that sandboxed too?

M1ndo:
Here's the image [LINK](https://imgur.com/a/cw8K3df) i'm using **xterm** ```unsandboxed```.

glitsj16:
I'm not able to fix this. There's a [tip](https://wiki.archlinux.org/index.php/MOC#Usage) on the Arch Linux wiki to use `Ctrl+z` instead of 'q' to detach but for me that isn't working either. My advise at this stage is to open an issue on it and I'll mark it as a bug. At least that way should give it better visibility as other collaborators probably don't follow this conversation on a PR/merged page.

M1ndo:
@glitsj16 Sorry For The Delay , here's the issue i've opened [Here](https://github.com/netblue30/firejail/issues/4033) discussing the problem with MOC and firejail .

PR #3438, opened 2020-05-27
harden mpg123.profile

glitsj16:
AFAIK mpg123 doesn't do video, so we can harden the profile accordingly.

### DIFF
--- a/etc/profile-m-z/mpg123.profile
+++ b/etc/profile-m-z/mpg123.profile
@@ -1,13 +1,13 @@
 # Firejail profile for mpg123
 # Description: MPEG audio player/decoder
 # This file is overwritten after every install/update
+quiet
 # Persistent local customizations
 include mpg123.local
 # Persistent global definitions
 include globals.local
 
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/mpg123.profile
```diff
@@ -23,19 +23,24 @@ include whitelist-var-common.inc
 apparmor
 caps.drop all
 netfilter
+no3d
+nodvd
```
rusty-snake, line 27: nodvd also includes (audio) cd.

glitsj16: @rusty-snake Indeed. I assumed that audio CD's are not in a compatible format for mpg123 to process directly. So a decoded WAV or MP3 file would reside on the regular filesystem instead. But I forgot archived audio files. Can't test due to having exchanged my laptop's CD/DVD drive for an extra SSD years ago. I'll remove the nodvd, it's a minor change anyway.

Issue #3389, opened 2020-04-29
Zeal profile not working


OndrejMalek:
**Describe the bug**
```firejail zeal``` does not display window or anything

**Behavior change on disabling firejail**
Zeal show up. But doesnot remember docsets
```
$ firejail --noprofile zeal
Parent pid 17556, child pid 17557
Child process initialized in 9.53 ms
qt5ct: using qt5ct plugin
Qt: Session management error: None of the authentication protocols specified are supported
zeal.core.applicationsingleton: Singleton ID: GNxAletmknejwIs_KXZiYaAzw9OzTsTq2ggd_l64LPQ
zeal.core.applicationsingleton: Starting as a primary instance. (PID: 6)
qt5ct: D-Bus global menu: no
libpng warning: iCCP: known incorrect sRGB profile
libpng warning: iCCP: known incorrect sRGB profile
qt5ct: D-Bus system tray: no
inotify_add_watch("/home/malekon/.config/qt5ct") failed: "No such file or directory"

Parent is shutting down, bye...
```

**To Reproduce**
``` firejail zeal```

**Desktop (please complete the following information):**
```
$ lsb_release -a && echo "" && firejail --version
No LSB modules are available.
Distributor ID:	LinuxMint
Description:	Linux Mint 19 Tara
Release:	19
Codename:	tara

firejail version 0.9.62

Compile time support:
	- AppArmor support is enabled
	- AppImage support is enabled
	- chroot support is enabled
	- file and directory whitelisting support is enabled
	- file transfer support is enabled
	- firetunnel support is enabled
	- networking support is enabled
	- overlayfs support is enabled
	- private-home support is enabled
	- seccomp-bpf support is enabled
	- user namespace support is enabled
	- X11 sandboxing support is enabled
```

**Additional context**
```
$ apt show zeal
Package: zeal
Version: 1:0.6.1-3ppa1~bionic1
```

**Checklist**
 - [x] The upstream profile (and redirect profile if exists) have no changes fixing it.

<details><summary> debug output </summary>
```
23:22:26 malekon@malekon:~
$ firejail --debug zeal > zeal.debug
Reading profile /etc/firejail/zeal.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-exec.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
[...]

PR #3440, opened 2020-06-01
Man pages: were missing info about .profile .local resolution

OndrejMalek:
Kind of hot fix to explained how .profile/.local resolution works. This was something I was confused about as someone who goes for man pages at first shot.

Partial info is copy pasted and linked from https://github.com/netblue30/firejail/wiki/Creating-Profiles#locations-and-types

This addresses larger issue that documentation is currently scattered cross man, github and web.

I have added links to SEE ALSO paragraph to somehow make user see that there is some more info.

### DIFF
--- a/src/man/firejail.txt
+++ b/src/man/firejail.txt
@@ -66,8 +66,8 @@ command line options. The default Firejail filesystem is based on the host files
 system directories mounted read-only. These directories are /etc, /var, /usr, /bin, /sbin, /lib, /lib32,
 /libx32 and /lib64. Only /home and /tmp are writable.
 .PP
-Upon execution Firejail first looks in ~/.config/firejail/ for a profile and if it doesn't find one, it looks in /etc/firejail/ .
-For profile resolution detail see https://github.com/netblue30/firejail/wiki/Creating-Profiles#locations-and-types .
+Upon execution Firejail first looks in ~/.config/firejail/ for a profile and if it doesn't find one, it looks in /etc/firejail/.
+For profile resolution detail see https://github.com/netblue30/firejail/wiki/Creating-Profiles#locations-and-types.
 If an appropriate profile is not found, Firejail will use a default profile.
 The default profile is quite restrictive. In case the application doesn't work, use --noprofile option
 to disable it. For more information, please see \fBSECURITY PROFILES\fR section below.
--- a/src/man/firejail.txt
+++ b/src/man/firejail.txt
@@ -66,7 +66,8 @@ command line options. The default Firejail filesystem is based on the host files
 system directories mounted read-only. These directories are /etc, /var, /usr, /bin, /sbin, /lib, /lib32,
 /libx32 and /lib64. Only /home and /tmp are writable.
 .PP
-Upon execution Firejail first looks in ~/.config/firejail/ for a profile and if it doesn't find one, it looks in /etc/firejail. For profile resolution detail see https://github.com/netblue30/firejail/wiki/Creating-Profiles#locations-and-types .
+Upon execution Firejail first looks in ~/.config/firejail/ for a profile and if it doesn't find one, it looks in /etc/firejail/ .
+For profile resolution detail see https://github.com/netblue30/firejail/wiki/Creating-Profiles#locations-and-types .
 If an appropriate profile is not found, Firejail will use a default profile.
 The default profile is quite restrictive. In case the application doesn't work, use --noprofile option
 to disable it. For more information, please see \fBSECURITY PROFILES\fR section below.
--- a/src/man/firejail.txt
+++ b/src/man/firejail.txt
@@ -66,7 +66,7 @@ command line options. The default Firejail filesystem is based on the host files
 system directories mounted read-only. These directories are /etc, /var, /usr, /bin, /sbin, /lib, /lib32,
 /libx32 and /lib64. Only /home and /tmp are writable.
 .PP
-As it starts up, Firejail tries to find a security profile based on the name of the application.
+Upon execution Firejail first looks in ~/.config/firejail/ for a profile and if it doesn't find one, it looks in /etc/firejail. For profile resolution detail see https://github.com/netblue30/firejail/wiki/Creating-Profiles#locations-and-types .
 If an appropriate profile is not found, Firejail will use a default profile.
 The default profile is quite restrictive. In case the application doesn't work, use --noprofile option
 to disable it. For more information, please see \fBSECURITY PROFILES\fR section below.
@@ -3107,5 +3107,9 @@ Homepage: https://firejail.wordpress.com
 \&\flfiremon\fR\|(1),
 \&\flfirecfg\fR\|(1),
 \&\flfirejail-profile\fR\|(5),
-\&\flfirejail-login\fR\|(5)
-\&\flfirejail-users\fR\|(5)
+\&\flfirejail-login\fR\|(5),
+\&\flfirejail-users\fR\|(5),
+.UR https://github.com/netblue30/firejail/wiki 
+.UE ,
+.UR https://github.com/netblue30/firejail 
+.UE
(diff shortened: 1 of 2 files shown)


src/man/firejail.txt
```diff
@@ -66,7 +66,7 @@ command line options. The default Firejail filesystem is based on the host files
 system directories mounted read-only. These directories are /etc, /var, /usr, /bin, /sbin, /lib, /lib32,
 /libx32 and /lib64. Only /home and /tmp are writable.
 .PP
-As it starts up, Firejail tries to find a security profile based on the name of the application.
+Upon execution Firejail first looks in ~/.config/firejail/ for a profile and if it doesn't find one, it looks in /etc/firejail. For profile resolution detail see https://github.com/netblue30/firejail/wiki/Creating-Profiles#locations-and-types .
```
rusty-snake, line 69: Maybe spilt over multiple lines, very long.

---

If firejail is self-compiled w/o `--prefix=/usr`, profiles are located under `/usr/local/etc` we've often issues where is is one point of confusion. Maybe a it is worth a note.

OndrejMalek: OK, but your additional info about self-compiled w/o `--prefix=/usr` , I would put into wiki not ot clutter man, when there is still wiki.

OndrejMalek: Well I can see that I was confused either. :)

OndrejMalek: @rusty-snake should not be self-compiled w/o `--prefix=/usr` rather somehow mitigated to not happen? Is there any reason to exist? I would be more elegant solution otherwise I should be mentioned everywhere where /etc/firejail/ is metioned. IMO

OndrejMalek: wiki edited 
https://github.com/netblue30/firejail/wiki/Creating-Profiles/8b5f3f2795f67b15f69dae94cb2eeba8afe6457c
https://github.com/netblue30/firejail/wiki/Creating-Profiles/878f36a671b223e126c05ec623d712983138aa17

rusty-snake: > should not be self-compiled w/o --prefix=/usr rather somehow mitigated to not happen? Is there any reason to exist? It would be more elegant solution otherwise it should be mentioned everywhere where /etc/firejail/ is metioned. IMO

sorry, but I don't get what you are asking.

`--prefix=/usr/local` is default for almost every buildsystem.
There is also the `${CFG}` macro in profiles, however it is so rarely used that it would be more confusing.

OndrejMalek: I am not C programmer thus I do not know about `--prefix`. There must be other contributors like me, who can contribute to .profiles/bugs/documentation. I have installed firajail from sources according to https://github.com/netblue30/firejail#compile-and-install . And now I have /etc/firejail (meybe from previous .deb installation) and /usr/local/etc/firejail. Should not be better to either mention it in https://github.com/netblue30/firejail#compile-and-install or I guess use something like`$ ./configure --prefix=/ && make && sudo make install-strip` to remove confusion? Thx for your time

rusty-snake: > I am not C programmer 

too

> meybe from previous .deb installation

yes, it is only remove with (`apt purge`) AFAIK

> I guess use something like$ ./configure --prefix=/ && make && sudo make install-strip to remove confusion?

If you install it from source (w/o .deb or .rpm) `./configure --prefix=/usr && make && sudo make install-strip`. `--prefix=/usr` implies `--sysconfdir=/etc` (for firejail).

See [this thread](https://github.com/netblue30/firejail/issues/3333#issuecomment-611623863) for how to install firejail from source.

rusty-snake: IIRC is `--enable-apparmor` broken w/o `--prefix=/usr`.

glitsj16: > And now I have /etc/firejail (meybe from previous .deb installation) and /usr/local/etc/firejail.

Perhaps it's not a bad idea to have proper `uninstall` instructions too. When installed using native OS tools it's simple: use the same tools to uninstall firejail. It potentially gets more complicated with a manual install. Especially if the original install directory no longer exists on the filsystem. In the latter case the conventional `sudo make uninstall` will fail. For that to work succesfully a user will have to redo the installation steps - hopefully remembering the exact ./configure options - prior to running the uninstall command.

This translates to:

(1) redo the manual installation exactly as before
- $ git clone https://github.com/netblue30/firejail.git
- $ cd firejail
- $ ./configure && make && sudo make install-strip

(2) uninstall
- $ sudo make uninstall

@OndrejMalek Tell us if you need assistance with 'unconfusing' your firejail installation.

rusty-snake: > hopefully remembering the exact ./configure options

hopefully it is in the shellhistory

> sudo make uninstall

In general using `.pkg.tar.xz`, `.deb`, `.rpm` is the prefered way to install software, maybe we should make this way some more prominent.

What do you guys thing about a "using firejail-git" wiki page?
I could maintain a Fedora section, @glitsj16 Arch and Ubuntu?

glitsj16: > In general using .pkg.tar.xz, .deb, .rpm is the prefered way to install software, maybe we should make this way some more prominent.

I agree and have made a WIP suggestion on how to achieve this in #3442.

> What do you guys thing about a "using firejail-git" wiki page?
> I could maintain a Fedora section, @glitsj16 Arch and Ubuntu?

Yes that's fine with me. I can always ask @reinerh for assistance on Debian/Ubuntu, although such a firejail-git wiki page shouldn't be that complicated.

OndrejMalek: @glitsj16 we are getting to different issue than this PR is about. I am not competent to solve --prefix thing. Please do it your PR.

But about situation a got into: 
1. On Linux Mint I have installed Firejail from repos  version - 0.9.52-2
2. Some profiles did not work.
3. I have probably unistalled firejail not properly
4. I have installed stable version 0.62.1 apparmor.deb from source forge.
5. Again some unexpected behavior ( https://github.com/netblue30/firejail/issues/3389 )
6. Again I have probably unistalled firejail not properly
7. Installed from sources -  things work right but can not make apparmor work - at this point I did not now about /usr/local/etc/firejail because a had still files in /etc/firejail - so when I was tweaking profiles a was oparating on wrong files at /etc/firejail, which is according to man pages right.
I can see I am quite BFU at this point :D. But this can happen.

Other thing is when I wanna uninstall sources I have to do:
```
sudo firecfg --clean # without this I have to manualy rm links, when I wanna do 'sudo firecfg' next time (probably due to changing --prefix=/usr)
sudo make uninstall
```

src/man/firejail.txt
```diff
@@ -66,7 +66,8 @@ command line options. The default Firejail filesystem is based on the host files
 system directories mounted read-only. These directories are /etc, /var, /usr, /bin, /sbin, /lib, /lib32,
 /libx32 and /lib64. Only /home and /tmp are writable.
 .PP
-As it starts up, Firejail tries to find a security profile based on the name of the application.
+Upon execution Firejail first looks in ~/.config/firejail/ for a profile and if it doesn't find one, it looks in /etc/firejail/ .
```
reinerh, line 69: in my opinion the spaces before the dots should be removed.

src/man/firejail.txt
rusty-snake: > things are working right but I can not make apparmor work 

Without `--prefix=/usr`, `firejail-default` (the AA-profile) is installed under /usr/local/etc/apparmor.d. I can't help you with AA, I never actively unsed AA.

src/man/firejail.txt
OndrejMalek: Thx, done.

src/man/firejail.txt
OndrejMalek: @rusty-snake thx dont bother with it.  I'll try some other time.

rusty-snake:
Everyone fine with it?

PR #3442, opened 2020-06-01
prioritize installing via OS

glitsj16:
First attempt to tackle some of the issues that came up in #3440.


rusty-snake:
@glitsj16 IMHO we can merge this and adding a link to the new wiki page.

Fred-Barclay:
My only concern is IIRC the firejail that ships with Ubuntu 18.04 is 0.9.52-LTS and tends to cause issues/confusion since it's missing a large portion of the features from the main branch. Perhaps we should favor using the [PPA](https://launchpad.net/~deki/+archive/ubuntu/firejail) from @reinerh for this distro?

rusty-snake:
… and for debian the backports package.

PR #3450, opened 2020-06-04
enable apparmor support by default in update_deb.sh

glitsj16:
IMO enabling apparmor support by default here ensures a git installation that is compatible with the OS repository and PPA packages (see the firejail-from-git wiki page).

One thing I'm not sure of is how we can patch firejail.config to accomodate https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=916920.

### DIFF
--- a/contrib/update_deb.sh
+++ b/contrib/update_deb.sh
@@ -6,11 +6,19 @@
 # Purpose: Fetch, compile, and install firejail from GitHub source. For
 # Debian-based distros only (Ubuntu, Mint, etc).
 set -e
+
 git clone --depth=1 https://github.com/netblue30/firejail.git
 cd firejail
 ./configure --enable-apparmor --prefix=/usr
+
+# Fix https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=916920
+sed -i \
+    -e "s/# cgroup .*/cgroup no/" \
+    -e "s/# restricted-network .*/restricted-network yes/" \
+    etc/firejail.config
+
 make deb
 sudo dpkg -i firejail*.deb
-echo "Firejail was updated!"
+echo "Firejail updated."
 cd ..
 rm -rf firejail
--- a/contrib/update_deb.sh
+++ b/contrib/update_deb.sh
@@ -8,7 +8,7 @@
 set -e
 git clone --depth=1 https://github.com/netblue30/firejail.git
 cd firejail
-./configure --prefix=/usr
+./configure --enable-apparmor --prefix=/usr
 make deb
 sudo dpkg -i firejail*.deb
 echo "Firejail was updated!"


rusty-snake:
https://github.com/netblue30/firejail/blob/master/mkdeb.sh.in#L45

```
sed -i \
    -e "s/# restricted-network .*/restricted-network yes/" \
    -e "s/# force-nonewprivs .*/force-nonewprivs yes/" \
    debian/etc/firejail/firejail.config
```

glitsj16:
@rusty-snake Yeah, using sed isn't the problem. I'm trying to create a script that creates deb files from the git code. https://github.com/netblue30/firejail/blob/master/mkdeb.sh is relevant for doing so with a tarball, so I'll need to dust of my Debian experience to try to integrate things nicely. I'm labelling this as WIP, my Ubuntu LTS is currently in a broken state due to https://github.com/netblue30/firejail/commit/52e24dbb3e89beca6d2f748eb3e5e49202270f03 and https://github.com/netblue30/firejail/commit/dae3933bc30a08e355083b67ee99e688bfa13b4e and I'll need some time to revive that. I use `checkinstall` in my regular build-firejail-from-git script on Ubuntu, and that didn't like those commits (make[1]: *** [Makefile:118: realinstall] Error 1). It wiped out all symlinks under /etc in the real filesystem, which caught me by surprise. But that's another matter entirely.

reinerh:
I'm not sure if it helps, but the .gitlab-ci.yml file (the debian_ci section) contains some rough instructions how to build a Debian package from a firejail upstream commit, with the build metadata/instructions from the current Debian package (except the patches; they are removed before the build, as they sometimes no longer apply cleanly, which would cause the CI to fail).

glitsj16:
@reinerh Thanks for the info, I'll look into that. While we're on the subject, I wonder how much extra work it would be to set up/maintain a PPA with daily builds or something similar? Do you think that's worth the effort? Not that I'm trying to shuffle my way out of anything, but I always considered you the best-placed collaborator for all things Debian/Ubuntu, so just asking for your input.

reinerh:
It's not really much effort to set up a PPA.
But with the current mkdeb.sh it's not so easy, as you upload Debian source packages into a PPA (.dsc files, which refer to a upstream source tarball and a Debian metadata tarball), not already built .deb packages (the Launchpad servers are actually building the packages).
One challenge I see for automated uploads is that the files need to be signed (with gpg), and I don't know what's the best way to integrate that into a build pipeline (while also keeping the keys secret).
An alternative to PPAs would be a Debian repository hosted somewhere else, though this has the same signing problem.

glitsj16:
> And I'm also not sure if that many users would actually be interested in daily builds (and are not able to build it for themselves).

Thanks for the feedback. I tend to agree. It's probably sufficient to provide clear instructions on how to build from git and add warnings where due, like how to avoid overwriting firejail.config and a link to the relevant bug(s).

PR #3469, opened 2020-06-17
corrections to the vlc profile

chrpinedo:
added the lost mkdir and whitelist rules.

### DIFF
--- a/etc/profile-m-z/vlc.profile
+++ b/etc/profile-m-z/vlc.profile
@@ -20,6 +20,14 @@ include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-xdg.inc
 
+mkdir ${HOME}/.cache/vlc
+mkdir ${HOME}/.config/vlc
+mkdir ${HOME}/.local/share/vlc
+whitelist ${HOME}/.cache/vlc
+whitelist ${HOME}/.config/vlc
+whitelist ${HOME}/.local/share/vlc
+whitelist ${MUSIC}
+whitelist ${VIDEOS}
 include whitelist-var-common.inc
 
 #apparmor - on Ubuntu 18.04 it refuses to start without dbus access


glitsj16:
> added the lost mkdir and whitelist rules.

How did these rules get **lost**? As far as I can see these were never part of the VLC profile. This could be just a (confusing) wording thing, in which case you can dismiss this remark.

etc/profile-m-z/vlc.profile
```diff
@@ -20,6 +20,14 @@ include disable-passwdmgr.inc
...
 include disable-xdg.inc
 
+mkdir ${HOME}/.cache/vlc
+mkdir ${HOME}/.config/vlc
+mkdir ${HOME}/.local/share/vlc
+whitelist ${HOME}/.cache/vlc
+whitelist ${HOME}/.config/vlc
+whitelist ${HOME}/.local/share/vlc
+whitelist ${MUSIC}
+whitelist ${VIDEOS}
```
glitsj16, line 30: Refactoring VLC as a whitelisting profile can definately harden it. Most if not all such profiles `include whitelist-common.inc` too (to cover potential UI breakage). Can you please test VLC's functionality with that included before we surprise users?

Fred-Barclay:
I agree, I'm a tad concerned about the impact since we don't use whitelisting for our other players (i.e. totem or xplayer). *However* whitelisting might be more important for VLC - i.e. since it can connect to the internet for streaming and etc (not sure if totem can).

chrpinedo:
Sorry, I had problems with VLC launched by Firefox, but it was due to VLC being launched inside the firefox profile. However, I though the problem was related to the lack of mkdir-whitelist sentences in the VLC profile. I was wrong. The solution to this problem is to add the required paths to file /etc/firejail/firefox-common-addons.local.

After reading https://github.com/netblue30/firejail/issues/1569 I think it senseless to include mkdir-whitelist rules in the VLC (or totem) profile because users might have multimedia files to reproduce in other directories.

I think you can discard this merge request. Sorry for the mess :)

glitsj16:
@Fred-Barclay Totem comes with apple-trailers and vimeo plugins on Arch Linux so I assume it does support streaming too. For what it's worth, I have been using mpv and mplayer with a hardened _whitelisting_ profile in my .local files for a very long time without any issues whatsoever. So we might consider refactoring all of these media player profiles as such. The only downside I see with doing so is that users who want to play media files located outside of the supported XDG dirs would need to whitelist those paths.

chrpinedo:
@glitsj16 very kind :) I also use Arch Linux with GNOME and I can confirm Totem streams. I use both, VLC and Totem, for HTTP/HTTPS streaming.

Fred-Barclay:
@glitsj16 We probably should consider whitelisting... I would suggest ${DOCUMENTS}, ${DOWNLOADS}, ${MUSIC}, and ${VIDEOS}. I wonder if @chrpinedo would like to take the lead on this/open an PR?

@chrpinedo thoughts? No pressure either way of course 😄 . If you would like to do this don't worry about it being perfect, we can easily discuss/fix in the PR.

Let's close this one for now - thanks @chrpinedo !!

glitsj16:
> We probably should consider whitelisting... I would suggest ${DOCUMENTS}, ${DOWNLOADS}, ${MUSIC}, and ${VIDEOS}.

@Fred-Barclay That should be easy, we just need to remove `include disable-xdg.inc` in that case.

chrpinedo:
@Fred-Barclay thanks for the trust :-D However, I don't completely understand firejail profiles and I only dare make small changes to profiles ! I think such a change needs a more deeper view.

Fred-Barclay:
No worries! I'll open a PR soon 😄

Issue #3424, opened 2020-05-20
Support bash or AppArmor like variables


curiosityseeker:
@rusty-snake 's [proposal](#3412 ) to support aliases gave me the idea to propose the support of variables in Firejail. 

I once saw this neat trick in an AppArmor profile for LibreOffice:

```
@{libo_user_dirs} = @{HOME} /mnt /media
@{libreoffice_ext} = [mM][mM][lL] [tT][iI][fF] [xX][mMsS][lL] {,f,F}[oO][dDtT][tTsSpPbBgGfF] [jJ][pP][eE][gG] [tT][iI][fF][fF] [sS][vV][gG] [pP][dD][fF] [sS][wW][fF] {,x,X}[hH][tT][mM]{,l,L} [pP][pP][tTsS]{,x,X} [rR][tT][fF] [tT][xX][tT] [sS][vV][gG][zZ] [dD][iIbB][fF] [jJ][pP][gG] [pP][nN][gG] [pP][sS][dD] [cCtT][sS][vV] [sS][lL][kK] [sS][dD][wW] [uU][oO][fFtTsSpP] [xX][lL][sSwWtT]{,x,X} [dD][oO][cCtT]{,x,X} [pP][oO][tT]{,m,M}

...
owner @{libo_user_dirs}/**.@{libreoffice_ext} rwk, #Open files rw with the right exts
owner @{libo_user_dirs}/**/ rw, #allow creating directories that we own
owner @{libo_user_dirs}/**~lock.* rw, #lock file support
```
Since then I'm using a similar approach in several of my AppArmor profiles.

I think supporting something similar in Firejail - in combination with globbing - would make writing profiles easier and, at the same time, more powerful. Above example illustrates how whitelisting files with specific extensions could be facilitated.

rusty-snake:
I just read @curiosityseeker variable proposal and felt remind to rpm's macros. At the end all three (aliases, variables, macros (not the static once we now have)) are the same, dynamical generated profiles.

rusty-snake:
FWIW: https://github.com/sailfishos/firejail/blob/master/rpm/0004-Implement-template-addition-for-replacing-keys-in-pr.patch

PR #3472, opened 2020-06-19
Use whitelisting for video players

Fred-Barclay:
Switches video players over to a whitelist basis.

I've tested celluloid, vlc, and xplayer and they seem to be fine - would really appreciate additional testing! 😃

### DIFF
--- a/etc/profile-a-l/celluloid.profile
+++ b/etc/profile-a-l/celluloid.profile
@@ -21,14 +21,13 @@ include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 
+read-only ${DESKTOP}
 mkdir ${HOME}/.config/celluloid
 mkdir ${HOME}/.config/gnome-mpv
 mkdir ${HOME}/.config/youtube-dl
 whitelist ${HOME}/.config/celluloid
 whitelist ${HOME}/.config/gnome-mpv
 whitelist ${HOME}/.config/youtube-dl
-whitelist ${DESKTOP}
-whitelist ${DOCUMENTS}
 whitelist ${DOWNLOADS}
 whitelist ${MUSIC}
 whitelist ${PICTURES}
--- a/etc/profile-a-l/celluloid.profile
+++ b/etc/profile-a-l/celluloid.profile
@@ -9,8 +9,6 @@ include globals.local
 noblacklist ${HOME}/.config/celluloid
 noblacklist ${HOME}/.config/gnome-mpv
 noblacklist ${HOME}/.config/youtube-dl
-noblacklist ${MUSIC}
-noblacklist ${VIDEOS}
 
 # Allow python (blacklisted by disable-interpreters.inc)
 include allow-python2.inc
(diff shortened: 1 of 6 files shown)


glitsj16:
Looks fine. As I have run with whitelisting profiles for mplayer and mpv for a long time I can confirm both work as expected.

On the matter of completely removing disable-xdg.inc I also never whitelist ${DESKTOP} and ${DOCUMENTS} in media player profiles.

Fred-Barclay:
Thanks @rusty-snake and @glitsj16 !

To address your thoughts...

1/2 my completely unproven gut feeling is that folks would be more likely to have videos on Desktop than Documents. Perhaps remove Documents but keep Desktop visible in the sandbox??

3. I believe we have to remove `disable-xdg.inc` or (alternatively) noblacklist everything but Documents (and potentially Desktop). At that rate it might be more clear to just blacklist Documents instead of having `include disable-xdg.inc` followed by a bunch of noblacklists.

4. Definitely will fix this one, thanks!

glitsj16:
> 1/2 my completely unproven gut feeling is that folks would be more likely to have videos on Desktop than Documents. Perhaps remove Documents but keep Desktop visible in the sandbox??

I can see your point. Personally I can live with removing ${DOCUMENTS} and keeping ${DESKTOP}.

About point 3. I have come across situations like these before a few times, meaning that whatever one opts for, users are going to have to decide on loosening/tightening individual profiles via *.local files. In such cases I think it makes more sense to make it harder on users to loosen sandboxes, instead of 'punishing' users that favor more tightened profiles. So in this case I would opt for **keeping** include disable-xdg.inc, _even if that means having a few more lines in the profiles_. Hope this makes sense, I might have one of those 'fuzzy logic' brain cells :)

curiosityseeker:
I also vote for removing ${DOCUMENTS} and keeping ${DESKTOP} - but what about making the latter _read-only_?

rusty-snake:
> but what about making the latter read-only?

IMHO we should add
```
read-only ${HOME}
read-write ${HOME}/.config/foo
read-write ${HOME}/.local/share/foo
```
to more profiles in general.

Fred-Barclay:
> I also vote for removing ${DOCUMENTS} and keeping ${DESKTOP} - but what about making the latter _read-only_?

My concern is with editing file metadata here - for example I believe VLC lets you easily change artist/album names? Maybe not too much an issue with videos but it's not unreasonable to expect people to play i.e. mp3s with video players and read-only access would be an issue here.


Also unrelated side note - I messed up by pushing to`master` on my fork for this PR - might have to close and send in a similar one, apologies for any inconvenience. 😄

curiosityseeker:
> My concern is with editing file metadata here - for example I believe VLC lets you easily change artist/album names? Maybe not too much an issue with videos but it's not unreasonable to expect people to play i.e. mp3s with video players and read-only access would be an issue here.

I understand, and I think you're right. Nevertheless I feel a bit uneasy. Write permission for the supported extensions would be probably preferable but it's a bit complicated to achieve. This is an example where something like what I suggested in #3424 would be helpful.

rusty-snake:
anyone fine?

Fred-Barclay:
Oops, sorry! I'll try to get this updated ASAP!

Fred-Barclay:
Updated! @rusty-snake as you mention mpv has a screenshot function that saves to $HOME I wonder if we should go back to blacklisting for this one?

Fred-Barclay:
@rusty-snake thanks, gotten a little rusty here, should have kept the whitelist as well...

rusty-snake:
https://github.com/netblue30/firejail/commit/deb6c12454191b7aeff3d259612a00427d1aa6a1#diff-74e11ae375b6951607bb9f53d646c4e4


`~/.config/mpv/foobar.conf`:
```
screenshot-directory=~/Pictures
```

PR #3487, opened 2020-07-03
clarify writing to /var/mail and /var/spool/mail in apparmor

glitsj16:
Thunderbird seems to be our only mail client profile that enables the `apparmor` option. Users need this when they follow instructions on how to allow reading local mail.

@Vincent43 Could you take a look at the rule I added please? Not sure if it is the correct syntax...

### DIFF
--- a/etc/apparmor/firejail-default
+++ b/etc/apparmor/firejail-default
@@ -51,7 +51,7 @@ owner /{,run/firejail/mnt/oroot/}{run,dev}/shm/** w,
 
 # Allow writing to /var/mail and /var/spool/mail (for mail clients)
 # Uncomment to enable
-#owner /{,var/}{mail,spool/mail}/** w,
+#owner /var/{mail,spool/mail}/** w,
 
 # Allow writing to removable media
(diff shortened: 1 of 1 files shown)


etc/apparmor/firejail-default
```diff
@@ -49,6 +49,10 @@ owner /{,run/firejail/mnt/oroot/}{,var/}run/firejail/mnt/trace w,
 owner /{,run/firejail/mnt/oroot/}{,var/}run/user/[0-9]*/** w,
 owner /{,run/firejail/mnt/oroot/}{run,dev}/shm/** w,
 
+# Allow writing to /var/mail and /var/spool/mail (for mail clients)
+# Uncomment to enable
+#owner /{,var/}{mail,spool/mail}/** w,
```
reinerh, line 54: this also allows /mail and /spool/mail, right? (as the var is optional)
consider changing it to `/var/{mail,spool/mail}/**`

glitsj16: Nice catch, thanks!

Vincent43:
Looks ok.

PR #3489, opened 2020-07-04
Okular profile fixes

Kishore96in:
The default okular profile doesn't whitelist some config files which are used to store keyboard shortcuts and toolbar configuration. This pull request whitelists those files in `okular.profile` and `firefox-common-addons.inc`

### DIFF
--- a/etc/inc/firefox-common-addons.inc
+++ b/etc/inc/firefox-common-addons.inc
@@ -17,7 +17,6 @@ noblacklist ${HOME}/.kde4/share/config/kgetrc
 noblacklist ${HOME}/.kde4/share/config/okularpartrc
 noblacklist ${HOME}/.kde4/share/config/okularrc
 noblacklist ${HOME}/.local/share/kget
-noblacklist ${HOME}/.local/share/kxmlgui5/okular
 noblacklist ${HOME}/.local/share/okular
 noblacklist ${HOME}/.local/share/qpdfview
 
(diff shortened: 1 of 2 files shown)


etc/inc/firefox-common-addons.inc
```diff
@@ -17,6 +17,7 @@ noblacklist ${HOME}/.kde4/share/config/kgetrc
 noblacklist ${HOME}/.kde4/share/config/okularpartrc
 noblacklist ${HOME}/.kde4/share/config/okularrc
 noblacklist ${HOME}/.local/share/kget
+noblacklist ${HOME}/.local/share/kxmlgui5/okular
```
rusty-snake, line 20: There is no `blacklist ${HOME}/.local/share/kxmlgui5/okular`.

etc/profile-m-z/okular.profile
```diff
@@ -15,6 +15,7 @@ noblacklist ${HOME}/.kde/share/config/okularrc
 noblacklist ${HOME}/.kde4/share/apps/okular
 noblacklist ${HOME}/.kde4/share/config/okularpartrc
 noblacklist ${HOME}/.kde4/share/config/okularrc
+noblacklist ${HOME}/.local/share/kxmlgui5/okular
```
rusty-snake, line 18: same here.

etc/inc/firefox-common-addons.inc
Kishore96in: Should I add it to `disable-programs.inc` or remove this line?

Kishore96in: The `.local/share/kxmlgui` folder actually contains configuration files for quite a few KDE programs. Should I remove this line from this pull request, and then make a separate one which blacklists `.local/share/kxmlgui` in `disable-programs.inc` and then adds `noblacklist ${HOME}/.local/share/kxmlgui/application_name` in the profiles for the relevant KDE applications?

Kishore96in: I've removed it for now and will be making a separate pull request soon.

rusty-snake:
Merged, Thanks.

PR #3490, opened 2020-07-04
noblacklist a config file in konversation profile

Kishore96in:
Without this, konversation doesn't remember the settings for notifications.

### DIFF
--- a/etc/profile-a-l/konversation.profile
+++ b/etc/profile-a-l/konversation.profile
@@ -7,6 +7,7 @@ include konversation.local
 include globals.local
 
 noblacklist ${HOME}/.config/konversationrc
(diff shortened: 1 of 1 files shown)


rusty-snake:
Merged, Thanks.

glitsj16:
@rusty-snake We don't blacklist ${HOME}/.config/konversation.notifyrc anywhere, so I'm a bit puzzled why it would need noblacklist. In any case that path should be blacklisted in disable-programs.inc now this is merged right?

rusty-snake:
Your right, something must go wrong while greping for the blacklist. 

> In any case that path should be blacklisted in disable-programs.inc now this is merged right?

Yes, adding more paths to disable-programs.inc is always good.

PR #3493, opened 2020-07-05
Blacklist .local/share/kxmlgui5 and allow access only for applications which use it.

Kishore96in:
Multiple KDE applications store their keyboard shortcut and toolbar config in subdirectories of `.local/share/kxmlgui5`. This pull request blacklists these directories in `disable-programs.inc`, and then noblacklists them in the profiles of the applications which require access to them.

### DIFF
--- a/etc/profile-a-l/akregator.profile
+++ b/etc/profile-a-l/akregator.profile
@@ -8,6 +8,7 @@ include globals.local
 
 noblacklist ${HOME}/.config/akregatorrc
 noblacklist ${HOME}/.local/share/akregator
+noblacklist ${HOME}/.local/share/kxmlgui5/akregator
 
 include disable-common.inc
 include disable-devel.inc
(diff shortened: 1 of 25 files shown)


etc/profile-a-l/akregator.profile
```diff
@@ -19,9 +20,11 @@ include disable-shell.inc
 
 mkfile ${HOME}/.config/akregatorrc
 mkdir ${HOME}/.local/share/akregator
+mkdir ${HOME}/.local/share/kxmlgui5/akregator
```
rusty-snake, line 23: If `${HOME}/.local/share/kxmlgui5/akregator` is a file, this should be `mkfile`.

Kishore96in: It's a directory on my system.

Kishore96in:
>     * does this cover _all_ kde-apps which use `.local/share/kxmlgui5/…` or just the one installed/used by you? (Of course you can ignore apps w/o firejail-profile)

Ah. Thanks for pointing that out. I'll check for the other KDE applications which have profiles.

Kishore96in:
@rusty-snake I don't understand your comment about the okular profile needing `mkdir+whitelist`. Currently, it looks like `okular.profile` doesn't whitelist anything in the home directory (current whitelist commands are for files in /usr/share).

Kishore96in:
I think I've caught all the relevant applications.

rusty-snake:
Merged, Thanks.

PR #3505, opened 2020-07-09
hardening some profiles

rusty-snake:
- harden and fix flameshot
 - wruc: frogatto, ghostwriter
 - harden gnome-latex
 - add whitelist opt-in note to keepassxc
 - add comment to minetest
 - harden openarena, tremulous, xonotic
 - add profile for xonotic-sdl-wrapper, openarena_ded

### DIFF
--- a/etc/profile-m-z/xonotic.profile
+++ b/etc/profile-m-z/xonotic.profile
@@ -14,12 +14,17 @@ include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
+include disable-xdg.inc
 
 mkdir ${HOME}/.xonotic
 whitelist ${HOME}/.xonotic
+whitelist /usr/share/xonotic
 include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
+apparmor
 caps.drop all
 netfilter
 nodvd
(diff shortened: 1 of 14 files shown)


etc/profile-m-z/xonotic.profile
```diff
@@ -32,12 +37,17 @@ novideo
...
+private-cache
+private-bin basename,bash,blind-id,cut,darkplaces-glx,darkplaces-sdl,dirname,glxinfo,grep,head,ldd,netstat,ps,readlink,sed,sh,uname,xonotic,xonotic-glx,xonotic-linux32-dedicated,xonotic-linux32-glx,xonotic-linux32-sdl,xonotic-linux64-dedicated,xonotic-linux64-glx,xonotic-linux64-sdl,xonotic-sdl,xonotic-sdl-wrapper,zenity
 private-dev
 private-etc alternatives,asound.conf,ca-certificates,crypto-policies,drirc,fonts,group,host.conf,hostname,hosts,ld.so.cache,ld.so.preload,localtime,machine-id,nsswitch.conf,passwd,pki,pulse,resolv.conf,ssl
 private-tmp
 
 dbus-user none
 dbus-system none
+
+read-only ${HOME}
```
SkewedZeppelin, line 52: What is the benefit of this since it is already whitelisted?

rusty-snake: Placing files under $HOME which are then executed. Sure, this is mitigated by disable-interpreters and disable-exec but as long as there are no drawback with a ro $HOME, we can add it.

PR #3518, opened 2020-07-17
Hardend Zoom profile

onovy:


etc/profile-m-z/zoom.profile
```diff
@@ -20,14 +22,20 @@ whitelist ${HOME}/.cache/zoom
...
+include whitelist-var-common.inc
 
 caps.drop all
 netfilter
 nodvd
+nogroups
 nonewprivs
 noroot
 notv
+nou2f
```
rusty-snake, line 34: Does zoom support u2f?

etc/profile-m-z/zoom.profile
```diff
@@ -20,14 +22,20 @@ whitelist ${HOME}/.cache/zoom
 whitelist ${HOME}/.config/zoomus.conf
 whitelist ${HOME}/.zoom
 include whitelist-common.inc
+include whitelist-var-common.inc
```
rusty-snake, line 25: Can we add `whitelist-runuser-common` and `whitelist-usr-share-common`?

etc/profile-m-z/zoom.profile
```diff
@@ -10,7 +10,9 @@ noblacklist ${HOME}/.zoom
 
 include disable-common.inc
 include disable-devel.inc
+include disable-exec.inc
 include disable-interpreters.inc
+include disable-passwdmgr.inc
 include disable-programs.inc
```
rusty-snake, line 16: Can we add `disable-shell` and `disable-xdg`?

etc/profile-m-z/zoom.profile
```diff
@@ -20,14 +22,20 @@ whitelist ${HOME}/.cache/zoom
...
 nonewprivs
 noroot
 notv
+nou2f
 protocol unix,inet,inet6,netlink
 seccomp !chroot
+shell none
 
+private-dev
+private-etc machine-id,resolv.conf,fonts
```
rusty-snake, line 40: https://github.com/netblue30/firejail/pull/3519#discussion_r456321068

etc/profile-m-z/zoom.profile
onovy: i don't think so. Only when you are using SSO, but it will open browser for auth then.

etc/profile-m-z/zoom.profile
onovy: we can, but why? It works fine without that :)

etc/profile-m-z/zoom.profile
onovy: I have firejail version 0.9.62 so I can't test disable-shell. I can add disable-xdg, looks harmless

etc/profile-m-z/zoom.profile
rusty-snake: Why using firejail? It works without it ;)

FYI: `include whitelist-runuser-common.inc` means only a few listed sockets in /run/user/UID are allowed. W/O it everything is allowed. Same for wusc.

etc/profile-m-z/zoom.profile
rusty-snake: You can copy it. disabel-shell works in 0.9.62 too.

(If `firejail --blacklist=/bin/bash zoom` fails you can skip it).

etc/profile-m-z/zoom.profile
rusty-snake: You can copy it, in 0.9.62 it may break XWayland on GNOME. But if you are still on X11 or KDE you can test it.

etc/profile-m-z/zoom.profile
onovy: tested, zoom fails with disable-shell

rusty-snake:
Based on its flatpak, zoom has this dbus-policy. IDK if someone with 0.9.63 can test it.
```
dbus-user filter
dbus-user.own us.zoom.Zoom
dbus-user.own org.kde.*
# dbus-user.talk ca.desrt.dconf
dbus-system none
```
_Why do they need to own `org.kde.*`?!_


We can probably add `disable-mnt`, `private-bin …`, `private-cache`, `tracelog`.

onovy:
> Based on its flatpak, zoom has this dbus-policy. IDK if someone with 0.9.63 can test it.
> 
> ```
> dbus-user filter
> dbus-user.own us.zoom.Zoom
> dbus-user.own org.kde.*
> # dbus-user.talk ca.desrt.dconf
> dbus-user.talk org.freedesktop.ScreenSaver
> dbus-system none
> ```
> 
> _Why do they need to own `org.kde.*`?!_

Because i can't test dbus-* I suggest to keep it without it now. It's still better/more hardended than old profile.

> We can probably add `disable-mnt`, `private-bin …`, `private-cache`, `tracelog`.

added disable-mnt, private-cache and tracelog. private-bin doesn't work for me, so omitted it.

rusty-snake:
Merged, Thanks.

Issue #2946, opened 2019-09-06
Electron & Chromium


rusty-snake:
There are serveral issues with the chromium sandbox (see below) which is also used in electron. If firejail breaks a electron-based program (or any other program internaly using chromium) and the problem can be fixed by adding `seccomp !chroot` to PROFILE.local, post here which program is affected. Note: If you are not using firejail lastet git, you must add the following to PROFILE.local to get the same effect:
```
ignore seccomp
seccomp.drop @clock,@cpu-emulation,@debug,@module,@obsolete,@raw-io,@reboot,@resources,@swap,acct,add_key,bpf,fanotify_init,io_cancel,io_destroy,io_getevents,io_setup,io_submit,ioprio_set,kcmp,keyctl,mount,name_to_handle_at,nfsservctl,ni_syscall,open_by_handle_at,personality,pivot_root,process_vm_readv,ptrace,remap_file_pages,request_key,setdomainname,sethostname,syslog,umount,umount2,userfaultfd,vhangup,vmsplice
```

If this doesn't work, but `firejail --ignore=nonewprivs --ignore=noroot --ignore=protocol --ignore=seccomp --ignore=caps.drop --caps.keep=sys_admin,sys_chroot PROGRAM` works, say it here. Otherwise open a new issue.

If none of the commands works, open a new issue.

----

Some issues about the chromium-sandbox:
#2933 - skypeforlinux 8.51.0.86 now requires SYS_ADMIN, SYS_CHROOT capabilities
#2912 - Skypeforlinux 8.51.0.72 crashes on startup since it's not permitted to use the chroot syscall
#2945 - Signal 1.27 Fails to Start
#2866 - new version of Slack Desktop (4.0) not working
#2854 - Standard notes not working
#2901 - [Teamspeak 3] crashes on opening options window if seccomp is enabled
#2821 - /usr/bin/riot-desktop: line 3: 8 Trace/breakpoint trap (core dumped) electron /usr/lib/riot/ "$@"
#2943 - firejail - Ubuntu 19.10 snap chromium incompatibility
#2944 - Firejail breaks Brave browser default sandboxing

Three new issues in 10 hours :scream: .
![](https://gist.githubusercontent.com/rusty-snake/9741684540c099c0677f41facf100da4/raw/2b853a00aca20f8a524af078c52c720745d820ca/screenshot.png)

daks:
Hi,

I'm again have a problem with slack after upgrading it to 4.1.1 on Debian 9.
[...]

PR #3519, opened 2020-07-17
Hardend Signal desktop profile

onovy:


etc/profile-m-z/signal-desktop.profile
```diff
@@ -32,12 +32,16 @@ caps.keep sys_admin,sys_chroot
 netfilter
 nodvd
 nogroups
+noroot
```
rusty-snake, line 35: AFAIK noroot break electron/chromium based program in newer versions (see #2946).

etc/profile-m-z/signal-desktop.profile
```diff
@@ -32,12 +32,16 @@ caps.keep sys_admin,sys_chroot
...
 notv
 nou2f
+novideo
+protocol unix,inet,inet6
```
rusty-snake, line 39: Same as noroot.

etc/profile-m-z/signal-desktop.profile
```diff
@@ -32,12 +32,16 @@ caps.keep sys_admin,sys_chroot
 netfilter
 nodvd
 nogroups
+noroot
 notv
 nou2f
+novideo
```
rusty-snake, line 38: IIRC signal has a video-chat. IDK if it is also supported by the desktop-version.

etc/profile-m-z/signal-desktop.profile
```diff
@@ -32,12 +32,16 @@ caps.keep sys_admin,sys_chroot
...
+protocol unix,inet,inet6
 shell none
 
 disable-mnt
 private-dev
+private-etc machine-id,resolv.conf,fonts
```
rusty-snake, line 44: IMHO also these should be added.
TLS/SSL: `ca-certificates,ssl,pki,crypto-policies,nsswitch.conf`
Others: `alternatives,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload`

Also sort it alphabetical.

etc/profile-m-z/signal-desktop.profile
onovy: desktop version don't have video-chat. Only Android/iOS version

rusty-snake:
Merged, Thanks.

PR #3521, opened 2020-07-18
integrate join(-or-start) with dbus options (partial fix)

smitsohu:
update D-Bus environment variables during `join`, so that
a joining process is able to use D-Bus, too.


smitsohu:
Note there are additional problems in the interaction between join and dbus options. 

See also #3522

PR #3525, opened 2020-07-19
New profile for homebank

kortewegdevries:
https://github.com/netblue30/firejail/issues/1139#issuecomment-660631307


etc/profile-a-l/homebank.profile
```diff
@@ -0,0 +1,65 @@
...
+private-tmp
+
+# dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+
+
+
+
```
rusty-snake, line 65: Remove (the blank lines)

etc/profile-a-l/homebank.profile
```diff
@@ -0,0 +1,65 @@
...
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+tracelog
+
+disable-mnt
+private-cache
```
rusty-snake, line 53: `private-bin` can stay if we don't need the browser (as @svc88 said).

Add `private-bin homebank`

etc/profile-a-l/homebank.profile
```diff
@@ -0,0 +1,65 @@
...
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
+# net none
+netfilter
+nodvd
+no3d
+nodbus
```
rusty-snake, line 38:
```suggestion

```

deprecated (if it works with this, you can unomment the `dbus-user none` below).

etc/profile-a-l/homebank.profile
```diff
@@ -0,0 +1,65 @@
...
+whitelist /usr/share/doc/homebank-data
+whitelist /usr/share/doc/homebank
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
```
rusty-snake, line 32:
```suggestion

```

It causes issues with legacy technologies such as X11.

etc/profile-a-l/homebank.profile
```diff
@@ -0,0 +1,65 @@
...
+disable-mnt
+private-cache
+private-dev
+private-etc alternatives,asound.conf,ca-certificates,crypto-policies,dconf,fonts,gtk-3.0,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale.alias,locale.conf,localtime,machine-id,mime.types,nsswitch.conf,pki,pulse,resolv.conf,selinux,ssl,X11
+private-tmp
+
+# dbus-user none
+dbus-system none
+
+memory-deny-write-execute
```
rusty-snake, line 61: Do you know the GUI toolkit. If it uses GTK this should be commented.

etc/profile-a-l/homebank.profile
```diff
@@ -0,0 +1,65 @@
+# Firejail profile for homebank
+# Description: Personal finance manager
+# This file is overwritten after every install/update
+# Persistent local customizations
+include homebank.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/homebank
```
rusty-snake, line 9: add `blacklist ${HOME}/.config/homebank` to disable-programs.inc

kortewegdevries:
> Add it to src/firecfg/firecfg.config.

Oops I forgot about that,I've to open a separate pull then?

etc/inc/disable-programs.inc
```diff
@@ -94,6 +94,7 @@ blacklist ${HOME}/.config/Gitter
 blacklist ${HOME}/.config/Google
 blacklist ${HOME}/.config/Google Play Music Desktop Player
 blacklist ${HOME}/.config/Gpredict
+blacklist ${HOME}/.config/homebank
```
rusty-snake, line 97: disable-programs is sorted alphabetically (as you saw) and case sensitive [with ~/.cache/* in a extra block]. So move it down (239:hexchat).

etc/profile-a-l/homebank.profile
```diff
@@ -0,0 +1,60 @@
...
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc 
+include disable-interpreters.inc
+include disable-programs.inc
+include disable-passwdmgr.inc
```
rusty-snake, line 16: We can add `include disable-shell.inc`.

etc/profile-a-l/homebank.profile
```diff
@@ -0,0 +1,60 @@
...
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/homebank
+whitelist ${DOWNLOADS}
+whitelist ${HOME}/.config/homebank
+whitelist /usr/share/homebank
+whitelist /usr/share/doc/homebank-data
+whitelist /usr/share/doc/homebank
```
rusty-snake, line 24: If it can not display it by it self. We should remove it.

kortewegdevries: No they don't.

etc/inc/disable-programs.inc
kortewegdevries: Thanks, I thought they were sorted by "updated".

svc88:
With the current https://github.com/netblue30/firejail/blob/7ae6c0a90b2db59effe1ad1d62dbcec431069be6/etc/profile-a-l/homebank.profile on my current v0.9.62 i get `Error: cannot access profile file: disable-shell.inc`

rusty-snake:
No, disable-shell.inc was added it 0.9.63.

svc88:
> No, disable-shell.inc was added it 0.9.63.

oh ok, how can i make this profile compatible with 0.9.62 ?

rusty-snake:
Copy disable-shell.inc from git (it works in 0.9.62 too) or comment its include.

svc88:
> Copy disable-shell.inc from git (it works in 0.9.62 too) or comment its include.

ok i tried both ways you said but in both ways it complains
`Error: cannot access profile file: whitelist-runuser-common.inc`

Should this also be copied to 0.9.62 ?

kortewegdevries:
@svc88 This should work for 0.9.62: [homebank.profile](https://github.com/netblue30/firejail/files/4944126/homebank.txt)

rusty-snake:
> Should this also be copied to 0.9.62 ?

You can, it may breaks XWayland under GNOME. However homebank uses GTK so it uses Wayland if possible.

> and now it complains:

0.9.62 is about a half year old. There were some change as you see. For this comment `dbus-* none` and add `nodbus`.

kortewegdevries:
Was trying to push to my testing,ended up here,sorry.

rusty-snake:
Merged, Thanks.

PR #3532, opened 2020-07-21
Added cawbird profile

kortewegdevries:
I think I needed to rebase it, Cawbird's final commit is df8c39a


etc/profile-a-l/cawbird.profile
```diff
@@ -0,0 +1,45 @@
...
+apparmor
+caps.drop all
+machine-id
+netfilter
+nodvd
+no3d
+nogroups
+nonewprivs
+noroot
+nosound
```
rusty-snake, line 29: Can it play videos?
---
Yes, please rebase.

kortewegdevries: I've tested it with two gifs and one video from WHO,I don't get a lot of videos on my timeline. No sound though, do we need sound and hyperlinking? I'll close this for now and rebase and open another PR later if that's okay.

rusty-snake: > I'll close this for now and rebase and open another PR later if that's okay.

You can also force push.

> do we need sound

If sound works w/o FJ and breaks with, we will get new issues.

kortewegdevries: And hyperlinking? Cawbird has some API restrictions,like polls/group DMs for example,not that I use those...

rusty-snake: If it is not such uncommon, we maybe should add a note which can be uncommented.

PR #3533, opened 2020-07-21
Added cawbird profile

kortewegdevries:
https://github.com/netblue30/firejail/pull/3532#discussion_r458142466

It has the ability to copy paste links... https://github.com/netblue30/firejail/pull/3532#discussion_r458149934 so it should be better this way?

The only real use of browser,imo comes from setting it up for the first time,ie, linking it to Twitter via OAuth which needs the link to opened in browser, but fortunately we can copy-paste: https://imgur.com/6iGonmb

### DIFF
--- a/etc/profile-a-l/cawbird.profile
+++ b/etc/profile-a-l/cawbird.profile
@@ -1,0 +1,45 @@
+# Firejail profile for cawbird
+# Description: Open-source Twitter client for Linux
+# This file is overwritten after every install/update
+# Persistent local customizations
+include cawbird.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/cawbird
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-programs.inc
+include disable-passwdmgr.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+apparmor
+caps.drop all
+machine-id
+netfilter
+nodvd
+no3d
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/cawbird.profile
```diff
@@ -0,0 +1,45 @@
...
+protocol unix,inet,inet6
+seccomp
+shell none
+tracelog
+
+disable-mnt
+private-bin cawbird
+private-cache
+private-dev
+private-tmp
```
Fred-Barclay, line 42: Can we add `private-etc`?

kortewegdevries: It works fine without. If we add `private-etc` the only two things it doesn't work without are: `ssl` and `resolv.conf` on my network.

Fred-Barclay: @kortewegdevries let's do it!

kortewegdevries: This? `alternatives,ca-certificates,crypto-policies,fonts,gtk-2.0,gtk-3.0,hosts,host.conf,hostname,mime.types,nsswitch.conf,pki,resolv.conf,ssl,xdg,X11`

Fred-Barclay: Seems right!

kortewegdevries: Done, thanks!

Fred-Barclay:
Squashed and merged, thanks!

Issue #2868, opened 2019-07-23
Virtualbox not able to write to /dev/vbox*


Niklas974:
I'm using firejail version 0.9.58.2 and VirtualBox Version 6.06 on Ubuntu 19.04

Running virtualbox in firejail using the default virtualbox profile provided by firejail, I run into networking issues when accessing host-only network adapters in the virtualbox "host-only network manager"

> Das Host-Netzwerkinterface konnte nicht erzeugt werden.
> 
> VBoxNetAdpCtl: Error while adding new interface: failed to open /dev/vboxnetctl: Operation not permitted.
> 
> Fehlercode:NS_ERROR_FAILURE (0x80004005)
> Komponente:HostNetworkInterfaceWrap
> Interface:IHostNetworkInterface {455f8c45-44a0-a470-ba20-27890b96dba9}

(translated: The Host-Network Interface could not be created)

interestingly, /dev/vbox* looks pretty similar from the outside and from inside the virtualbox firejail:

```
ls -al /dev/vbox*
crw------- 1 root root      10, 53 Jul 22 15:09 /dev/vboxdrv
crw-rw-rw- 1 root root      10, 52 Jul 22 15:09 /dev/vboxdrvu
crw------- 1 root root      10, 51 Jul 22 15:09 /dev/vboxnetctl

/dev/vboxusb:
insgesamt 0
drwxr-x---  5 root vboxusers  100 Jul 22 15:09 .
drwxr-xr-x 23 root root      5080 Jul 22 15:10 ..
drwxr-x---  2 root vboxusers  100 Jul 22 15:09 001
drwxr-x---  2 root vboxusers   60 Jul 22 15:09 002
drwxr-x---  2 root vboxusers  160 Jul 22 15:10 003
 ~ firejail --join=16421
Switching to pid 16422, the first child process inside the sandbox
Child process initialized in 12.40 ms
ubuntu% ls -al /dev/vbox*
crw------- 1 root root      10, 53 Jul 22 15:09 /dev/vboxdrv
crw-rw-rw- 1 root root      10, 52 Jul 22 15:09 /dev/vboxdrvu
crw------- 1 root root      10, 51 Jul 22 15:09 /dev/vboxnetctl

/dev/vboxusb:
insgesamt 0
drwxr-x---  5 root vboxusers  100 Jul 22 15:09 .
drwxr-xr-x 23 root root      5080 Jul 22 15:10 ..
[...]

PR #3537, opened 2020-07-22
Update virtualbox.profile

Neo00001:


etc/profile-m-z/virtualbox.profile
```diff
@@ -14,19 +14,34 @@ noblacklist /usr/lib/virtualbox
...
 mkdir ${HOME}/VirtualBox VMs
 whitelist ${HOME}/.config/VirtualBox
 whitelist ${HOME}/VirtualBox VMs
 whitelist ${DOWNLOADS}
 include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
-caps.keep net_raw,sys_admin,sys_nice
```
rusty-snake, line 29: Sure this does not break? On which distros have you tested it?

etc/profile-m-z/virtualbox.profile
```diff
@@ -14,19 +14,34 @@ noblacklist /usr/lib/virtualbox
...
 netfilter
 nodvd
+#nogroups
 notv
+shell none
+tracelog
+
+#disable-mnt
+private-cache
+private-etc alsa,asound.conf,ca-certificates,dconf,fonts,hostname,hosts,ld.so.cache,pulse,localtime,machine-id,conf.d,resolv.conf,ssl,X11
```
rusty-snake, line 44: Add crypto-policies and pki.

etc/profile-m-z/virtualbox.profile
```diff
@@ -14,19 +14,34 @@ noblacklist /usr/lib/virtualbox
...
+
+dbus-user none
```
rusty-snake, line 46: This breaks things such as notifications and tray-icons, has virtualbox and (usefull) usage of such things?

Neo00001: Virtualbox doesn't have such features.

etc/profile-m-z/virtualbox.profile
Neo00001: done

etc/profile-m-z/virtualbox.profile
Neo00001: I have tested on manjaro(KDE). Kernel version: 5.7.9. you can even use `caps.drop all` . Do you have any profile testing community or group?

kortewegdevries:
Have you tested these? https://github.com/netblue30/firejail/issues/2868#issuecomment-518647630

Neo00001:
> Have you tested these? [#2868 (comment)](https://github.com/netblue30/firejail/issues/2868#issuecomment-518647630)

I tested with NAT.  NAT is virtualbox's default.But it seems without `sys_admin` host-only network doesn't work. So,my suggestion would be rather than enabling `sys_admin` by default, use it in comment. `sys_admin` is overloaded.

Neo00001:
@kortewegdevries  hey can you take a look into this [vmware](https://github.com/Neo00001/frjlprfls/blob/master/vmware) profile.

rusty-snake:
Merged, Thanks.

etc/profile-m-z/virtualbox.profile
Neo00001: `caps.keep net_raw,sys_admin,sys_nice` 
it's needed to be in comment

PR #3548, opened 2020-07-25
fix typo in multicast CIDR

EmilGedda:
The current rule for dropping outgoing traffic to multicast addresses in nolocal.net does not properly prohibit multicasting due to a typo. The multicast CIDR should be 224.0.0.0/4 and not 244.0.0.0/4.

### DIFF
--- a/etc/net/nolocal.net
+++ b/etc/net/nolocal.net
@@ -32,5 +32,5 @@
 -A OUTPUT -d 172.16.0.0/12 -j DROP
 
 # drop multicast traffic
--A OUTPUT -d 244.0.0.0/4 -j DROP
+-A OUTPUT -d 224.0.0.0/4 -j DROP
 COMMIT


EmilGedda:
Example: 
GNU netcat successfully writes to a multicast address with current config.
```bash
$ firejail --quiet --noprofile --net=eth0 --netfilter=/etc/firejail/nolocal.net nc -vu 224.1.1.1 5001 -c 1 <<< "foo"
$ echo $?
0
```
If nolocal.net is updated with the proper CIDR, netcat will fail.
```bash
$ firejail --quiet --noprofile --net=eth0 --netfilter=/etc/firejail/nolocal.net nc -vu 224.1.1.1 5001 -c 1 <<< "foo"
write(net): Operation not permitted
$ echo $?
1
```

rusty-snake:
Merged, Thanks.

PR #3553, opened 2020-07-26
Added Nuclear profile

kortewegdevries:
[Nuclear](https://github.com/nukeop/nuclear/)


kortewegdevries:
Does `--build` fail sometimes? The program doesn't even open if I try with that option.tracelog prvents it from starting. Once they upstream the fix for the bug, the seccomp filter needs to be changed.It works for now.

rusty-snake:
> Does --build fail sometimes?

Uses internally --trace which uses the same technic as --tracelog. Build also fails if `--nonewprivs` or `--caps.drop all` breaks.

rusty-snake:
Merged, Thanks.

PR #3555, opened 2020-07-27
Added minitube profile

kortewegdevries:
Request: # 1139

This needs to be merged before I push profiles for apps from the same author. They all have common directory `Flavio Tordini`

### DIFF
--- a/etc/profile-m-z/minitube.profile
+++ b/etc/profile-m-z/minitube.profile
@@ -38,13 +38,13 @@ include whitelist-var-common.inc
 apparmor
 caps.drop all
 netfilter
-no3d
 nodvd
 nogroups
 nonewprivs
 noroot
 notv
 nou2f
+novideo
 protocol unix,inet,inet6,netlink
 seccomp
 shell none
--- a/etc/profile-m-z/minitube.profile
+++ b/etc/profile-m-z/minitube.profile
@@ -1,0 +1,61 @@
+# Firejail profile for minitube
+# Description: Native Youtube viewer for Linux
+# This file is overwritten after every install/update
+# Persistent local customizations
+include minitube.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${PICTURES}
+noblacklist ${HOME}/.cache/Flavio Tordini
+noblacklist ${HOME}/.config/Flavio Tordini
+noblacklist ${HOME}/.local/share/Flavio Tordini
+
+include allow-lua.inc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc 
+include disable-xdg.inc
+
+mkdir ${HOME}/.cache/Flavio Tordini
+mkdir ${HOME}/.config/Flavio Tordini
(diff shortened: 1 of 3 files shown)


etc/profile-m-z/minitube.profile
```diff
@@ -0,0 +1,61 @@
...
+whitelist /usr/share/minitube
+include whitelist-common.inc 
+include whitelist-runuser-common.inc 
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+netfilter
+no3d
```
rusty-snake, line 41: "Disable 3D hardware acceleration.", could cause performance issues.

etc/profile-m-z/minitube.profile
```diff
@@ -0,0 +1,61 @@
...
+nodvd
+nogroups
+nonewprivs
+noroot
+notv
+nou2f
```
rusty-snake, line 47: `novideo`?

kortewegdevries: Wouldn't video break with it?

etc/profile-m-z/minitube.profile
kortewegdevries: Yeah,even without it I get Mesa terminal spams, I added it at the last moment. I'll remove it.

etc/profile-m-z/minitube.profile
rusty-snake: I thought it is to watch and don't need a webcam.

kortewegdevries: I mixed up the two,pushed.

kortewegdevries:
Off-topic: Is there a way to monitor dbus accesses of a program with something, other than, filtering `dbus-monitor` or flatpak`?

rusty-snake:
> Is there a way to monitor dbus accesses of a program with something, other than, filtering dbus-monitor or 'flatpak'?

If the policy is `filter`, you can use `--dbus-log`. I plan to write a python-script which can parse the output and generate rules.

> Also how do other distros like Arch,SUSE handle pop-up notifications,pretty sure notify-send is common across Ubuntu (atleast Xubuntu),

`notify-send` exists also in Fedora, Arch, …
If there would be own standards they would be part of a DE not a distro. 

> do we allow pop-ups?

`org.freedesktop.Notifications` can be used to escape the sandbox. I should be avoided on sandboxed with high-attack-surface / strict profile / whitelist profile. If it is essential for a program or the program has "crash" as "error-handling" it must be allowed.

kortewegdevries:
Thanks.

rusty-snake:
Not yet complete https://github.com/netblue30/firejail/wiki/Restrict-D-Bus

rusty-snake:
`org.freedesktop.Notifications`:
- sway: safe (AFAIK)
- gnome <= 3.36.0: unsafe
- gnome >= 3.36.1: safe
- kde: unsafe

kortewegdevries:
Idk why I think screensaver may be used to steal user credentials since it the login screen is one of the places to snoop on and be sure that a password is entered. Is this possible? https://gitlab.gnome.org/GNOME/gnome-keyring/-/issues/5

rusty-snake:
Merged, Thanks.

PR #3556, opened 2020-07-27
Added lyx profile

kortewegdevries:
Request: # 1139

I had to create this on a _very_ _temporary_ LaTeX setup, would appreciate some additional testing. Thanks.

`disable-shell` breaks luatex etc from converting files, `private-tmp` breaks temporary copies of files,too many programs to add to `private-bin`.

### DIFF
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -101,6 +101,7 @@ blacklist ${HOME}/.config/Jitsi Meet
 blacklist ${HOME}/.config/Kid3
 blacklist ${HOME}/.config/Kingsoft
 blacklist ${HOME}/.config/Luminance
+blacklist ${HOME}/.config/LyX
 blacklist ${HOME}/.config/Mattermost
 blacklist ${HOME}/.config/Meltytech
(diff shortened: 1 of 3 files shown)


kortewegdevries:
I'll push a few fixes later: The `texmf` dirctory seems to vary by distro. I made this on Arch where it is in `/usr/share/texmf/` but on Ubuntu it's in `/usr/share/texlive/texmf/`, also perl needs to be allowed (`perl-tk` is used in latex).

> Editors have usually blacklisting profiles

Example?

glitsj16:
> Example?

Things like emacs, vim for example. And then there are gedit, kate and several others. Whitelisting text editors just doesn't make sense, unless you really want to make things very dysfunctional for the user. If you have a specific workflow and know exactly where you'd want to edit files, than you can tighten things up by using a whitelisting profile. But that's very hard to predict for every user, so we tend to use blacklisting profiles as @rusty-snake indicated above.

rusty-snake:
Merged, Thanks.

kortewegdevries:
>  If I include allow-perl /usr/share/perl* needs to be whitelisted too, here or like on YouTube-viewer commit.

What about this?

PR #3559, opened 2020-07-29
harden bandwidth command

smitsohu:
* add extra checks to defend against command injection (strings are already controlled by Firejail, so this should be redundant and only for the paranoid)
* run shell in a minimal sandbox


smitsohu:
If there is no protest I'll merge this in the next couple of days

glitsj16:
No objections here to merge this.

PR #3569, opened 2020-08-05
seccomp: logging

topimiettinen:
Allow `log` as an alternative seccomp error action instead of killing
or returning an errno code.

Signed-off-by: Topi Miettinen <toiwoton@gmail.com>


topimiettinen:
Enabling logging may produce a lot of audit entries, so perhaps there should be a mechanism for system administrator to disable the feature.

startx2017:
Should work this way, they can always rotate the logs disable logging.

PR #3571, opened 2020-08-06
add --include

### LINKED ISSUE #2923: include on the commandline
for debuging it would be usefull to have the opportunity to use `include` on the command line.
Example:

```
firejail --noprofile --include=disable-programs --include=disable-common bash
firejail --profile=firefox --include=$HOME/temp-firefox.local
```

rusty-snake:
closes #2923

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -1776,11 +1776,16 @@ int main(int argc, char **argv, char **envp) {
 			while (*ptr != '/' && *ptr != '\0')
 				ptr++;
 			if (*ptr == '\0') {
-				// ppath contains no '/', assume it's a name
-				int rv = profile_find_firejail(ppath, 0);
-				if (rv == 0) {
-					// It's not a valid name, but maybe it's a file in the working director
+				if (access(ppath, R_OK)) {
 					profile_read(ppath);
+				}
+				else {
+					// ppath contains no '/' and is not a local file, assume it's a name
+					int rv = profile_find_firejail(ppath, 0);
+					if (!rv) {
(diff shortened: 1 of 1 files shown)


smitsohu:
Looks good. I still wonder if `--include` and `--profile` in the end should do the profile search the same way, including handling of file extensions.

Then the only difference would be that `--profile` overrides automatic profile loading, whereas `--include` just loads additional instructions from a file.

src/firejail/main.c
```diff
@@ -1767,6 +1767,29 @@ int main(int argc, char **argv, char **envp) {
...
+		else if (strncmp(argv[i], "--include=", 10) == 0) {
+			char *ppath = expand_macros(argv[i] + 10);
+			if (!ppath)
+				errExit("strdup");
+
+			char *ptr = ppath;
+			while (*ptr != '/' && *ptr != '\0')
+				ptr++;
+			if (*ptr == '\0') {
+				// ppath contains no '/', assume it's a name
```
smitsohu, line 1779: Would it be possible to somehow prioritize search in the current working directory?

rusty-snake: Done.

> Looks good to me. I still wonder if --include and --profile in the end should do the profile search the same way, including handling of file extensions.
>
>Then the only difference would be that --profile overrides automatic profile loading, whereas --include just loads additional instructions from a file.

I tried first, but I don't completely understand the `--profile` code. What does `has_colon`.

netblue30:
@rusty-snake -  looks fine, are you merging it?

PR #3572, opened 2020-08-06
hardening: run plugins with dumpable flag cleared

smitsohu:
The kernel clears the dumpable flag if a user has no read permission on an executable and it is owned by another user (root in this case). I omitted faudit, fbuilder and ftee for now as they are not used to configure the sandbox itself, and as this commit is going to complicate debugging efforts to some extent. By design some processes are not covered this way, most notably xauth and shell instances created by system(3). But still it should be an improvement over the current state.


smitsohu:
As xauth is quite small, if we want it to be non-dumpable as well, maybe we could just copy it in the sandbox.

smitsohu:
Appears to work quite well with the copied xauth, so I'm adding more commits.

Along the way add a minimal sandbox for xauth to run in and fix the issue that `--x11=xorg` didn't work with `--net` (because of the hidden X11 socket; now the new Xauthority file is created in the runtime directory, so there is no need to mount a tmpfs on /tmp anymore)

smitsohu:
As file permissions are exposed to the user, the various plugins should detect if they are dumpable and print a (debug) message in that case.

I'll remove the commit dealing with system(3) and handle it in a separate pull request.

PR #3573, opened 2020-08-07
mkdeb.sh should not use files outside $CODE_DIR

dandelionred:
With this change you can build firejail deb with only `firejail-*.tar.xz` and `mkdeb.sh` in current dir.

Currently (as well as in 9.62) such build fails:
```
install: cannot stat ‘platform/debian/copyright’: No such file or directory
```


startx2017:
Thanks for the patch!

PR #3574, opened 2020-08-08
annotate some functions as non-returning

reinerh:
I noticed some comments in the code like "// it will never get here", so I added some annotations to make this assumption also clear to the compiler.


netblue30:
Fine with me, check it in!

PR #3583, opened 2020-08-11
Fix nomacs

kortewegdevries:
Probably very recently `ImageLounge 3.17.0`
```
Aug 11 16:32:32 korte audit[29004]: SECCOMP auid=1000 uid=1000 gid=1000
ses=2 subj==firejail-default (enforce) pid=29004 comm="nomacs"
exe="/usr/bin/nomacs" sig=31 arch=c000003e syscall=9 compat=0
ip=0x7fa2a1cc98c6 code=0x0

```


Fred-Barclay:
merged, thanks!

PR #3588, opened 2020-08-15
Fix private-etc of electron-mail, fix geary,minitube

kortewegdevries:
https://imgur.com/nKwR29x

Does someone want to make this a redirect to electron too, it'd be easier to manage that way since most of them are such...?

### DIFF
--- a/etc/profile-m-z/minitube.profile
+++ b/etc/profile-m-z/minitube.profile
@@ -46,7 +46,7 @@ notv
 nou2f
 novideo
 protocol unix,inet,inet6,netlink
-seccomp
+seccomp !kcmp
 shell none
 tracelog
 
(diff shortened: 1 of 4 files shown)


kortewegdevries:
Geary has redirect to firefox:

```
```

etc/profile-m-z/minitube.profile
```diff
@@ -46,7 +46,7 @@ notv
 nou2f
 novideo
 protocol unix,inet,inet6,netlink
-seccomp
+seccomp !kcmp
```
rusty-snake, line 49: Is this still necessary? Have you set `seccomp-error-action kill`?

kortewegdevries: In `firejail.config`? No. The screen freezes without, on Arch...

rusty-snake: In globals.local or other included overrides.

> The screen freezes without, on Arch...

Ahh, freeze not crash. :+1:

rusty-snake (approved):

rusty-snake:
Thanks for this contribution!

PR #3592, opened 2020-08-17
Allow video for Signal profile

onovy:
Signal is adding support for video calls on desktop, see
https://signal.org/blog/desktop-calling-beta/


Fred-Barclay:
Merged, thanks!

PR #3607, opened 2020-08-28
Switch mails to whitelisting

kortewegdevries:
I might do kmail too, if I can. Will be same format, pushed in this PR.

### DIFF
--- a/etc/profile-a-l/evolution.profile
+++ b/etc/profile-a-l/evolution.profile
@@ -7,8 +7,7 @@ include evolution.local
 include globals.local
 
 noblacklist ${HOME}/.bogofilter
-# Uncomment for gpg
-# noblacklist ${HOME}/.gnupg
+noblacklist ${HOME}/.gnupg
 noblacklist ${HOME}/.pki
 noblacklist ${HOME}/.cache/evolution
 noblacklist ${HOME}/.config/evolution
@@ -27,28 +26,24 @@ include disable-shell.inc
 include disable-xdg.inc
 
 mkdir ${HOME}/.bogofilter
-# Uncomment for gpg
-# mkdir ${HOME}/.gnupg
+mkdir ${HOME}/.gnupg
 mkdir ${HOME}/.pki
 mkdir ${HOME}/.cache/evolution
 mkdir ${HOME}/.config/evolution
 mkdir ${HOME}/.local/share/evolution
 mkdir ${HOME}/.local/share/pki
 whitelist ${HOME}/.bogofilter
-# Uncomment for gpg
-# whitelist ${HOME}/.gnupg
+whitelist ${HOME}/.gnupg
 whitelist ${HOME}/.pki
 whitelist ${HOME}/.cache/evolution
 whitelist ${HOME}/.config/evolution
 whitelist ${HOME}/.local/share/evolution
 whitelist ${HOME}/.local/share/pki
 whitelist ${DOWNLOADS}
-# Uncomment for gpg
-# whitelist ${RUNUSER}/gnupg
+whitelist ${RUNUSER}/gnupg
 whitelist /usr/share/evolution
-# Uncomment for gpg
-# whitelist /usr/share/gnupg
-# whitelist /usr/share/gnupg2
+whitelist /usr/share/gnupg
+whitelist /usr/share/gnupg2
 whitelist /var/mail
 whitelist /var/spool/mail
 include whitelist-common.inc
@@ -74,9 +69,8 @@ seccomp
 shell none
 tracelog
 
-disable-mnt
-# Add "gpg,gpg2,gpg-agent,pinentry-curses,pinentry-emacs,pinentry-fltk,pinentry-gnome3,pinentry-gtk,pinentry-gtk2,pinentry-gtk-2,pinentry-qt,pinentry-qt4,pinentry-tty,pinentry-x2go,pinentry-kwallet" for gpg
-private-bin evolution
+# disable-mnt
+# private-bin evolution
 private-cache
 private-dev
 private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,gcrypt,gtk-2.0,gtk-3.0,groups,hostname,hosts,mailname,passwd,pki,resolv.conf,selinux,ssl,xdg
@@ -89,8 +83,7 @@ dbus-user.own org.gnome.Evolution
 dbus-user.talk ca.desrt.dconf
 # Uncomment to have keyring access
 # dbus-user.talk org.freedesktop.secrets
+dbus-user.talk org.gnome.keyring.SystemPrompter
+dbus-user.talk org.gnome.OnlineAccounts
 dbus-user.talk org.freedesktop.Notifications
 dbus-system none
-
-# Comment to use gpg
-read-only ${HOME}/.gnupg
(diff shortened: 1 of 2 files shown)


etc/profile-a-l/evolution.profile
```diff
@@ -40,7 +72,25 @@ novideo
...
+private-bin evolution
+private-cache
 private-dev
+private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,gcrypt,gtk-2.0,gtk-3.0,groups,hostname,hosts,mailname,passwd,pki,resolv.conf,selinux,ssl,xdg
 private-tmp
+writable-run-user
 writable-var
+
+dbus-user filter
+dbus-user.own org.gnome.Evolution
```
rusty-snake, line 88: The flatpak has also
```
org.gnome.keyring.SystemPrompter=talk
org.gnome.OnlineAccounts=talk
```
I'm not sure if they are needed.

etc/profile-a-l/evolution.profile
```diff
@@ -40,7 +72,25 @@ novideo
...
+dbus-user.talk ca.desrt.dconf
+# Uncomment to have keyring access
+# dbus-user.talk org.freedesktop.secrets
+dbus-user.talk org.freedesktop.Notifications
+dbus-system none
+
+# Comment to use gpg
+read-only ${HOME}/.gnupg
```
rusty-snake, line 96: ~/.gnupg is blacklisted by default in your current commit. Why these?

etc/profile-a-l/evolution.profile
```diff
@@ -40,7 +72,25 @@ novideo
 protocol unix,inet,inet6
 seccomp
 shell none
+tracelog
 
+disable-mnt
+# Add "gpg,gpg2,gpg-agent,pinentry-curses,pinentry-emacs,pinentry-fltk,pinentry-gnome3,pinentry-gtk,pinentry-gtk2,pinentry-gtk-2,pinentry-qt,pinentry-qt4,pinentry-tty,pinentry-x2go,pinentry-kwallet" for gpg
+private-bin evolution
```
rusty-snake, line 79: Probably breaks others things too?

etc/profile-a-l/evolution.profile
```diff
@@ -40,7 +72,25 @@ novideo
 protocol unix,inet,inet6
 seccomp
 shell none
+tracelog
 
+disable-mnt
```
rusty-snake, line 77: Is it common to attach files from USB-Sticks?

rusty-snake (commented):
- Whitelisting: +1
- Disabling gpg support: I'm not sure whether this is good. I mean if the most users will need to enable it, it would be better to keep it and the users who don't need it can remove it if they want.

etc/profile-a-l/evolution.profile
kortewegdevries: The prompter is for the password box I think, in this regard seahorse needs to be fixed, exporting secret key doesn't work with current dbus-filters, it's probably x.x.x.internalprompter something. I don't know about online accounts...

etc/profile-a-l/evolution.profile
kortewegdevries: Copy/paste error...

etc/profile-a-l/evolution.profile
kortewegdevries: I haven't seen except browsers,I used plain-text emails. I'll comment bin and remove the gpg line?

etc/profile-a-l/evolution.profile
kortewegdevries: Is it? Idk...

etc/profile-a-l/evolution.profile
rusty-snake: I would say add the gpg stuff to the commented private-bin for references if someone wants to use private-bin.

kortewegdevries:
So I will rewrite the geary profile since it was decided not to have a firefox redirect (ref. https://github.com/netblue30/firejail/issues/  1773#issuecomment-364792590) The dbus-filters needs to be added, since I could find one issue regarding dbus on Arch.

rusty-snake:
> is it common to open a link in mail?

yes

kortewegdevries:
So do you want to add profiles.ini,dbus talk,commented private-bin bash,sh,which,firefox to them?

rusty-snake:
If the profile has no private-bin, add the lines from thunderbird.
```
noblacklist ${HOME}/.mozilla
```
If it has a private-bin add it commented and a note about private-bin that you need to ignore or extend.

kortewegdevries:
If there aren't any corrections, I'm done on both mails: (2+3)...

etc/profile-a-l/evolution.profile
```diff
@@ -71,7 +73,8 @@ shell none
 tracelog
 
 # disable-mnt
-# Add "gpg,gpg2,gpg-agent,pinentry-curses,pinentry-emacs,pinentry-fltk,pinentry-gnome3,pinentry-gtk,pinentry-gtk2,pinentry-gtk-2,pinentry-qt,pinentry-qt4,pinentry-tty,pinentry-x2go,pinentry-kwallet" for gpg 
+# Add "gpg,gpg2,gpg-agent,pinentry-curses,pinentry-emacs,pinentry-fltk,pinentry-gnome3,pinentry-gtk,pinentry-gtk2,pinentry-gtk-2,pinentry-qt,pinentry-qt4,pinentry-tty,pinentry-x2go,pinentry-kwallet" for gpg
+# Add "sh,bash,which,firefox" for hyperlinks
```
rusty-snake, line 77: The private-bin for firefox is distro-specific, this works only for Arch.

kortewegdevries: xdg-open or something else? I use this on various profiles on Xubuntu with 0.9.64-2,assuming a firefox is running...

kortewegdevries: Since it's distro dependent, I just remove that line?

netblue30:
@rusty-snake should we pull it in?

rusty-snake:
IDK, I didn't see something against it. Testing it would be the best.

etc/profile-a-l/evolution.profile
```diff
@@ -40,7 +70,27 @@ novideo
 protocol unix,inet,inet6
 seccomp
 shell none
+tracelog
 
+# disable-mnt
+# Add "gpg,gpg2,gpg-agent,pinentry-curses,pinentry-emacs,pinentry-fltk,pinentry-gnome3,pinentry-gtk,pinentry-gtk2,pinentry-gtk-2,pinentry-qt,pinentry-qt4,pinentry-tty,pinentry-x2go,pinentry-kwallet" for gpg
```
Fred-Barclay, line 76: If private-bin doesn't support gpg by default (which really I think we should support) then shouldn't we block access to ~/.gnupg ?

rusty-snake: `private-bin` is disabled (see https://github.com/netblue30/firejail/pull/3607#discussion_r479185506), this means gpg is in $PATH.

Fred-Barclay: @rusty-snake 🤦 yep!

hyiltiz:
I can test out kmail if/when it is ready.

rusty-snake:
Should we merge now?

netblue30:
all in, thanks!

PR #3622, opened 2020-09-05
New disable include: disable-write-mnt.inc

rusty-snake:
It is for profiles which have a reasonable mnt access (we can not add
disable-mnt), but no edit function (e.g. any kind of viewer).

Added to
 - profile.template
 - default.profile
 - eo-common.profile


reinerh (approved):

SkewedZeppelin:
Should it really be added to default.profile?

rusty-snake:
I did not expect many breaktages and the default profile is a profile that might work or not, a own profile is always better. However, the default profile has no other options/includes which restricts user file, I remove it.

PR #3652, opened 2020-10-02
added configure option to disable man pages

monich:
Those are unnecessary in embedded environment. Besides, in our particular build environment, man page generation simply fails.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -14,12 +14,18 @@ HAVE_APPARMOR=@HAVE_APPARMOR@
 HAVE_CONTRIB_INSTALL=@HAVE_CONTRIB_INSTALL@
 BUSYBOX_WORKAROUND=@BUSYBOX_WORKAROUND@
 HAVE_SUID=@HAVE_SUID@
+NEED_MAN=@NEED_MAN@
 
-all: all_items mydirs man filters
+ifneq ($(NEED_MAN),no)
+MAN_TARGET = man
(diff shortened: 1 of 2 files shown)


netblue30:
Thanks!

reinerh:
Thanks @monich. Can you please also make the changes to configure.ac?
configure is a file generated by autotools out of configure.ac.
This means your change would be overwritten when the file is re-generated the next time.

monich:
> Thanks @monich. Can you please also make the changes to configure.ac?
> configure is a file generated by autotools out of configure.ac.
> This means your change would be overwritten when the file is re-generated the next time.

That was actually my original plan but `configure` turned out to be already out of sync with `configure.ac` because of https://github.com/netblue30/firejail/commit/61295ce AFAICT.

I can separately submit my changes to `configure.ac` but keep in mind that if you run autoreconf it would overwrite those `runstatedir` changes. Either way it's a mess ))

rusty-snake:
https://github.com/netblue30/firejail/commit/228be74d59e54290f4a8f4e061112dd67ba2778f#diff-67e997bcfdac55191033d57a16d1408a

monich:
Ah, never mind, just noticed that configure.ac has already been updated. `runstatedir` stuff still isn't there, though. Or it could be that I have a different version of autohell. Whatever))

rainemak:
Thanks @monich :+1:

PR #3653, opened 2020-10-05
fix command test in jail_prober.py

briaeros:
In jail_prober.py, a bug prevent a correct testing :
Argument of the firejail command  aren't tested when they are presented to the user, but one loop after (once they're added to "goodargs" array).
This also imply that they're not removed if they doesn't work 

This little modification correct these two problems.

### DIFF
--- a/contrib/jail_prober.py
+++ b/contrib/jail_prober.py
@@ -114,7 +114,10 @@ def run_firejail(program, allArgs):
     print('Attempting to run %s in Firejail' % program)
     for arg in allArgs:
         print('Running with', arg)
-        subprocess.call(goodArgs)
+        #We are adding the argument in a copy of the actual list to avoid modify it now.
+        myargs=goodArgs.copy()
+        myargs.insert(-1,arg)
+        subprocess.call(myargs)
         ans = input('Did %s run correctly? [y]/n ' % program)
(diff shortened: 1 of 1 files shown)


Fred-Barclay:
@briaeros good eye! The first thing is (I believe) actually intentional though implemented incorrectly. Basically I'd like this to make the first iteration with `--noprofile` only as a sanity check. 

Would you like to make the changes so that this occurs? `--noprofile` the first time, `--noprofile` and argument 1 the second, `--noprofile` and arguments 1 and 2 the third, and so on...
No worries of course if you don't feel like doing this! I need to clean the file up anyways and fix up some style issues.

Cheers and thanks!
Fred

briaeros:
Hello,
I've create a little logic so that we have the first pass without argument.
It's probably not the briefest way to do it, but it works.

I didn't remove the "\r" there are at the end of each line, as I don't know if they are intentional or if this (this PR) is the right place to do it.

Fred-Barclay:
Thanks! Looks good, I'll merge this in! :) No worries on line endings (did you copy this to a Windows machine by chance?), I'm going to refactor the file soon anyways.

Cheers and thanks!
Fred

Fred-Barclay (approved):

PR #3656, opened 2020-10-06
Update vmware.profile

Neo00001:
With private-etc enabled vmware-tools doesn't get installed. Existing VM with an installed vmware-tools works as usual. For the time being keep it commented.

### DIFF
--- a/etc/profile-m-z/vmware.profile
+++ b/etc/profile-m-z/vmware.profile
@@ -34,6 +34,6 @@ shell none
 tracelog
 
 #disable-mnt
-private-etc alsa,asound.conf,ca-certificates,conf.d,crypto-policies,dconf,fonts,gtk-2.0,gtk-3.0,hostname,hosts,ld.so.cache,localtime,machine-id,pki,pulse,resolv.conf,ssl,vmware,vmware-installer,vmware-vix
+#private-etc alsa,asound.conf,ca-certificates,conf.d,crypto-policies,dconf,fonts,gtk-2.0,gtk-3.0,hostname,hosts,ld.so.cache,localtime,machine-id,pki,pulse,resolv.conf,ssl,vmware,vmware-installer,vmware-vix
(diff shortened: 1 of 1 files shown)


rusty-snake:
LGTM, what about adding a note like `Uncomment if vmware-tools is installed`?

Neo00001:
> LGTM, what about adding a note like `Uncomment if vmware-tools is installed`?

As of now,I can for sure say that 'vmware-tools' is not being installed because of private-etc.There can many other problems also. So if you have to put a note,don't make it 'vmware-tools' specific. What about, "private-etc may break certain basic features,uncomment at your own risk."

I will try to fix it but problem is vmware logs doesn't hint at anything. If you have any suggestion please let me know.

rusty-snake:
Ah, I had understand that it only breaks installing vmware-tools, but works fine after vmware-tools are installed. If it has general issues, that's fine.

rusty-snake:
> I will try to fix it but problem is vmware logs doesn't hint at anything. If you have any suggestion please let me know.

`firejail --trace=vmware.trace vmware`

Neo00001:
> > I will try to fix it but problem is vmware logs doesn't hint at anything. If you have any suggestion please let me know.
> 
> `firejail --trace=vmware.trace vmware`, if it works.

got nothing useful.Looks like, have to fiddle with it a bit more.Need time for that.btw,what's the difference between `firejail --trace` & `firejail --debug`


> Ah, I had understand that it only breaks installing vmware-tools, but works fine after vmware-tools are installed.

yea,you are correct in case of `vmware-tools`. I'm suspecting some other features breakage may also happen because of `private-etc`.

rusty-snake:
--trace: trace all files opened/readed/create/... by the program
--debug: print a verbose output what firejail does (e.g blacklist /foo/bar, set nonewprivs, ...)

PR #3660, opened 2020-10-10
Fix AppArmor 3.0 support (closes #3659)

### LINKED ISSUE #3659: AppArmor profile fails to load with AppArmor 3.0.0 installed
**Bug and expected behavior**
Expected behaviour: The AppArmor profile loads without an error (i.e. the command `sudo apparmor_parser -r /etc/apparmor.d/firejail-default` exits with exit code 0), allowing myself to use AppArmor integration in firejail.

Actual behaviour: When loading the AppArmor profile using `sudo apparmor_parser -r /etc/apparmor.d/firejail-default`, I get `Found reference to variable run, but is never declared` and the process exits with exit code 1. As a result, Firejail cannot use AppArmor to confine applications. When running a program in Firejail using the `apparmor` option enabled, I get:

```
```

The program runs normally except there is no AppArmor confinement. Running `sudo aa-status` confirms this, it only says the module is loaded.

Probable cause: The `apparmor` package in my distribution has been upgraded from 2.13.4 to 3.0.0. A change in the syntax or something may have broken the profile.

Probable fix: Update AppArmor profile maybe? not sure how older apparmor versions will be managed then

Temporary fix for those affected: I simply downgraded the AppArmor package as described in the [ArchWiki](https://wiki.archlinux.org/index.php/firejail), this is only for Arch users of course. I had no side effects of downgrading, also I was able to immediately use AppArmor without need to reboot, but I would still recommend rebooting.

**Reproduce**
Install AppArmor 3.0.0 in your distribution and try to load the firejail-default AppArmor profile.

**Environment**
Distro: Arch Linux rolling, latest version as of 2020-10-09

Firejail version: git 5549f89325c324e27a0ba356d02b76591a2fdf43

Should have the same issue if used with latest stable release.

**Additional context**
The error I get when trying to load the apparmor profile using `sudo apparmor_parser -r /etc/apparmor.d/firejail-default`:

```
Found reference to variable run, but is never declared
```

Happens when AppArmor 3.0.0 is used with Firejail.

Loading the apparmor profile using `sudo aa-enforce firejail-default`:

```
Setting /etc/apparmor.d/firejail-default to enforce mode.

ERROR: Found reference to variable run, but is never declared

```
Error code 1.

The loading profile using apparmor_parser is told from the archwiki, the loading profile using aa-enforce is told from firejail docs/program. I don't think it should matter

kris7t:
AppArmor introduces the @{run} variable, which is used in
<abstractions/dbus-strict> and <abstractions/dbus-session-strict> among
other places. Thus, we must #include <tunables/run> to be able to call
these abstractions.

Standard profiles rely on <tunables/global> to include <tunables/run>, since it exists in previous AppArmor versions, too.

~As an attempt at backwards compatibility, we #include if exists instead
of #include, since there is no <tunables/run> in AppArmor 2.x.~

~However, if exists is a relatively new feature, see e.g.
https://phabricator.kde.org/D14526
(However, do note that if exists does not appear as a new feature in
https://gitlab.com/apparmor/apparmor/-/wikis/Release_Notes_2.13).
Therefore, this commit restricts our compatibility to relatively new
(<10 months) old AppArmor releases only.~

~As an alternative, we could detect the AppArmor version at configure
time, and emit a firejail-default profile based on that.~

~Here, I opted for the simpler approach, as distributions likely to ship >10 months old AppArmor (Debian) support Firejail on their own via backports (according to https://github.com/netblue30/firejail/security/policy), anyways.~

### DIFF
--- a/etc/apparmor/firejail-default
+++ b/etc/apparmor/firejail-default
@@ -2,6 +2,10 @@
 # Generic Firejail AppArmor profile
 #########################################
 
+# AppArmor 3.0 uses the @{run} variable in <abstractions/dbus-strict>
+# and <abstractions/dbus-session-strict>.
+#include <tunables/global>
+
 ##########
 # A simple PID declaration based on Ubuntu's @{pid}
 # Ubuntu keeps it under tunables/kernelvars and include it via tunables/global.


reinerh:
> AppArmor introduces the @{run} variable, which is used in
> <abstractions/dbus-strict> and <abstractions/dbus-session-strict> among
> other places. Thus, we must #include <tunables/run> to be able to call
> these abstractions.

If the abstractions are using the @{run} variable, isn't that a bug in these abstractions and would be better addressed there?
When they start using the tunable or drop the variable completely in the future, we have an unnecessary import in our profile.

> As an attempt at backwards compatibility, we #include if exists instead
> of #include, since there is no <tunables/run> in AppArmor 2.x.
> 
> However, if exists is a relatively new feature, see e.g.
> https://phabricator.kde.org/D14526
> (However, do note that if exists does not appear as a new feature in
> https://gitlab.com/apparmor/apparmor/-/wikis/Release_Notes_2.13).
> Therefore, this commit restricts our compatibility to relatively new
> (<10 months) old AppArmor releases only.

Ubuntu 18.04 LTS still ships with 2.12. Though I could patch it when uploading to the PPA.

kris7t:
@reinerh The two examples of `<abstraction/dbus>` use I could find in the built-in profiles were `usr.sbin.avahi-daemon` and `usr.sbin.dnsmasq`. Both of these include `<tunables/global>` before including `<abstractions/dbus>`, which in turn includes `<tunables/run>`.

So most likely `<abstractions/dbus-strict>` is not broken, we just have to follow convention by including `<tunables/global>`. Luckily, `<tunables/global>` is pretty old, so we don't need `if exists` when including it. This should clear up the backwards compatibility issue. I'll update my patch accordingly.

reinerh (approved):

PR #3667, opened 2020-10-13
Update wire-desktop.profile (again)

Neo00001:
On arch,wire-desktop is now depending on electron9. Using wildcard for this sorta packages would be better.

### DIFF
--- a/etc/profile-m-z/wire-desktop.profile
+++ b/etc/profile-m-z/wire-desktop.profile
@@ -27,7 +27,7 @@ seccomp !chroot
 shell none
 
 disable-mnt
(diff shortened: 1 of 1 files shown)


rusty-snake:
You can use wildcards in private-bin. Search for `private-bin .*python\*.*` .

Neo00001:
So, should I change it to `electron*`? or keep it as it is.I saw you people used exact same name,that's why I wanted to know whether I should use wildcard or not.

rusty-snake:
I would say yes. If we don't use globing, we will have this repeatedly. Maybe we could restrict the globbing to `electron,electron[0-9],electron[0-9][0-9]` or `electron??`.

rusty-snake:
Merged, thanks.

PR #3672, opened 2020-10-14
Updated fix_private-bin.py shebang to use env python3 like other contrib/*.py scripts

hlein:
This will simplify some distro packagings of firejail.


reinerh:
Thanks!

hlein:
Awesome, thanks!

PR #3674, opened 2020-10-15
Apply --rmenv immediately to help to avoid the env var length check

### LINKED ISSUE #3673: bug? --rmenv seems to fire after check for length of environment variables, which makes long variables impossible to remove from firejail side
First of all, thank you for this amazing tool, really enjoyed playing with it recently!

As for issue, there is not much to describe aside from the title, so here is

### how to reproduce

<details><summary> steps </summary>

Let's set 2 short environment variables:

```sh
~ ❯ export AAAA=bbbb

~ ❯ export AAAA2=bbbb
```

```sh
~ ❯ firejail --rmenv="AAAA" env | grep -i aaaa

Parent pid 1352757, child pid 1352764
Child process initialized in 88.72 ms
AAAA2=bbbb

~ ❯ firejail --noprofile --rmenv="AAAA" env | grep -i aaaa
Parent pid 1354494, child pid 1354497
Child process initialized in 26.68 ms
AAAA2=bbbb

```

As expected, `AAAA` is removed, and `AAAA2` is available. 

But now let's see what happens if we set `AAAA` to something really long.

```sh
~ ❯ export AAAA=$(tr -dc 'a-zA-Z0-9' < /dev/urandom | head -c 8000)

~ ❯ firejail --noprofile --rmenv="AAAA" env | grep -i aaaa
Error: too long environment variables
```

</details>

I expected same as in the first case to happen, but, as you can see, it just exits with error.

### why even do this

My actual use case for it is removing LS_COLORS, which is generated by [vivid](https://github.com/sharkdp/vivid) and about is 8000 characters long.

I know about a workaround with alias as shown [here](https://github.com/netblue30/firejail/pull/3325#issuecomment-612673155), but LS_COLORS is also used by e.g. [lf](https://github.com/gokcehan/lf) (which sometimes i launch just by itself, not from interactive shell), and my various scripts (which i usually call from [sxhkd](https://github.com/baskerville/sxhkd)). 

Of course, it's fixable even from my side, but **i think it makes more sense for --rmenv to work before the length check.**

That way, i could just add 

```
rmenv LS_COLORS
```

to `~/.config/firejail/globals.local`

And have just one LS_COLORS in .profile, instead of a bunch of local `env LS_COLORS="$(vivid ...)" whatever` or similar.

So what do you think? Is this expected behavior?

**Environment**
 - Arch linux, linux-zen 5.8.14, everything up-to-date.
 - Firejail version: 0ab64e277fc48c56b842e0639d450b3311038333 (because i want new dbus stuff for firefox, because #3290 )

**Checklist**
- [ ] ~~The upstream profile (and redirect profile if exists) have no changes fixing it.~~
- [ ] ~~The program has a profile. (If not, request one in [# 1139](https://github.com/netblue30/firejail/issues/1139))~~
- [ ] ~~Programs needed for interaction are listed in the profile.~~
- [x] A short search for duplicates was performed.
- [ ] ~~If it is a AppImage, `--profile=PROFILENAME` is used to set the right profile.~~

...Behavior is not related to particular program, in regards to duplicates - i haven't  seen anything about length check / --rmenv order.

<details><summary> debug output (with long AAAA) </summary>

```sh
~ ❯ firejail --debug --rmenv="AAAA" env
Error: too long environment variables
```

(yes, there is nothing else)

</details>

## EDIT: 

Ops, i didn't actually need to use latest master to fix #3290. 

Turns out, it's an error on my side - what i wanted to do is to have 2 separate isolated profiles with different configs and stuff, and naturally i used firefox with -no-remote flag, so two instances could run simultaneously.

And when i tried to open new links in one of them with `firejail firefox -no-remote -p pr1 *link*` - i was getting the same error, as in #3290. 

After messing around in the middle of the night with a launch script (in which i wrapped conditional whitelisting and stuff) and upgrading to the latest master it eventually work, so i assumed that master fixed it and forgot about now missing -no-remote.

Turns out, when you run firefox in firejail, you can run two instances simultaneously without -no-remote.

Well, that's embarrassing. :sweat_smile:

But i think original question about --rmenv / length checking order is still valid, since i (and probably other vivid users too) eventually will have to deal with it.

In case anyone wants the launching script - here it is:

<details><summary> fjfirefox </summary>

```sh
#!/usr/bin/env sh

unset LS_COLORS

while [ "$#" -gt 0 ]; do
    case "$1" in
        -p)
            profile="$2"
            shift
            ;;
        -u)
            url="$2"
            shift
            ;;
        *)
            exit 1
            ;;
    esac
    shift
done

profile="${profile:-stable}"
# note that it assumes that profiles are in non default location.
# that way, you can whitelist only one profile and nothing else.
# cache also stored here, as you can check with about:cache
wlprof="$(fd -t d -d 1 "\.$profile$" . ~/.config/firefox-profiles)"

if [ -z "$url" ]; then
    firejail \
        --whitelist="$wlprof" \
        firefox-nightly -p "$profile"
else
    firejail \
        --whitelist="$wlprof" \
        firefox-nightly -p "$profile" "$url"
fi
```

</details>

topimiettinen:
Remove environment variables with --rmenv (and profile rmenv)
immediately. This fixes removing long environment variables (LS_COLORS
generated by vivid), previously the length filter would trip before
the command was processed.

This changes user visible behavior slightly, for example --rmenv=LANG
now applies also to Firejail, while earlier it would only apply to
sandboxed program.

Closes #3673.

Signed-off-by: Topi Miettinen <toiwoton@gmail.com>

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -1004,17 +1004,21 @@ int main(int argc, char **argv, char **envp) {
 			fprintf(stderr, "Error: too long arguments\n");
 			exit(1);
 		}
+		// Also remove requested environment variables
+		// entirely to avoid tripping the length check below
+		if (strncmp(argv[i], "--rmenv=", 8) == 0)
+			unsetenv(argv[i] + 8);
 	}
 
 	// sanity check for environment variables
 	for (i = 0, ptr = envp; ptr && *ptr && i < MAX_ENVS; i++, ptr++) {
 		if (strlen(*ptr) >= MAX_ENV_LEN) {
-			fprintf(stderr, "Error: too long environment variables\n");
+			fprintf(stderr, "Error: too long environment variables, please use --rmenv\n");
 			exit(1);
 		}
 	}
 	if (i >= MAX_ENVS) {
-		fprintf(stderr, "Error: too many environment variables\n");
+		fprintf(stderr, "Error: too many environment variables, please use --rmenv\n");
 		exit(1);
 	}
 
--- a/src/firejail/profile.c
+++ b/src/firejail/profile.c
@@ -912,6 +912,7 @@ int profile_check_line(char *ptr, int lineno, const char *fname) {
 		return 0;
 	}
 	if (strncmp(ptr, "rmenv ", 6) == 0) {
+		unsetenv(ptr + 6); // Remove also immediately from Firejail itself
 		env_store(ptr + 6, RMENV);
 		return 0;
 	}


netblue30:
All in, thanks!

PR #3675, opened 2020-10-16
Strip out \r's.

hlein:
As is, this will not execute because env attempts to locate the
executable 'python3\r', which does not exist.


rusty-snake:
:+1: I want to do this for several weeks already, because rpmlint complains.

PR #3676, opened 2020-10-16
Allow --tmpfs and --bind inside $HOME for unprivileged users

rusty-snake:
--tmpfs was added in 0.9.14 and restricted to root only in 0.9.38
due to priv-esc CVE-2016-10117 (e.g. --tmpfs=/etc and modify
/etc/sudoers). This commit reintroduce it for normal users, if the
realpath of it is inside users-home.

BTW: Could we not allow tmpfs complete if nnp is enforced?

### DIFF
--- a/RELNOTES
+++ b/RELNOTES
@@ -1,3 +1,6 @@
+firejail (0.9.65) baseline; urgency=low
+  * allow --tmpfs inside $HOME for unprivileged users
+
 firejail (0.9.64) baseline; urgency=low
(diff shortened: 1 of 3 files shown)


netblue30:
We go for it immediately after the release middle of next week. You are right, we shouldn't restrict it to root in home directory!

rusty-snake:
Rebased and added `--bind` as well.

netblue30:
Let's try it out!

PR #3683, opened 2020-10-20
Fix blu-ray playback with libaacs

### LINKED ISSUE #3682: No blu-ray playback with vlc using libaacs
**Bug and expected behavior**
Using firejail with the default vlc.profile breaks blu-ray playback of encrypted blu-ray discs depending on libaacs:

Sandboxed VLC output:

```
keydbcfg.c:701: No valid AACS configuration files found
aacs.c:121: No usable AACS libraries found!
dec.c:197: aacs_open() failed: -2!
```
This happens because ~/.config/aacs (where the key database resides) is not accessible.

**No profile and disabling firejail**
Issue can't be reproduced when running with --noprofile

**Reproduce**
1. Run in bash `firejail vlc`
2. Try to play a protected blu-ray disc
3. blu-ray disc does not play, vlc complains about missing AACS database

**Environment**
Arch Linux with firejail version 0.9.64rc1

**Additional context**
This issue is fixed by creating /etc/firejail/vlc.local and whitelisting ~/.config/aacs
```
noblacklist ${HOME}/.config/aacs
whitelist ${HOME}/.config/aacs
```
Pull request to fix the default profile is attached.

**Checklist**
 - [X] The upstream profile (and redirect profile if exists) have no changes fixing it.
 - [X] The program has a profile. (If not, request one in [# 1139](https://github.com/netblue30/firejail/issues/1139))
 - [X] Programs needed for interaction are listed in the profile.
 - [X] A short search for duplicates was performed.

jmetrius:
This fixes #3682 by allowing access to ~/.config/aacs, where the key database for libaacs resides.

### DIFF
--- a/etc/profile-m-z/vlc.profile
+++ b/etc/profile-m-z/vlc.profile
@@ -8,6 +8,7 @@ include globals.local
 
 noblacklist ${HOME}/.cache/vlc
 noblacklist ${HOME}/.config/vlc
+noblacklist ${HOME}/.config/aacs
 noblacklist ${HOME}/.local/share/vlc
 
 include disable-common.inc
@@ -23,6 +24,7 @@ mkdir ${HOME}/.config/vlc
 mkdir ${HOME}/.local/share/vlc
 whitelist ${HOME}/.cache/vlc
 whitelist ${HOME}/.config/vlc
+whitelist ${HOME}/.config/aacs
 whitelist ${HOME}/.local/share/vlc
 include whitelist-common.inc
 include whitelist-players.inc
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -68,6 +68,7 @@ blacklist ${HOME}/.cliqz
 blacklist ${HOME}/.clonk
 blacklist ${HOME}/.config/0ad
 blacklist ${HOME}/.config/2048-qt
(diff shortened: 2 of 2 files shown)


etc/profile-m-z/vlc.profile
```diff
@@ -8,6 +8,7 @@ include globals.local
 
 noblacklist ${HOME}/.cache/vlc
 noblacklist ${HOME}/.config/vlc
+noblacklist ${HOME}/.config/aacs
```
rusty-snake, line 11: There is no such blacklist. Either remove this, or better add a blacklist statement to disable-programs.inc.

rusty-snake (commented):
Great.

Does it need to be writeable or could we add `read-only ${HOME}/.config/aacs`?

etc/profile-m-z/vlc.profile
jmetrius: Thanks! I will add the path to disable.programs.inc

jmetrius:
Unfortunately, `read-only ${HOME}/.config/aacs` in vlc.local does not resolve the issue for me. I don't think that write access should be necessary, but I also have no real idea what libaacs is exactly doing. I tried to poke vlc a bit with ltrace but it seems that doesn't really work well with the default binaries in Arch.

rusty-snake:
thanks

PR #3692, opened 2020-10-24
let tests fail build

reinerh:
This change simplifies the test targets, as the test names need to be specified only once, instead of repeating the targets for each test.
It will now also check the tests for errors and cause a build failure, as they currently seem to not get noticed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,16 +7,20 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
+  build_and_test:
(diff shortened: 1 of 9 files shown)


reinerh:
Had to keep some tests disabled for now (test-profiles, test-fs, test-filter) as they have failing sub-tests.

netblue30:
merged, thanks!

reinerh:
I think you accidentally closed instead of merging. I merged it manually now. :-)

PR #3698, opened 2020-10-25
Update celluloid.profile

Neo00001:
`liblua` is needed for celluloid & otherwise at least on arch it's showing this error - "celluloid: error while loading shared libraries: liblua5.2.so.5.2: cannot open shared object file: Permission denied."

### DIFF
--- a/etc/profile-a-l/celluloid.profile
+++ b/etc/profile-a-l/celluloid.profile
@@ -9,6 +9,7 @@ include globals.local
 noblacklist ${HOME}/.config/celluloid
 noblacklist ${HOME}/.config/gnome-mpv
 noblacklist ${HOME}/.config/youtube-dl
+noblacklist /usr/lib/liblua*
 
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/celluloid.profile
```diff
@@ -9,6 +9,7 @@ include globals.local
 noblacklist ${HOME}/.config/celluloid
 noblacklist ${HOME}/.config/gnome-mpv
 noblacklist ${HOME}/.config/youtube-dl
+noblacklist /usr/lib/liblua*
```
reinerh, line 12: I think for allowing Lua it's better to include allow-lua.inc (which also has this line, but can be easily changed in the future).

Neo00001: done. didn't notice `allow-lua.inc` at first, otherwise would have included that.

reinerh (approved):

rusty-snake:
merged, thanks.

PR #3710, opened 2020-10-30
Exec failure logging

sfc-gh-dbrown:
Add logging to stderr if exec of subprocess fails.


sfc-gh-dbrown:
I see that there was a previous update that already added logging.

PR #3716, opened 2020-11-01
Remove nou2f in ssh profile

Svaag:
[OpenSSH 8.2 added support for U2F](https://www.openssh.com/txt/release-8.2) but the current Firejail SSH profile has nou2f configured. 

My Yubikeys are working fine with the option commented out and I see no reason why U2F shouldn't be enabled in the profile now.


rusty-snake (approved):

netblue30:
Thanks for the patch!

Issue #3690, opened 2020-10-24
test failure: mkdir.exp


reinerh:
```
$ cd test/fs/
$ ./mkdir.exp
spawn /bin/bash
firejail --profile=mkdir.profile find ~/.firejail_test
  $ firejail --profile=mkdir.profile find ~/.firejail_test
Reading profile mkdir.profile
Warning: cannot create user directory
TESTING ERROR 1.2
```
The test fails because of the warning "cannot create user directory".
I'm not sure if this is expected and can be ignored (so the test needs to be adapted) or an actual issue.

netblue30:
I think we can ignore them, if something goes wrong the test will catch it anyway. Try it out.

reinerh:
It's still failing, but at a later point. See log in #3719.
```
TESTING: recursive mkdir (test/fs/mkdir.exp)
spawn /bin/bash
rm -fr ~/.firejail_test
runner@fv-az108-477:~/work/firejail/firejail/test/fs$ rm -fr ~/.firejail_test
runner@fv-az108-477:~/work/firejail/firejail/test/fs$ 
<ejail --profile=mkdir.profile find ~/.firejail_test  
Reading profile mkdir.profile
Warning: cannot create 1001 directory
Parent pid 5169, child pid 5176
Warning: cannot create 1001 directory
]0;firejail find /home/runner/.firejail_test Child process initialized in 13.30 ms
/home/runner/.firejail_test
/home/runner/.firejail_test/a
/home/runner/.firejail_test/a/b
/home/runner/.firejail_test/a/b/c
/home/runner/.firejail_test/a/b/c/d.txt
rm -rf ~/.firejail_test
Warning: removing 24 bytes from stdin

Parent is shutting down, bye...
runner@fv-az108-477:~/work/firejail/firejail/test/fs$ 
<il --profile=mkdir.profile find /tmp/.firejail_test  
Reading profile mkdir.profile
Warning: cannot create 1001 directory
Parent pid 5181, child pid 5184
Warning: cannot create 1001 directory
]0;firejail find /tmp/.firejail_test Child process initialized in 10.48 ms
/tmp/.firejail_test
/tmp/.firejail_test/a
/tmp/.firejail_test/a/b
/tmp/.firejail_test/a/b/c
/tmp/.firejail_test/a/b/c/d.txt
rm -rf /tmp/.firejail_test
Warning: removing 27 bytes from stdin

Parent is shutting down, bye...
[...]

PR #3719, opened 2020-11-01
ci: enable test-fs tests on github-ci

reinerh:
Should work after #3690 has been fixed.


netblue30:
I'm doing some testing for mkdir.exp

PR #3742, opened 2020-11-12
Add profile for straw-viewer

kortewegdevries:


rusty-snake (approved):
LGTM

Are the video-players listed in `private-bin` (e.g. mpv, vlc, ...) started by straw-viewer or gtk-straw-viewer? If they are started by straw-viewer the `blacklist /tmp/.X11-unix` will break.

What about adding `noblacklist` and `whitelist` for the config files of mpv,smplayer,vlc,youtube-dl?

etc/profile-a-l/gtk-straw-viewer.profile
```diff
@@ -0,0 +1,18 @@
+# Firejail profile for gtk-straw-viewer
+# Description: Gtk front-end to straw-viewer
+# This file is overwritten after every install/update
+# Persistent local customizations
+include gtk-straw-viewer.local
+# Persistent global definitions
+# include globals.local
```
glitsj16, line 7: Please use our format for explaining why globals.local is not included here:
```
# added by included profile
#include globals.local
```

kortewegdevries:
Those are the three players it currently supports by default but you can add any player in the config file. It passes some parameters to ignore existing configs so I don't think adding the player's config paths are necessary.

I'll make rest of the changes.

rusty-snake:
merged, thanks.

PR #3748, opened 2020-11-13
evince.profile: noblacklist bookmark folder

samtinel:
bookmarks are saved unter $HOME/.local/share/gvfs-metadata
since evince is the primary pdf reader, a firejailed evince can't read
or write those
this commit makes them available in the sandbox

### DIFF
--- a/etc/profile-a-l/evince.profile
+++ b/etc/profile-a-l/evince.profile
@@ -6,6 +6,10 @@ include evince.local
 # Persistent global definitions
 include globals.local
 
+# Uncomment this line and the bottom ones to use bookmarks
+# NOTE: This possibly exposes information, including file history from other programs.
+#noblacklist ${HOME}/.local/share/gvfs-metadata
+
 noblacklist ${HOME}/.config/evince
 noblacklist ${DOCUMENTS}
 
@@ -53,5 +57,8 @@ private-lib evince,gcc/*/*/libgcc_s.so.*,gcc/*/*/libstdc++.so.*,gconv,gdk-pixbuf
 private-tmp
 
 # might break two-page-view on some systems
-dbus-user none
+dbus-user filter
+# Also uncomment these two lines if you want to use bookmarks
+#dbus-user.talk org.gtk.vfs.Daemon
+#dbus-user.talk org.gtk.vfs.Metadata
 dbus-system none


etc/profile-a-l/evince.profile
```diff
@@ -7,6 +7,7 @@ include evince.local
 include globals.local
 
 noblacklist ${HOME}/.config/evince
+noblacklist ${HOME}/.local/share/gvfs-metadata
```
glitsj16, line 10: We don't seem to blacklist that path anywhere AFAICT, so this doesn't make much sense. Are you by any chance needing this somehow because of options in a evince.local file?

SkewedZeppelin:
> We don't seem to blacklist that path anywhere AFAICT

Wow, we don't.
Crazy!
gvfs-metadata contains information on all storage, static or removable
and also file access history from some programs.

Just run
strings .local/share/gvfs-metadata/*
and you'll see file you recently used.

etc/profile-a-l/evince.profile
samtinel: hmm, that line I have added is the only line in my .local for evince. It doesn't work without it.
I will do more digging next week, maybe I have changed some global settings I can't remember.

rusty-snake:
Fedora 32; Evince 3.36.7; Firejail 0.9.65;

This is the minimal evince.local I could find which allowed me to view/add/remove bookmarks. gvfs/gio is likely doing the same as dconf/gsettings: read via file-system (performance) and write via D-Bus service (avoid corruptions).

```
ignore dbus-user none

noblacklist ${HOME}/.local/share/gvfs-metadata

dbus-user filter
dbus-user.talk org.gtk.vfs.Daemon
dbus-user.talk org.gtk.vfs.Metadata
```

samtinel:
> Fedora 32; Evince 3.36.7; Firejail 0.9.65;
> 
> This is the minimal evince.local I could find
Updated the PR with your changes  - but judging from the first two reactions, I doubt this will be merged?
If that is the case, please close

rusty-snake (changes requested):
remarks just on the profile ordering:

- Now there are two `noblacklist ${HOME}/.local/share/gvfs-metadata`, remove one.
- The D-Bus stuff should go at the end.

thoughts:
Maybe it would be good to make this opt-in.

samtinel:
> remarks just on the profile ordering:
fixed
> thoughts:
> Maybe it would be good to make this opt-in.
This is probably the better idea, I commented out the changes and added an explaining comment - if someone wants it, they can uncomment the lines to use bookmarks.

etc/profile-a-l/evince.profile
```diff
@@ -55,3 +59,9 @@ private-tmp
 # might break two-page-view on some systems
 dbus-user none
 dbus-system none
+
+# Also uncomment these four lines if you want to use bookmarks
+#ignore dbus-user none
+#dbus-user filter
+#dbus-user.talk org.gtk.vfs.Daemon
+#dbus-user.talk org.gtk.vfs.Metadata
```
rusty-snake, line 67: A `ignore` after the line which should be ignored did not work.

```suggestion
# might break two-page-view on some systems
dbus-user filter
# Also uncomment (or put in your evince.local) these four lines if you want to use bookmarks
#dbus-user.talk org.gtk.vfs.Daemon
#dbus-user.talk org.gtk.vfs.Metadata
dbus-system none
```

etc/profile-a-l/evince.profile
```diff
@@ -6,6 +6,10 @@ include evince.local
 # Persistent global definitions
 include globals.local
 
+# Uncomment this line and the bottom ones to use bookmarks
+# NOTE: This possibly exposes information, including file history from other programs.
```
rusty-snake, line 10: The allowed D-Bus names below
- allow editing of all metadata and (u)mounting (if I'm correct)
- maybe allow filesystem access outside of the sandbox (=escape)

glitsj16:
@rusty-snake Any progress here? I see a 'Changes requested' message that looks to be resolved.

rusty-snake:
What about?
```
# might break two-page-view on some systems
dbus-user filter
# Also uncomment these two lines if you want to use bookmarks
#dbus-user.talk org.gtk.vfs.Daemon
#dbus-user.talk org.gtk.vfs.Metadata
dbus-system none
```

rusty-snake (approved):

rusty-snake:
merged, thanks.

0x29a:
@rusty-snake, does this profile still work for you? I've uncommented all specified lines, but can't make evince to store bookmarks.

rusty-snake:
I use evince now as flatpak (to have document portal).

Does it work if you add `dbus-user.talk org.gtk.vfs.*`?

0x29a:
>Does it work if you add dbus-user.talk org.gtk.vfs.*?

@rusty-snake, nope, same.

PR #3751, opened 2020-11-14
disable-shell.inc: add mksh shell

Ypnose:


rusty-snake:
:1st_place_medal:  thanks.

glitsj16:
For other shells we make user rc files read-only in disable-common.inc (section 'Startup files'). I'm not familiar with mksh, but online documentation seems to suggest it uses ${HOME}/.mkshrc. Should we add that path as read-only to disable-common.inc?

rusty-snake:
yes. additional: have we any ro command for ksh startup file? I see no.

glitsj16:
@rusty-snake good catch, that seems to be missing too

Ypnose:
Thanks for merging and sorry for `*rc` files. I forgot it.

glitsj16:
> Thanks for merging and sorry for *rc files. I forgot it.

Not a problem, it made us aware of a few omissions :-) Thanks again for the PR!

PR #3752, opened 2020-11-14
reimplement --get using --cat

### LINKED ISSUE #2301: File transfer fails on large files
Xubuntu 18.04 64-bit
firejail self build from https://github.com/netblue30/firejail/commit/9eca281abbb24f97fdb92c06018cd2bd0a5b215e
```
$ firejail --version
firejail version 0.9.57

Compile time support:
        - AppArmor support is enabled
        - AppImage support is enabled
        - chroot support is enabled
        - seccomp-bpf support is enabled


```

**Steps to reproduce**

1) check free space available in `/run` with `df -Th`, and create or find a file larger than that (On my system I used a 2 GB file)

2) run `firejail --noprofile --overlay-tmpfs`

3) Try to get the large file from (1) using `firejail --get=<pid> path/to/the/file`

**Expected results**

At (3), file should be copied from the overlay-tmpfs to the working directory

**Actual results**

At (3) it sits for a short time, then exits with status 1 without anything seeming to happen.  I think it's related to the file being bigger than free space in `/run` because of this -

https://github.com/netblue30/firejail/blob/9eca281abbb24f97fdb92c06018cd2bd0a5b215e/src/firejail/ls.c#L316-L345

And indeed, a file does appear at `/run/firejail/tmpget-...` while the `firejail --get=...` command is running.

smitsohu:
fixes #2301

### DIFF
--- a/src/firejail/ls.c
+++ b/src/firejail/ls.c
@@ -26,6 +26,7 @@
 #include <dirent.h>
 #include <pwd.h>
 #include <grp.h>
+#include <fcntl.h>
 //#include <dirent.h>
 //#include <stdio.h>
 //#include <stdlib.h>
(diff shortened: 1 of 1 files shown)


netblue30:
merged, thanks!

Issue #3417, opened 2020-05-15
seccomp.block-secondary


rusty-snake:
Are there any reasons why we don't use `seccomp.block-secondary`? It should work for almost all profiles (except wine and steam).

glitsj16:
I have been running with `seccomp.block-secondary` in globals.local for as long as it exists without issues. But I don't use wine nor steam. The only other exception I encountered is when firejailing an **appimage** (which generally I try to avoid). We would need to determine how appimages react to the seccomp.block-secondary option, which I for one am not particularly looking forward to.

rusty-snake:
> I have been running with seccomp.block-secondary in globals.local

I too

> The only other exception I encountered is when firejailing an appimage

Just this one? or also other appimages?

glitsj16:
> Just this one? or also other appimages?

This is the only appimage I use because I couldn't create a reasonable profile for its non-appimage counterpart. It's not something I'm happy with, in general I avoid using snap/flatpak/appimage technology. So I cannot determine if it is something specific to this application or something in general with appimages. I'll try to test a few others, but to be conclusive we might ask @ [AppImageKit](https://github.com/AppImage/AppImageKit) I suppose.

rusty-snake:
Tested KeePassXC and Cool-Retro-Term AI, both worked with sbs. We can go and add sbs to the most profiles.

PR #3757, opened 2020-11-16
from my overrides

rusty-snake:
- add seccomp.block-secondary to a lot profiles
 - add wruc to firefox-common and ignore it in TB and
   firefox-common-addons
 - harden dia, gnome-keyring, libreoffice, megaglest, pngquant,
   ghostwriter, rhythmbox, sqlitebrowser

### DIFF
--- a/etc/inc/firefox-common-addons.inc
+++ b/etc/inc/firefox-common-addons.inc
@@ -2,6 +2,8 @@
 # Persistent customizations should go in a .local file.
 include firefox-common-addons.local
 
+ignore include whitelist-runuser-common.inc
+
 noblacklist ${HOME}/.config/kgetrc
 noblacklist ${HOME}/.config/okularpartrc
 noblacklist ${HOME}/.config/okularrc
--- a/etc/profile-a-l/0ad.profile
+++ b/etc/profile-a-l/0ad.profile
@@ -16,6 +16,7 @@ include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
+include disable-xdg.inc
 
 mkdir ${HOME}/.cache/0ad
 mkdir ${HOME}/.config/0ad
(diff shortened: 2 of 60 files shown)


SkewedZeppelin:
@rusty-snake
lgtm

glitsj16:
LGTM. Just one question/remark. I've been running with `seccomp.block-secondary` in `globals.local` for quite a while without any issues. Only when running appimages that option needs to be ignored. So there might still be other profiles that could have that option. Are there any downsides I'm not seeing to using `seccomp.block-secondary` like this?

SkewedZeppelin:
@glitsj16
> Are there any downsides I'm not seeing to using `seccomp.block-secondary`
[globally]?


steam
qemu?
maybe some compilers?

glitsj16:
@SkewedZeppelin Thans for pointing those out. I guess I can get away with it globally because I don't use steam etc. Sorry for the noise. Looks ready to commit!

topimiettinen:
On amd64, `seccomp.block-secondary` blocks 32-bit i386 system calls. They should be only used by Wine and Steam, possibly old proprietary applications. Compilers and qemu don't need them, unless you have a native 32-bit compiler or 32-bit qemu for some reason. I don't know about other architectures, like aarch64 vs. 32 bit arm. I'd make it global if it's easy to override it in the few profiles that need it.

Though in same logic and depending on the ratio of user/non-user cases, other flags like `no3d`, `nodvd`, `nosound`, `novideo`, `notv, `nou2f` etc., or even `caps.drop=all`, `net=none`, `shell=none` could be global default but overridden for the apps that need access.

rusty-snake:
> So there might still be other profiles that could have that option. 

It works for nearly all programs. Exceptions: steam, wine, debuggers/IDEs when debugging such programs. I use it in pre-globals.local for a long time. I even already tried to set `SystemCallArchitectures=native` in `/etc/systemd/system.conf` and `gtk-query-immodules-3.0-32` was the only program which violated this.

> Only when running appimages that option needs to be ignored.

https://github.com/netblue30/firejail/issues/3417#issuecomment-684921438

>

PR #3762, opened 2020-11-20
reimplement --private-cache using --tmpfs

smitsohu:
Use new `--tmpfs` functionality to implement `--private-cache`. This has the added advantage that Firejail can now deal with symlinked ~/.cache directories, at least as long as the symlink's target lives in user home.

The new `--private-cache` is less verbose and now behaves like `--blacklist`, `--read-only` and so on, which is why the test had to be adapted.

### DIFF
--- a/src/firejail/fs.c
+++ b/src/firejail/fs.c
@@ -162,11 +162,12 @@ static void disable_file(OPERATION op, const char *filename) {
 	}
 	else if (op == MOUNT_TMPFS) {
 		if (S_ISDIR(s.st_mode)) {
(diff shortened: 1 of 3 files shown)


Fred-Barclay:
Cool!
For those of us (aka me, 😆 ) who don't speak C very well, what was the old behavior?

smitsohu:
@Fred-Barclay So far users get a warning when `--private-cache` was specified, but `~/.cache` does not exist or is a link. This warning is dropped with the patch.

netblue30:
merged, thanks!

PR #3766, opened 2020-11-22
Miscellaneous whitelist-runuser-common fixes

kris7t:
1. We must ignore include `whitelist-runuser-common.profile` because it breaks Enigmail (TB 68) and GnuPG smartcard (TB 78) support. The current `thunderbird.profile` had a small typo, so the include wasn't ignored.

However, since the update to Thunderbird 78, Thunderbird does not call the gnupg agent for any GPG public key operations, and only calls it for private key operations if `mail.openpgp.allow_external_gnupg` is set (to allow the use of smart cards). So, this being a quite niche use-case, we might think about including `whitelist-runuser-common.profile` anyways, and leaving only a comment for gnupg agent users.

Alternatively, I had some success with just `whitelist ${RUNUSER}/gnupg` (in conjunction with `writable-run-user` already in `thunderbird.profile`), but that only works if `${RUNUSER}/gnupg` already exists, I think. Maybe `mkdir ${RUNUSER}/gnupg` could help here?

2. If the GDM display manager runs with Wayland support, and it starts a desktop environment other than (?) GNOME, the desktop environment will use the `wayland-1` socket instead of the `wayland-0` socket. Here, I just allow `wayland-1`, too. 

Situations where `wayland-2` or higher ends up being the default socket seem much rarer (but might be possible to trigger with nested compositors, see e.g. sway's wayland backend). So we might think about allowing higher numbers, too.

I'm not sure about the `pipewire-0` (and possibly higher) sockets. Screen sharing with browsers under Wayland (with `wlr-desktop-portal`) might rely on them, but I haven't had the chance to experiment.

Note that exposing multiple Wayland sockets might be a security hole if multiple sockets are created (e.g. with nested compositors, proxies à la dbus-proxy, or security features provided by compositors) for different application. Such setups are probably extremely niche currently. If they become popular, we might try to add direct support for them inside firejail (like `dbus-user` and `dbus-system`). In the meantime, those require manual configuration, but we should document which Wayland sockets are allowed by default (so they can be blocked when desired).

### DIFF
--- a/etc/inc/whitelist-runuser-common.inc
+++ b/etc/inc/whitelist-runuser-common.inc
@@ -10,4 +10,5 @@ whitelist ${RUNUSER}/ICEauthority
 whitelist ${RUNUSER}/.mutter-Xwaylandauth.*
 whitelist ${RUNUSER}/pulse/native
 whitelist ${RUNUSER}/wayland-0
+whitelist ${RUNUSER}/wayland-1
 whitelist ${RUNUSER}/xauth_*
--- a/etc/profile-m-z/thunderbird.profile
+++ b/etc/profile-m-z/thunderbird.profile
@@ -6,7 +6,7 @@ include thunderbird.local
 # Persistent global definitions
 include globals.local
 
-ignore whitelist-runuser-common.inc
+ignore include whitelist-runuser-common.inc
 
 # writable-run-user and dbus are needed by enigmail
 ignore dbus-user none


rusty-snake (approved):
1. I think `${RUNUSER}/gnupg` is created on session start up, so before thunderbird is started. (We have other programs with that line too). Or can this not be the case with minimal sessions e.g. sway?

2. When I wrote wruc, was the only case I could imagine to run a nested composite for e.g. testing. Which is such a edge case and likely only done by users who can edit profiles. If there's a common case where wayland-1 is used, we should add it.

> I'm not sure about the pipewire-0 (and possibly higher) sockets. Screen sharing with browsers under Wayland (with wlr-desktop-portal) might rely on them, but I haven't had the chance to experiment.

Actually the will use it, but at least firefox has no wayland screen-sharing at all ATM IIRC.

kris7t:
> 1. I think ${RUNUSER}/gnupg is created on session start up, so before thunderbird is started. 

I believe (at least on Arch) gpg-agent is socket-activated by the user systemd daemon, so it only starts up on demand. However, thinking about it more, this _does_ mean that the `gnupg` directory (and the sockets inside it!) already exist when the session starts, they're just proxied by systemd.

But in more minimal environments (a window manager without a systemd user daemon), there is no socket activation. Thus gpg-agent must be started manually when the session starts, creating the `gnupg` directory. If it is restarted (e.g., to forget any cached passphrases), I believe the gnupg directory will stay in place (only the sockets disappear).

So thanks for the insights! Looks like we can still get away with just allowing the `gnupg` directory.

I wonder if I could solve the KPXC-Browser wruc situation by similarly socket-activating a firejailed KPXC...

> 2. If there's a common case where wayland-1 is used, we should add it.

GDM-on-Wayland starting a non-GNOME desktop environment might be somewhat common. After all, GDM is the only Wayland-native display manager I know of. So if someone is using a Wayland desktop environment other than GNOME with a Wayland display manager (practically, that's `sway` and `wayfire` at the moment, KDE users are much more likely to use SDDM on X11 even if they use KDE on Wayland), they'd end up with a `wayland-1` socket.

Curiously, I still only have a `pipewire-0` socket and no `piperwire-1`, even though I use `pipewire-pulse` to emulate PulseAudio, on which GDM also depends. So it'd look like GDM starts a Pipewire instance, and later my systemd user daemon also socket-activates another Pipewire instance, but, unlike with Wayland, I don't end up with a situation when two sockets are active at the same time.

netblue30:
all in, thanks!

rusty-snake:
> >I'm not sure about the pipewire-0 (and possibly higher) sockets. Screen sharing with browsers under Wayland (with wlr-desktop-portal) might rely on them, but I haven't had the chance to experiment.
>
> Actually the will use it, but at least firefox has no wayland screen-sharing at all ATM IIRC.

Using Mozilla Firefox for Fedora (i.e. firefox from fedoras repos), screen-sharing under wayland  works with the command below. However fedora adds a lot own patches for firefox' wayland support, which may not in the offical firefox.

```
firejail --ignore=noroot --dbus-user.talk=org.freedesktop.portal.Desktop --whitelist='${RUNUSER}/pipewire-0' /usr/bin/firefox
```

PR #3768, opened 2020-11-22
add gnome-shell search-provider file to firefox.profile

glitsj16:
Firefox has gnome-shell search-provider support since version 78:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1239694
- https://mastransky.wordpress.com/2020/09/25/firefox-gnome-shell-search-provider/

Let's add a comment on how to enable this in firejail.

### DIFF
--- a/etc/profile-a-l/firefox.profile
+++ b/etc/profile-a-l/firefox.profile
@@ -16,6 +16,8 @@ whitelist ${HOME}/.mozilla
 
 whitelist /usr/share/doc
 whitelist /usr/share/firefox
+# Uncomment or put in your firefox.local to enable gnome-shell search provider support
+#whitelist /usr/share/gnome-shell/search-providers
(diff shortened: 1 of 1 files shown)


rusty-snake (commented):
`whitelist /usr/share/gnome-shell/search-providers/firefox-search-provider.ini` should work too, should it not? Then we can enable it by default IMO.

On Fedora it contains `BusName=org.mozilla.Firefox.SearchProvider` I'm wondering why we need no additional `dbus-user.own`.

----

OT: gnome-shell search-provider are started via D-Bus and are always unsandboxed. Even firecfg.py has only partial support for it ATM.

glitsj16:
> whitelist /usr/share/gnome-shell/search-providers/firefox-search-provider.ini should work too, should it not? Then we can enable it by default IMO.

@rusty-snake Yep, that's a cleaner way to do it. Thanks!

PR #3772, opened 2020-11-24
use openat2 syscall when available

smitsohu:
Kernel 5.6 added a new system call [openat2](https://man7.org/linux/man-pages/man2/openat2.2.html) which can help us in dealing with symbolic links.

`safe_fd()` also had a tiny bug where the root directory would be opened with the wrongs flags. While not very relevant to current Firejail, fixing it allowed sorting instructions in a different and imho slightly more logic way.


netblue30:
all in!

PR #3778, opened 2020-11-28
Fix building C# projects in Godot

RandomVoid:
Since version 3.0 Godot is supporting C# as a language for writing scripts. The C# solution can be built directly in Godot editor using MSBuild, which requires access to directory /etc/mono. This directory contains configuration of Mono enviroment. If MSBuild don't have access to this directory, it's not able to determine location of DLL files and it's throwing System.DllNotFoundException at beginning of the build process.

### DIFF
--- a/etc/profile-a-l/godot.profile
+++ b/etc/profile-a-l/godot.profile
@@ -38,7 +38,7 @@ tracelog
 # private-bin godot
 private-cache
 private-dev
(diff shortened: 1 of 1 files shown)


rusty-snake (approved):

rusty-snake:
Thanks for this contribution!

PR #3779, opened 2020-11-28
Update build.yml

rusty-snake:
Currently we run all actions for all commits. This is not resource friendly. Let's tweak this a bit.

This commit adds support for "skip ci" tags (i.e. if the head commit contains `[skip   ci]`, no jobs are executed.
In addition are all commits which modify non-code files (e.g. README) only excluded.

Furthermore we should not run cppcheck and scan-build if only profiles are changed and sort.py need only to be execute if profiles are changed.


Fred-Barclay (approved):
Sounds good to me!

glitsj16:
Very nice indeed. I had noticed things tend to take a long while for a simple profile change to get the green light.

reinerh:
Nice changes! I like it.

Does `github.event.commits[0].message` check the whole message or only the summary line (I noticed in some earlier commits you added it to the summary)?
(IMHO having this kind of metadata in the commit body looks nicer when reading the log)

reinerh (approved):

rusty-snake:
> Does github.event.commits[0].message check the whole message or only the summary line

AFAICTY it's the whole message (the docs aren't really helpfull).

> IMHO having this kind of metadata in the commit body looks nicer when reading the log

agree

PR #3791, opened 2020-12-07
Improvements to balsa,fractal,gajim,trojita

bbhtt:


etc/profile-a-l/gajim.profile
```diff
@@ -20,7 +20,6 @@ include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
-# Comment the following line if you need to whitelist other folders than ~/Downloads
```
rusty-snake, line 23: Why removing?

etc/profile-a-l/gajim.profile
```diff
@@ -47,9 +49,17 @@ shell none
 tracelog
 
 disable-mnt
-private-bin bash,gajim,gajim-history-manager,gpg,gpg2,paplay,python,python3,sh,zsh
+private-bin bash,gajim,gajim-history-manager,paplay,python,python3,sh,zsh
```
rusty-snake, line 52: Has gajim removed gpg support?

etc/profile-a-l/gajim.profile
```diff
@@ -47,9 +49,17 @@ shell none
...
+private-cache
 private-dev
-private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,group,hostname,hosts,ld.so.cache,ld.so.conf,localtime,machine-id,passwd,pki,pulse,resolv.conf,ssl
+private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,group,hostname,hosts,ld.so.cache,ld.so.conf,localtime,machine-id,passwd,pki,pulse,resolv.conf,selinux,ssl,xdg
```
rusty-snake, line 55: Are they needed?

etc/profile-m-z/trojita.profile
```diff
@@ -57,7 +57,10 @@ private-dev
 private-etc alternatives,ca-certificates,crypto-policies,fonts,hostname,hosts,pki,resolv.conf,selinux,ssl,xdg
 private-tmp
 
-dbus-user none
+dbus-user filter
+dbus-user.talk org.freedesktop.Notifications
```
rusty-snake, line 61: `org.freedesktop.Notifications` is insecure (allows sandbox escapes, code execution,  screenshots, keylogger, ...). It should be avoided if the program works w/o it and the profile is not complete open.

etc/profile-a-l/balsa.profile
```diff
@@ -58,9 +61,9 @@ shell none
 tracelog
 
 # disable-mnt
-# Add "gpg,gpg2,gpg-agent,pinentry-curses,pinentry-emacs,pinentry-fltk,pinentry-gnome3,pinentry-gtk,pinentry-gtk2,pinentry-gtk-2,pinentry-qt,pinentry-qt4,pinentry-tty,pinentry-x2go,pinentry-kwallet" for gpg 
+# Add "pinentry-curses,pinentry-emacs,pinentry-fltk,pinentry-gnome3,pinentry-gtk,pinentry-gtk2,pinentry-gtk-2,pinentry-qt,pinentry-qt4,pinentry-tty,pinentry-x2go,pinentry-kwallet" for gpg 
 # Add "ignore private-bin" for hyperlinks or have a look at the private-bins in firefox.profile and firefox-common.profile.
-private-bin balsa,balsa-ab
+private-bin balsa,balsa-ab,gpg,gpg-agent,gpg2,gpgsm
```
rusty-snake, line 66: Running the gpg-agent inside the sandbox is often a bad idea (and if, it will need pinetry I think).

etc/profile-a-l/gajim.profile
kortewegdevries: I'll add it back.

etc/profile-a-l/gajim.profile
kortewegdevries: No but it comes with only plugin installer addon and gpg is a separate addon that needs to be installed unless the distro packages it as a dependency; if we are supporting every addon (or the more important ones like two gpg addons,omemo,indicator support etc.) the profile needs to be loosened I think. Maybe we can make a separate gajim-addons.profile to resolve it?

etc/profile-a-l/gajim.profile
kortewegdevries: I saw it accessing xdg at one point; don't think it makes a difference on the user side unless I missed it.

etc/profile-m-z/trojita.profile
kortewegdevries: It works but won't display the notification on tray I think. 
I can remove it make it like this :

```
dbus-user none
# Uncomment both to allow keyring access 
# dbus-user filter
# Uncomment to use system keyring
# dbus-user.talk org.freedesktop.secrets
dbus-system none
```
Not sure how to resolve it; just don't want it to have a secrets access for no reason if the user chooses plain-text storage.

etc/profile-a-l/balsa.profile
kortewegdevries: I don't think signing or encryption will work since gpg-agent is the one that sets the homedir for gpg to source keys. Divided on the pinentry one. Do I list every binary in private-bin or keep it as is and the user can put the one for their system from the commented list?

etc/profile-a-l/gajim.profile
rusty-snake: Until now the most plugins in gajim where supported. If you want to create a gajim-addons-common.proile/.inc, go for it.

Theres a differences between addons in e.g. firefox and gajim or eog for example. 

Gajim (and eog and some other) have "core-plugins". Some features are implemented via plugins.

kortewegdevries: I'll see it later, if it is not too much of a change then I'll add them to gajim.profile. The current profile in master won't allow to use gpg because of no access to .gnupg etc.

kortewegdevries: I checked all of the plugins out the most problematic was gpg; added support for it; rest of them didn't need anything extra in the profile. And netlink socket is not needed.

rusty-snake (commented):
I'm still unsure about gajim.profile, the dbus-policy maybe break and `netlink` has likely a reason. All the other profiles look good.

Gajims flatpak dbus policy. (WTF why is there a rule for `org.freedesktop.portal.*`?)
```
[Session Bus Policy]
org.freedesktop.portal.Fcitx=talk
org.gnome.Mutter.IdleMonitor=talk
org.kde.kwalletd5=talk
org.mpris.MediaPlayer2.*=talk
org.freedesktop.secrets=talk

[System Bus Policy]
org.freedesktop.login1=talk
org.freedesktop.GeoClue2=talk
```

bbhtt:
> Gajims flatpak dbus policy.

Didn't realize it had a flatpak. I get the rest of them but not this one `org.freedesktop.login1=talk`. Maybe we should do it like gnome-calendar...

etc/profile-a-l/fractal.profile
```diff
@@ -8,6 +8,9 @@ include globals.local
 
 noblacklist ${HOME}/.cache/fractal
 
+include allow-python2.inc
+include allow-python3.inc
```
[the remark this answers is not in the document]
bbhtt: Hm yea, adding it.

etc/profile-a-l/gajim.profile
```diff
@@ -48,8 +57,25 @@ tracelog
...
 private-tmp
+writable-run-user
+
+dbus-user filter
+dbus-user.own org.gajim.Gajim
+dbus-user.talk org.gnome.Mutter.IdleMonitor
+dbus-user.talk ca.desrt.dconf
+# Uncomment for notifcation support
+# dbus-user.talk org.freedesktop.Notifications
+dbus-user.talk org.freedesktop.portal.Fcitx
```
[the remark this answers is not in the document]
bbhtt: This one is need for a plugin I think.

rusty-snake (approved):

rusty-snake:
Thanks for this contribution!

bbhtt:
I forgot to push the last commit.

Issue #3806, opened 2020-12-10
Refactor electron.profile and electron based programs


rusty-snake:
- [x] Allow chrome-sandbox

Add `caps.keep sys_admin,sys_chroot` to electron.profile and remove the following.
Additional add `#include chromium-common-hardened.inc` with a note.
```
caps.drop all
nonewprivs
noroot
protocol unix,inet,inet6,netlink
seccomp
```

- [x] Move `include whitelist-common.inc` to electron.profile 
- [x] Discover which command are common enough to be part of electron.profile. Consider these:
   - [x] `include disable-devel.inc`
   - [x] `include disable-exec.inc`
   - [x] `include disable-interpreters.inc`
   - [x] `include disable-xdg.inc`
   - [x] `shell none`
   - [x] ~`private-bin electron[0-9],electron[0-9][0-9]`~
   - [x] `nou2f`
   - [x] `novideo`
   - [x] `private-tmp`
   - [x] `include whitelist-var-common.inc`
   - [x] `include whitelist-usr-share-common.inc`
   - [x] `include whitelist-runuser-common.inc`
   - [x] `disable-mnt`
   - [x] `private-cache`
   - [x] `private-dev`
- [x] consistent `include globals.local`
- [x] Make all electron based programs, electron redirect profiles. Current redirect profiles are below. If you know any missing post it.

```
$ grep -l "include electron.profile" /etc/firejail/*.profile
/etc/firejail/beaker.profile
/etc/firejail/freetube.profile
/etc/firejail/jitsi-meet-desktop.profile
/etc/firejail/nuclear.profile
/etc/firejail/riot-web.profile
/etc/firejail/rocketchat.profile
/etc/firejail/teams-for-linux.profile
/etc/firejail/teams.profile
/etc/firejail/twitch.profile
/etc/firejail/whalebird.profile
/etc/firejail/wire-desktop.profile
/etc/firejail/youtubemusic-nativefier.profile
/etc/firejail/youtube.profile
/etc/firejail/ytmdesktop.profile
```

Needs update:
- [x] beaker.profile
- [x] freetube.profile
- [x] jitsi-meet-desktop.profile
- [x] nuclear.profile
- [x] riot-web.profile
- [x] rocketchat.profile
- [x] teams-for-linux.profile
- [x] teams.profile
- [x] twitch.profile
- [x] whalebird.profile
- [x] wire-desktop.profile
- [x] youtubemusic-nativefier.profile
- [x] youtube.profile
[...]

Issue #3586, opened 2020-08-14
element-desktop not starting


reinerh:
I received a [bug report](https://bugs.debian.org/968382) that element-desktop is not starting with firejail (0.9.62) and was able to reproduce it.
The problem (at least for me) was "nodbus" from electron.profile.
With "ignore nodbus" it's successfully starting.

With the new dbus changes in master I'm currently not sure what the best fix would be.
"ignore dbus-user" and "ignore dbus-system", or is there a better way to allow narrower dbus access?

kortewegdevries:
It launches fine under current profiles of 0.9.62.2-1 (Ubuntu groovy proposed) _without_ `ignore nodus`. 

This 

> ```
50:0814/095134.384213:FATAL:platform_shared_memory_region_posix.cc(254)] This is frequently caused by incorrect permissions on /dev/shm.  Try 'sudo chmod 1777 /dev/shm' to fix.
```
appears  only when `--no-sandbox` is used and even with `ignore nodbus` included in Element profile it doesn't launch with `--no-sandbox` (the screen stays white).

reinerh:
Sorry, I forgot to mention that `--no-sandbox` is used.

kortewegdevries:
#https://github.com/vector-im/element-web/issues/13171#issuecomment-650806459

reinerh:
@kortewegdevries I also see the white screen after starting.
But with `nodbus` it's not even fully starting up here (there is no tray icon). Did you observe the same?

kortewegdevries:
> @kortewegdevries I also see the white screen after starting.
> But with `nodbus` it's not even fully starting up here (there is no tray icon). Did you observe the same?

No in my case it starts in tray, doesn't expand to windowed/full screen until toggled from tray and stays white thereafter.

reinerh:
It might be related to my window manager, but with `nodbus` I don't get any tray icon (so I can't open the element window).

> (element-desktop:12): libappindicator-WARNING **: 10:20:24.734: Unable to get the session bus: Could not connect: No such file or directory
> 
> (element-desktop:12): LIBDBUSMENU-GLIB-WARNING **: 10:20:24.734: Unable to get session bus: Could not connect: No such file or directory

But maybe there is also a different issue, as the bug reporter had different messages.

kortewegdevries:
Those two messages appear for me regardless of `--no-sandbox` under firejail; I don't know if they are harmless or actually break some functionality (I don't use element), I'm under Xfce...

[...]

PR #3807, opened 2020-12-10
Refactor electron.profile and electron based programs

rusty-snake:
see  #3806


etc/profile-a-l/freetube.profile
```diff
@@ -6,26 +6,22 @@ include freetube.local
 # Persistent global definitions
 include globals.local
 
+# Disabled until someone reported positive feedback
```
glitsj16, line 9: I have been running freetube without issues having these in my freetube.local:

include whitelist-common.inc

nou2f
novideo

The nou2f is untested due to not having such hardware.

rusty-snake: If I'm right freetube has no login support at all.

etc/profile-m-z/riot-web.profile
```diff
@@ -4,14 +4,28 @@
 # Persistent local customizations
 include riot-web.local
 # Persistent global definitions
-# added by included profile
-#include globals.local
+include globals.local
+
+# Disabled until someone reported positive feedback
```
glitsj16, line 9: Tested with element-{desktop,web} from Arch community repo. Works fine with the below in riot-web.local:

```

# WIP 3807
include disable-devel.inc
# shell needed for tray icon
noblacklist ${PATH}/bash
noblacklist ${PATH}/sh
ignore noexec /tmp
whitelist /usr/share/chromium
whitelist /usr/share/webapps/element
include disable-exec.inc
include disable-xdg.inc
nou2f
novideo
# shell needed for tray icon
#shell none
disable-mnt
private-cache
private-dev
private-tmp
```

rusty-snake: two questions
- electron does not `include disable-shell.inc`, is it in one of your locals?
- `shell none` just mean start direct and w/o a shell, but a shell is still present inside the sandbox. Does it really need to be ignored?

glitsj16: > electron does not include disable-shell.inc, is it in one of your locals?

It isn't, nice catch. The relatively long include chain (element-desktop --> riot-desktop --> riot-web --> electron) threw me off here apparently.

> shell none just mean start direct and w/o a shell, but a shell is still present inside the sandbox. Does it really need to be ignored?

Ditto as above. I suspect it was a leftover from testing permutations. I can now confirm adding 'shell none' works fine.

To put things straight, element-desktop is working with the below in my riot-web.local:

# WIP 3807
include disable-devel.inc
ignore noexec /tmp
whitelist /usr/share/chromium
whitelist /usr/share/webapps/element
include disable-exec.inc
include disable-xdg.inc
nou2f
novideo
shell none
disable-mnt
private-cache
private-dev
private-tmp

mYnDstrEAm:
Should it be sufficient to change /etc/firejail/electron.profile riot-web.profile and riot-desktop.profile to get element-desktop working with the latest firejail version in backports in Debian 10 (0.9.64)? I changed these 3 files as well as element-desktop.profile to match the latest versions in this repo but now I get this error when running `firejail element-desktop`:
```
Child process initialized in 1135.90 ms
The setuid sandbox is not running as root. Common causes:
  * An unprivileged process using ptrace on it, like a debugger.
  * A parent process set prctl(PR_SET_NO_NEW_PRIVS, ...)

```

rusty-snake:
Using profiles from git master with a stable release (even the latest) is nothing what you should expect to work. Anyway _ATM_ it works to use git master profiles in 0.9.64.

debian + element-desktop -> #3586.

PR #3808, opened 2020-12-11
integrate relevant options into server.profile

glitsj16:
Try to uplift our server.profile by bringing in relevant options that got introduced recently. As an experiment in hardening a few options are flipped on by default:

- include disable-write-mnt.inc
- include disable-xdg.inc
- apparmor
- machine-id
- disable-mnt
- dbus-user none

### DIFF
--- a/etc/profile-m-z/server.profile
+++ b/etc/profile-m-z/server.profile
@@ -45,10 +45,17 @@ include disable-common.inc
 # include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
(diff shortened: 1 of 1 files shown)


rusty-snake (commented):
:rocket: LGTM

Thoughts:

- mdwe break interpreter languages (python, perl, php, ...).
- `dbus-system none` is maybe to strict.

glitsj16:
@rusty-snake Thanks for the suggestions, exactly the kind of feedback I was looking for.

PR #3819, opened 2020-12-14
rename whitelist-players.inc to whitelist-player-common.inc

glitsj16:
whitelist-player-common.inc fits in with the naming-scheme used throughout our codebase

IMPORTANT NOTE:
users will have to rename their whitelist-players.local to whitelist-player-common.local to ensure stuff stays working as expected

### DIFF
--- a/etc/inc/whitelist-player-common.inc
+++ b/etc/inc/whitelist-player-common.inc
@@ -1,6 +1,6 @@
 # This file is overwritten during software install.
 # Persistent customizations should go in a .local file.
-include whitelist-players.local
+include whitelist-player-common.local
 
 # common whitelist for all media players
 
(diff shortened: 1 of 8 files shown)


rusty-snake:
OT: We could also rename `softmaker-common.inc` to `softmaker-common.profile`. We use `*-common.profile` for commons based on same code-base and `*-common.inc` for commons based on same usage target (e.g. wusc or players).

glitsj16:
> OT: We could also rename softmaker-common.inc to softmaker-common.profile. We use *-common.profile for commons based on same code-base and *-common.inc for commons based on same usage target (e.g. wusc or players).

@rusty-snake That's a great idea, thanks for pointing it out. Adding it to the to-do list, unless you beat me to it :-)

PR #3820, opened 2020-12-14
Refactor archivers

glitsj16:
Introduce `archive-common.inc`: common file for all archivers.
For now this WIP simply recreates existing archiver profiles.
It's intention is to:
- simplify adding new archiver profile(s);
- review potential inconsistencies in existing archiver profile(s).
E.g. most have nogroups and have noroot commented, but a few divert from this pattern. I intend to do a follow-up review after extended testing, so for now this is a WIP.

### DIFF
--- a/etc/inc/archiver-common.inc
+++ b/etc/inc/archiver-common.inc
@@ -4,7 +4,6 @@ include archiver-common.local
 
 # common profile for archiver/compression tools
 
-blacklist ${RUNUSER}/wayland-*
 blacklist ${RUNUSER}
 
 include disable-common.inc
--- a/etc/inc/archiver-common.inc
+++ b/etc/inc/archiver-common.inc
@@ -2,7 +2,7 @@
 # Persistent customizations should go in a .local file.
 include archiver-common.local
 
-# common settings for all archiver/compression tool profiles
+# common profile for archiver/compression tools
 
 blacklist ${RUNUSER}/wayland-*
 
--- a/etc/inc/archiver-common.inc
+++ b/etc/inc/archiver-common.inc
@@ -5,6 +5,7 @@ include archiver-common.local
 # common profile for archiver/compression tools
 
 blacklist ${RUNUSER}/wayland-*
+blacklist ${RUNUSER}
 
 include disable-common.inc
 include disable-devel.inc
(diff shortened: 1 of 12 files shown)


etc/inc/archiver-common.inc
```diff
@@ -0,0 +1,44 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+include archiver-common.local
+
+# common profile for archiver/compression tools
+
+blacklist ${RUNUSER}/wayland-*
```
rusty-snake, line 7: I think we can `blacklist ${RUNUSER}`.

rusty-snake (commented):
LGTM

etc/inc/archiver-common.inc
glitsj16: @rusty-snake Thanks for having a look at this draft. I'm in the middle of testing all my archivers including their .local overrides. Will add `blacklist ${RUNUSER}` to the tests but indeed, I agree this shouldn't pose any problems. The plan is to bring existing profiles in an easier-to-change state using archive-common.inc as baseline and do specific changes in seperate PR's once this is merged. Shouldn't take all that long to get the first part finished.

glitsj16: @rusty-snake I added `blacklist ${RUNUSER}` to archiver-common.inc. What's the point in having that in combination with `blacklist ${RUNUSER}/wayland-*` again? I see such a doubling in several other profiles (curl, dig, drill, file, wget to name a few) but I can't say I actually understand the why. If ${RUNUSER} is blacklisted, shouldn't that imply ${RUNUSER}/wayland-* (and any other subdir for that matter) is blacklisted too?

rusty-snake: `blacklist ${RUNUSER}/wayland-*` is useless with `blacklist ${RUNUSER}` and can be simplified, that's right.

The profiles are written with this logic:

```
If not requires(X11) then
    If 'net none' then
        add 'x11 none'
    Else
        add 'blacklist /tmp/.X11-unix'
    Fi
Fi

If 'x11 none' or 'blacklist /tmp/.X11-unix' then
    add 'blacklist ${RUNUSER}/wayland-*'
Fi

If possible_to_add(wruc) then
    add wruc
Fi

If get_used_commands(wruc) == nil then
    replace wruc with 'blacklist ${RUNUSER}'
Fi
```

glitsj16: @rusty-snake Thanks for the explanation! I'll do the needed changes in a minute, before I forget. It sure is a luxury to have you on board :-)

PR #3821, opened 2020-12-15
Fix sound in games using FMOD

fenuks:
I've noticed that profile template has `alsa` in `private-etc`, and `steam.profile` does not, which breaks sound in games using FMOD (Pillars of Eternity in my case, but it should also fix problems mentioned in #3508).


rusty-snake:
Thanks for this contribution!

fenuks:
That was a trivial fix, but thank you!

PR #3810, opened 2020-12-11
Dc add ldns

kmk3:
disable-common.inc: blacklist ldns tools

drill(1) from ldns is the first tool suggested on the Arch Wiki for DNS
lookup:
https://wiki.archlinux.org/index.php/Domain_name_resolution#Lookup_utilities

Home page: https://www.nlnetlabs.nl/projects/ldns/about/

    $ pacman -Q ldns
    ldns 1.7.1-2
    $ pacman -Qlq ldns | grep bin
    /usr/bin/
    /usr/bin/drill
    /usr/bin/ldns-chaos
    /usr/bin/ldns-compare-zones
    /usr/bin/ldns-config
    /usr/bin/ldns-dane
    /usr/bin/ldns-dpa
    /usr/bin/ldns-gen-zone
    /usr/bin/ldns-key2ds
    /usr/bin/ldns-keyfetcher
    /usr/bin/ldns-keygen
    /usr/bin/ldns-mx
    /usr/bin/ldns-notify
    /usr/bin/ldns-nsec3-hash
    /usr/bin/ldns-read-zone
    /usr/bin/ldns-resolver
    /usr/bin/ldns-revoke
    /usr/bin/ldns-rrsig
    /usr/bin/ldns-signzone
    /usr/bin/ldns-test-edns
    /usr/bin/ldns-testns
    /usr/bin/ldns-update
    /usr/bin/ldns-verify-zone
    /usr/bin/ldns-version
    /usr/bin/ldns-walk
    /usr/bin/ldns-zcat
    /usr/bin/ldns-zsplit
    /usr/bin/ldnsd

PR #3828, opened 2020-12-15
Dc add dns

kmk3:
* disable-common.inc: split bind and knot tools
* disable-common.inc: blacklist missing bind tools
* disable-common.inc: blacklist missing knot tools
* disable-common.inc: blacklist unbound tools
* dnscrypt-proxy.profile: fix section structure
* disable-common.inc: add misc dns tools

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -517,16 +517,19 @@ blacklist /proc/config.gz
 blacklist ${PATH}/dig
 blacklist ${PATH}/dlint
 blacklist ${PATH}/dns2tcp
+blacklist ${PATH}/dnssec-*
 blacklist ${PATH}/dnswalk
 blacklist ${PATH}/drill
 blacklist ${PATH}/host
 blacklist ${PATH}/iodine
 blacklist ${PATH}/kdig
+blacklist ${PATH}/khost
 blacklist ${PATH}/knsupdate
 blacklist ${PATH}/ldns-*
 blacklist ${PATH}/ldnsd
 blacklist ${PATH}/nslookup
 blacklist ${PATH}/resolvectl
+blacklist ${PATH}/unbound-host
 
 # rest of ${RUNUSER}
 blacklist ${RUNUSER}/*.lock


kmk3:
I'm unsure about the latest commit though; I don't know if these are used
elsewhere:

    $ find etc -name '*dns*' | sort
    etc/profile-a-l/dnscrypt-proxy.profile
    etc/profile-a-l/dnsmasq.profile
    etc/profile-a-l/fdns.profile

etc/inc/disable-common.inc
```diff
@@ -514,19 +514,49 @@ blacklist /proc/config.gz
...
+blacklist ${PATH}/host
+blacklist ${PATH}/mdig
+blacklist ${PATH}/named
+blacklist ${PATH}/named-*
+blacklist ${PATH}/nsec3hash
+blacklist ${PATH}/nslookup
+blacklist ${PATH}/nsupdate
+blacklist ${PATH}/rndc
+blacklist ${PATH}/rndc-confgen
+blacklist ${PATH}/tsig-keygen
```
reinerh, line 532: are files like tsig-keygen and nsec3hash (and couple of others mentioned below) really used for communicating with DNS servers?
and are they commonly used? (at least I don't have them installed on my system)

i think the idea of the file is not to blacklist any binary in existence, but the ones commonly installed and where it makes sense to have them blacklisted.

what is your opinion @rusty-snake?

kmk3: >     +blacklist ${PATH}/nsupdate
>     +blacklist ${PATH}/rndc
>     +blacklist ${PATH}/rndc-confgen
>     +blacklist ${PATH}/tsig-keygen

> are files like tsig-keygen and nsec3hash (and couple of others mentioned
> below) really used for communicating with DNS servers?

Looking at the man pages for `tsig-keygen` / `nsec3hash`, they do look safe
indeed.  But looking at the one for `nsupdate`, it is not obvious to me that it
cannot communicate with DNS servers.

Do you know which ones can be safely ignored?  Should I try to come up with a
list?

> and are they commonly used? (at least I don't have them installed on my
> system)

Well, they are part of `bind`.  I didn't have it installed prior to this PR
either.  But if there exists a section for DNS tools, then wouldn't leaving it
incomplete give a false sense of security?  Alternatively, how about moving the
section to a new `disable-dns.inc`?  I had considered doing the latter, but I
thought that it would be a too invasive change for this PR.  I can still do it
if it makes sense.

> i think the idea of the file is not to blacklist any binary in existence, but
> the ones commonly installed and where it makes sense to have them
> blacklisted.

For reference, I just followed what was said on
https://github.com/netblue30/firejail/pull/3810#issuecomment-742962106:

@glitsj16 commented 5 days ago

> > Would it be better to check each program or to merely
> > blacklist all binaries for safety/simplicity?
> 
> For now it might be a good idea to keep things simple and just add the
> suggested binaries to disable-common.inc under the "# prevent DNS malware
> attempting to communicate with the server..." section. We can always add
> `noblacklist ${PATH}/foo` to relevant profiles later on. At first glance this
> shouldn't break anything badly. But I'll do some more extended testing after
> the weekend.

glitsj16:
> i think the idea of the file is not to blacklist any binary in existence, but the ones commonly installed and where it makes sense to have them blacklisted.

I fully agree. There seems to be a bit of a misunderstanding regarding my [earlier comment](https://github.com/netblue30/firejail/pull/3810#issuecomment-742920539). To be clear, I was only refering to `khost`, `unbound-host` and the `dnssec-*` tools from bind, nothing else. Whether or not we should have a dedicated disable-dns.inc is better left for a seperate discussion / PR IMHO. Apologies for the potential confusion.

kmk3:
> > i think the idea of the file is not to blacklist any binary in existence,
> > but the ones commonly installed and where it makes sense to have them
> > blacklisted.
> 
> I fully agree. There seems to be a bit of a misunderstanding regarding my
> [earlier
> comment](https://github.com/netblue30/firejail/pull/3810#issuecomment-742920539).
> To be clear, I was only refering to `khost`, `unbound-host` and the
> `dnssec-*` tools from bind, nothing else. Whether or not we should have a
> dedicated disable-dns.inc is better left for a seperate discussion / PR IMHO.
> Apologies for the potential confusion.

Ah, I see now.  When I read "the other executables mentioned on that Arch wiki"
I started looking at the packages and thought that the ones you mentioned were
only the most obvious examples that you noticed that were missing, not the full
list.  I suppose I got carried away, sorry about that.

I have force-pushed adding only these 3 entries.  Also, I'm not sure if `ldnsd`
can communicate with DNS servers.  If not, should I remove it?

glitsj16 (approved):

PR #3829, opened 2020-12-16
disable-shell.inc: add oksh

kmk3:
"Portable OpenBSD ksh, based on the Public Domain Korn Shell (pdksh)."

Project page: https://github.com/ibara/oksh

    $ pacman -Q oksh
    oksh 6.8.1-1
    $ pacman -Qlq oksh | grep bin/
    /usr/bin/
    /usr/bin/oksh

### DIFF
--- a/etc/inc/disable-shell.inc
+++ b/etc/inc/disable-shell.inc
@@ -8,6 +8,7 @@ blacklist ${PATH}/dash
 blacklist ${PATH}/fish
 blacklist ${PATH}/ksh
 blacklist ${PATH}/mksh
+blacklist ${PATH}/oksh
 blacklist ${PATH}/sh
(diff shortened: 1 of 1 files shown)


rusty-snake:
Thanks for this contribution!

kmk3:
Hmm I noticed that GitHub butchered the author name and author date of the
original commit.

@rusty-snake Did you use rebase-merge or something like that?  I think I'll
test later what option causes this.

P.S.: It also set the committer name to simply "GitHub".  IMO ideally it would
be set to whoever merged it, like in a normal non-fast-forward merge, though
that's secondary.

rusty-snake:
> Did you use rebase-merge or something like that?

"squash and merge"

PR #3832, opened 2020-12-16
Archiver fixes - drop private-bin

glitsj16:
Archivers should be able to compress binaries (see discussion in https://github.com/netblue30/firejail/commit/95ad89d24e8e75f2f52defbb80d0d4ee0f27d31e). Let's drop private-bin in relevant profiles.

### DIFF
--- a/etc/profile-a-l/7z.profile
+++ b/etc/profile-a-l/7z.profile
@@ -10,5 +10,3 @@ include globals.local
 noblacklist ${PATH}/bash
 noblacklist ${PATH}/sh
 include archiver-common.inc
-
-private-bin 7z,7z*,bash,p7zip,sh
--- a/etc/profile-a-l/ar.profile
+++ b/etc/profile-a-l/ar.profile
@@ -8,5 +8,3 @@ include ar.local
 include globals.local
 
 include archiver-common.inc
-
-private-bin ar
--- a/etc/profile-a-l/bsdtar.profile
+++ b/etc/profile-a-l/bsdtar.profile
@@ -8,6 +8,4 @@ include globals.local
 
 include archiver-common.inc
 
-# support compressed archives
-private-bin bash,bsdcat,bsdcpio,bsdtar,bzip2,compress,gtar,gzip,lbzip2,libarchive,lz4,lzip,lzma,lzop,sh,xz
 private-etc alternatives,group,localtime,passwd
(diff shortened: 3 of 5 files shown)


glitsj16:
@reinerh I seem to have broken a test again. These are my first baby-steps in the GitHub UI trying to find what broke exactly, so I can fix it. Getting nowhere though. If you could spare the time to offer some insights/info on how to proceed without breakin even more stuff, that would be very much appreciated.

reinerh:
@glitsj16 Failing tests are found by grepping the test logs (output) for the string "TESTING ERROR".
So to locate the errors, I use the search bar on the right side to search for "TESTING ERROR" (see screenshot).
github

In this case this error can be found:
```
TESTING: tar
spawn /bin/bash
firejail /bin/tar -cjvf firejail_t2 /usr/share/doc/firejail
runner@fv-az137-563:~/work/firejail/firejail/test/sysutils$ 
< /bin/tar -cjvf firejail_t2 /usr/share/doc/firejail        
/bin/tar: Removing leading `/' from member names
/usr/share/doc/firejail/
/usr/share/doc/firejail/README
/usr/share/doc/firejail/syscalls.txt
/usr/share/doc/firejail/profile.template
/usr/share/doc/firejail/redirect_alias-profile.template
/usr/share/doc/firejail/COPYING
bzip2: error while loading shared libraries: libbz2.so.1.0: cannot open shared object file: No such file or directory
/bin/tar: firejail_t2: Wrote only 4096 of 10240 bytes
/bin/tar: Child returned status 127
/bin/tar: Error is not recoverable: exiting now
runner@fv-az137-563:~/work/firejail/firejail/test/sysutils$ 
<test/sysutils$ stat -c '|%s|' firejail_t2; uname -s        
|0|
TESTING ERROR 2.2
```

So it looks like it can't access `libbz2.so.1.0`.

reinerh:
I *think* the reason for the failure is that you removed private-bin. private-bin also copies the shared libraries into the jail to make sure the binaries can be run. But without private-bin but still having private-lib, the libs needed by some of the compression tools are not available.

glitsj16:
> I think the reason for the failure is that you removed private-bin. private-bin also copies the shared libraries into the jail to make sure the binaries can be run. But without private-bin but still having private-lib, the libs needed by some of the compression tools are not available.

@reinerh That makes perfect sense. Thanks for your speedy reply and the explanation!

The more I look at the archivers the less sense existing profiles make. I'm starting to wonder why we restrict access to the filesystem at all, as this would defeat the purpose of creating an archive of (parts of) it from a user standpoint. Realizing it would loosen them up IMO we should drop private-lib too where existing. I'll open an issue on iy shortly to collect input, thoughts and opinions on this topic.

Issue #3833, opened 2020-12-17
archiver profiles and limiting file system access


glitsj16:
While working on our archiver profiles I started to notice a few things I don't fully understand, or that don't seem to make sense IMO. This all started when I broke a test in a [recent PR](https://github.com/netblue30/firejail/commit/95ad89d24e8e75f2f52defbb80d0d4ee0f27d31e) which prompted @reinerh to pose the question 'why is it not allowed to archive binaries?'. This should be possible - archiving binaries is a legitimate use-case IMO. So I started to drop `private-bin` from our archivers in #3832.

That triggered the more generic question: `why do we impose file system limits in archiver profiles to begin with?` Isn't doing so going to break (legitimate) user expectations when archiving stuff for a quick backup of their ${HOME}/.config for example? And what about backing-up /etc?

Realizing it may be a bit over-the-top, I want to propose to drop a few options from the archiver tools to avoid such scenario's:

- include disable-foo.inc
- private-bin
- private-etc
- private-lib

All input, thoughts, opinions etcetera on this topic is obviously wanted and very welcome.

rusty-snake:
We can do it like we did with meld. Meld has a legitimate use-case on the most parts of the file-system too.

glitsj16:
I created a preliminary PR implementing the suggested meld approach. Also added a note to explain the difference between loosening/tightening **all** archivers at once via archiver-common.local OR doing so **per archiver** in the relevant <foo>.local. Let's keep this open for a while so other people can voice their opinions.

kmk3:
glitsj16 commented 11 days ago

> That triggered the more generic question: `why do we impose file system
> limits in archiver profiles to begin with?` Isn't doing so going to break
> (legitimate) user expectations when archiving stuff for a quick backup of
> their ${HOME}/.config for example?

Indeed, and considering that this is the default profile and not a hardened
alternative, I personally would only block things that aren't meant to be read
directly, like character devices in /dev.  I have archived `${HOME}/.config`
and what not many times and having something like that fail without me noticing
(e.g.: due to a `whitelist`) could mean loss of data.

Ideally, it would be at least as simple as adding `ignore whitelist *` +
`ignore blacklist *` to the relevant profiles.  I was thinking about a similar
problem this week: if there was a profile for `ls` enabled by default, any
blacklist/whitelist would result in subtly wrong output being produced (like
`x11 none` == `~/.Xauthority` appears to be owned by root), which could result
[...]

PR #3834, opened 2020-12-17
archivers: limiting file system access

glitsj16:
Preliminary work based on discussion in #3833. As suggested there we take a similar approach as in our meld profile.


rusty-snake (commented):
LGTM

Maybe a note that not only your workflow, but also other programs can be broken by to tight profiles. example: Just because I only need to (un)compress files in DOWNLOADS, some programs may still need to (un)compress files in .local/share or something else.

PR #3839, opened 2020-12-19
x11=none: don't fail on abstract socket if netns …

### LINKED ISSUE #3838: --x11=none --netns=isolated invalidly errors on the abstract X11 socket being accessible
**Scenario + use case**

I'm want to run a native Wayland program under firejail. Since it uses Wayland only I want to disable all X11 access so it can't possibly talk to XWayland to get information about running XWayland programs. My sandbox is using a separate network namespace so it can't reach other localhost sockets in any way and is also forced out via a VPN.

**Problem**

```
$ firejail --noprofile --netns=isolated --x11=none
ERROR: --x11=none specified, but abstract X11 socket still accessible.
Additional setup required. To block abstract X11 socket you can either:
 * use network namespace in firejail (--net=none, --net=...)
 * add "-nolisten local" to xserver options
   (eg. to your display manager config, or /etc/X11/xinit/xserverrc)
```

But since I'm assigning it a separate namespace the abstract socket **is not accessible**:
```
$ firejail --noprofile --netns=isolated
$ netstat -a | grep X11
$
```

**Expected outcome**

When `--netns=something` is given, `--x11=none` should not complain about the abstract X11 socket being accessible to the jail.

**Environment**
Fedora 33 Workstation
```
$ firejail --version
firejail version 0.9.64

Compile time support:
	- AppImage support is enabled
	- chroot support is enabled
	- SELinux support is enabled
```

**Checklist**
 - [x] A short search for duplicates was performed.

rusty-snake:
…is used.

fix #3838 -- --x11=none --netns=isolated invalidly errors on the abstract X11 socket being accessible

### DIFF
--- a/src/firejail/x11.c
+++ b/src/firejail/x11.c
@@ -1368,7 +1368,7 @@ void fs_x11(void) {
 void x11_block(void) {
 #ifdef HAVE_X11
 	// check abstract socket presence and network namespace options
-	if ((!arg_nonetwork && !cfg.bridge0.configured && !cfg.interface0.configured)
+	if ((!arg_nonetwork && !arg_netns && !cfg.bridge0.configured && !cfg.interface0.configured)
 	&& x11_abstract_sockets_present()) {
 		fprintf(stderr, "ERROR: --x11=none specified, but abstract X11 socket still accessible.\n"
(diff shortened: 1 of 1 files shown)


glitsj16 (approved):
LGTM

smitsohu:
Merged!

PR #3847, opened 2020-12-28
Small fixes

bbhtt:


rusty-snake (commented):
- Without `quiet` these programs break (have ugly output) I assume.
- Are these D-Bus permissions really needed by default? Or can we add "uncumment if you need foo".

bbhtt:
> Without quiet these programs break (have ugly output) I assume.

The ugly output is an issue when called from a script or other terminal programs such as file managers.

> Are these D-Bus permissions really needed by default? Or can we add "uncomment if you need foo".

Element is needed, it stores an access token in secret storage. I'm commenting the other too since they are only for notification popups.

rusty-snake (approved):

PR #3848, opened 2020-12-28
Add profiles for MS Edge dev build for Linux and Librewolf

bbhtt:


rusty-snake (approved):

bbhtt:
I'm setting an empty redirect right now; when MS releases the stable build of Edge, the microsoft-edge.profile needs to be changed. Right now they have a microsoft-dev build with `.../bin/microsoft-edge` and `.../bin/microsoft-edge-dev`. I assume the former will point to the stable release when it is released.

Issue #3527, opened 2020-07-19
[Question] SkypeforLinux - or General .deb Security


svc88:
This might be considered a noob question here.
If i use firejail to secure skypeforlinux (or any proprietary .deb file), whats the point to secure it, if its already installed in your system?
Once you install that .deb file, isnt there a slight possibility that any one of those files can "self-execute" or rather skypeforlinux install a backdoor as a service or some other way that in essence firejail will not evidently help before even running the actual app?
How do you know its actually secured when app isnt running?

In this case, are appimages more secure than .deb files to be used with Firejail?  (If yes, then i havent found any appimage files for skypeforlinux)

rusty-snake:
A .deb can execute scripts while (un)installing, override system binaries, install autostart and systemd-unit files, install suids, ...

TLTR: Do install software you don't trust. Nothing (technical) can help.

svc88:
@rusty-snake so would appimages be preferred to be used with Firejail if available (from a security standpoint) ?

rusty-snake:
Then the part with the installing is gone, but if you don't trust it it (i.e. you think it has a backdoor) it can escape the sandbox if it want.

svc88:
If the appimage has a backdoor and you run it in a tight sandbox firejail profile, can it **really** escape the sandbox?

rusty-snake:
It can use
- D-Bus if not blocked (as it should be in a tight sandbox)
- various things like `~/.bashrc` (mitigation: `include disable-common.inc`, protection: `whitelist ${HOME}/foo`, `read-only ${HOME}`, `private-home foo,bar`, `private`, `private foobar`)
- talk to other abstract unix sockets (protection: `net none` (break internet) or `net eth0` (very unhandy)
- use X11 for bad things
- talk to other TCP-Services on 127.0.0.1
- talk to sockets in /run/quux (mitigation: find all and blacklist them)
- talk/edit/touch/… in /tmp (protection: `private-tmp`)
- …
- some others I don't know about

Or short: if the program you run in the sandbox wants to escape, it can escape.

If you don't trust the software you run, use a VM (and even a VM can be escaped).

rusty-snake:
[...]

Issue #3447, opened 2020-06-02
Replace `whitelist` and `blacklist` commands with better terms


ClemMakesApps:
Since `Whitelist`/`Blacklist` have connotations about value and map to racial terms, we should replace those commands in this project with `Allowlist` and `Denylist` or something similar.

netblue30:
Are you kidding?

Fred-Barclay:
@ClemMakesApps I appreciate the concern but let's don't go down this path. Our terminology is based upon the behavior/functionality of the code, nothing more, and reflect long-standing terminology and usage of the coding community. There's no need to change things which are intrinsically neutral based upon non-existent connotations to the ratio of melanin in people's skin.

NicoleSchwartz:
@Fred-Barclay It doesn't matter what the original intent was. People do have a negative reaction to these terms. There is no justification for keeping it. 

Over time, although you may intend them neutrally they do have negative impact. Many projects and groups have moved to the equally descriptive terms that avoid making people uncomfortable but are still equally descriptive. see https://github.com/rails/rails/issues/33677

"To compound the issue, it is also striking how often the term “whitelist” is used for a supposedly good, respectable, or safe list of publishers [20, 22, 27]. The racism in such “black is bad, white is good” metaphors is inappropriate and needs to cease. The black-white dualism explicit in these binary terms is often associated with Western thinking that is usually traced back to the work of Rene Descartes. Although the epistemological dualism of Descartes may be seen in earlier works by Plato and Aristotle, this way of thinking is often associated with the Enlightenment and the subsequent scientific revolution and industrial development [28–30]. Thus, a foundational ontological dualism accepted by many people in Western cultures includes the supposedly “natural” divides between subject-object, body-spirit, human-nature, and self-other. Such dualism extends into our conceptions of good-evil, sacred/divine-profane, and civilized-heathen/barbarian [31]." https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6148600/

glitsj16:
> People do have a negative reaction to these terms. There is no justification for keeping it.

Some do and some don't, it's all in the 'Eye Of The Beholder'. There's no justification for not keeping it either, if this is nothing more than a plea to change a few words and call it quits. The conclusions drawn, and the subsequent (disputable) history lesson on ontological dualism feel out of place. Quoting research examining the emerging literature surrounding `predatory publishing` might buy one some 'social capital' (to quote another famous French male theorist), but what good does that bring besides a few likes?

[...]

Issue #318, opened 2016-02-23
firejail is losing shell special characters in args


dionorgua:
Imagine that we've firefox that can be invoked like:
`firefox "http://example.com/test?a=b&c=d"`
Now we decided to isolate it:
`firefox --join=firefox -- firefox "http://example.com/test?a=b&c=d"`
This will open url without c=d parameter.  

And even worse. If user puts 'firejail' wrapper
This opens a way to execute some shell commands in jail like:
"http://example.com/test?a=b&ls -la /"

netblue30:
Yes, there seem to be a problem with --join command. I'll bring a fix in. Thanks.

netblue30:
Fixed! Give it a try and I'll reopen the bug if necessary.

PR #3849, opened 2020-12-28
Email part (2)

bbhtt:
Continuation of https://github.com/netblue30/firejail/pull/3607

Mutt/Neomutt needs some testing I may not have went through all use cases, leaving a (neo)muttrc if anyone wants to do some testing https://termbin.com/tryb

### DIFF
--- a/etc/profile-a-l/email-common.profile
+++ b/etc/profile-a-l/email-common.profile
@@ -71,6 +71,8 @@ private-tmp
 # encrypting and signing email
 writable-run-user
 
+dbus-system none
+
 # If you want to read local mail stored in /var/mail, add the following to email-common.local:
 #noblacklist /var/mail
 #noblacklist /var/spool/mail
--- a/etc/profile-a-l/email-common.profile
+++ b/etc/profile-a-l/email-common.profile
@@ -35,7 +35,6 @@ whitelist ${DOCUMENTS}
 whitelist ${DOWNLOADS}
 # when storing mail outside the default ${HOME}/Mail path, 'whitelist' the custom path in your email-common.local
 whitelist ${HOME}/Mail
-
 whitelist ${RUNUSER}/gnupg
 whitelist /usr/share/gnupg
 whitelist /usr/share/gnupg2
@@ -68,7 +67,6 @@ private-cache
 private-dev
 private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,gcrypt,gnupg,groups,gtk-2.0,gtk-3.0,hostname,hosts,hosts.conf,mailname,nsswitch.conf,passwd,pki,resolv.conf,selinux,ssl,xdg
 private-tmp
-
 # encrypting and signing email
 writable-run-user
 
--- a/etc/profile-a-l/email-common.profile
+++ b/etc/profile-a-l/email-common.profile
@@ -24,9 +24,9 @@ include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-xdg.inc
 
+mkdir ${HOME}/.gnupg
 mkfile ${HOME}/.config/mimeapps.list
 mkfile ${HOME}/.signature
-mkdir ${HOME}/.gnupg
 whitelist ${HOME}/.config/mimeapps.list
 whitelist ${HOME}/.mozilla/firefox/profiles.ini
 whitelist ${HOME}/.gnupg
@@ -59,6 +59,7 @@ nou2f
 novideo
 protocol unix,inet,inet6
 seccomp
+seccomp.block-secondary
 shell none
 tracelog
 
--- a/etc/profile-a-l/email-common.profile
+++ b/etc/profile-a-l/email-common.profile
@@ -8,6 +8,7 @@ include email-common.local
 #include globals.local
 
 noblacklist ${HOME}/.gnupg
+noblacklist ${HOME}/.mozilla
 noblacklist ${HOME}/.signature
 # when storing mail outside the default ${HOME}/Mail path, 'noblacklist' the custom path in your email-common.local
 # and 'blacklist' it in your disable-common.local too so it is  kept hidden from other applications
@@ -17,28 +18,35 @@ noblacklist ${DOCUMENTS}
 
 include disable-common.inc
 include disable-devel.inc
+include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-xdg.inc
 
-whitelist ${DOCUMENTS}
-whitelist ${DOWNLOADS}
 mkfile ${HOME}/.config/mimeapps.list
-mkdir ${HOME}/.gnupg
 mkfile ${HOME}/.signature
+mkdir ${HOME}/.gnupg
 whitelist ${HOME}/.config/mimeapps.list
+whitelist ${HOME}/.mozilla/firefox/profiles.ini
 whitelist ${HOME}/.gnupg
 whitelist ${HOME}/.signature
+whitelist ${DOCUMENTS}
+whitelist ${DOWNLOADS}
 # when storing mail outside the default ${HOME}/Mail path, 'whitelist' the custom path in your email-common.local
 whitelist ${HOME}/Mail
+
+whitelist ${RUNUSER}/gnupg
 whitelist /usr/share/gnupg
 whitelist /usr/share/gnupg2
 include whitelist-common.inc
+include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
+apparmor
 caps.drop all
+machine-id
 netfilter
 no3d
 nodvd
@@ -54,12 +62,11 @@ seccomp
 shell none
 tracelog
 
+# disable-mnt
 private-cache
 private-dev
+private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,gcrypt,gnupg,groups,gtk-2.0,gtk-3.0,hostname,hosts,hosts.conf,mailname,nsswitch.conf,passwd,pki,resolv.conf,selinux,ssl,xdg
 private-tmp
-
-dbus-user none
-dbus-system none
 
 # encrypting and signing email
 writable-run-user
@@ -70,3 +77,6 @@ writable-run-user
 #whitelist /var/mail
 #whitelist /var/spool/mail
 #writable-var
+
+read-only ${HOME}/.mozilla/firefox/profiles.ini
+read-only ${HOME}/.signature
--- a/etc/profile-a-l/geary.profile
+++ b/etc/profile-a-l/geary.profile
@@ -66,11 +66,11 @@ shell none
 tracelog
 
 # disable-mnt
-# Add ignore private-bin to geary.local for hyperlink support
+# Add 'ignore private-bin' to geary.local for hyperlink support
 private-bin geary
 private-cache
 private-dev
-private-etc alternatives,ca-certificates,crypto-policies,fonts,hostname,hosts,pki,resolv.conf,selinux,ssl,xdg
+private-etc alternatives,ca-certificates,crypto-policies,fonts,hostname,hosts,pki,resolv.conf,ssl,xdg
 private-tmp
 
 dbus-user filter
--- a/etc/profile-a-l/geary.profile
+++ b/etc/profile-a-l/geary.profile
@@ -6,8 +6,12 @@ include geary.local
 # Persistent global definitions
 include globals.local
 
+noblacklist ${HOME}/.cache/evolution
+noblacklist ${HOME}/.cache/folks
 noblacklist ${HOME}/.cache/geary
+noblacklist ${HOME}/.config/evolution
 noblacklist ${HOME}/.config/geary
+noblacklist ${HOME}/.local/share/evolution
 noblacklist ${HOME}/.local/share/geary
 noblacklist ${HOME}/.mozilla
 
@@ -20,11 +24,19 @@ include disable-programs.inc
 include disable-shell.inc
 include disable-xdg.inc
 
+mkdir ${HOME}/.cache/evolution
+mkdir ${HOME}/.cache/folks
 mkdir ${HOME}/.cache/geary
+mkdir ${HOME}/.config/evolution
 mkdir ${HOME}/.config/geary
+mkdir ${HOME}/.local/share/evolution
 mkdir ${HOME}/.local/share/geary
+whitelist ${HOME}/.cache/evolution
+whitelist ${HOME}/.cache/folks
 whitelist ${HOME}/.cache/geary
+whitelist ${HOME}/.config/evolution
 whitelist ${HOME}/.config/geary
+whitelist ${HOME}/.local/share/evolution
 whitelist ${HOME}/.local/share/geary
 whitelist ${HOME}/.mozilla/firefox/profiles.ini
 whitelist ${DOWNLOADS}
@@ -36,6 +48,7 @@ include whitelist-var-common.inc
 
 apparmor
 caps.drop all
+machine-id
 netfilter
 no3d
 nodvd
@@ -62,6 +75,10 @@ private-tmp
 dbus-user filter
 dbus-user.own org.gnome.Geary
 dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.gnome.Contacts
+dbus-user.talk org.gnome.OnlineAccounts
+dbus-user.talk org.gnome.evolution.dataserver.AddressBook10
+dbus-user.talk org.gnome.evolution.dataserver.Sources5
 dbus-user.talk org.freedesktop.secrets
 dbus-system none
 
--- a/etc/profile-a-l/geary.profile
+++ b/etc/profile-a-l/geary.profile
@@ -61,6 +61,7 @@ nou2f
 novideo
 protocol unix,inet,inet6
 seccomp
+seccomp.block-secondary
 shell none
 tracelog
 
--- a/etc/profile-a-l/geary.profile
+++ b/etc/profile-a-l/geary.profile
@@ -4,19 +4,21 @@
 # Persistent local customizations
 include geary.local
 # Persistent global definitions
-# added by included profile
-#include globals.local
-
-# Users have Geary set to open a browser by clicking a link in an email
-# We are not allowed to blacklist browser-specific directories
-
-ignore dbus-user filter
-ignore dbus-system none
-ignore private-tmp
+include globals.local
 
 noblacklist ${HOME}/.cache/geary
 noblacklist ${HOME}/.config/geary
 noblacklist ${HOME}/.local/share/geary
+noblacklist ${HOME}/.mozilla
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
 
 mkdir ${HOME}/.cache/geary
 mkdir ${HOME}/.config/geary
@@ -24,8 +26,43 @@ mkdir ${HOME}/.local/share/geary
 whitelist ${HOME}/.cache/geary
 whitelist ${HOME}/.config/geary
 whitelist ${HOME}/.local/share/geary
+whitelist ${HOME}/.mozilla/firefox/profiles.ini
+whitelist ${DOWNLOADS}
 whitelist /usr/share/geary
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
 
-# allow Mozilla browsers
-# Redirect
-include firefox.profile
+apparmor
+caps.drop all
+netfilter
+no3d
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+tracelog
+
+# disable-mnt
+# Add ignore private-bin to geary.local for hyperlink support
+private-bin geary
+private-cache
+private-dev
+private-etc alternatives,ca-certificates,crypto-policies,fonts,hostname,hosts,pki,resolv.conf,selinux,ssl,xdg
+private-tmp
+
+dbus-user filter
+dbus-user.own org.gnome.Geary
+dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.freedesktop.secrets
+dbus-system none
+
+read-only ${HOME}/.mozilla/firefox/profiles.ini
--- a/etc/profile-a-l/geary.profile
+++ b/etc/profile-a-l/geary.profile
@@ -31,6 +31,7 @@ mkdir ${HOME}/.config/evolution
 mkdir ${HOME}/.config/geary
 mkdir ${HOME}/.local/share/evolution
 mkdir ${HOME}/.local/share/geary
+whitelist ${DOWNLOADS}
 whitelist ${HOME}/.cache/evolution
 whitelist ${HOME}/.cache/folks
 whitelist ${HOME}/.cache/geary
@@ -39,7 +40,6 @@ whitelist ${HOME}/.config/geary
 whitelist ${HOME}/.local/share/evolution
 whitelist ${HOME}/.local/share/geary
 whitelist ${HOME}/.mozilla/firefox/profiles.ini
-whitelist ${DOWNLOADS}
 whitelist /usr/share/geary
 include whitelist-common.inc
 include whitelist-runuser-common.inc
@@ -76,11 +76,11 @@ private-tmp
 dbus-user filter
 dbus-user.own org.gnome.Geary
 dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.freedesktop.secrets
 dbus-user.talk org.gnome.Contacts
 dbus-user.talk org.gnome.OnlineAccounts
 dbus-user.talk org.gnome.evolution.dataserver.AddressBook10
 dbus-user.talk org.gnome.evolution.dataserver.Sources5
-dbus-user.talk org.freedesktop.secrets
 dbus-system none
 
 read-only ${HOME}/.mozilla/firefox/profiles.ini
--- a/etc/profile-m-z/mutt.profile
+++ b/etc/profile-m-z/mutt.profile
@@ -38,9 +38,11 @@ noblacklist ${HOME}/sent
 blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}/wayland-*
 
-include allow-perl.inc
-include allow-python2.inc
-include allow-python3.inc
+# Uncomment or put them in mutt.local for oauth.py,S/MIME
+
+#include allow-perl.inc
+#include allow-python2.inc
+#include allow-python3.inc
 
 include disable-common.inc
 include disable-devel.inc
@@ -142,7 +144,6 @@ dbus-user none
 dbus-system none
 
 memory-deny-write-execute
-
 read-only ${HOME}/.elinks
 read-only ${HOME}/.nanorc
 read-only ${HOME}/.signature
--- a/etc/profile-m-z/mutt.profile
+++ b/etc/profile-m-z/mutt.profile
@@ -55,16 +55,17 @@ mkdir ${HOME}/.bogofilter
 mkdir ${HOME}/.cache/mutt
 mkdir ${HOME}/.config/mutt
 mkdir ${HOME}/.config/nano
+mkdir ${HOME}/.elinks
 mkdir ${HOME}/.emacs.d
 mkdir ${HOME}/.gnupg
 mkdir ${HOME}/.mail
 mkdir ${HOME}/.mutt
 mkdir ${HOME}/.vim
+mkdir ${HOME}/.w3m
 mkdir ${HOME}/Mail
 mkdir ${HOME}/mail
 mkdir ${HOME}/postponed
 mkdir ${HOME}/sent
-mkfile ${HOME}/.elinks
 mkfile ${HOME}/.emacs
 mkfile ${HOME}/.mailcap
 mkfile ${HOME}/.msmtprc
@@ -73,7 +74,6 @@ mkfile ${HOME}/.nanorc
 mkfile ${HOME}/.signature
 mkfile ${HOME}/.viminfo
 mkfile ${HOME}/.vimrc
-mkfile ${HOME}/.w3m
 whitelist ${DOCUMENTS}
 whitelist ${DOWNLOADS}
 whitelist ${HOME}/.Mail
@@ -126,6 +126,7 @@ nou2f
 novideo
 protocol unix,inet,inet6
 seccomp
+seccomp.block-secondary
 shell none
 tracelog
 
@@ -139,6 +140,8 @@ writable-var
 
 dbus-user none
 dbus-system none
+
+memory-deny-write-execute
 
 read-only ${HOME}/.elinks
 read-only ${HOME}/.nanorc
--- a/etc/profile-m-z/mutt.profile
+++ b/etc/profile-m-z/mutt.profile
@@ -1,7 +1,7 @@
 # Firejail profile for mutt
 # Description: Text-based mailreader supporting MIME, GPG, PGP and threading
-quiet
 # This file is overwritten after every install/update
+quiet
 # Persistent local customizations
 include mutt.local
 # Persistent global definitions
@@ -9,8 +9,10 @@ include globals.local
 
 noblacklist /var/mail
 noblacklist /var/spool/mail
+noblacklist ${DOCUMENTS}
 noblacklist ${HOME}/.Mail
 noblacklist ${HOME}/.bogofilter
+noblacklist ${HOME}/.cache/mutt
 noblacklist ${HOME}/.config/mutt
 noblacklist ${HOME}/.config/nano
 noblacklist ${HOME}/.elinks
@@ -37,7 +39,8 @@ blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}/wayland-*
 
 include allow-perl.inc
-include allow-python.inc
+include allow-python2.inc
+include allow-python3.inc
 
 include disable-common.inc
 include disable-devel.inc
@@ -60,6 +63,7 @@ mkfile ${HOME}/.vimrc
 mkfile ${HOME}/.w3m
 mkdir ${HOME}/.Mail
 mkdir ${HOME}/.bogofilter
+mkdir ${HOME}/.cache/mutt
 mkdir ${HOME}/.config/mutt
 mkdir ${HOME}/.config/nano
 mkdir ${HOME}/.emacs.d
@@ -73,6 +77,7 @@ mkdir ${HOME}/postponed
 mkdir ${HOME}/sent
 whitelist ${HOME}/.Mail
 whitelist ${HOME}/.bogofilter
+whitelist ${HOME}/.cache/mutt
 whitelist ${HOME}/.config/mutt
 whitelist ${HOME}/.config/nano
 whitelist ${HOME}/.elinks
@@ -133,3 +138,15 @@ writable-var
 
 dbus-user none
 dbus-system none
+
+read-only ${HOME}/.elinks
+read-only ${HOME}/.emacs
+read-only ${HOME}/.mailcap
+read-only ${HOME}/.msmtprc
+read-only ${HOME}/.muttrc
+read-only ${HOME}/.nanorc
+read-only ${HOME}/.signature
+read-only ${HOME}/.vimrc
+read-only ${HOME}/.viminfo
+read-only ${HOME}/.vimrc
+read-only ${HOME}/.w3m
--- a/etc/profile-m-z/mutt.profile
+++ b/etc/profile-m-z/mutt.profile
@@ -50,17 +50,6 @@ include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-xdg.inc
 
-mkfile ${HOME}/.elinks
-mkfile ${HOME}/.emacs
-mkfile ${HOME}/.mailcap
-mkfile ${HOME}/.msmtprc
-mkfile ${HOME}/.muttrc
-mkfile ${HOME}/.nanorc
-mkfile ${HOME}/.signature
-mkfile ${HOME}/.vimrc
-mkfile ${HOME}/.viminfo
-mkfile ${HOME}/.vimrc
-mkfile ${HOME}/.w3m
 mkdir ${HOME}/.Mail
 mkdir ${HOME}/.bogofilter
 mkdir ${HOME}/.cache/mutt
@@ -75,6 +64,17 @@ mkdir ${HOME}/Mail
 mkdir ${HOME}/mail
 mkdir ${HOME}/postponed
 mkdir ${HOME}/sent
+mkfile ${HOME}/.elinks
+mkfile ${HOME}/.emacs
+mkfile ${HOME}/.mailcap
+mkfile ${HOME}/.msmtprc
+mkfile ${HOME}/.muttrc
+mkfile ${HOME}/.nanorc
+mkfile ${HOME}/.signature
+mkfile ${HOME}/.vimrc
+mkfile ${HOME}/.viminfo
+mkfile ${HOME}/.vimrc
+mkfile ${HOME}/.w3m
 whitelist ${HOME}/.Mail
 whitelist ${HOME}/.bogofilter
 whitelist ${HOME}/.cache/mutt
@@ -140,13 +140,6 @@ dbus-user none
 dbus-system none
 
 read-only ${HOME}/.elinks
-read-only ${HOME}/.emacs
-read-only ${HOME}/.mailcap
-read-only ${HOME}/.msmtprc
-read-only ${HOME}/.muttrc
 read-only ${HOME}/.nanorc
 read-only ${HOME}/.signature
-read-only ${HOME}/.vimrc
-read-only ${HOME}/.viminfo
-read-only ${HOME}/.vimrc
 read-only ${HOME}/.w3m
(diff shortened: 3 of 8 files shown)


etc/profile-m-z/mutt.profile
```diff
@@ -34,14 +36,77 @@ noblacklist ${HOME}/sent
 blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}/wayland-*
 
+include allow-perl.inc
+include allow-python.inc
```
rusty-snake, line 40: There's no `allow-python.inc`, use `allow-python{2,3}.inc`:
```suggestion
include allow-python2.inc
include allow-python3.inc
```

etc/profile-m-z/mutt.profile
```diff
@@ -10,13 +11,14 @@ noblacklist /var/mail
 noblacklist /var/spool/mail
 noblacklist ${HOME}/.Mail
 noblacklist ${HOME}/.bogofilter
-noblacklist ${HOME}/.cache/mutt
+noblacklist ${HOME}/.config/mutt
```
rusty-snake, line 14: There's a `blacklist ${HOME}/.cache/mutt` in `disable-programs.inc`. We should keep the `noblacklist` for cases like the users adds `ignore private-cache`. Or is `${HOME}/.cache/mutt` not used?

etc/profile-m-z/mutt.profile
```diff
@@ -34,14 +36,77 @@ noblacklist ${HOME}/sent
...
+include allow-perl.inc
+include allow-python.inc
+
 include disable-common.inc
 include disable-devel.inc
+include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
+include disable-xdg.inc
```
rusty-snake, line 48: You need to add `noblacklist ${DOCUMENTS}` in order to make `whitelist ${DOCUMENTS}` work.

etc/profile-m-z/mutt.profile
```diff
@@ -34,14 +36,77 @@ noblacklist ${HOME}/sent
...
+whitelist ${HOME}/.mail
+whitelist ${HOME}/.mailcap
+whitelist ${HOME}/.msmtprc
+whitelist ${HOME}/.mutt
+whitelist ${HOME}/.muttrc
+whitelist ${HOME}/.nanorc
+whitelist ${HOME}/.signature
+whitelist ${HOME}/.vim
+whitelist ${HOME}/.viminfo
+whitelist ${HOME}/.vimrc
```
rusty-snake, line 91: Should we make all these editor configurations (vim,emacs,nano,...) read-only?

rusty-snake (commented):
### mutt

- `whitelist` :rocket: 
- some nitpicks and future suggestions
- does it break w/o `quiet`?

### neomutt

Wouldn't it make sense to make it a mutt redirect?

rusty-snake (commented):
### geary

I've a unfinished and **untested** version of this in my local profiles directory.

```
# Uncomment if you want to ...
#noblacklist /var/mail
#noblacklist /var/spool/mail
#whitelist /var/mail
#whitelist /var/spool/mail
#writable-var
```

```
machine-id
seccomp.block-secondary
```

What have other email programs?
```
disable-mnt
```

From flatpak and to permissive:
```
dbus-user filter
dbus-user.own org.gnome.Geary
dbus-user.talk ca.desrt.dconf
dbus-user.talk org.gnome.Contacts
dbus-user.talk org.gnome.OnlineAccounts
dbus-user.talk org.gnome.evolution.dataserver.AddressBook10
dbus-user.talk org.gnome.evolution.dataserver.Sources5
dbus-system none
```

etc/profile-m-z/mutt.profile
bbhtt: On first run it didn't create a cache directory. I assume it is for mail cache which the users can specify the directory. For that use case I can add it back.

etc/profile-m-z/mutt.profile
bbhtt: I can only comment on what I've used/use nano,vim.signature do not need write. I'm adding them read-only.

bbhtt:
> I've a unfinished and untested version of this in my local profiles directory.

I don't think Geary has the option to utilize a local mailbox.

> From flatpak and to permissive:

Thanks.

> `machine-id`

Makes sense.

> `seccomp.block-secondary`

I'll check this.

> What have other email programs?

Not sure what you mean. `disable-mnt` was un-commented in all mail clients post-whitelisting.

kmk3:
>     diff --git a/etc/profile-m-z/mutt.profile b/etc/profile-m-z/mutt.profile
>     index 1ce12f54f..87e7c7f06 100644
>     --- a/etc/profile-m-z/mutt.profile
>     +++ b/etc/profile-m-z/mutt.profile
>     [...]
>     +whitelist ${HOME}/.config/nano
>     +whitelist ${HOME}/.elinks
>     +whitelist ${HOME}/.emacs
>     +whitelist ${HOME}/.emacs.d
>     +whitelist ${HOME}/.gnupg
>     +whitelist ${HOME}/.mail
>     +whitelist ${HOME}/.mailcap
>     +whitelist ${HOME}/.msmtprc
>     +whitelist ${HOME}/.mutt
>     +whitelist ${HOME}/.muttrc
>     +whitelist ${HOME}/.nanorc
>     +whitelist ${HOME}/.signature
>     +whitelist ${HOME}/.vim
>     +whitelist ${HOME}/.viminfo
>     +whitelist ${HOME}/.vimrc

MUAs like mutt are also used as [development tools][1], especially on projects
where the reviews center around mailing lists (e.g.: the Linux kernel), so
using whitelists by default seems a bit much.

As a featureful example, aerc has a built-in terminal emulator and applying
patches, building the project and pushing the result [is part of the indended
workflow][2].  So a complete whitelist to support that would have to include at
least git, every cli text editor and potentially every adjacent dev tool and
language/runtime if you build projects from inside aerc (e.g.: using `git am` +
`make test`).

Mutt is also not too far off since it supports macros for running arbitrary
commands, such as applying patches with `git am` (see [3], [4]).

Besides that, there's the effort of mantaining the whitelists for every email
client x text editor combination.  Examples off the top of my head:

* Text editors: nano, vim, neovim, emacs
* Email clients: mutt, neomutt, aerc, alpine

These alone would mean maintaining 4 sets of paths on 4 different profiles.

---

On a related note, there's already some text editor exceptions scattered in a
few profiles, but I've been working on a PR (WIP) to move all of that to a
single .inc.  And also related, in there I've stumbled upon some whitelist
entries (e.g.: those on `git.profile`) that make it harder to organize things.

The main problem with whitelists is that there's no way to tell firejail to
only apply a given whitelist rule if a whitelist is already in effect (e.g.: if
there was a `whitelist-enable /usr/share` command).  If it worked differently
in this regard, it would be possible to add whitelist command to .inc files
without worrying about breakage, just like the noblacklist commands.

[1]: https://git-send-email.io/
[2]: https://drewdevault.com/2019/06/03/Announcing-aerc-0.1.0.html
[3]: https://github.com/dvhart/mailtogit
[4]: http://kroah.com/log/blog/2019/08/14/patch-workflow-with-mutt-2019/

etc/profile-m-z/mutt.profile
kmk3: > Should we make all these editor configurations (vim,emacs,nano,...) read-only?

Isn't that already handled by `include disable-common.inc`?
https://github.com/netblue30/firejail/blob/508dc3ae991e6f0418eaf42babaf5de6db4f7cd9/etc/inc/disable-common.inc#L296

rusty-snake:
> MUAs like mutt are also used as development tools, especially on projects
where the reviews center around mailing lists (e.g.: the Linux kernel), so
using whitelists by default seems a bit much.

`whitelist`ing is a essential security feature, especial for programs with such a high attack vector as email-clients. Blacklisting profiles can be escaped easily in the most cases. Enabling `whitelist` for allusers is good IMHO. If you need arbitrary file access in $HOME, just use `ignore whitelist ${HOME}` .

Users who use a tui-mail-client and develop, are expected to have the knowledge to read documentation and customize there profiles. Therefore the question is IMHO not "Does this work with all workflows?" but "Does it work with the workflow of the (huge) majority of the users?". IDK how many mutt+firejail users develop on the kernel or similar, although I think that there are not that many.

> The main problem with whitelists is that there's no way to tell firejail to
only apply a given whitelist rule if a whitelist is already in effect (e.g.: if
there was a whitelist-enable /usr/share command). If it worked differently
in this regard, it would be possible to add whitelist command to .inc files
without worrying about breakage, just like the noblacklist commands.

Conditions would also be a solution (e.g. `?HAS_USR_SHARE_WHITELIST: whitelist /usr/share/spam`). However #3447 is a better solution I think because it solves this by removing complexity instead of adding. (Talking about the complexity of profiles, not the one of the code to implement this.)

etc/profile-m-z/mutt.profile
bbhtt: Right. These 4 are missing from `disable-common.inc` : `.elinks,.nanorc,.signature,.w3m`. `.nanorc` is safe, `.signature` is used by some mail clients to write a signature and by Mugshot (xfce); I don't use the other two.

kmk3:
> > MUAs like mutt are also used as development tools, especially on projects
> > where the reviews center around mailing lists (e.g.: the Linux kernel), so
> > using whitelists by default seems a bit much.
> 
> `whitelist`ing is a essential security feature, especial for programs with
> such a high attack vector as email-clients.

Agreed.

> Blacklisting profiles can be escaped easily in the most cases.

Sounds interesting; do you know of any docs/articles/papers on this?

> Enabling `whitelist` for allusers is good IMHO. If you need arbitrary file
> access in $HOME, just use `ignore whitelist ${HOME}` .

Thanks, I didn't know that disabling the whitelist was that simple. 
 
> Users who use a tui-mail-client and develop, are expected to have the
> knowledge to read documentation and customize there profiles.  Therefore the
> question is IMHO not "Does this work with all workflows?" but "Does it work
> with the workflow of the (huge) majority of the users?".

If all it takes is `ignore whitelist ${HOME}` / `ignore whitelist /usr/share` /
etc instead of `ignore whitelist ${HOME}/foo/bar/*`, then sure.

But to nitpick a bit: speaking of documentation, it didn't occur to me that
this was the way to do it and it doesn't seem to be documented, so it seems a
little unfair to expect such knowledge just because one uses a tui-mail-client:

```console
$ git show -q --pretty='%h %s' master
862f6820 manpage: /bin/bash -> user's perferred shell
$ grep -Fnr 'ignore whitelist' | grep -v '^etc/profile-'
```

It's not exactly clear to me which paths are treated specially by `blacklist` /
`whitelist`.  For example, I know that `blacklist /` (I used that for testing)
does not affect `${HOME}`, so `blacklist ${HOME}` would be required as well.

Anyways, my other point of contention is the amount of junk files that would
dumped at the doorstep of `${HOME}` through `mkdir` / `mkfile` by merely
opening a program.  I know that they have to exist before applying the
namespace, but there has to be a better way...

Well, if `ignore whitelist ${HOME}` works, so should `ignore mkfile ${HOME}`,
right?

```firejail
ignore mkdir ${HOME}
ignore mkfile ${HOME}
ignore whitelist ${HOME}
```

But I did some testing and unfortunately that's not the case:

```console
$ cat /tmp/firejail/test.profile
ignore mkdir ${HOME}
ignore mkfile ${HOME}
ignore whitelist ${HOME}

mkdir ${HOME}/.spam
mkfile ${HOME}/.spamrc
whitelist ${HOME}/.spam
whitelist ${HOME}/.spamrc
$ ls .spam .spamrc | cat
ls: cannot access '.spam': No such file or directory
ls: cannot access '.spamrc': No such file or directory
$ firejail --quiet --profile=/tmp/firejail/test.profile ls .spam .spamrc | cat
.spamrc

.spam:
$ ls .spam .spamrc | cat
.spamrc

.spam:
```

So `ignore whitelist ${HOME}` is not enough to avoid the side effects of
profiles with `whitelist`s; it seems that there would need to be an exception
for every `mkdir` / `mkfile` entry as well, which would be a hassle to maintain
locally, especially for every text editor x email client combination.

---

On a related side note, it would be nice to create only the xdg-base-dir paths
so that at least there's less clutter:

```diff
-mkfile ${HOME}/.muttrc
-mkfile ${HOME}/.nanorc
 [...]
 mkdir ${HOME}/.config/mutt
 mkdir ${HOME}/.config/nano
```

But since `whitelist` fails when a path doesn't exist, there would need to be a
`whitelist-if-exists`...  Problem case: `${HOME}/.nanorc` exists, but the
profile has only `mkfile ${HOME}/.config/nano` -> user config is not present in
the namespace -> nano uses an empty config.

---

> IDK how many mutt+firejail users develop on the kernel or similar, although I
> think that there are not that many.

I personally see both mutt and firejail as relatively medium-advanced user
tools (or at least not something that a new Linux user would go about
installing), so I think that there is above average overlap in the audiences
(potential or otherwise).

I mentioned the kernel because it's the most known example, but perhaps not a
very good one.  

A more acessible instance would be [SourceHut](https://sourcehut.org/).  It is
a software forge like GitHub, but every project gets a mailing list.  The web
and email interfaces are linked together; whatever you do on one shows up in
the other.  You can send a patchset, open issues, and reply to them by email
(all without needing an account) and everything will show up in the web ui.  If
you do something in the web ui, people get emails and can reply to them and
perform actions (e.g.: close an issue) through them.  Incidentally, there's an
article from today comparing the workflows:
<https://blog.brixit.nl/git-email-flow-versus-github-flow/>

The point being that this is not something only used by high-profile or
high-complexity projects.

(I'll reply to the rest on a new issue as that's mostly about a feature
request)

etc/profile-m-z/mutt.profile
```diff
@@ -34,14 +36,77 @@ noblacklist ${HOME}/sent
...
+mkfile ${HOME}/.mailcap
+mkfile ${HOME}/.msmtprc
+mkfile ${HOME}/.muttrc
+mkfile ${HOME}/.nanorc
+mkfile ${HOME}/.signature
+mkfile ${HOME}/.vimrc
+mkfile ${HOME}/.viminfo
+mkfile ${HOME}/.vimrc
+mkfile ${HOME}/.w3m
+mkdir ${HOME}/.Mail
```
kmk3, line 64: Considering `etc/templates/profile.template`, shouldn't `mkdir` come
before `mkfile`?

rusty-snake:
> So ignore whitelist ${HOME} is not enough to avoid the side effects of
profiles with whitelists; it seems that there would need to be an exception
for every mkdir / mkfile entry as well, which would be a hassle to maintain
locally, especially for every text editor x email client combination.

The other option would be to provide a commented opt-in whitelist. (Again) users of such software aren't expected to be completely unskilled, so they can look at the profile and uncomment/copy this part. ... or we have it opt-out and add a "If you need less restricted $HOME access, add 'ignore whitelist ${HOME}' to your mutt.local" note.

> But to nitpick a bit: speaking of documentation, it didn't occur to me that
this was the way to do it and it doesn't seem to be documented, so it seems a
little unfair to expect such knowledge just because one uses a tui-mail-client:

Agreed. (As in the most OS-projects, there's a lower interest in writing and fine-tuning documentation and no one is paid to do that). Here you need either some playing with `ignore` or reading it's full code.

> For example, I know that blacklist / (I used that for testing)
does not affect ${HOME}, so blacklist ${HOME} would be required as well.

/home | ${HOME} is usually treated special (and some others). Again undocumented.

Commands to `/` often don't work as expected (e.g. `read-only /` doesn't make everything ro). On the other hand the command like `blacklist` or `noexec` does make sens to be used on `/`. Again undocumented.

> > Blacklisting profiles can be escaped easily in the most cases.
>
> Sounds interesting; do you know of any docs/articles/papers on this?

1. Of course there's no blacklisting of unknown files. How should it work.
2. Blacklisting is only applied to existing files. Example: `~/.config/systemd` is read by `systemd --user` (which is used by Debian, Ubuntu, Mint, Fedora, Arch, OpenSUSE, ...) but is not present on the most systems. Or `~/.profile` isn't in `/etc/skel` on some distros (I know Fedora, as I use it) but it is read by `sh -l`. Or `~/.pam_environment` or `~/.config/environment.d/` and so on. So the issue are files which can be execute or change the environment (e.g. `PATH=$HOME/dir_with_bad_coreutils:$PATH`).

FYI: #3527. In my opinion a skilled attacker who can run any code in the sandbox, can escape any sandbox. The question is how much work it is to escape (without exploits). If the NSA wants to hack *you*, they can. If some ransomware uses `/usr/bin/openssl` for encryption of ~/Documents, `read-only ${HOME}` `whitelist ${HOME}/foo` `blacklist ${PATH}/openssl` (`disable-common.inc`) and `blacklist ${DOCUMENTS}` (`disable-xdg.inc`) all break it and why should a ransomware author do much extra work for edge cases, there are hundred system where it works.

----

becomes OT. If you have more question it would be the best to open an issues with quotes from here.

etc/profile-m-z/mutt.profile
rusty-snake: profile.template has it alphabetical (with two lines there're only two options) which is not bad for small list. With longer lists like here it's fine in my opinion to have `noblacklist`, `mk{dir,file}` and `whitelist` listed the files/dirs in the same order on all three.

bbhtt (commented):
I added Sylpheed and Claws-mail too. (1) gpg (2) hyperlinks (3) dbus permissions

kmk3:
> The other option would be to provide a commented opt-in whitelist. (Again)
> users of such software aren't expected to be completely unskilled, so they
> can look at the profile and uncomment/copy this part. ... or we have it
> opt-out and add a "If you need less restricted $HOME access, add 'ignore
> whitelist ${HOME}' to your mutt.local" note.

My point isn't so much about the defaults or knowing how to write profiles, but
about UX / how complex it is profile-wise to create exceptions and maintain
them.  If configuring it away is simple enough (e.g.: the 3 lines to ignore
mkdir/mkfile/whitelist in `${HOME}`), then whitelisting by default is fine by
me.

Something which would work right now and partly achieve that is to put the
editor paths on e.g.: `whitelist-text-editors.inc` instead of hard-coding them
on these profiles.  That way, it's at least possible to override the .inc with
only the editor(s) that I actually use and that would also work for every
profile that includes that.

Or rather, to avoid derailing this further, this can be merged without that and
I'll do the changes in the WIP branch that I mentioned.  In that case, @bbhtt
after this is merged, would you mind waiting for a PR of that branch before
creating new ones which whitelist editors?

> > But to nitpick a bit: speaking of documentation, it didn't occur to me that
> > this was the way to do it and it doesn't seem to be documented, so it seems
> > a little unfair to expect such knowledge just because one uses a
> > tui-mail-client:
> 
> Agreed. (As in the most OS-projects, there's a lower interest in writing and
> fine-tuning documentation and no one is paid to do that). Here you need
> either some playing with `ignore` or reading it's full code.
> 
> > For example, I know that blacklist / (I used that for testing) does not
> > affect ${HOME}, so blacklist ${HOME} would be required as well.
> 
> /home | ${HOME} is usually treated special (and some others). Again
> undocumented.
> 
> Commands to `/` often don't work as expected (e.g. `read-only /` doesn't make
> everything ro). On the other hand the command like `blacklist` or `noexec`
> does make sens to be used on `/`. Again undocumented.

Yeah, maintaining documentation is not exactly the most fun activity.  As long
as people aren't expected to dive deep into the source for creating profiles,
that's alright by me.  Anyways, thanks for the details.

> > > Blacklisting profiles can be escaped easily in the most cases.
> > 
> > 
> > Sounds interesting; do you know of any docs/articles/papers on this?
> 
>     1. Of course there's no blacklisting of unknown files. How should it
>        work.
> 
>     2. Blacklisting is only applied to existing files. Example:
>        `~/.config/systemd` is read by `systemd --user` (which is used by
>        Debian, Ubuntu, Mint, Fedora, Arch, OpenSUSE, ...) but is not present
>        on the most systems. Or `~/.profile` isn't in `/etc/skel` on some
>        distros (I know Fedora, as I use it) but it is read by `sh -l`. Or
>        `~/.pam_environment` or `~/.config/environment.d/` and so on. So the
>        issue are files which can be execute or change the environment (e.g.
>        `PATH=$HOME/dir_with_bad_coreutils:$PATH`).
> 
> 
> FYI: #3527.

Thanks for the explanation/link.  I thought you had meant escape in the sense
of bypassing the `root:root` ownership/file permissions imposed by firejail or
something.  I wasn't aware that blacklisting only works if the file already
exists; that's crucial information.  Now whitelisting sounds a lot more
important.

Some testing for future reference:

```console
$ cat .emacs
cat: .emacs: No such file or directory
$ firejail --noprofile --quiet --blacklist='${HOME}/.emacs' \
  bash -c 'echo evil >~/.emacs; echo $?' 2>/dev/null
0
$ cat .emacs
evil
$ firejail --noprofile --quiet --blacklist='${HOME}/.emacs' \
  bash -c 'echo evil2 >>~/.emacs; echo $?' 2>/dev/null
1
$ cat .emacs
evil
```

> In my opinion a skilled attacker who can run any code in the sandbox, can
> escape any sandbox. The question is how much work it is to escape (without
> exploits). If the NSA wants to hack _you_, they can. If some ransomware uses
> `/usr/bin/openssl` for encryption of ~/Documents, `read-only ${HOME}`
> `whitelist ${HOME}/foo` `blacklist ${PATH}/openssl` (`disable-common.inc`)
> and `blacklist ${DOCUMENTS}` (`disable-xdg.inc`) all break it and why should
> a ransomware author do much extra work for edge cases, there are hundred
> system where it works.

Indeed, and that's where I believe firejail shines the most, by mitigating the
damage from untargeted userland exploits (even if it's just by being an edge
case) and by protecting against abuse from proprietary programs.  IMO it's what
antivirus programs should have been.  I don't really expect it to be a life
saver if e.g.: kernel exploits are involved.  Either way, the overall cost
increase is a net win.

> becomes OT. If you have more question it would be the best to open an issues
> with quotes from here.

Agreed.

etc/profile-m-z/mutt.profile
kmk3: > profile.template has it alphabetical (with two lines there're only two
> options) which is not bad for small list. With longer lists like here it's
> fine in my opinion to have `noblacklist`, `mk{dir,file}` and `whitelist`
> listed the files/dirs in the same order on all three.

I mean, the order of files/dirs would still be the same in both lists, the only
difference when sorting from the first `mkfile` to the last `mkdir` is that the
`mkdir` entries would come before the `mkfile` ones.  Having it all sorted
means being able to add/edit anything anywhere in the range of [`mkdir`,
`whitelist`] and then just run `:sort u` on it to make it all sorted, without
having to visually scan things (also, creating dirs before files seems a bit
more natural to me).  Anyways, these were sorted by @bbhtt, but there's a few
others; I'll add a few code suggestions to demonstrate; feel free to
accept/reject.

etc/profile-a-l/geary.profile
```diff
@@ -4,28 +4,83 @@
...
 mkdir ${HOME}/.local/share/geary
+whitelist ${HOME}/.cache/evolution
+whitelist ${HOME}/.cache/folks
 whitelist ${HOME}/.cache/geary
+whitelist ${HOME}/.config/evolution
 whitelist ${HOME}/.config/geary
+whitelist ${HOME}/.local/share/evolution
 whitelist ${HOME}/.local/share/geary
+whitelist ${HOME}/.mozilla/firefox/profiles.ini
+whitelist ${DOWNLOADS}
```
kmk3, line 42: sort

```suggestion
whitelist ${DOWNLOADS}
whitelist ${HOME}/.cache/evolution
whitelist ${HOME}/.cache/folks
whitelist ${HOME}/.cache/geary
whitelist ${HOME}/.config/evolution
whitelist ${HOME}/.config/geary
whitelist ${HOME}/.local/share/evolution
whitelist ${HOME}/.local/share/geary
whitelist ${HOME}/.mozilla/firefox/profiles.ini
```

etc/profile-a-l/geary.profile
```diff
@@ -4,28 +4,83 @@
...
+private-tmp
+
+dbus-user filter
+dbus-user.own org.gnome.Geary
+dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.gnome.Contacts
+dbus-user.talk org.gnome.OnlineAccounts
+dbus-user.talk org.gnome.evolution.dataserver.AddressBook10
+dbus-user.talk org.gnome.evolution.dataserver.Sources5
+dbus-user.talk org.freedesktop.secrets
```
kmk3, line 83: sort

```suggestion
dbus-user filter
dbus-user.own org.gnome.Geary
dbus-user.talk ca.desrt.dconf
dbus-user.talk org.freedesktop.secrets
dbus-user.talk org.gnome.Contacts
dbus-user.talk org.gnome.OnlineAccounts
dbus-user.talk org.gnome.evolution.dataserver.AddressBook10
dbus-user.talk org.gnome.evolution.dataserver.Sources5
```

etc/profile-m-z/mutt.profile
```diff
@@ -34,14 +38,80 @@ noblacklist ${HOME}/sent
...
+mkdir ${HOME}/postponed
+mkdir ${HOME}/sent
+mkfile ${HOME}/.elinks
+mkfile ${HOME}/.emacs
+mkfile ${HOME}/.mailcap
+mkfile ${HOME}/.msmtprc
+mkfile ${HOME}/.muttrc
+mkfile ${HOME}/.nanorc
+mkfile ${HOME}/.signature
+mkfile ${HOME}/.vimrc
```
kmk3, line 74: duplicate

```suggestion
```

etc/profile-m-z/mutt.profile
```diff
@@ -34,14 +38,80 @@ noblacklist ${HOME}/sent
...
+whitelist ${HOME}/.vim
+whitelist ${HOME}/.viminfo
+whitelist ${HOME}/.vimrc
+whitelist ${HOME}/.w3m
+whitelist ${HOME}/Mail
+whitelist ${HOME}/mail
+whitelist ${HOME}/postponed
+whitelist ${HOME}/sent
+whitelist ${DOCUMENTS}
+whitelist ${DOWNLOADS}
```
kmk3, line 103: sort

```suggestion
whitelist ${DOCUMENTS}
whitelist ${DOWNLOADS}
whitelist ${HOME}/.Mail
whitelist ${HOME}/.bogofilter
whitelist ${HOME}/.cache/mutt
whitelist ${HOME}/.config/mutt
whitelist ${HOME}/.config/nano
whitelist ${HOME}/.elinks
whitelist ${HOME}/.emacs
whitelist ${HOME}/.emacs.d
whitelist ${HOME}/.gnupg
whitelist ${HOME}/.mail
whitelist ${HOME}/.mailcap
whitelist ${HOME}/.msmtprc
whitelist ${HOME}/.mutt
whitelist ${HOME}/.muttrc
whitelist ${HOME}/.nanorc
whitelist ${HOME}/.signature
whitelist ${HOME}/.vim
whitelist ${HOME}/.viminfo
whitelist ${HOME}/.vimrc
whitelist ${HOME}/.w3m
whitelist ${HOME}/Mail
whitelist ${HOME}/mail
whitelist ${HOME}/postponed
whitelist ${HOME}/sent
```

etc/profile-m-z/neomutt.profile
```diff
@@ -0,0 +1,149 @@
...
+noblacklist ${HOME}/.neomuttrc
+noblacklist ${HOME}/.signature
+noblacklist ${HOME}/.vim
+noblacklist ${HOME}/.viminfo
+noblacklist ${HOME}/.vimrc
+noblacklist ${HOME}/.w3m
+noblacklist ${HOME}/Mail
+noblacklist ${HOME}/mail
+noblacklist ${HOME}/postponed
+noblacklist ${HOME}/sent
```
kmk3, line 38: sort

```suggestion
noblacklist ${DOCUMENTS}
noblacklist ${HOME}/.Mail
noblacklist ${HOME}/.bogofilter
noblacklist ${HOME}/.config/mutt
noblacklist ${HOME}/.config/nano
noblacklist ${HOME}/.config/neomutt
noblacklist ${HOME}/.elinks
noblacklist ${HOME}/.emacs
noblacklist ${HOME}/.emacs.d
noblacklist ${HOME}/.gnupg
noblacklist ${HOME}/.mail
noblacklist ${HOME}/.mailcap
noblacklist ${HOME}/.msmtprc
noblacklist ${HOME}/.mutt
noblacklist ${HOME}/.muttrc
noblacklist ${HOME}/.nanorc
noblacklist ${HOME}/.neomutt
noblacklist ${HOME}/.neomuttrc
noblacklist ${HOME}/.signature
noblacklist ${HOME}/.vim
noblacklist ${HOME}/.viminfo
noblacklist ${HOME}/.vimrc
noblacklist ${HOME}/.w3m
noblacklist ${HOME}/Mail
noblacklist ${HOME}/mail
noblacklist ${HOME}/postponed
noblacklist ${HOME}/sent
noblacklist /var/mail
noblacklist /var/spool/mail
```

etc/profile-m-z/neomutt.profile
```diff
@@ -0,0 +1,149 @@
...
+mkdir ${HOME}/sent
+mkfile ${HOME}/.elinks
+mkfile ${HOME}/.emacs
+mkfile ${HOME}/.mailcap
+mkfile ${HOME}/.msmtprc
+mkfile ${HOME}/.muttrc
+mkfile ${HOME}/.nanorc
+mkfile ${HOME}/.neomuttrc
+mkfile ${HOME}/.signature
+mkfile ${HOME}/.vimrc
```
kmk3, line 76: duplicate

```suggestion
```

etc/profile-m-z/neomutt.profile
```diff
@@ -0,0 +1,149 @@
...
+whitelist ${HOME}/.vim
+whitelist ${HOME}/.viminfo
+whitelist ${HOME}/.vimrc
+whitelist ${HOME}/.w3m
+whitelist ${HOME}/Mail
+whitelist ${HOME}/mail
+whitelist ${HOME}/postponed
+whitelist ${HOME}/sent
+whitelist ${DOCUMENTS}
+whitelist ${DOWNLOADS}
```
kmk3, line 107: sort

```suggestion
whitelist ${DOCUMENTS}
whitelist ${DOWNLOADS}
whitelist ${HOME}/.Mail
whitelist ${HOME}/.bogofilter
whitelist ${HOME}/.config/mutt
whitelist ${HOME}/.config/nano
whitelist ${HOME}/.config/neomutt
whitelist ${HOME}/.elinks
whitelist ${HOME}/.emacs
whitelist ${HOME}/.emacs.d
whitelist ${HOME}/.gnupg
whitelist ${HOME}/.mail
whitelist ${HOME}/.mailcap
whitelist ${HOME}/.msmtprc
whitelist ${HOME}/.mutt
whitelist ${HOME}/.muttrc
whitelist ${HOME}/.nanorc
whitelist ${HOME}/.neomutt
whitelist ${HOME}/.neomuttrc
whitelist ${HOME}/.signature
whitelist ${HOME}/.vim
whitelist ${HOME}/.viminfo
whitelist ${HOME}/.vimrc
whitelist ${HOME}/.w3m
whitelist ${HOME}/Mail
whitelist ${HOME}/mail
whitelist ${HOME}/postponed
whitelist ${HOME}/sent
```

etc/profile-m-z/sylpheed.profile
```diff
@@ -13,5 +13,15 @@ whitelist ${HOME}/.sylpheed-2.0
 
 whitelist /usr/share/sylpheed
 
+# private-bin curl,gpg,gpg2,gpg-agent,gpgsm,pinentry,pinentry-gtk-2,sylpheed
+
+dbus-user filter
+dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.gnome.keyring.SystemPrompter
+dbus-user.talk org.freedesktop.secrets
```
kmk3, line 21: sort

```suggestion
dbus-user filter
dbus-user.talk ca.desrt.dconf
dbus-user.talk org.freedesktop.secrets
dbus-user.talk org.gnome.keyring.SystemPrompter
```

bbhtt:
> In that case, @bbhtt after this is merged, would you mind waiting for a PR of that branch before creating new ones which whitelist editors?

Sure.

About the sorts except dbus: IMO it is better to consider macros as their own section. Otherwise if the whitelists get too long it'll be easy to loose them since Downloads/Documents will be in the top then Home/... then Pictures,Videos etc. people most often need to tweak these.

etc/profile-m-z/neomutt.profile
```diff
@@ -0,0 +1,148 @@
...
+writable-run-user
+writable-var
+
+dbus-user none
+dbus-system none
+
+read-only ${HOME}/.elinks
+read-only ${HOME}/.nanorc
+read-only ${HOME}/.signature
+read-only ${HOME}/.w3m
```
rusty-snake, line 148: Add `read-only ${HOME}/.vimrc`

etc/profile-m-z/neomutt.profile
```diff
@@ -0,0 +1,148 @@
...
+whitelist /var/mail
+whitelist /var/spool/mail
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+netfilter
```
rusty-snake, line 119: We could add `ipc-namespace`.

etc/profile-m-z/mutt.profile
```diff
@@ -56,7 +125,20 @@ novideo
...
 writable-run-user
 writable-var
+
+dbus-user none
+dbus-system none
+
+read-only ${HOME}/.elinks
+read-only ${HOME}/.nanorc
+read-only ${HOME}/.signature
+read-only ${HOME}/.w3m
```
rusty-snake, line 144: `.vimrc`

etc/profile-a-l/email-common.profile
```diff
@@ -54,13 +62,12 @@ seccomp
...
 tracelog
 
+# disable-mnt
 private-cache
 private-dev
+private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,gcrypt,gnupg,groups,gtk-2.0,gtk-3.0,hostname,hosts,hosts.conf,mailname,nsswitch.conf,passwd,pki,resolv.conf,selinux,ssl,xdg
 private-tmp
 
-dbus-user none
-dbus-system none
```
rusty-snake, line 62: We can keep `dbus-system none` here.

etc/profile-m-z/neomutt.profile
bbhtt: It is part of disable-common.inc #318?

rusty-snake: your right :100:

kmk3:
> About the sorts except dbus: IMO it is better to consider macros except Home/
> as their own section. Otherwise if the whitelists get too long it'll be easy
> to loose them since Downloads/Documents will be in the top then Home/... then
> Pictures,Videos etc. people most often need to tweak these. I'm fine either
> way.

Ah, that's right; I forgot about those.

Considering that, putting `${HOME}` entries first makes more sense now, as
`${DOCUMENTS}`, `${PICTURES}`, etc don't need `mkdir`, so the sorting would
still work from the first `mkdir` until the last `whitelist ${HOME}`.

Anyway, there was one last attempted sort on `email-common.profile`, but GitHub
failed with:

> Applying suggestions on deleted lines is not supported.

So here is the sorted version:

```
mkdir ${HOME}/.gnupg
mkfile ${HOME}/.config/mimeapps.list
mkfile ${HOME}/.signature
whitelist ${HOME}/.config/mimeapps.list
whitelist ${HOME}/.gnupg
whitelist ${HOME}/.signature
whitelist ${DOCUMENTS}
whitelist ${DOWNLOADS}
# when storing mail outside the default ${HOME}/Mail path, 'whitelist' the custom path in your email-common.local
whitelist ${HOME}/Mail
```

Other than that, LGTM.

kmk3:
```console
$ grep -Fnr 'mkfile ${HOME}/.elinks' etc
etc/profile-m-z/neomutt.profile:68:mkfile ${HOME}/.elinks
etc/profile-m-z/mutt.profile:67:mkfile ${HOME}/.elinks

$ grep -Fnr 'mkfile ${HOME}/.w3m' etc
etc/profile-m-z/neomutt.profile:78:mkfile ${HOME}/.w3m
etc/profile-m-z/mutt.profile:76:mkfile ${HOME}/.w3m
```

These should be `mkdir`.

kmk3:
Hello, I have a pile of changes that are based on this branch (the WIP branch
mentioned [here][1]) and are almost ready for a PR, so I'm wondering what is
the current status of this PR.

[1]: https://github.com/netblue30/firejail/pull/3849#issuecomment-752222345

etc/profile-m-z/neomutt.profile
```diff
@@ -0,0 +1,153 @@
...
+private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,gcrypt,gnupg,hostname,hosts,hosts.conf,mail,mailname,Mutt,Muttrc,Muttrc.d,neomuttrc,neomuttrc.d,nntpserver,nsswitch.conf,passwd,pki,resolv.conf,ssl,xdg
+private-tmp
+writable-run-user
+writable-var
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+
```
rusty-snake, line 149:
```suggestion
```

etc/profile-m-z/mutt.profile
```diff
@@ -55,8 +126,24 @@ nou2f
...
+private-etc alternatives,ca-certificates,crypto-policies,fonts,gai.conf,gcrypt,gnupg,gnutls,hostname,hosts,hosts.conf,mail,mailname,Mutt,Muttrc,Muttrc.d,nntpserver,nsswitch.conf,passwd,pki,resolv.conf,ssl,terminfo,xdg
+private-tmp
 writable-run-user
 writable-var
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+
```
rusty-snake, line 145:
```suggestion
```

etc/profile-m-z/mutt.profile
```diff
@@ -34,15 +38,82 @@ noblacklist ${HOME}/sent
 blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}/wayland-*
 
+include allow-perl.inc
+include allow-python2.inc
+include allow-python3.inc
```
rusty-snake, line 43: Are they needed? mutt.profile has `disable-interpreters.inc` since 3 years.

etc/profile-a-l/geary.profile
```diff
@@ -4,28 +4,83 @@
...
+seccomp.block-secondary
+shell none
+tracelog
+
+# disable-mnt
+# Add ignore private-bin to geary.local for hyperlink support
+private-bin geary
+private-cache
+private-dev
+private-etc alternatives,ca-certificates,crypto-policies,fonts,hostname,hosts,pki,resolv.conf,selinux,ssl,xdg
```
rusty-snake, line 73:
```suggestion
private-etc alternatives,ca-certificates,crypto-policies,fonts,hostname,hosts,pki,resolv.conf,ssl,xdg
```

etc/profile-a-l/geary.profile
```diff
@@ -4,28 +4,83 @@
...
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+
+# disable-mnt
+# Add ignore private-bin to geary.local for hyperlink support
```
rusty-snake, line 69:
```suggestion
# Add 'ignore private-bin' to geary.local for hyperlink support
```

etc/profile-a-l/email-common.profile
```diff
@@ -51,22 +59,27 @@ nou2f
...
 seccomp
+seccomp.block-secondary
 shell none
 tracelog
 
+# disable-mnt
 private-cache
 private-dev
+private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,gcrypt,gnupg,groups,gtk-2.0,gtk-3.0,hostname,hosts,hosts.conf,mailname,nsswitch.conf,passwd,pki,resolv.conf,selinux,ssl,xdg
 private-tmp
```
rusty-snake, line 71:
```suggestion
```

etc/profile-a-l/email-common.profile
```diff
@@ -17,28 +18,35 @@ noblacklist ${DOCUMENTS}
...
 mkfile ${HOME}/.signature
 whitelist ${HOME}/.config/mimeapps.list
+whitelist ${HOME}/.mozilla/firefox/profiles.ini
 whitelist ${HOME}/.gnupg
 whitelist ${HOME}/.signature
+whitelist ${DOCUMENTS}
+whitelist ${DOWNLOADS}
 # when storing mail outside the default ${HOME}/Mail path, 'whitelist' the custom path in your email-common.local
 whitelist ${HOME}/Mail
+
```
rusty-snake, line 38:
```suggestion
```

rusty-snake (commented):
Some suggestions (mostly styling). I still think a redirect profile for neomutt would be good, but this can be done in a future PR.

bbhtt:
> I still think a redirect profile for neomutt would be good, but this can be done in a future PR.

Yea this one is too long I'll add it in the next one.

etc/profile-m-z/mutt.profile
bbhtt: These are some optional functionality for Oauth etc. I will comment them.

netblue30:
All going in, thanks!

kmk3:
@bbhtt

Hello, I'm working on adding more paths to private-etc groups (see #6400) and I
noticed that most profiles use `host.conf` in `private-etc`, while
email-related profiles use `hosts.conf` (which seems to have been added in this
PR).

Is this a typo or do these programs actually use `hosts.conf`?

bbhtt:
Most likely a typo, but it's been a long time and I don't remember much.

kmk3:
> Most likely a typo, but it's been a long time and I don't remember much.

That makes sense; I didn't find any mention of `hosts.conf` anywhere, but you
never know.

I'll convert them to `host.conf`.

Thanks for the quick reply.

Issue #2934, opened 2019-08-28
join fails with private-bin and an alternate (non-bash/sh) shell as default


veloute:
as per https://github.com/netblue30/firejail/pull/2633#issuecomment-515280729
> i've run into the problem of joining working about 40% of the time; out of the few i randomly just tested, keepassxc, mpv, hexchat and flameshot won't allow me to join (always with "execvp: No such file or directory" as the error) with zsh as the default shell.

glitsj16:
@veloute The profiles you mention (keepassxc, mpv, hexchat and flameshot) all have a restricted `private-bin`, i.e. without including binaries zsh needs (or any other shell for that matter). Have you tried adding `private-bin <whatever-zsh-needs>` in foo.local yet? Another option is to add `ignore private-bin` (also in a foo.local file), but that will make the profile(s) slightly less tight... Perhaps it's not a bad idea to implement this whenever `join` is used with a profile containing `private-bin`.

veloute:
that works, but we 
should zsh be added to every private-bin line that has sh and bash or could something else be done? adding zsh to every line is the messier but only way i could think of that fixes the issue and doesn't potentially cause major holes in the profiles.

glitsj16:
> should zsh be added to every private-bin line that has sh and bash or could something else be done?

Until we fix this in the join option itself, that is your best bet I'm afraid. Marking this as a bug, so it might get more close attention. Perhaps you could change the issue to something like 'join fails with restrictive private-bin' or whatever you feel is appropriate. IMHO it's not only a zsh issue per se.

veloute:
>IMHO it's not only a zsh issue per se.
that's what i was thinking - users using alternate shells such as fish, csh, etc. would all (most likely) experience the same problem.

how about "join fails with private-bin and an alternate (non-bash/sh) shell as default"?

glitsj16:
That sounds perfect :+1: .

rusty-snake:
My shell:  zsh
Here is an workaround:
```
[ Terminal 1 ]$ firejail --noprofile --name=test --private-bin=sleep,sl --shell=none sleep 5m
[ Terminal 2 ]$ firejail --shell=none --join=test sl
                         (  ) (@@) ( )  (@)  ()    @@    O     @     O     @      O
                     (@@@)
                 (    )
              (@@@@)

            (   )
        ====        ________                ___________
    _D _|  |_______/        \__I_I_____===__|_________|
[...]

PR #3850, opened 2020-12-28
join: add fexecve fallback for shells

smitsohu:
Allows users to join a sandbox and get a shell even if there is none in the sandbox mount namespace.

Limitations are described in https://github.com/netblue30/firejail/issues/2934#issuecomment-747819719


netblue30:
merged. thanks!

PR #3852, opened 2020-12-29
Implement netns in profiles, closes #3846

### LINKED ISSUE #3846: Not able to use `netns` configuration directive in .profile or .local files
**Bug and expected behavior**
```bash
$ cat /home/faern/.config/firejail/foobar.profile
netns whatever
$ firejail --profile=foobar
Reading profile /home/faern/.config/firejail/foobar.profile
Error: line 1 in /home/faern/.config/firejail/foobar.profile is invalid
```

I can use `--netns=whatever` from the command line. But if I put `netns anythinggoes` in a profile it says that line of the configuration is invalid. I would expect to be able to set up a profile that persistently joins a network namespace. For many of my jailed programs I want them to always join a given namespace, but if I can't specify that in my `program.local` override I will have to have separate launcher scripts for them, which is inconvenient and complicates the setup.

**Environment**
Fedora 33 Workstation
```
$ firejail --version
firejail version 0.9.64

Compile time support:
	- AppImage support is enabled
	- chroot support is enabled
	- SELinux support is enabled
```

rusty-snake:


netblue30:
merged, thanks!

Issue #3809, opened 2020-12-11
Relationship between disable-mnt and disable-write.inc


glitsj16:
I am wondering if I understand the difference/relationship between `disable-mnt` and (including) `disable-write.inc` correctly. The former blacklists /media, /mnt, /run/media and /run/mount access, while the latter makes those exact same directories read-only. Shouldn't it make sense to ONLY include `disable-write.inc` in profiles that DO NOT have `disable-mnt`?

If so, we might have to look at a few profiles that mix them up non-sensically, at least for me:

- BAD:
  - dbus-send.profile
  - notify-send.profile
  - qrencode.profile

- GOOD
  - eo-common.profile

rusty-snake:
>  I am wondering if I understand the difference/relationship between disable-mnt and (including) disable-write.inc correctly. The former blacklists /media, /mnt, /run/media and /run/mount access, while the latter makes those exact same directories read-only.

Yup, that's it.

> Shouldn't it make sense to ONLY include disable-write.inc in profiles that DO NOT have disable-mnt?

This makes the progiles easier, as disable-write-mnt has no effect.

glitsj16:
> This makes the progiles easier, as disable-write-mnt has no effect.

Makes more sense now. Thanks!

PR #3853, opened 2020-12-29
New profile for CoyIM

botherder:
I have created here a first profile for CoyIM. I have tried to make it as restrictive as I could and follow the guidelines of the template. It is however the first firejail profile I create, so any input or feedback would be very welcome.

### DIFF
--- a/etc/profile-a-l/coyim.profile
+++ b/etc/profile-a-l/coyim.profile
@@ -20,6 +20,9 @@ include disable-xdg.inc
 mkdir ${HOME}/.config/coyim
 whitelist ${HOME}/.config/coyim
 include whitelist-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-runuser-common.inc
+include whitelist-var-common.inc
 
 caps.drop all
 netfilter
--- a/etc/profile-a-l/coyim.profile
+++ b/etc/profile-a-l/coyim.profile
@@ -15,11 +15,11 @@ include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-shell.inc
-include disable-write-mnt.inc
 include disable-xdg.inc
 
 mkdir ${HOME}/.config/coyim
 whitelist ${HOME}/.config/coyim
+include whitelist-common.inc
 
 caps.drop all
 netfilter
@@ -37,10 +37,10 @@ tracelog
 disable-mnt
 private-cache
 private-dev
-private-etc alternatives,ca-certificates,crypto-policies,fonts,ssl
+private-etc alternatives,ca-certificates,crypto-policies,fonts,machine-id,pki,ssl
 private-tmp
 
 dbus-user none
 dbus-system none
 
-memory-deny-write-execute
+#memory-deny-write-execute
--- a/etc/profile-a-l/coyim.profile
+++ b/etc/profile-a-l/coyim.profile
@@ -39,3 +39,8 @@ private-cache
 private-dev
 private-etc alternatives,ca-certificates,crypto-policies,fonts,ssl
 private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
--- a/etc/profile-a-l/coyim.profile
+++ b/etc/profile-a-l/coyim.profile
@@ -1,0 +1,41 @@
+# Firejail profile for coyim
+# Description: GTK Jabber client written in Go
+# This file is overwritten after every install/update
+# Persistent local customizations
+include coyim.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/coyim
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/coyim.profile
```diff
@@ -0,0 +1,46 @@
...
+disable-mnt
+private-cache
+private-dev
+private-etc alternatives,ca-certificates,crypto-policies,fonts,ssl
+private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
```
rusty-snake, line 46: mdwe + GTK3 + Wayland bites. Probably better to comment.

```suggestion
#memory-deny-write-execute
```

etc/profile-a-l/coyim.profile
```diff
@@ -0,0 +1,46 @@
...
+nou2f
+protocol unix,inet,inet6
+seccomp
+shell none
+tracelog
+
+disable-mnt
+private-cache
+private-dev
+private-etc alternatives,ca-certificates,crypto-policies,fonts,ssl
```
rusty-snake, line 40: - add `pki`. If one of `ca-certificates,crypto-policies,pki,ssl` is required, all should be used because tls-config differs between distros.
- add `machine-id`, pulseaudio does not work without it.
```suggestion
private-etc alternatives,ca-certificates,crypto-policies,fonts,machine-id,pki,ssl
```

etc/profile-a-l/coyim.profile
```diff
@@ -0,0 +1,46 @@
...
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/coyim
+whitelist ${HOME}/.config/coyim
```
SkewedZeppelin, line 22: add
`include whitelist-common.inc`

test if these work too
```
```

etc/profile-a-l/coyim.profile
```diff
@@ -0,0 +1,46 @@
...
+noblacklist ${HOME}/.config/coyim
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
```
rusty-snake, line 18: see discussion in #3809
```suggestion

```

etc/profile-a-l/coyim.profile
rusty-snake, line 22: `include whitelist-common.inc` should always be used in whitelisting profiles.
```suggestion
whitelist ${HOME}/.config/coyim
include whitelist-common.inc
```

etc/profile-a-l/coyim.profile
```diff
@@ -0,0 +1,46 @@
+# Firejail profile for coyim
+# Description: GTK Jabber client written in Go
+# This file is overwritten after every install/update
+# Persistent local customizations
+include coyim.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/coyim
```
SkewedZeppelin, line 9: `blacklist ${HOME}/.config/coyim` needs to be added to `etc/disable-programs.inc`

SkewedZeppelin (changes requested):
mostly good aside from minor things

rusty-snake (commented):
> I have tried to make it as restrictive as I could and follow the guidelines of the template. It is however the first firejail profile I create, so any input or feedback would be very welcome.

It's always better to make more review steps and have a profile :smile: . So here're my inputs.

- Can we use `include whitelist-runuser-common.inc`, `include whitelist-usr-share-common.inc`, `include whitelist-var-common.inc`?
- Add `blacklist ${HOME}/.config/coyim` to `disable-programs.inc`
- Add `coyim` to `src/firecfg/firecfg.config`.

botherder:
I have added the suggested changes.

botherder:
> It's always better to make more review steps and have a profile  . So here're my inputs.

Thank you for your feedback.

> * Can we use `include whitelist-runuser-common.inc`, `include whitelist-usr-share-common.inc`, `include whitelist-var-common.inc`?

What's the rulebook regarding the inclusion of whitelist-*? Isn't it preferable to avoid adding additional inclusions if they are not necessary to the proper functioning of the application?

rusty-snake:
> What's the rulebook regarding the inclusion of whitelist-*? Isn't it preferable to avoid adding additional inclusions if they are not necessary to the proper functioning of the application?

`whitelist-common.inc` (wc): Should always be used for whitelisting profiles. It provides things such as bookmarks in file-selection dialogs and custom GTK/QT themes, icon themes, .... (exception: commandline-program which only read `~/.foorc`)
`whitelist-runuser-common.inc` (wruc), `whitelist-usr-share-common.inc` (wusc) and `whitelist-var-common.inc` (wvc): including these profiles add restrictions because they enabled whitelisting in these locations.

rusty-snake:
Had you found time to test CoyIM with wruc, wusc and wvc.

botherder:
Yep, works fine.

rusty-snake (approved):

netblue30:
merged, thanks!

PR #3859, opened 2020-12-31
really fix running kernel config check

glitsj16:
Our archiver-common.inc includes `disable-shell.inc`, breaking running kernel config check with zcat or zgrep.

### DIFF
--- a/etc/profile-m-z/zcat.profile
+++ b/etc/profile-m-z/zcat.profile
@@ -8,6 +8,8 @@ include zcat.local
 #include globals.local
 
 # Allow running kernel config check
+noblacklist ${PATH}/bash
+noblacklist ${PATH}/sh
 noblacklist /proc/config.gz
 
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/zcat.profile
```diff
@@ -8,6 +8,8 @@ include zcat.local
 #include globals.local
 
 # Allow running kernel config check
+noblacklist ${PATH}/bash
```
reinerh, line 11: it looks like that assumes that bash is the shell used with /bin/sh. shouldn't also other shells be allowed? because Debian uses dash as default shell for /bin/sh.

glitsj16: I forgot about Debian using dash. To cover all setups it's better to ignore including disable-shell.inc IMO. Thanks for catching this! A fix is waiting for the checks to finish.

Issue #3741, opened 2020-11-11
Error: failed to run /run/firejail/lib/fcopy


reinerh:
A user [reported](https://bugs.debian.org/973756) issues with starting transmission-cli with firejail, and I'm able to reproduce it.
```
$ firejail transmission-cli 
/run/firejail/lib/fcopy: error while loading shared libraries: libpcre2-8.so.0: cannot open shared object file: No such file or directory
Error: failed to run /run/firejail/lib/fcopy
Error: proc 29091 cannot sync with peer: unexpected EOF
Peer 29092 unexpectedly exited with status 1

$ firejail --profile=transmission-cli /bin/bash
/run/firejail/lib/fcopy: error while loading shared libraries: libpcre2-8.so.0: cannot open shared object file: No such file or directory
Error: failed to run /run/firejail/lib/fcopy
Error: proc 29376 cannot sync with peer: unexpected EOF
Peer 29377 unexpectedly exited with status 1
```

I also see the same error a couple of times with different profiles when running the test-profiles part of the test suite.

reinerh:
Seems to be related to a combination of private-bin and private-etc.
When commenting out either of them, the process is starting.

reinerh:
Commenting out private-lib in transmission-common.profile also helps.

kortewegdevries:
Yea I observed the same on Sid. I suggest to comment out `private-lib` in `transmission-common.profile` since it works fine on Arch without any modification.

reinerh:
That would workaround it for transmission only. But the same issue happens for other profiles as well.

rusty-snake:
I've `private-lib libpcre2-8.so.0` in less.local and whois.local. Actually we should always add it to `private-lib` via it's implementation.

reinerh:
`libpcre2-8.so.0` seems to come via the linking to `libselinux.so.1`.
maybe `libpcre2-8.so.` needs to be added to `src/firejail/fs_lib2.c`?

kortewegdevries:
Ah, ok I wasn't aware of that.

reinerh:
@smitsohu This issue seems to be back.
I upgraded to 0.9.64.2 and now have this problem again:
```
$ firejail transmission-cli 
/run/firejail/lib/fcopy: error while loading shared libraries: libpcre2-8.so.0: cannot open shared object file: No such file or directory
Error: failed to run /run/firejail/lib/fcopy
Error: proc 6495 cannot sync with peer: unexpected EOF
[...]

PR #3867, opened 2021-01-04
return to non-dumpable plugins

smitsohu:
(hopefully) fixes the issues that led to reverting
commits 6abb65d328af61d67361890743190bd4c57f8e3c and 98e42dc6da4e4b1e47ed2aa020012d4dedc1e80e


smitsohu:
The tests uncovered a weird bug that exists in Debian/Ubuntu but as it seems not in Fedora.

smitsohu:
A minimal reproducer looks like this:
`firejail --private-etc=passwd --private-bin=doesntexist --private-lib anything`
or
`firejail --debug --private-bin=doesntexist --private-lib anything`
with fsec-print failing instead of fcopy.

Most probably a problem created by `private-lib`. 

Unfortunately this also causes a regression on #3741, as fldd cannot work without read permission on the binary.

PR #3873, opened 2021-01-06
harden liferea

glitsj16:
- enable novideo
- dbus filtering

### DIFF
--- a/etc/profile-a-l/liferea.profile
+++ b/etc/profile-a-l/liferea.profile
@@ -42,7 +42,7 @@ noroot
 # nosound
 notv
 nou2f
-# novideo
+novideo
 protocol unix,inet,inet6
 seccomp
 shell none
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/liferea.profile
```diff
@@ -51,3 +51,11 @@ tracelog
 disable-mnt
 private-dev
 private-tmp
+
+dbus-user filter
+dbus-user.own net.sourceforge.liferea
+dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.freedesktop.Notifications
+dbus-user.talk org.freedesktop.secrets
+dbus-user.talk org.gnome.Shell.Notifications
```
rusty-snake, line 60: `org.gnome.Shell.Notifications`? `org.freedesktop.Notifications` should be enough, it don't?

glitsj16: The legend in https://github.com/netblue30/firejail/wiki/Restrict-D-Bus confused me here. It has a note that org.freedesktop.Notifications is ￼ for GNOME >= 3.36.1 and I wanted to make sure liferea notifications work regardless of GNOME version. Let me do a quick test without `dbus-user.talk org.gnome.Shell.Notifications`

glitsj16: @rusty-snake org.gnome.Shell.Notifications is indeed unneeded. I also opted to tighten the dbus filtering and adding comments. What do you think?

PR #3876, opened 2021-01-08
refactor nodejs applications (npm & yarn)

glitsj16:
This PR does the following:
- add new yarn.profile
- add new nodejs-common.profile (common for npm & yarn)
- refactor npm.profile

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,4 +195,4 @@ Stats:
 
 ### New profiles:
 
-spectacle, chromium-browser-privacy, gtk-straw-viewer, gtk-youtube-viewer, gtk2-youtube-viewer, gtk3-youtube-viewer, straw-viewer, lutris, dolphin-emu, authenticator-rs, servo, tutanota-desktop, npm, marker
+spectacle, chromium-browser-privacy, gtk-straw-viewer, gtk-youtube-viewer, gtk2-youtube-viewer, gtk3-youtube-viewer, straw-viewer, lutris, dolphin-emu, authenticator-rs, servo, tutanota-desktop, npm, marker, yarn
--- a/etc/inc/allow-common-devel.inc
+++ b/etc/inc/allow-common-devel.inc
@@ -11,6 +11,15 @@ noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.gradle
 noblacklist ${HOME}/.java
 
+# Node.js
+noblacklist ${HOME}/.node-gyp
+noblacklist ${HOME}/.npm
+noblacklist ${HOME}/.npmrc
+noblacklist ${HOME}/.yarn
+noblacklist ${HOME}/.yarn-config
+noblacklist ${HOME}/.yarncache
+noblacklist ${HOME}/.yarnrc
+
 # Python
 noblacklist ${HOME}/.pylint.d
 noblacklist ${HOME}/.python-history
@@ -25,7 +34,3 @@ noblacklist ${HOME}/.cargo/registry
 noblacklist ${HOME}/.cargo/.crates.toml
 noblacklist ${HOME}/.cargo/.crates2.json
 noblacklist ${HOME}/.cargo/.package-cache
-
-# npm
-noblacklist ${HOME}/.npm
-noblacklist ${HOME}/.npmrc
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -310,6 +310,7 @@ read-only ${HOME}/.msmtprc
 read-only ${HOME}/.mutt/muttrc
 read-only ${HOME}/.muttrc
 read-only ${HOME}/.nano
+read-only ${HOME}/.npmrc
 read-only ${HOME}/.pythonrc.py
 read-only ${HOME}/.reportbugrc
 read-only ${HOME}/.tmux.conf
@@ -318,6 +319,7 @@ read-only ${HOME}/.viminfo
 read-only ${HOME}/.vimrc
 read-only ${HOME}/.xmonad
 read-only ${HOME}/.xscreensaver
+read-only ${HOME}/.yarnrc
 read-only ${HOME}/_exrc
 read-only ${HOME}/_gvimrc
 read-only ${HOME}/_vimrc
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -760,6 +760,7 @@ blacklist ${HOME}/.neverball
 blacklist ${HOME}/.newsbeuter
 blacklist ${HOME}/.newsboat
 blacklist ${HOME}/.nicotine
+blacklist ${HOME}/.node-gyp
 blacklist ${HOME}/.npm
 blacklist ${HOME}/.npmrc
 blacklist ${HOME}/.nv
@@ -848,6 +849,10 @@ blacklist ${HOME}/.xmr-stak
 blacklist ${HOME}/.xonotic
 blacklist ${HOME}/.xournalpp
 blacklist ${HOME}/.xpdfrc
+blacklist ${HOME}/.yarn
+blacklist ${HOME}/.yarn-config
+blacklist ${HOME}/.yarncache
+blacklist ${HOME}/.yarnrc
 blacklist ${HOME}/.zoom
 blacklist /tmp/akonadi-*
 blacklist /tmp/ssh-*
--- a/etc/profile-m-z/nodejs-common.profile
+++ b/etc/profile-m-z/nodejs-common.profile
@@ -1,0 +1,55 @@
+# Firejail profile for Node.js
+# Description: Common profile for npm/yarn
+quiet
+# This file is overwritten after every install/update
+# Persistent local customizations
+include nodejs-common.local
+# Persistent global definitions
+# added by caller profile
+#include globals.local
+
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}
+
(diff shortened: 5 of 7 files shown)


kmk3 (commented):
> ```diff
> diff --git a/etc/inc/allow-common-devel.inc b/etc/inc/allow-common-devel.inc
> index 68e91a09b..41643657d 100644
> --- a/etc/inc/allow-common-devel.inc
> +++ b/etc/inc/allow-common-devel.inc
> @@ -11,6 +11,15 @@ noblacklist ${HOME}/.git-credentials
>  noblacklist ${HOME}/.gradle
>  noblacklist ${HOME}/.java
> 
> +# Node.js
> +noblacklist ${HOME}/.node-gyp
> +noblacklist ${HOME}/.npm
> +noblacklist ${HOME}/.npmrc
> +noblacklist ${HOME}/.yarn
> +noblacklist ${HOME}/.yarn-config
> +noblacklist ${HOME}/.yarncache
> +noblacklist ${HOME}/.yarnrc
> +
>  # Python
>  noblacklist ${HOME}/.pylint.d
>  noblacklist ${HOME}/.python-history
> @@ -25,7 +34,3 @@ noblacklist ${HOME}/.cargo/registry
>  noblacklist ${HOME}/.cargo/.crates.toml
>  noblacklist ${HOME}/.cargo/.crates2.json
>  noblacklist ${HOME}/.cargo/.package-cache
> -
> -# npm
> -noblacklist ${HOME}/.npm
> -noblacklist ${HOME}/.npmrc
> ```

Suggestion (untested):

```diff
diff --git a/etc/inc/disable-common.inc b/etc/inc/disable-common.inc
index d88506d90..0de539d57 100644
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -310,6 +310,7 @@ read-only ${HOME}/.msmtprc
 read-only ${HOME}/.mutt/muttrc
 read-only ${HOME}/.muttrc
 read-only ${HOME}/.nano
+read-only ${HOME}/.npmrc
 read-only ${HOME}/.pythonrc.py
 read-only ${HOME}/.reportbugrc
 read-only ${HOME}/.tmux.conf
@@ -318,6 +319,7 @@ read-only ${HOME}/.viminfo
 read-only ${HOME}/.vimrc
 read-only ${HOME}/.xmonad
 read-only ${HOME}/.xscreensaver
+read-only ${HOME}/.yarnrc
 read-only ${HOME}/_exrc
 read-only ${HOME}/_gvimrc
 read-only ${HOME}/_vimrc
diff --git a/etc/profile-m-z/npm.profile b/etc/profile-m-z/npm.profile
index 29bab4cb9..50b7498e5 100644
--- a/etc/profile-m-z/npm.profile
+++ b/etc/profile-m-z/npm.profile
@@ -21,5 +21,8 @@ noblacklist ${HOME}/.npmrc
 #whitelist ${HOME}/Projects
 #include whitelist-common.inc
 
+ignore read-only ${HOME}/.npm-packages
+ignore read-only ${HOME}/.npmrc
+
 # Redirect
 include nodejs-common.profile
diff --git a/etc/profile-m-z/yarn.profile b/etc/profile-m-z/yarn.profile
index 06a5e0fc2..e507214d3 100644
--- a/etc/profile-m-z/yarn.profile
+++ b/etc/profile-m-z/yarn.profile
@@ -23,5 +23,7 @@ noblacklist ${HOME}/.yarnrc
 #whitelist ${HOME}/Projects
 #include whitelist-common.inc
 
+ignore read-only ${HOME}/.yarnrc
+
 # Redirect
 include nodejs-common.profile
```

Not sure about the rest of the paths.

Adding `ignore read-only` instead of `read-write` because it comes before
disable-common.inc (included in nodejs-common.profile), which uses `read-only`
and the last `read-` wins.

glitsj16:
@kmk3 Thanks for the suggestions! Added.

rusty-snake (commented):
npm.profile has no `quiet`.

kmk3:
@glitsj16 Thanks!

Nitpick: I screwed up the entry location on the suggestion (I had just done
`s/read-write/ignore read-only/`).  Now considering the order on
etc/templates/profile.template:

> ```firejail
> # Sections structure
> #   HEADER
> #   COMMENTS
> #   IGNORES
> #   NOBLACKLISTS
> #   ALLOW INCLUDES
> #   BLACKLISTS
> #   DISABLE INCLUDES
> #   NOWHITELISTS
> #   MKDIRS
> #   WHITELISTS
> #   WHITELIST INCLUDES
> #   OPTIONS (caps*, net*, no*, protocol, seccomp*, shell none, tracelog)
> #   PRIVATE OPTIONS (disable-mnt, private-*, writable-*)
> #   DBUS FILTER
> #   SPECIAL OPTIONS (mdwx, noexec, read-only, join-or-start)
> #   REDIRECT INCLUDES
> ```

Suggestion:

```diff
diff --git a/etc/profile-m-z/npm.profile b/etc/profile-m-z/npm.profile
index 20d3d75d7..e95e875be 100644
--- a/etc/profile-m-z/npm.profile
+++ b/etc/profile-m-z/npm.profile
@@ -7,11 +7,12 @@ include npm.local
 # Persistent global definitions
 include globals.local

+ignore read-only ${HOME}/.npm-packages
+ignore read-only ${HOME}/.npmrc
+
 noblacklist ${HOME}/.node-gyp
 noblacklist ${HOME}/.npm
 noblacklist ${HOME}/.npmrc
-ignore read-only ${HOME}/.npm-packages
-ignore read-only ${HOME}/.npmrc

 # If you want whitelisting, change ${HOME}/Projects below to your npm projects directory
 # and uncomment the lines below.
diff --git a/etc/profile-m-z/yarn.profile b/etc/profile-m-z/yarn.profile
index a67e76f9b..f20225050 100644
--- a/etc/profile-m-z/yarn.profile
+++ b/etc/profile-m-z/yarn.profile
@@ -6,11 +6,12 @@ include yarn.local
 # Persistent global definitions
 include globals.local

+ignore read-only ${HOME}/.yarnrc
+
 noblacklist ${HOME}/.yarn
 noblacklist ${HOME}/.yarn-config
 noblacklist ${HOME}/.yarncache
 noblacklist ${HOME}/.yarnrc
-ignore read-only ${HOME}/.yarnrc

 # If you want whitelisting, change ${HOME}/Projects below to your yarn projects directory and uncomment the lines below.
 #mkdir ${HOME}/.yarn
```

Many profiles seem to use it this way as well.  Example from
etc/profile-m-z/youtube-dl.profile:

```firejail
include globals.local

# breaks when installed under ${HOME} via `pip install --user` (see #2833)
ignore noexec ${HOME}

noblacklist ${HOME}/.cache/youtube-dl
noblacklist ${HOME}/.netrc
noblacklist ${MUSIC}
noblacklist ${VIDEOS}
```

glitsj16:
> Many profiles seem to use it this way as well.

@kmk3 Indeed they do, so I'll move things around for consistency's sake, even though it makes more sense to me to follow the same (vertically trickling up/down) order in overrides as in what the profile includes. But I've been called neurotic before :dancers: on several occasions by people close to me.

kmk3:
> > Many profiles seem to use it this way as well.
> 
> @kmk3 Indeed they do, so I'll move things around for consistency's sake, even
> though it makes more sense to me to follow the same (vertically trickling
> up/down) order in overrides as in what the profile includes.

To me inconsistency is usually a bigger distraction than the quality of the
ordering :p

It makes me ponder whether the deviation is intentional or not, and if so,
whether it has functional differences, or if it's just stylistic.  And in the
latter's case, when/why is the exception acceptable and if there are other
similar instances, so that if it's implicitly part of the coding standard, I
can make sure to follow it as well.

Anyways, if the structure ordering is bad enough (I have no opinion on this
particular case), especially if it interferes functionally, then I'd consider
making a case for changing it on the template and on every profile.

> But I've been called neurotic before dancers on several occasions by people
> close to me.

I'm sure some people would say the same about me, but if no one cared about
cleanliness, then the profiles would probably look a lot worse and be a pain to
work with.  I have ran into bumps before when following the ordering of the
template, but oftentimes that also helped me understand why some directives are
placed before others.  Additionally, the fact that it exists and is followed
enough is a blessing to me.

I think "neurotic" is a matter of how (personal/interpersonal) conflicts are
handled.  There were times when I have certainly pushed for some things more
than it made sense to as a maintainer on a company project.  I guess it takes
practice and effort to get closer to the ideal spot.  But either way, I always
try to point out if something does not look quite right, however small it may
be.  If the suggestion doesn't get through, at least I feel like I did my part.
In that sense, being a reviewer (i.e.: pointing out problems) is easy, being a
maintainer (i.e.: choosing what to veto) is not.

rusty-snake:
> Many profiles seem to use it this way as well.

https://github.com/netblue30/firejail/blob/37452ef1a71473b87431c3c708d3b31ca1b7a25f/etc/templates/profile.template#L25-L26

> It makes me ponder whether the deviation is intentional or not

There is just no CI that raps you on the knuckles. 

> whether it has functional differences, or if it's just stylistic. 

As long as the `ignore foobar` comes before `foobar` it's just stylistic.

> And in the latter's case, when/why is the exception acceptable

If you have "an own logical block" which would split comments at (more then two) different places in the profile. Examples are thunderbird.profile or spectacle.profile

https://github.com/netblue30/firejail/blob/37452ef1a71473b87431c3c708d3b31ca1b7a25f/etc/profile-m-z/spectacle.profile#L9-L13

PR #3879, opened 2021-01-10
Whitelist Bohemia Interactive config dir for Steam

aidalgol:
At least Arma 3 stores its config directory under
~/.local/share/bohemiainteractive

Without this, Arma 3's configuration and game save do not persist across Steam firejail sessions.

### DIFF
--- a/etc/profile-m-z/steam.profile
+++ b/etc/profile-m-z/steam.profile
@@ -9,6 +9,7 @@ include globals.local
 noblacklist ${HOME}/.killingfloor
 noblacklist ${HOME}/.local/share/3909/PapersPlease
 noblacklist ${HOME}/.local/share/aspyr-media
+noblacklist ${HOME}/.local/share/bohemiainteractive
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/steam.profile
```diff
@@ -9,6 +9,7 @@ include globals.local
 noblacklist ${HOME}/.killingfloor
 noblacklist ${HOME}/.local/share/3909/PapersPlease
 noblacklist ${HOME}/.local/share/aspyr-media
+noblacklist ${HOME}/.local/share/bohemiainteractive
```
SkewedZeppelin, line 12: can you add the corresponding
`blacklist ${HOME}/.local/share/bohemiainteractive`
to etc/disable-programs.inc

aidalgol: Added.

SkewedZeppelin:
Merged, thanks!

Issue #3845, opened 2020-12-27
On squashing commits and git workflows


kmk3:
Recently I opened two Pull Requests which were integrated using GitHub's
"squash and merge" method.  This causes GitHub to replace the author date
(i.e.: when the changes were made) of the original commit with "the date that
the PR was merged" (which is what the commit date is for) and to integrate the
resulting commit.  I find historical value in preserving the original commit
date, as the registry of when the patch was made and it can also be considered
part of its context.  Thus, replacing it with an unrelated value pretending to
be the actual author date bothers me.  As an extreme example, imagine merging a
PR this way a year after it was opened.  Not only would the date be way off
than what was intended, but also what made sense at the time of change might
not make as much sense at a later time (e.g.: using functions which later
became deprecated).  Moreover, note that this is a GitHub-specific deficiency;
`git merge --squash` at least puts the metadata in the commit message.  But
still, ideally the author would be the one to squash their own commits.

Further, when a branch has more than 1 commit, there are additional problems
with the method:

* GitHub mangles the commit subject (uses PR title)
* Commits that were once a single logical change each become a big blob of
  changes

(This hasn't happened to any of my PRs, but since it's not documented and since
it's been done to single commits, I assume that it can happen to larger
branches)

I see that squash-merge has been used in a few PRs as a quick way to turn many
amend commits into a singular one (like a web-based `git rebase`).  Which is
fine when used in your own branches (and I can understand why it would be done
if a PR is opened with many amend commits), but doing it to someone else's
branch without notice would be very rude, as in my case you'd be integrating
under my name a patch other than the ones I've sent, as if I had written and
submitted that tweaked version myself.

I personally try to write "[good][1]" commits (e.g.: clarify context, ensure 1
commit per logical change), so that 6 months later I can quickly understand the
what/why by just reading each commit locally, without depending on external
sources (and possibly refer to it when I need to do something similar).  That
usually entails many local rebases before sending the patches, so if such
branch-squashing happened to me, it would be a punch to the face.

With that said, I ask three things:

1. Ask for changes instead of directly changing someone else's commits (yes, I
   know it's tempting to simply do that)
2. Use "squash and merge" only after the author(s) agreed to it
3. Document on e.g.: `CONTRIBUTING.md` that "squash and merge" is part of the
   workflow so that whoever sends patches knows what to expect

Perhaps the latter 2 could be simplified to adding an "I agree to have all of
[...]

PR #3885, opened 2021-01-12
ssh: Refactor, fix bugs & harden

kmk3:
```console
* etc: add allow-ssh.inc
* allow-ssh.inc: allow /etc/ssh/ssh_config
* ssh: deny access to the rest of /etc/ssh/*
* allow-ssh.inc: allow access to ssh-agent(1)
* disable-common.inc: add missing openssh paths
* disable-common.inc: mark ~/.ssh as read-only
```

Note: As mentioned on #3845, please don't squash any commits; use normal merge
or rebase instead.

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -347,6 +347,9 @@ read-only ${HOME}/.local/share/mime
 # Write-protection for thumbnailer dir
 read-only ${HOME}/.local/share/thumbnailers
 
+# prevent access to ssh-agent
+blacklist /tmp/ssh-*
+
 # top secret
 blacklist ${HOME}/*.kdb
 blacklist ${HOME}/*.kdbx
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -291,7 +291,15 @@ read-only ${HOME}/.zshrc
 read-only ${HOME}/.zshrc.local
 
 # Remote access
-read-only ${HOME}/.ssh/authorized_keys
+blacklist ${HOME}/.rhosts
+blacklist ${HOME}/.shosts
+blacklist ${HOME}/.ssh/authorized_keys
+blacklist ${HOME}/.ssh/authorized_keys2
+blacklist ${HOME}/.ssh/environment
+blacklist ${HOME}/.ssh/rc
+blacklist /etc/hosts.equiv
+read-only ${HOME}/.ssh/config
+read-only ${HOME}/.ssh/config.d
 
 # Initialization files that allow arbitrary command execution
 read-only ${HOME}/.caffrc
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -396,6 +396,7 @@ blacklist /etc/shadow
 blacklist /etc/shadow+
 blacklist /etc/shadow-
 blacklist /etc/ssh
+blacklist /etc/ssh/*
 blacklist /home/.ecryptfs
 blacklist /home/.fscrypt
 blacklist /var/backup
--- a/etc/profile-a-l/android-studio.profile
+++ b/etc/profile-a-l/android-studio.profile
@@ -10,11 +10,13 @@ noblacklist ${HOME}/.android
 noblacklist ${HOME}/.jack-server
 noblacklist ${HOME}/.jack-settings
 noblacklist ${HOME}/.local/share/JetBrains
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.tooling
 
 # Allows files commonly used by IDEs
 include allow-common-devel.inc
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 include disable-common.inc
 include disable-passwdmgr.inc
--- a/etc/profile-a-l/aosp.profile
+++ b/etc/profile-a-l/aosp.profile
@@ -11,11 +11,13 @@ noblacklist ${HOME}/.jack-server
 noblacklist ${HOME}/.jack-settings
 noblacklist ${HOME}/.repo_.gitconfig.json
 noblacklist ${HOME}/.repoconfig
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.tooling
 
 # Allows files commonly used by IDEs
 include allow-common-devel.inc
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 include disable-common.inc
 include disable-passwdmgr.inc
--- a/etc/profile-a-l/clion.profile
+++ b/etc/profile-a-l/clion.profile
@@ -11,8 +11,10 @@ noblacklist ${HOME}/.gitconfig
 noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.java
 noblacklist ${HOME}/.local/share/JetBrains
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.tooling
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 include disable-common.inc
 include disable-passwdmgr.inc
--- a/etc/profile-a-l/filezilla.profile
+++ b/etc/profile-a-l/filezilla.profile
@@ -8,11 +8,13 @@ include globals.local
 
 noblacklist ${HOME}/.config/filezilla
 noblacklist ${HOME}/.filezilla
-noblacklist ${HOME}/.ssh
 
 # Allow python (blacklisted by disable-interpreters.inc)
 include allow-python2.inc
 include allow-python3.inc
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 include disable-common.inc
 include disable-devel.inc
--- a/etc/profile-a-l/git-cola.profile
+++ b/etc/profile-a-l/git-cola.profile
@@ -11,7 +11,6 @@ ignore noexec ${HOME}
 noblacklist ${HOME}/.gitconfig
 noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.gnupg
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.subversion
 noblacklist ${HOME}/.config/git
 noblacklist ${HOME}/.config/git-cola
@@ -21,6 +20,9 @@ noblacklist ${HOME}/.config/git-cola
 # Allow python (blacklisted by disable-interpreters.inc)
 include allow-python2.inc
 include allow-python3.inc
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 include disable-common.inc
 include disable-devel.inc
--- a/etc/profile-a-l/git-cola.profile
+++ b/etc/profile-a-l/git-cola.profile
@@ -18,6 +18,7 @@ noblacklist ${HOME}/.config/git-cola
 # Put your editor,diff viewer config path below and uncomment to load settings
 # noblacklist ${HOME}/
 
+# Allow python (blacklisted by disable-interpreters.inc)
 include allow-python2.inc
 include allow-python3.inc
 
--- a/etc/profile-a-l/git.profile
+++ b/etc/profile-a-l/git.profile
@@ -15,9 +15,11 @@ noblacklist ${HOME}/.gitconfig
 noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.gnupg
 noblacklist ${HOME}/.nanorc
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.vim
 noblacklist ${HOME}/.viminfo
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}/wayland-*
--- a/etc/profile-a-l/gitg.profile
+++ b/etc/profile-a-l/gitg.profile
@@ -10,7 +10,9 @@ noblacklist ${HOME}/.config/git
 noblacklist ${HOME}/.gitconfig
 noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.local/share/gitg
-noblacklist ${HOME}/.ssh
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 include disable-common.inc
 include disable-devel.inc
--- a/etc/profile-a-l/idea.sh.profile
+++ b/etc/profile-a-l/idea.sh.profile
@@ -10,11 +10,13 @@ noblacklist ${HOME}/.android
 noblacklist ${HOME}/.jack-server
 noblacklist ${HOME}/.jack-settings
 noblacklist ${HOME}/.local/share/JetBrains
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.tooling
 
 # Allows files commonly used by IDEs
 include allow-common-devel.inc
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 include disable-common.inc
 include disable-passwdmgr.inc
--- a/etc/profile-m-z/meld.profile
+++ b/etc/profile-m-z/meld.profile
@@ -18,7 +18,6 @@ noblacklist ${HOME}/.config/git
 noblacklist ${HOME}/.gitconfig
 noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.local/share/meld
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.subversion
 
 # Allow python (blacklisted by disable-interpreters.inc)
@@ -26,6 +25,9 @@ include allow-python3.inc
 
 # Python 2 is EOL (see #3164). Uncomment the next line (or put it into your meld.local) if you understand the risks but want python 2 support for older meld versions.
 #include allow-python2.inc
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 # Uncomment the next line (or put it into your meld.local) if you don't need to compare files in disable-common.inc.
 #include disable-common.inc
--- a/etc/profile-m-z/seahorse.profile
+++ b/etc/profile-m-z/seahorse.profile
@@ -9,7 +9,6 @@ include globals.local
 blacklist /tmp/.X11-unix
 
 noblacklist ${HOME}/.gnupg
-noblacklist /tmp/ssh-*
 
 # Allow ssh (blacklisted by disable-common.inc)
 include allow-ssh.inc
--- a/etc/profile-m-z/seahorse.profile
+++ b/etc/profile-m-z/seahorse.profile
@@ -9,8 +9,10 @@ include globals.local
 blacklist /tmp/.X11-unix
 
 noblacklist ${HOME}/.gnupg
-noblacklist ${HOME}/.ssh
 noblacklist /tmp/ssh-*
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 include disable-common.inc
 include disable-devel.inc
--- a/etc/profile-m-z/ssh-agent.profile
+++ b/etc/profile-m-z/ssh-agent.profile
@@ -6,7 +6,6 @@ include ssh-agent.local
 # Persistent global definitions
 include globals.local
 
-noblacklist /etc/ssh/*
 noblacklist /tmp/ssh-*
 
 # Allow ssh (blacklisted by disable-common.inc)
--- a/etc/profile-m-z/ssh-agent.profile
+++ b/etc/profile-m-z/ssh-agent.profile
@@ -6,7 +6,7 @@ include ssh-agent.local
 # Persistent global definitions
 include globals.local
 
-noblacklist /etc/ssh
+noblacklist /etc/ssh/*
 noblacklist /tmp/ssh-*
 
 # Allow ssh (blacklisted by disable-common.inc)
--- a/etc/profile-m-z/ssh-agent.profile
+++ b/etc/profile-m-z/ssh-agent.profile
@@ -8,7 +8,9 @@ include globals.local
 
 noblacklist /etc/ssh
 noblacklist /tmp/ssh-*
-noblacklist ${HOME}/.ssh
+
+# Allow ssh (blacklisted by disable-common.inc)
+include allow-ssh.inc
 
 blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}/wayland-*
--- a/etc/profile-m-z/ssh-agent.profile
+++ b/etc/profile-m-z/ssh-agent.profile
@@ -6,8 +6,6 @@ include ssh-agent.local
 # Persistent global definitions
 include globals.local
 
-noblacklist /tmp/ssh-*
-
 # Allow ssh (blacklisted by disable-common.inc)
 include allow-ssh.inc
 
--- a/etc/profile-m-z/ssh.profile
+++ b/etc/profile-m-z/ssh.profile
@@ -7,7 +7,6 @@ include ssh.local
 # Persistent global definitions
 include globals.local
 
-noblacklist /tmp/ssh-*
 # nc can be used as ProxyCommand, e.g. when using tor
 noblacklist ${PATH}/nc
 noblacklist ${PATH}/ncat
--- a/etc/profile-m-z/ssh.profile
+++ b/etc/profile-m-z/ssh.profile
@@ -7,7 +7,7 @@ include ssh.local
 # Persistent global definitions
 include globals.local
 
-noblacklist /etc/ssh
+noblacklist /etc/ssh/*
 noblacklist /tmp/ssh-*
 # nc can be used as ProxyCommand, e.g. when using tor
 noblacklist ${PATH}/nc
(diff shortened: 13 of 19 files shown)


glitsj16:
LGTM. PR title seems to imply the current ssh profiles are broken. If this is not the case, please consider changing it to something like 'Refactor & harden ssh' or whatever you prefer.

glitsj16 (approved):
Nice job. I've been running with these extra options added to my ssh-agent.local:
- private-cache
- private-dev
- memory-deny-write-execute
While you're working on our SSH profiles, can you test if those can be integrated into ssh-agent.profile? Especially mdrw would be a nice hardening to have IMO.

rusty-snake (approved):
Looks great so far. Can you add it to etc/tempaltes/profile.template too.

kmk3:
(I just noticed that there are some things missing, so I converted it to draft for now)

kmk3:
> (I just noticed that there are some things missing, so I converted it to
> draft for now)

I'm working on adding the missing paths to disable-programs.inc.

@glitsj16 commented 22 hours ago:

> LGTM. PR title seems to imply the current ssh profiles are broken. If this is
> not the case, please consider changing it to something like 'Refactor &
> harden ssh' or whatever you prefer.

Ah, I didn't know what to put, so I just left in the branch name.  The "fix"
part is mostly about allowing ssh to access `/etc/ssh/ssh_config`.  But yeah,
the title did sound a bit dramatic, so I added your suggestion as part of it.

@glitsj16 left a comment:

> Nice job.

Thanks.

> I've been running with these extra options added to my ssh-agent.local:
> 
> * private-cache
> * private-dev
> * memory-deny-write-execute
>
> While you're working on our SSH profiles, can you test if those can be
> integrated into ssh-agent.profile? Especially mdrw would be a nice hardening
> to have IMO.

I had started adding those and then I noticed that there a few more that can be
added to both that and ssh.profile.  Also, it probably makes sense to move most
of the options to a new ssh-common.profile.  Mind if I do that in a new PR?  As
what I did in this branch was essentially just reading the man pages and
copying and moving paths around.  These new restrictions seem to be more likely
to break something.

@rusty-snake left a comment:

> Looks great so far. Can you add it to etc/tempaltes/profile.template too.

Thanks.  Good catch; I didn't know about that list.  I added that to the first
commit and force-pushed.

glitsj16:
> Mind if I do that in a new PR?

Please do, we can always open a new thread to discuss potential hardening later on.

kmk3:
Sorry for the delay, I was busy with IRL stuff.

(rambling below)

On my last reply I was stunned because I hadn't noticed the ssh auth socket
blacklist on disable-programs.inc; I thought that all the ssh blocking was done
only on disable-common.inc.  So I was trying to figure out what paths would it
make sense to copy over to disable-programs.inc.  It was especially confusing
because there are no /etc paths on it (I might ask about this in a new issue
later).  But then I noticed that all of the socket blocking appears to live on
disable-common.inc and that the other ssh-related paths are already there.  So
I just moved the remaining ssh path to it.

Just in case that there was a reason for it to be on disable-programs.inc, I'm
pinging the author of commit e93fbf3bd ("disable ssh-agent sockets in
disable-programs.inc"): @netblue30.

[Main changes][1]:

* move ssh auth socket path to disable-common.inc
* misc reordering of `include allow-ssh.inc`

I didn't rebase it to master to make it easier to see the changes, but I can do
it if needed.

> > Mind if I do that in a new PR?
> 
> Please do, we can always open a new thread to discuss potential hardening
> later on.

Alright, thanks.

[1]: https://github.com/netblue30/firejail/compare/b32bd093a270461ed3dc4c437b078a646e151f4e..9ef67f7ea568a7d7bd90d0ebb9d415bbbbded5e4

etc/inc/disable-common.inc
```diff
@@ -291,7 +291,16 @@ read-only ${HOME}/.zshrc
 read-only ${HOME}/.zshrc.local
 
 # Remote access
-read-only ${HOME}/.ssh/authorized_keys
```
glitsj16, line 294: I realize this file only exists on SSH server, and our default stance on servers is to advise users to use systemd hardening instead of firejail. Still, it might avoid users shootig themself in the foot by leaving this file unprotected. Not sure on whether we'd like to keep it in, just marking this to gather input from other collaborators.

glitsj16 (approved):
LGTM

etc/inc/disable-common.inc
kmk3: For context, this is what the PR changes on that line:

```diff
-read-only ${HOME}/.ssh/authorized_keys
+blacklist ${HOME}/.ssh/authorized_keys
```

Note: By your response, I'm not sure if you have missed the second line in the
full diff (which would mean that the PR just removes it) or that you're
suggesting removing all of these:

```diff
+blacklist ${HOME}/.rhosts
+blacklist ${HOME}/.shosts
+blacklist ${HOME}/.ssh/authorized_keys
+blacklist ${HOME}/.ssh/authorized_keys2
+blacklist ${HOME}/.ssh/environment
+blacklist ${HOME}/.ssh/rc
+blacklist /etc/hosts.equiv
```

I'm this writing assuming the latter.

> I realize this file only exists on SSH server

I'm not sure if by "SSH server" you imply that this only applies to
virtual/hardware servers (i.e.: as opposed to personal computers), but I think
that it is a common use case to have sshd running on e.g.: a home computer, for
using it as a remote file server or as a home VPN.

> and our default stance on servers is to advise users to use systemd hardening
> instead of firejail.

Interesting, I didn't know about that.  But why is that?  Also, do you mean
using alternative hardening solutions just for server programs (like sshd) or
refraining from using firejail at all on virtual servers?

I have certainly considered using firejail on virtual servers, especially for
non-server/standard cli programs such as text editors and what not.

Also, what about when the virtual server uses openrc/runit/s6/etc?

> Still, it might avoid users shootig themself in the foot by leaving this file
> unprotected. Not sure on whether we'd like to keep it in, just marking this
> to gather input from other collaborators.

The main problem here is that the files for the client (i.e.: ssh) live on the
same directory (i.e.: ~/.ssh) as the files for the server (i.e.: sshd).  So
simply by using the client (either directly or through e.g.: git) and leaving
the server files accessible (i.e.: access to everything in ~/.ssh), a
vulnerability might result in unrestricted remote code execution:

With the blacklists:

* Exploit -> RCE under firejail -> damage is limited to what the profile allows

Without the blacklists:

* Exploit -> RCE under firejail -> write to `~/.ssh/authorized_keys` -> remote
  shell under sshd (not firejailed by default)

On a related note, I have been working on a followup PR for further hardening
these paths on whitelisting profiles (e.g.: seahorse.profile).

glitsj16: > Note: By your response, I'm not sure if you have missed the second line in the
full diff (which would mean that the PR just removes it) or that you're
suggesting removing all of these:
blacklist ${HOME}/.rhosts
blacklist ${HOME}/.shosts
blacklist ${HOME}/.ssh/authorized_keys
blacklist ${HOME}/.ssh/authorized_keys2
blacklist ${HOME}/.ssh/environment
blacklist ${HOME}/.ssh/rc
blacklist /etc/hosts.equiv

> I'm this writing assuming the latter.

I was wondering why we don't simply `blacklist ${HOME}/.ssh` instead of the 4 files in there individually. What's the logic behind the latter exactly?


> I'm not sure if by "SSH server" you imply that this only applies to
virtual/hardware servers (i.e.: as opposed to personal computers), but I think
that it is a common use case to have sshd running on e.g.: a home computer, for
using it as a remote file server or as a home VPN.

I was indeed thinking about sshd.


>    and our default stance on servers is to advise users to use systemd hardening
    instead of firejail.

> Interesting, I didn't know about that. But why is that?

Because we hardly have any (tried-and-tested) server-related profiles, [fdns](https://github.com/netblue30/fdns) being the obvious exception. Sure, one can run firejail on a VPS just fine. But you're pretty much on your own regarding community-tested profiles. IMO firejail primarily shines on desktops due to the number of available profiles we offer and by the 'easy' desktop integration we offer. Servers were never the focus or explicit use-case as far as I'm aware of.

> On a related note, I have been working on a followup PR for further hardening
these paths on whitelisting profiles (e.g.: seahorse.profile).

Yep, I can see why seahorse could use some extra hardening. Let's get this PR moving so you can do the hardening soonish once this is merged.

kmk3:
Sorry, I just noticed a big problem:

`read-only ${HOME}/.ssh` (on disable-common.inc) makes everything in ~/.ssh
always read-only, regardless of the `read-write`s inside of it, so all attempts
to write to ~/.ssh/known_hosts fail.

And `read-only ${HOME}/.ssh/*` wouldn't work either because allow-ssh.inc is
included before that, so `read-write ${HOME}/.ssh/known_hosts` would be
overridden by the former command.  With that, I'm trying to think of a working
alternative.

I suppose that I hadn't encountered this problem before because on my (not very
thorough) tests and normal usage, nothing tried to write to ~/.ssh/known_hosts
(until today).

glitsj16:
> Sorry, I just noticed a big problem

I think you're making this hard on yourself. Suggestion:
- deal with moving (no)blacklist items first and keep an eye on the rule that anything noblacklist'ed implies it being blacklist'ed somewhere else
  for example, I see `noblacklist /etc/ssh/ssh_config` in _allow-ssh.inc_ yet there isn't any `blacklist /etc/ssh/ssh_config` in any of the other files in this PR
  

- bring in any read-only/read-write restrictions afterwards

etc/inc/disable-common.inc
kmk3: Note: I originally wrote a much more detailed reply to this (with ssh config
examples and a firejail bypass POC), but without the `read-only` changes, it
doesn't feel quite right.  So for now I'll just copy and paste the parts that
still make sense (and strikethrough some parts that don't):

---

> I was wondering why we don't simply `blacklist ${HOME}/.ssh` instead of the 4
> files in there individually. What's the logic behind the latter exactly?

The first blacklist is already done on disable-common.inc.  But some programs
use ssh.  And ssh writes to ~/.ssh/known_hosts and might also need access to
~/.ssh/config, ~/.ssh/config.d, ~/.ssh/known_hosts.d (since 8.4p1 [1] [2]) and
the key pairs, all of which can have custom names and paths ~~(more on that
later)~~.  But it doesn't use (and shouldn't have access to)
~/.ssh/authorized_keys et al.  Writing to them ~~(e.g.: `read-write
${HOME}/.ssh` in seahorse.profile)~~ allows remote access and reading them
(i.e.: `noblacklist ${HOME}/.ssh/*`) leaks information about which users and
hosts have shell access to the local machine (and the type and size of each key
pair).

Note2: The `read-write ${HOME}/.ssh` mentioned above is part of the `read-only`
commit.

> > I'm not sure if by "SSH server" you imply that this only applies to
> > virtual/hardware servers (i.e.: as opposed to personal computers), but I
> > think that it is a common use case to have sshd running on e.g.: a home
> > computer, for using it as a remote file server or as a home VPN.
> 
> I was indeed thinking about sshd.
>
> > and our default stance on servers is to advise users to use systemd
> > hardening instead of firejail.
> 
> > Interesting, I didn't know about that. But why is that?
> 
> Because we hardly have any (tried-and-tested) server-related profiles,
> [fdns](https://github.com/netblue30/fdns) being the obvious exception. Sure,
> one can run firejail on a VPS just fine. But you're pretty much on your own
> regarding community-tested profiles. IMO firejail primarily shines on
> desktops due to the number of available profiles we offer and by the 'easy'
> desktop integration we offer. Servers were never the focus or explicit
> use-case as far as I'm aware of.

I see, thanks.  I might follow up on this on the TBD "/etc" question that I
mentioned.

> > On a related note, I have been working on a followup PR for further
> > hardening these paths on whitelisting profiles (e.g.: seahorse.profile).
> 
> Yep, I can see why seahorse could use some extra hardening. Let's get this PR
> moving so you can do the hardening soonish once this is merged.

Cheers to that.

[1]: https://www.openssh.com/txt/release-8.4
[2]: https://bugzilla.mindrot.org/show_bug.cgi?id=1654

kmk3:
> > Sorry, I just noticed a big problem
> 
> I think you're making this hard on yourself. Suggestion:

Yeah, I guess so.  The read-only part was an idea that I had long after I had
done most of the other changes.  Though it does provide much more security (as
in: fixes a pre-existing security problem), so that's why I was trying to get
it in this PR.

> * deal with moving (no)blacklist items first and keep an eye on the rule that
>   anything noblacklist'ed implies it being blacklist'ed somewhere else for
>   example, I see `noblacklist /etc/ssh/ssh_config` in _allow-ssh.inc_ yet
>   there isn't any `blacklist /etc/ssh/ssh_config` in any of the other files
>   in this PR

There is a `blacklist /etc/ssh/*` on disable-common.inc.  Please see the
explanation on commit 516332d89 ("allow-ssh.inc: allow /etc/ssh/ssh_config").
I don't know if you normally check the commits, but I usually use something
like this (along with [tig][1])  to review them in a given PR:

```sh
git fetch origin
git log -p --reverse master..origin/pr/3885
```

Or if you don't fetch PRs by default:

```sh
git fetch origin pull/3885/head
git log -p --reverse master..FETCH_HEAD
```

I find it much easier to review things this way than through the web UI,
especially when using [diff-so-fancy][2].

> * bring in any read-only/read-write restrictions afterwards

Agreed; I just removed the last commit and force-pushed.

[1]: https://github.com/jonas/tig
[2]: https://github.com/so-fancy/diff-so-fancy

etc/profile-a-l/android-studio.profile
```diff
@@ -10,11 +10,11 @@ noblacklist ${HOME}/.android
 noblacklist ${HOME}/.jack-server
 noblacklist ${HOME}/.jack-settings
 noblacklist ${HOME}/.local/share/JetBrains
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.tooling
 
 # Allows files commonly used by IDEs
 include allow-common-devel.inc
+include allow-ssh.inc
```
glitsj16, line 17: Placing: without separation it suggests include allow-ssh.inc is part of the "# Allows files commonly used by IDEs" block. Add the "# Allow ssh (blacklisted by disable-common.inc)" line cfr. your template. We do this all over to avoid confusion/improve readability of our profiles. Same goes for all other profiles where allow-ssh.inc gets included. I'll refer to this remark in the other files to keep things simple.

etc/profile-a-l/aosp.profile
```diff
@@ -11,9 +11,10 @@ noblacklist ${HOME}/.jack-server
 noblacklist ${HOME}/.jack-settings
 noblacklist ${HOME}/.repo_.gitconfig.json
 noblacklist ${HOME}/.repoconfig
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.tooling
 
+include allow-ssh.inc
```
glitsj16, line 16: Add "# Allow ssh (blacklisted by disable-common.inc)" above.

etc/profile-a-l/clion.profile
```diff
@@ -11,9 +11,10 @@ noblacklist ${HOME}/.gitconfig
 noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.java
 noblacklist ${HOME}/.local/share/JetBrains
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.tooling
 
+include allow-ssh.inc
```
glitsj16, line 16: Add "# Allow ssh (blacklisted by disable-common.inc)" above.

etc/profile-a-l/filezilla.profile
```diff
@@ -8,11 +8,11 @@ include globals.local
 
 noblacklist ${HOME}/.config/filezilla
 noblacklist ${HOME}/.filezilla
-noblacklist ${HOME}/.ssh
 
 # Allow python (blacklisted by disable-interpreters.inc)
 include allow-python2.inc
 include allow-python3.inc
+include allow-ssh.inc
```
glitsj16, line 15: Placing: cfr. android-studio remark.

etc/profile-a-l/git-cola.profile
```diff
@@ -20,6 +19,7 @@ noblacklist ${HOME}/.config/git-cola
 
 include allow-python2.inc
 include allow-python3.inc
+include allow-ssh.inc
```
glitsj16, line 22: Placing: cfr. android-studio remark.

etc/profile-a-l/git.profile
```diff
@@ -15,10 +15,11 @@ noblacklist ${HOME}/.gitconfig
 noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.gnupg
 noblacklist ${HOME}/.nanorc
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.vim
 noblacklist ${HOME}/.viminfo
 
+include allow-ssh.inc
```
glitsj16, line 21: Add "# Allow ssh (blacklisted by disable-common.inc)" above.

etc/profile-a-l/gitg.profile
```diff
@@ -10,7 +10,8 @@ noblacklist ${HOME}/.config/git
 noblacklist ${HOME}/.gitconfig
 noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.local/share/gitg
-noblacklist ${HOME}/.ssh
+
+include allow-ssh.inc
```
glitsj16, line 14: Add "# Allow ssh (blacklisted by disable-common.inc)" above.

etc/profile-a-l/idea.sh.profile
```diff
@@ -10,11 +10,11 @@ noblacklist ${HOME}/.android
 noblacklist ${HOME}/.jack-server
 noblacklist ${HOME}/.jack-settings
 noblacklist ${HOME}/.local/share/JetBrains
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.tooling
 
 # Allows files commonly used by IDEs
 include allow-common-devel.inc
+include allow-ssh.inc
```
glitsj16, line 17: Placing: cfr. android-studio remark.

etc/profile-m-z/meld.profile
```diff
@@ -18,11 +18,11 @@ noblacklist ${HOME}/.config/git
 noblacklist ${HOME}/.gitconfig
 noblacklist ${HOME}/.git-credentials
 noblacklist ${HOME}/.local/share/meld
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.subversion
 
 # Allow python (blacklisted by disable-interpreters.inc)
 include allow-python3.inc
+include allow-ssh.inc
```
glitsj16, line 25: Placing: cfr. android-studio remark.

etc/profile-m-z/seahorse.profile
```diff
@@ -9,8 +9,8 @@ include globals.local
 blacklist /tmp/.X11-unix
 
 noblacklist ${HOME}/.gnupg
-noblacklist ${HOME}/.ssh
-noblacklist /tmp/ssh-*
+
+include allow-ssh.inc
```
glitsj16, line 13: Add "# Allow ssh (blacklisted by disable-common.inc)" above.

etc/profile-m-z/ssh-agent.profile
```diff
@@ -6,9 +6,7 @@ include ssh-agent.local
 # Persistent global definitions
 include globals.local
 
-noblacklist /etc/ssh
-noblacklist /tmp/ssh-*
-noblacklist ${HOME}/.ssh
+include allow-ssh.inc
```
glitsj16, line 9: Add "# Allow ssh (blacklisted by disable-common.inc)" above.

etc/profile-m-z/ssh.profile
```diff
@@ -7,13 +7,12 @@ include ssh.local
...
 include globals.local
 
-noblacklist /etc/ssh
-noblacklist /tmp/ssh-*
-noblacklist ${HOME}/.ssh
 # nc can be used as ProxyCommand, e.g. when using tor
 noblacklist ${PATH}/nc
 noblacklist ${PATH}/ncat
 
+include allow-ssh.inc
```
glitsj16, line 14: Add "# Allow ssh (blacklisted by disable-common.inc)" above.

etc/profile-m-z/webstorm.profile
```diff
@@ -8,11 +8,11 @@ include globals.local
 noblacklist ${HOME}/.WebStorm*
 noblacklist ${HOME}/.android
 noblacklist ${HOME}/.local/share/JetBrains
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.tooling
 
 # Allows files commonly used by IDEs
 include allow-common-devel.inc
+include allow-ssh.inc
```
glitsj16, line 15: Placing: cfr. android-studio remark.

etc/profile-m-z/x2goclient.profile
```diff
@@ -6,10 +6,11 @@ include x2goclient.local
 # Persistent global definitions
 include globals.local
 
-noblacklist ${HOME}/.ssh
 noblacklist ${HOME}/.x2go
 noblacklist ${HOME}/.x2goclient
 
+include allow-ssh.inc
```
glitsj16, line 12: Add "# Allow ssh (blacklisted by disable-common.inc)" above.

etc/profile-a-l/android-studio.profile
kmk3: Added the comments, blank lines and also did some sorting:

<https://github.com/netblue30/firejail/compare/2bf2581d6f4c233b84a3d6a475799d2ae5efa013..83ac0239722f85ffed15e3b6b6088bfff547ac1b>

glitsj16: Thanks, those commits look great!

etc/inc/disable-common.inc
glitsj16: I must admit it took me some time to see the real gain here, but I do now. Apologies for the potential confusion raised before. Looking forward to your follow-up PR(s) :+1:

glitsj16 (approved):
Looks ready for merging to me. Do you want to add anything here or would you prefer seeing this merged and do other work in separate PR('s)?

etc/inc/disable-common.inc
kmk3: > I must admit it took me some time to see the real gain here, but I do now.
> Apologies for the potential confusion raised before.

It's all good.  Sometimes the changes seem more obvious than they are when you
have all the context in your head.

> Looking forward to your follow-up PR(s) +1

Thanks.

kmk3:
> Looks ready for merging to me. Do you want to add anything here or would you
> prefer seeing this merged and do other work in separate PR('s)?

It's been a while since I've been focused on the remaining commits, so I'd be
glad to just see this merged.  I also have some branches that are based on
this, so that would make things easier.

By the way, could you mark the (from second to last) reviews about the missing
comments as resolved (so that GH collapses them to reduce noise)?  As to me
sometimes it looks confusing when someone other than the reviewer marks them as
resolved (e.g.: were there changes made or was the review dismissed for some
reason?), especially when doing so without replying and I'd rather not spam a
dozen threads hehe.

> By now you've seen that my git fu is **very basic** and I do use the Web UI
> mostly.

> I really appreciate the git tips you gave but it'll take me some time to dig
> deeper into using git from CLI before I feel comfortable.

No worries, git takes a while to learn.  I think the CLI is mostly about
practice and those tools make it a much more pleasant experience.  I would just
create a dummy project/clone and try commands to see what happens.  Here are
some basic ones:

1. `status` / `add [-p | -u] [<path>]` / `reset [-p] [<path>]` / `commit`
   (avoid -m)
2. `branch [-d | -D] <branchname>` / `checkout <branchname>`
3. `merge [--ff-only | --no-ff] <ref>` / `tag [-a | -s | -d] <tagname>`

Some that are good to know:

4. `reset [--mixed | --hard] <ref>` / `stash [list | push [-m] | pop]` /
   `reflog [<ref>]`

Networking commands:

5. `git clone <url>` / `git remote [add <name> <url> | set-url <name> <url>]`
6. `fetch [--prune] [<repository>]` / `pull [<repository>]` / `push [-d] [-u]
   [<repository>]`

And lastly the "dangerous" (and very useful) commands that "change history",
such as:

7. `commit --amend` / `rebase [-i] [--autosquash] <ref>` (together with `commit
   --fixup=<commit>` and `commit -C (<commit> | ORIG_HEAD)`)
8. `pull --rebase` / `push --force-with-lease`

It sounds like a lot (and there are inconsistent commands/flags), but some
commands are usually just used without arguments to get information (like
status/reflog), and others I at least don't use very often (like
tag/stash/reflog/remote).  The really hard part is when using fetch/pull/push
with amend/rebase (including squash), as that "changes (public) history", which
may end up pulling the rug from under someone else.

From a high level perspective, for me there were two important milestones:

1. When you really get that it's all just a graph
2. When you learn and get used to `git rebase -i` (and `git add -p` / `git
   reset -p`)

Anyway, if you get stuck feel free to mention or email me.

> Together with your earlier remarks about not squashing your commits I would
> like to suggest asking a more experienced collaborator with merge rights to
> proceed with this if that's all right.

> And I don't want to mess up your nice work here. @rusty-snake What do you
> think, could you take care of merging this if you're satisfied with the
> latest changes?

Sure; thanks for the consideration.

The key point is to avoid the "squash and merge" option in the web UI.  I'd
mostly use the standard/no-ff merge (also in the web UI, for commit message
consistency).  Can't go wrong with that one, as it's the only one that uses the
PR branch as is (i.e.: without duplicating all of the original commits and
changing the metadata).

rusty-snake (approved):

glitsj16:
@kmk3 Resolved the outstanding items as per your request. We encountered some CI issues lately but this one seems to have gone thru as expected. Thanks again for your work and the extra tips on using git from CLI, much appreciated.

PR #3896, opened 2021-01-15
Add new allow include allow-bin-sh.inc

rusty-snake:
/bin/sh is usually just a symlink to bash. However this is not the case
for every distro, debian for example uses dash. bash,dash and sh have a
blacklist command in disable-shell.inc. An own allow-*.inc for it
enusres usage of all necessary nolacklists.

For private-bin sh is enough because it follows symlinks.


glitsj16 (approved):
Nice and future-proof addition!

Issue #3733, opened 2020-11-08
telegram.profile need netlink protocol


IlyaIndigo:
When running telegram via firejail, it issues a warning.
Could not create AF_NETLINK socket (Operation not supported)
Please change telegram.profile to enable netlink protocol.

-protocol unix,inet,inet6
+protocol unix,inet,inet6,netlink

glitsj16:
Duplicate of #3614.

You can make the needed change in a telegram.local file until your OS packages the firejail version that has that option.

IlyaIndigo:
Thank you!

Issue #3638, opened 2020-09-20
Why are all the $HOME dirs and files visible in Telegram and not jailed?


svc88:
Currently, If you open Telegram and click on the attachment (paper clip) icon to send a file, i can see all files and dirs under $HOME.  Why isnt there only one folder like ${DOWNLOADS} shown and the rest blacklisted?
Im not 100% sure that Telegram is jailed properly.

firejail version 0.9.62

SkewedZeppelin:
Does telegram process show up on "firejail --list" command?

svc88:
> Does telegram process show up on "firejail --list" command?
Yes

Is it supposed to blacklist all folders? Because i dont see "whitelist" anywhere in telegram.profile

rusty-snake:
There's a lot of options which can be added to telegram.profile.
- More `disable-*.inc` includes (especially `include disable-xdg.inc`)
- `include whitelist-*.inc`
- DBus policy
- Others such as nogroups, novideo, nosound, tracelog likely/obviously break

Flatpak permissions:
```
[Context]
shared=network;ipc;
sockets=x11;wayland;pulseaudio;
devices=all;
filesystems=xdg-download;xdg-config/kdeglobals:ro;

[Session Bus Policy]
org.freedesktop.Notifications=talk
com.canonical.indicator.application=talk
org.freedesktop.portal.Fcitx=talk
org.gnome.Mutter.IdleMonitor=talk
org.kde.StatusNotifierWatcher=talk
org.freedesktop.ScreenSaver=talk
org.ayatana.indicator.application=talk
com.canonical.AppMenu.Registrar=talk
com.canonical.Unity=talk
```

First draft (dbus-policy can be future hardened):

```
include disable-passwdmgr.inc
include disable-shell.inc
include disable-xdg.inc

mkdir ${HOME}/<telegram>
whitelist ${HOME}/<telegram>
whitelist ${DOWNLOADS}
whitelist /usr/share/<telegram>
include whitelist-common.inc
include whitelist-runuser-common.inc
include whitelist-usr-share-common.inc
include whitelist-var-common.inc

dbus-user filter
dbus-user.own org.telegram.desktop
dbus-user.talk org.freedesktop.Notifications
dbus-user.talk com.canonical.indicator.application
dbus-user.talk org.gnome.Mutter.IdleMonitor
dbus-user.talk org.kde.StatusNotifierWatcher
dbus-user.talk org.freedesktop.ScreenSaver
[...]

Issue #3614, opened 2020-08-30
telegram.profile needs netlink protocol


reagentoo:
Since v2.3.0 the Telegram Desktop supports video calls.
https://github.com/telegramdesktop/tdesktop/releases/tag/v2.3.0
This feature does not work w/o netlink protocol in the profile. Tested on v2.3.2.
Please change the next [line](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/telegram.profile#L23)
`protocol unix,inet,inet6`
to
`protocol unix,inet,inet6,netlink`

rusty-snake:
Fixed, Thanks.

PR #3897, opened 2021-01-17
Update telegram.profile

nidamanx:
Allow Telegram ONLY in .TelegramDesktop, .local/share/TelegramDesktop and Downloads

If your PR isn't about profiles or you have no idea how to do one of these, skip the following and go ahead with this PR.

If you make a PR for new profiles or changeing profiles please do the following:
 - The ordering of options follow the rules descripted in [/usr/share/doc/firejail/profile.template](https://github.com/netblue30/firejail/blob/master/etc/templates/profile.template).  
   > Hint: The profile-template is very new, if you install firejail with your package-manager, it maybe missing, therefore, and to follow the latest rules, it is recommended to use the template from the repository.
 - Order the arguments of options alphabetical, you can easy do this with the [sort.py](https://github.com/netblue30/firejail/tree/master/contrib/sort.py).  

   | Distro | Path |
   | ------ | ---- |

   Note also that the sort.py script exists only since firejail `0.9.61`.

See also [CONTRIBUTING.md](/CONTRIBUTING.md).

### DIFF
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -19,7 +19,7 @@ mkdir ${HOME}/.local/share/TelegramDesktop
 whitelist ${DOWNLOADS}
 whitelist ${HOME}/.TelegramDesktop
 whitelist ${HOME}/.local/share/TelegramDesktop
-include /etc/firejail/whitelist-common.inc
+include whitelist-common.inc
 
 caps.drop all
 netfilter
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -14,6 +14,13 @@ include disable-exec.inc
 include disable-interpreters.inc
 include disable-programs.inc
 
+mkdir ${HOME}/.TelegramDesktop
+mkdir ${HOME}/.local/share/TelegramDesktop
+whitelist ${DOWNLOADS}
+whitelist ${HOME}/.TelegramDesktop
+whitelist ${HOME}/.local/share/TelegramDesktop
+include /etc/firejail/whitelist-common.inc
+
 caps.drop all
 netfilter
 nodvd
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -12,26 +12,22 @@ include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
+include disable-passwdmgr.inc
 include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
 
 mkdir ${HOME}/.TelegramDesktop
 mkdir ${HOME}/.local/share/TelegramDesktop
-whitelist ${DOWNLOADS}
 whitelist ${HOME}/.TelegramDesktop
 whitelist ${HOME}/.local/share/TelegramDesktop
-
+whitelist ${DOWNLOADS}
 include whitelist-common.inc
+include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
-include whitelist-runuser-common.inc
-include disable-shell.inc
-include disable-passwdmgr.inc
-include disable-xdg.inc
 
 apparmor
-shell none
-private-dev
-
 caps.drop all
 netfilter
 nodvd
@@ -40,8 +36,10 @@ noroot
 notv
 protocol unix,inet,inet6,netlink
 seccomp
+shell none
 
 disable-mnt
 private-cache
+private-dev
 private-etc alsa,alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.cache,localtime,machine-id,os-release,pki,pulse,resolv.conf,ssl,xdg
 private-tmp
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -19,7 +19,18 @@ mkdir ${HOME}/.local/share/TelegramDesktop
 whitelist ${DOWNLOADS}
 whitelist ${HOME}/.TelegramDesktop
 whitelist ${HOME}/.local/share/TelegramDesktop
+
 include whitelist-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+include whitelist-runuser-common.inc
+include disable-shell.inc
+include disable-passwdmgr.inc
+include disable-xdg.inc
+
+apparmor
+shell none
+private-dev
 
 caps.drop all
 netfilter


rusty-snake (commented):
LGTM, except that `include`s should go by name and not by path.

BTW: There's a lot more we can do on telegram.profile, see https://github.com/netblue30/firejail/issues/3638#issuecomment-695988562.

OT: TelegramDesktop is a electron too if I'm right. We should refactor it too.

etc/profile-m-z/telegram.profile
```diff
@@ -14,6 +14,13 @@ include disable-exec.inc
 include disable-interpreters.inc
 include disable-programs.inc
 
+mkdir ${HOME}/.TelegramDesktop
+mkdir ${HOME}/.local/share/TelegramDesktop
+whitelist ${DOWNLOADS}
+whitelist ${HOME}/.TelegramDesktop
+whitelist ${HOME}/.local/share/TelegramDesktop
+include /etc/firejail/whitelist-common.inc
```
rusty-snake, line 22:
```suggestion
include whitelist-common.inc
```

NDMann: @rusty-snake Tried your suggested code. It works. Thanks!

rusty-snake (approved):

Neo00001:
> OT: TelegramDesktop is a electron too if I'm right. We should refactor it too.

TelegramDesktop is Qt-based desktop client afaik.

Neo00001:
@NDMann 
```
include disable-passwdmgr.inc
include disable-xdg.inc

include whitelist-common.inc
```
try these.Hopefully it will work

NDMann:
> @NDMann
> 
> ```
> include disable-passwdmgr.inc
> include disable-xdg.inc
> 
> include whitelist-usr-share-common.inc
> include whitelist-var-common.inc
> include whitelist-runuser-common.inc
> ```
> 
> try these.Hopefully it will work

Thanks @Neo00001 
one submitted setting was already approved
https://github.com/netblue30/firejail/pull/3897/files

I found very interesting your suggestion about 
include disable-passwdmgr.inc
Are you using it?

Neo00001:
> Are you using it?

yes.& these two also
```
include disable-shell.inc
apparmor
shell none
```
 my `dbus-policy`  is

```
dbus-user filter
dbus-user.talk org.kde.StatusNotifierWatcher

dbus-system none
```
I'm on kde, for other DE it will be different though.

if you have some other folders/programs in `opt` you may create an empty folder `foo` & add

`private-opt foo` to telegram's profile.

depending upon your usecase, you may restrict much more.

NDMann:
> > Are you using it?
> 
> yes.& these also
> 
> ```
> include disable-shell.inc
> apparmor
> shell none
> ```
> 
> my `dbus-policy` is
> 
> ```
> dbus-user filter
> dbus-user.talk org.freedesktop.Notifications
> dbus-user.talk org.kde.StatusNotifierWatcher
> 
> dbus-system none
> ```
> 
> I'm on kde, for other DE it will be different though.
> 
> if you have some other folders/programs in `opt` you may create an empty folder `foo` & add
> 
> `private-opt foo` to telegram's profile.
> 
> depending upon your usecase, you may restrict much more.

That's also interesting!
```
include disable-shell.inc
apparmor
shell none
```
Could be useful to find a basic config ready for all DE.
I'll try your added lines 
Thanks!

Neo00001:
> All was exposed

what were exposed? 
btw,you had to add those lines into your profile along with your already submitted changes. otherwise home folder won't be properly restricted.

NDMann:
> > All was exposed
> 
> what were exposed?
> btw,you had to add those lines into your profile along with your already submitted changes. otherwise home folder won't be properly restricted.

Ah, ADDED to my submitted lines!
So, I think, all will be absolutely better!
I thought your setup could be a better way to have the same results of the code i submitted
...btw i already stated to integrate your lines

NDMann:
So, @Neo00001,  you are on KDE.
I'm on Gnome and Xfce.
Let's try a config for more DEs

My install:

- telegram in /opt/telegram
- alias in /usr/local/bin/telegram
- launcher in /usr/local/share/applications/telegram-messenger.desktop
- firejail profile in: ~/.config/firejail/telegram.profile

The following lines work perfect to me (Gnome and Xfce) and probably in KDE too.
Adding all the lines you suggested didn't allow my telegram to start from launcher
```
mkdir ${HOME}/.TelegramDesktop
mkdir ${HOME}/.local/share/TelegramDesktop
whitelist ${DOWNLOADS}
whitelist ${HOME}/.TelegramDesktop
whitelist ${HOME}/.local/share/TelegramDesktop
include whitelist-common.inc

include disable-passwdmgr.inc
include disable-xdg.inc
shell none
apparmor
```

NDMann:
Adding the following lines, seems absolutely better:
now we're in private-dev too

```
mkdir ${HOME}/.TelegramDesktop
mkdir ${HOME}/.local/share/TelegramDesktop
whitelist ${DOWNLOADS}
whitelist ${HOME}/.TelegramDesktop
whitelist ${HOME}/.local/share/TelegramDesktop
include whitelist-common.inc
include disable-passwdmgr.inc
include disable-xdg.inc
shell none
apparmor
private-dev
```

In audit: I can see a good protection (nothing marked as BAD)

Only, if I run Telegram from terminal, I have these few errors
```
[ALSOFT] (EE) Failed to set real-time priority for thread: not allowed
Could not create AF_NETLINK socket not supported
```

But Telegram seems to work fine.
Maybe we have to investigate more about those errors?
Thanks

Neo00001:
> The following lines work perfect to me (Gnome and Xfce) and probably in KDE too.

yes.It works for kde also.

> [ALSOFT] (EE) Failed to set real-time priority for thread: not allowed

this probably occurs without firejail also. try to run `telegram` with `firejail --noprofile` to check

> Could not create AF_NETLINK socket not supported

 #3733

>     * telegram in /opt/telegram

try all of my suggestions except `private-opt`

btw,what's the output of ` ls -al /opt/telegram`

NDMann:
> this probably occurs without firejail also. try to run telegram with firejail --noprofile to check

You're right!

>Could not create AF_NETLINK socket not supported

fixed as in [#3614](https://github.com/netblue30/firejail/issues/3614)

```
- unix,inet,inet6
+ unix,inet,inet6,netlink
```

> try all of my suggestions except private-opt

All seems fine, but not the launcher (fails)
Maybe we also need to whitelist launcher location.
_Same result if I move launcher in ~/.local/share/applications/telegram-messenger.desktop_

```
cat /usr/local/share/applications/telegram-messenger.desktop 
[Desktop Entry]
Encoding=UTF-8
Name=Telegram Messenger
Comment=Telegram Messenger for Desktop
Exec=firejail /opt/telegram/Telegram
Icon=/usr/local/share/icons/telegram-messenger-icon.png
Type=Application
Categories=Application;
```

> btw,what's the output of ` ls -al /opt/telegram`

```
drwxrwxr-x 2 user group   4096 gen 17 11:58 .
drwxr-xr-x 6 root root    4096 gen 16 18:54 ..
-rwxr-xr-x 1 user group   93162336 dic 23 13:43 Telegram
-rwxr-xr-x 1 user group   1810344 dic 23 13:43 Updater
```

Neo00001:
> All seems fine in terminal, but not the launcher (fails)
> Seems the only parameter I can add is

So when you run `telegram` from `terminal`, all those parameters work but in case of launcher that ain't the case. I'm a bit puzzled here tbh. @rusty-snake  may help us out here.

> -rwxr-xr-x 1 user group   93162336 dic 23 13:43 Telegram

`private-opt` won't work for u


> Maybe we also need to whitelist launcher location.

`/usr/local/share/` is not blocked in your case. Something else is the problem. & btw firejail's `apparmor` profile also supports execution from `/usr/local`  or `/opt`

- when you run telegram from terminal, do you notice any error?


> [Desktop Entry]
> Encoding=UTF-8

did you use any software to create this desktop entry? `Encoding` is a deprecated key.

NDMann:
>     * So when you run telegram from terminal, all those parameters work

exactly. It's a bit strange

>     * when you run telegram from terminal, do you notice any error?

no, no errors. if all params in _TESTING_ are enables (see below the full lines added to the default one) .desktop file just doesn't work

>     * did you use any software to create this desktop entry? Encoding is a deprecated key.

done by myself. Now `encoding` is removed, but same result.

Here is the actual working config:

```
# STABLE BEGIN
mkdir ${HOME}/.TelegramDesktop
mkdir ${HOME}/.local/share/TelegramDesktop
whitelist ${DOWNLOADS}
whitelist ${HOME}/.TelegramDesktop
whitelist ${HOME}/.local/share/TelegramDesktop
include whitelist-common.inc
include disable-passwdmgr.inc
include disable-xdg.inc
shell none
apparmor
private-dev
# STABLE END

# TESTING BEGIN
#include disable-shell.inc
# TESTING END
```

nidamanx:
All is now fixed using the Debian Backport Package (0.9.64-1~bpo10+1)
Tested on KDE, Gnome, Xfce.
I'm going to submit the code

rusty-snake (changes requested):
Can you follow our profile ordering/sorting from profile.template.

nidamanx:
> Can you follow our profile ordering/sorting from profile.template.

Done

rusty-snake (approved):

rusty-snake:
merged, thanks.

nidamanx:
You're welcome! :-)

PR #3900, opened 2021-01-18
Add $PATH expansion to private-lib

smitsohu:
Support is already quite good if `private-bin` and `private-lib` are combined, or if a full application path is provided. Add $PATH expansion to `private-lib` in order to support simple application names at a similar level.

For example `firejail --noprofile --private-lib gedit` or `firejail --noprofile --private-lib okular` work this way without specifiying particular libraries.


glitsj16 (approved):
Nice!

PR #3902, opened 2021-01-18
add new profile: pkglog

glitsj16:


etc/profile-m-z/pkglog.profile
```diff
@@ -0,0 +1,59 @@
+# Firejail profile for pklog
+# Description: Reports log of package updates
+# This file is overwritten after every install/update
+# Persistent local customizations
+include pkglog.local
+# Persistent global definitions
+include globals.local
+
+whitelist /var/log/apt/history.log
+whitelist  /var/log/dnf.rpm.log
```
rusty-snake, line 10: `whitelist` should go below `disable-*.inc`

```suggestion
whitelist /var/log/dnf.rpm.log
```

glitsj16: Thx! Fixed it.

Issue #3907, opened 2021-01-22
Bibletime profile does not work, should add a new whitelist


hhzek0014:
Bibletime does not work with the default profile, but works with `--noprofile` option.

Output:

```
$ firejail --debug --profile=/etc/firejail/bibletime.profile bibletime
...
(BibleTime 3.0.0) WARNING: Cannot find documentation directory relative to "/usr"
(BibleTime 3.0.0) FATAL: Error initializing directory cache!
Please report this bug at https://github.com/bibletime/bibletime/issues
...
```

After some check I found `/usr/share/doc/bibletime` should be whitelisted. I will send pull request. This report is used to track the issue only.

PR #3908, opened 2021-01-22
Update bibletime.profile, add new whitelist

hhzek0014:
To fix issue #3907, doc directory of the bibletime has to be whitelisted. Otherwise, it always fails to start.


glitsj16 (approved):

glitsj16:
@hhzek0014 Thank you for contributing!

PR #3854, opened 2020-12-29
profiles: add redirect from matrix-mirage to mirage

reinerh:
In Debian mirage is packaged as matrix-mirage (as mirage was already taken by a different program).
This creates a redirect profile.

It also adds ldconfig to private-bin (and noblacklists /sbin), as otherwise some Python/Qt related modules were not loading:
```
! 18:17:41 | QSystemTrayIcon::setVisible: No Icon set
~ 18:17:42 | "PyOtherSide error: Traceback (most recent call last):\n\n  File \"qrc:/src/backend/qml_bridge.py\", line 27, in <module>\n    from .pyotherside_events import CoroutineDone, LoopException\n\n  File \"qrc:/src/backend/pyotherside_events.py\", line 8, in <module>\n    from .utils import serialize_value_for_qml\n\n  File \"qrc:/src/backend/utils.py\", line 26, in <module>\n    import magic\n\n  File \"/usr/lib/python3/dist-packages/magic/__init__.py\", line 361, in <module>\n    add_compat(globals())\n\n  File \"/usr/lib/python3/dist-packages/magic/__init__.py\", line 325, in add_compat\n    from magic import compat\n\n  File \"/usr/lib/python3/dist-packages/magic/compat.py\", line 61, in <module>\n    _open = _libraries['magic'].magic_open\n\n  File \"/usr/lib/python3.9/ctypes/__init__.py\", line 387, in __getattr__\n    func = self.__getitem__(name)\n\n  File \"/usr/lib/python3.9/ctypes/__init__.py\", line 392, in __getitem__\n    func = self._FuncPtr((name_or_ordinal, self))\n\nAttributeError: matrix-mirage: undefined symbol: magic_open\n"
! 18:17:42 | Unhandled PyOtherSide error: Cannot import module: backend.qml_bridge (Traceback (most recent call last):

  File "qrc:/src/backend/qml_bridge.py", line 27, in <module>
    from .pyotherside_events import CoroutineDone, LoopException

  File "qrc:/src/backend/pyotherside_events.py", line 8, in <module>
    from .utils import serialize_value_for_qml

  File "qrc:/src/backend/utils.py", line 26, in <module>
    import magic

  File "/usr/lib/python3/dist-packages/magic/__init__.py", line 361, in <module>
    add_compat(globals())

  File "/usr/lib/python3/dist-packages/magic/__init__.py", line 325, in add_compat
    from magic import compat

  File "/usr/lib/python3/dist-packages/magic/compat.py", line 61, in <module>
[...]

PR #3917, opened 2021-01-24
Fix cheese,authenticator; add wruc,private-cache,etc to liferea

bbhtt:


etc/profile-a-l/authenticator-rs.profile
```diff
@@ -51,5 +51,7 @@ private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,gtk-2.0,gtk
 private-tmp
 
 dbus-user filter
+dbus-user.own uk.co.grumlimited.authenticator-rs
```
bbhtt, line 54: It doesn't work without this, I've rechecked again and dbus shows a string communicating.

rusty-snake (commented):
Cheese works for me without this. Is there something broken?

lifefrea: :heavy_check_mark:

bbhtt:
> Cheese works for me without this. Is there something broken?

It exits out on Arch Linux with this for me: 
```
```
and on Xubuntu it doesn't exit but shows a bunch of warnings about `org.gnome.Cheese.`  Also w/o allow-python it shows missing PYTHONHOME,PYTHONPATH modules not found etc. 

Same with auth-rs: 
```
```

etc/profile-m-z/mirage.profile
```diff
@@ -20,7 +20,6 @@ include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
-include disable-shell.inc
```
glitsj16, line 23: We have [allow-bin-sh.inc](https://github.com/netblue30/firejail/commit/f18124baa4290e658babc6dd0716232e9a474335) now to handle this.

bbhtt: Ah thanks, I just saw it.

etc/profile-a-l/liferea.profile
```diff
@@ -49,7 +50,10 @@ shell none
 tracelog
 
 disable-mnt
+#private-bin liferea
+private-cache
 private-dev
+private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,gtk-2.0,gtk-3.0,hostname,hosts,mime.types,nsswitch.conf,pki,pulse,resolv.conf,ssl,X11,xdg
```
glitsj16, line 56: On Arch Linux with a user ID of 1001 I get:
```
Error: cannot read UID_MIN and/or GID_MIN from /etc/login.defs, using 1000 by default
```
Please add `login.defs` to private-etc.

etc/profile-a-l/liferea.profile
```diff
@@ -49,7 +50,10 @@ shell none
 tracelog
 
 disable-mnt
+#private-bin liferea
+private-cache
```
glitsj16, line 54: Liferea drops several files into ${HOME}/.cache/liferea. Enabling private-cache will discard those on sandbox shutdown. Is this intentional here? Amongst other things this implies feed favicons would have to be redownloaded on each invocation. Personally I don't see much harm in dropping private-cache in this particular case. Just wondering what your motivation is to add it to the profile...

glitsj16 (changes requested):

etc/profile-a-l/liferea.profile
bbhtt: Right, I was testing with a small number of feeds synced to my ttrss instance and didn't notice the difference. Tried on my main profile >200 feeds it it is an issue with slow networks and feed numbers.

etc/profile-m-z/mirage.profile
```diff
@@ -11,6 +11,8 @@ noblacklist ${HOME}/.config/mirage
 noblacklist ${HOME}/.local/share/mirage
 noblacklist /sbin
 
+include allow-bin-sh.inc
```
rusty-snake, line 14: This can not work because `private-bin` does not contain `sh`. Was this tested?

rusty-snake (changes requested):

etc/profile-m-z/mirage.profile
bbhtt: Yes, I'm going to add it back, I had some other small changes for matrix apps. It was a conflict.

bbhtt:
Nheko: See (Databse location) https://github.com/Nheko-Reborn/nheko/commit/fc76a939bb3c2eff1d90819d369328df1fa43499#diff-c8d6667e13434c86eaef0d2d4ce3610b3aafda388c2c56adf0afcf9b88c382c5R179 
(Keychain) https://github.com/Nheko-Reborn/nheko/releases/tag/v0.8.0

etc/profile-m-z/nheko.profile
```diff
@@ -48,9 +50,9 @@ private-dev
 private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,gtk-2.0,gtk-3.0,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,mime.types,nsswitch.conf,pki,pulse,resolv.conf,selinux,ssl,X11,xdg
 private-tmp
 
-dbus-user none
+dbus-user filter
+dbus-user.talk org.freedesktop.secrets
 # Comment the above line and uncomment below lines for notification popups
```
rusty-snake, line 55:
```suggestions
# Uncomment the lines below for native notification and tray icons
```

rusty-snake, line 55:
```suggestions
# Uncomment the lines below for native notification and tray icons
```

bbhtt: Thanks!

etc/profile-m-z/nheko.profile
```diff
@@ -8,6 +8,7 @@ include globals.local
 
 noblacklist ${HOME}/.config/nheko
 noblacklist ${HOME}/.cache/nheko
+noblacklist ${HOME}/.local/share/nheko
```
rusty-snake, line 11: Add `blacklist ${HOME}/.local/share/nheko` to disable-programs.inc.

bbhtt:
Why is the CI failing?

etc/profile-m-z/mirage.profile
bbhtt: --build on the appimage says it tries to access ` x86_64-linux-gnu-ld.bfd,x86_64-linux-gnu-gcc-10` in private-bin amongs others.

bbhtt: The previous profile seems to work fine for me but I'm adding the changes except that if in case it doesn't work.

etc/profile-m-z/nheko.profile
```diff
@@ -22,6 +23,7 @@ mkdir ${HOME}/.config/nheko
 mkdir ${HOME}/.cache/nheko/nheko
 whitelist ${HOME}/.config/nheko
 whitelist ${HOME}/.cache/nheko
+whitelist ${HOME}/.local/share/nheko
```
rusty-snake, line 26: Needs a `mkdir`/`mkfile` to work.

etc/profile-m-z/mirage.profile
```diff
@@ -11,6 +11,8 @@ noblacklist ${HOME}/.config/mirage
 noblacklist ${HOME}/.local/share/mirage
 noblacklist /sbin
```
rusty-snake, line 12: Site note: This break for systems with a unified file-system-hirachy w/o `nobalcklist /usr/sbin`.

bbhtt: I didn't introduce it; it was by reinerh for the `matrix-mirage`  Debian package https://github.com/netblue30/firejail/pull/3854/files#diff-a1f8494c6f1dd528c0cc8bba9fd3d8739afcd1e3baef0e1f112fa4d91c6d6b08R12; maybe it can be moved to that profile or add a note here?

rusty-snake (commented):
IMHO we can merge after these 3 tasks are solved:
- [ ] rebase due to merge conflicts
- [ ] fix the `/{usr/,}sbin` issue, it's just one line (`noblacklist /usr/sbin`)
- [ ] update all comments according to #4165

etc/profile-m-z/quaternion.profile
```diff
@@ -44,11 +46,21 @@ shell none
...
 
-dbus-user none
+dbus-user.filter
+dbus-user.talk org.freedesktop.secrets
+# Add below line to 'quaternion.local' for native notifications
+# dbus-user.talk org.freedesktop.Notifications
+dbus-user.talk org.kde.kwalletd5
+# Add below lines to 'quaternion.local' for KDE tray notifications
+# dbus-user.talk org.kde.StatusNotifierWatcher
+# dbus-user.talk org.kde.StatusNotifierItem-13-1
```
rusty-snake, line 63: - StatusNotifierItem must be a own rule, it must not?
- The pid and id in StatusNotifierItem are not stable, use wildcards.

```suggestion
# dbus-user.own org.kde.*
# dbus-user.talk org.kde.StatusNotifierWatcher
```

etc/profile-m-z/quaternion.profile
```diff
@@ -44,11 +46,21 @@ shell none
...
 private-dev
 private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,gtk-2.0,gtk-3.0,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,mime.types,nsswitch.conf,pki,pulse,resolv.conf,selinux,ssl,X11,xdg
 private-tmp
 
-dbus-user none
+dbus-user.filter
+dbus-user.talk org.freedesktop.secrets
+# Add below line to 'quaternion.local' for native notifications
+# dbus-user.talk org.freedesktop.Notifications
+dbus-user.talk org.kde.kwalletd5
```
rusty-snake, line 60: I would move the kwalletd5 up:

```
dbus-user.talk org.kde.kwalletd5
# Add below line to 'quaternion.local' for native notifications
# dbus-user.talk org.freedesktop.Notifications
```

rusty-snake (dismissed):
Do these firefox rules work for firefox as native wayland client?

etc/profile-m-z/quaternion.profile
bbhtt: Yes, it is a own rule.

bbhtt:
> Do these firefox rules work for firefox as native wayland client?

Seems to be working but `passwd` and `login.defs` in `private-etc` are needed and the whitelist mozilla needs to be changed. Other than that FF seems to be getting caught in the `seccomp` and `disable-shell.inc` (Arch Linux..) of `quaternion.profile`. Not sure if I should put the `!chroot` in there too.

rusty-snake:
Is firefox running in the quaternion sandbox?

bbhtt:
> the quaternion sandbox?

Other than that only `seccomp` needs to be commented I don't know if that's something specific to my setup or `!chroot` needs to be whitelisted lik in `firefox-common.profile`; it can follow links from quaternion I haven't checked if anything else with firefox is broken.

bbhtt:
I reverted the Firefox specific commits since it's getting complicated to tackle in this PR due to different configurations in distribution/DE but in the future these need to have a proper hyperlinking support.

The minimum of 

```
noblacklist ${HOME}/.mozilla
ignore private-bin # or add which,sh,bash,firefox,env etc.
include allow-bin-sh.inc
```
is still creating issues like 1> Firefox reporting corrupted profile or not found 2>Launching needs specific directories sometimes under /usr/share, ~/.config/mimeapps.list etc. 3> dbus errors and `private-tmp` related issues 4> shell specific issues 5> Sometimes browser opens but links don't follow 5> Programs passing directly /usr/bin/firefox to open the link etc.

rusty-snake:
The open hyperlinks from sandboxed programs is a topic for Issues rather then PRs, so we can find out what necessary and what's the best way to deal with it.

PR #3918, opened 2021-01-24
Add profile for kdiff3

Neo00001:


etc/profile-a-l/kdiff3.profile
```diff
@@ -0,0 +1,43 @@
...
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+seccomp
+shell none
```
rusty-snake, line 38: Can we add `tracelog`?

etc/profile-a-l/kdiff3.profile
```diff
@@ -0,0 +1,43 @@
...
+
+disable-mnt
+private-bin  kdiff3
+private-cache
+private-dev
```
rusty-snake, line 43: Can we add `dbus-{user,system} {none,filter}`?

etc/profile-a-l/kdiff3.profile
```diff
@@ -0,0 +1,43 @@
...
+net none
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+seccomp
```
rusty-snake, line 37: Can we add `seccomp.block-secondary`.

rusty-snake (changes requested):
kdiff3 has a very similar usage as meld. We should do the same filesystem restrictions as in [meld.profile](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/meld.profile).

Can you add all the 
```
# Uncomment the next line (or put it into your meld.local) if you don't need to compare files in …
```
comments to make this restrictions opt-in.

etc/profile-a-l/kdiff3.profile
Neo00001: yep.

etc/profile-a-l/kdiff3.profile
Neo00001: for some reason,I always tend to forget about `tracelog`.

etc/profile-a-l/kdiff3.profile
Neo00001: added

rusty-snake (approved):

netblue30:
merged, thanks!

Issue #3906, opened 2021-01-21
Google Earth Pro not working on Archlinux


X6B:
Firejail 0.9.64 version running Google Earth Pro 7.3.3.7786-4 installed from AUR repository:

>google-earth-pro
Reading profile /etc/firejail/google-earth.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-exec.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-common.inc
Parent pid 31220, child pid 31221
Private /opt installed in 195.92 ms
7 programs installed in 4.55 ms
Warning: cleaning all supplementary groups
Warning: cleaning all supplementary groups
Warning: cleaning all supplementary groups
Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Warning: cleaning all supplementary groups
Child process initialized in 238.68 ms
Error: no suitable /usr/bin/google-earth-pro executable found
Parent is shutting down, bye...>

And this is my local profile to make the program work:

>private-bin google-earth-pro,readlink
include /etc/firejail/google-earth.profile

rusty-snake:
What does `firejail --ignore=private-bin --profile=google-earth ls -l /usr/bin/google-earth` show?

X6B:
> firejail --ignore=private-bin --profile=google-earth ls -l /usr/bin/google-earth
Reading profile /etc/firejail/google-earth.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-exec.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-common.inc
Parent pid 77280, child pid 77281
Private /opt installed in 197.85 ms
Warning: cleaning all supplementary groups
Warning: cleaning all supplementary groups
Warning: cleaning all supplementary groups
Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Warning: cleaning all supplementary groups
Child process initialized in 245.52 ms
[...]

PR #3923, opened 2021-01-28
add extensive comment on sandboxing google-earth-pro

glitsj16:


glitsj16:
Forgot to mention this relates to #3906, allthough its in the comments.

Issue #3647, opened 2020-09-27
webkit2gtk-4.0 requires bwrap


SkewedZeppelin:
The new webkit2gtk-4.0 seems to hard require bubblewrap.
This causes firejailed programs such as evolution and epiphany fail to start.
The profiles will need to be updated ala chromium probably.

reinerh:
surf is another one using webkitgtk. though after a quick test it was still working in firejail.

rusty-snake:
FWIW: https://blogs.gnome.org/mcatanzaro/2020/03/31/sandboxing-webkitgtk-apps/

---

Fedoras kernel has unprivileged user-ns always enabled. The non-suid bwrap binary which is therefore used as default in Fedora works with these minimal changes:

```
$ firejail '--noblacklist=${PATH}/bwrap' '--seccomp=!mount,!pivot_root,!umount2' bwrap --dev-bind / / ls
```

However the suid variant will need more permissions, such as no nonewprivs.

ckotte:
Just adding ```noblacklist /usr/bin/bwrap``` works with evolution 3.38 on Arch Linux

rusty-snake:
> How can I put those seccomp and dev-bind options into the profile?

1. `--dev-bind / /` is a bwrap option.
2. Just create a evolution.local (in e.g. ~/config/firejail) with this content
```
noblacklist ${PATH}/bwrap
seccomp !mount,!pivot_root,!umount2
```
However, from the error message you likely need to add `protocol unix,inet,inet6,netlink` (adding `netlink`) likewise.

ckotte:
Now I get this ```bwrap: Can't mount proc on /newroot/proc: Operation not permitted```. Probably better to not use firejail with evolution anymore..

rusty-snake:
Does Arch ship the suid variant? (check `ls -l /usr/bin/bwrap`)
Does it work with `--noprofile`? If not we'll need to remove all those programs from firecfg (or make it work).

ckotte:
> Does Arch ship the suid variant? (check `ls -l /usr/bin/bwrap`)

No suid:
```
ls -l /usr/bin/bwrap
-rwxr-xr-x 1 root root 55352 Mar 30  2020 /usr/bin/bwrap
```

> Does it work with `--noprofile`? If not we'll need to remove all those programs from firecfg (or make it work).

Doesn't work with ```--noprofile```. I always get ```bwrap: Can't mount proc on /newroot/proc: Operation not permitted```

[...]

PR #3926, opened 2021-01-29
Disable the webkit2gtk-4.0 sandbox in bijiben

rusty-snake:
webkit2gtk uses a bwrap based sandbox by default since 4.0, see #3647.
This is good as it means more security by default on for linux system.
Unfortunately is it not possible to run bwrap inside firejail if bwrap
is started with --unshare-pid --proc /proc at all. In general we should
exclude a program from firecfg until a final solution is found. But
bijiben is special, while epiphany or evolution display random stuff
from the internet is webkit2gtk in bijiben used to display local files
create by the user. Bijiben has a thight profile (net none, whitelist,
private-bin, ...) therefore my decision here was to disable the
webkit2gtk sandbox rather then firejail.

### DIFF
--- a/etc/profile-a-l/bijiben.profile
+++ b/etc/profile-a-l/bijiben.profile
@@ -57,3 +57,5 @@ dbus-user.own org.gnome.Notes
 dbus-user.talk ca.desrt.dconf
 dbus-user.talk org.freedesktop.Tracker1
 dbus-system none
+
+env WEBKIT_FORCE_SANDBOX=0


glitsj16 (approved):

glitsj16:
@rusty-snake LGTM, including the reasoning behind it :+1:. On a side note, CI is failing also in #3924. I'm not very good with CI so I went over the PR several times but couldn't spot anything fishy that would explain things. Now that I see this one is also failing it looks like it's currently 'kaput'. #3924 is strictly cosmetic and can wait until CI returns to normal. But this one is more important IMO so I'm not sure on how to proceed in cases like this. Feel free to decide whatever you think is best.

PR #3930, opened 2021-01-30
manpage: clarify seccomp behavior

fissie:
Behavior of "seccomp" and "seccomp @default" differs in a very non-obvious way (generating kills vs. EPERM), which is hard to see from the code or other places in the documentation.

### DIFF
--- a/src/man/firejail.txt
+++ b/src/man/firejail.txt
@@ -2148,6 +2148,13 @@ installed with \-\-seccomp.32.
 .br
 
 .br
+Note that there is a semantic difference between using plain "seccomp" and
+using for example "seccomp @default". "seccomp" will default to killing the
+process, while "seccomp @default" will default to returning EPERM. This may be
+adjusted using \-\-seccomp-error-action.
+.br
+
+.br
 Firejail will print seccomp violations to the audit log if the kernel was compiled with audit support (CONFIG_AUDIT flag).
 .br
 


rusty-snake:
bug or feature? I would say bug.

rusty-snake:
This kills! my test program:

```
firejail --noprofile --seccomp-error-action=EPERM --seccomp unshare -r ./chroot
```

while this return EPERM:

```
firejail --noprofile --seccomp-error-action=EPERM --seccomp=chroot unshare -r ./chroot
```

looks like `--seccomp` ignores `seccomp-error-action`.

----

> seccomp @default

makes no sense, only `seccomp some_thing_no_in_@default` or `seccomp.drop @default` make sense.

fissie:
Ok, I can rewrite this a bit, but the point was that using seccomp <nothing> and seccomp <something> yields different results in terms of how seccomp filter is generated. But you are right, it may be that it's actually a bug, and the result should be the same in both.

fissie:
The way I found the difference was with audit. The stock profile for 'psi' had
seccomp !chroot (generates EPERM, respects the flag) 
While my old personal profile which I'm getting rid of had plain seccomp (generates KILL, doesn't respect the flag). 

Running audit with seccomp is happy (because --audit relies on the program being killed, and doesn't check for the return value), while running audit with "seccomp !chroot" is sad, because this will start returning EPERM instead of killing the program.

The fact that the audit binary should probably take EPERM into consideration, instead of relying on killing the program, is also true, and something I wanted to fix afterwards ;-).

topimiettinen:
It's a bug, so let's not document it as a feature...

fissie:
Sounds reasonable. Do you want me to report it into the tracker?

topimiettinen:
Please do.

The problem could be `fsec-optimize`, which always uses KILL action instead of honoring the seccomp error action setting.

fissie:
Closing pull request, as this is a bug. Will move the details into a bug.

PR #3935, opened 2021-01-30
Add profile for avidemux

Neo00001:


rusty-snake (commented):
LGTM. One thing, the whitelist block (line 37-45) and the options block (line 21-35) should be flipped.

Neo00001:
> the whitelist block (line 37-45) and the options block (line 21-35) should be flipped.

done.

rusty-snake (approved):

rusty-snake:
mergde, thanks.

PR #3936, opened 2021-01-31
Fix minor typo in firecfg's manual page

irandms:
I believe this to be a minor typo, given the precedence of `/usr` as a directory on many Linux distributions, and the lines following this which explicitly reference `/usr/local/bin `instead of `/user/local/bin` .


rusty-snake:
merged, thanks.

PR #3947, opened 2021-02-05
add a /usr/share whitelist item for uim

antonv6:
uim is a multilingual input method framework, so any program that takes user input potentially needs it to work.


rusty-snake:
merged, thanks.

PR #3950, opened 2021-02-06
disable-interpreters.inc: blacklist the other libmozjs

kmk3:
And sort the paths on allow-gjs.inc.

    $ pacman -Q js78
    js78 78.6.0-1
    $ pacman -Qlq js78 | grep -v /usr/include/
    /usr/
    /usr/bin/
    /usr/bin/js78
    /usr/bin/js78-config
    /usr/lib/
    /usr/lib/libmozjs-78.so
    /usr/lib/pkgconfig/
    /usr/lib/pkgconfig/mozjs-78.pc

This appears to be the only counterpart path missing when looking at the
current lib64 entries with:

    $ grep -Fnr lib64 etc

---

Note: As mentioned on #3845, please don't squash any commits; use normal merge
or rebase instead.

### DIFF
--- a/etc/inc/allow-gjs.inc
+++ b/etc/inc/allow-gjs.inc
@@ -5,7 +5,8 @@ include allow-gjs.local
 noblacklist ${PATH}/gjs
 noblacklist ${PATH}/gjs-console
 noblacklist /usr/lib/gjs
-noblacklist /usr/lib64/gjs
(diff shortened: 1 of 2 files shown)


glitsj16 (approved):
LGTM

rusty-snake:
Thanks.

PR #3970, opened 2021-02-10
small man fixes

glitsj16:


src/man/firejail-profile.txt
```diff
@@ -266,7 +266,7 @@ Mount new /root and /home/user directories in temporary
 filesystems. All modifications are discarded when the sandbox is
 closed.
 .TP
-\fBprivate directory
+\fBprivate=directory
```
rusty-snake, line 269: The profile syntax is indeed `private ${HOME}/.foo/bar` rather then `private=${HOME}/.foo/bar`.

glitsj16: Hmmm, I thought the man pages use the command line syntax, that's why I changed it. Without "=" that breaks. Confused!

glitsj16:
```
$ cd ~/Downloads
$ mkdir test-pr3970
$ firejail --noprofile --private ${HOME}/Downloads/test-pr3970
Error: invalid command name
$ firejail --noprofile --private=${HOME}/Downloads/test-pr3970
Parent pid 32695, child pid 32696
Child process initialized in 58.31 ms
[glitsj16@lab16 ~]$ exit
exit

```
@rusty-snake Are you seeing anything different?

rusty-snake:
```bash
$ cat > test.profile <<EOF
private=${HOME}/my_home
EOF
$ firejail --profile=test.profile
Reading profile test.profile
Error: line 1 in test.profile is invalid
$ sed "s/=/ /g" -i test.profile
$ firejail --profile=test.profile
Reading profile test.profile
Error: invalid private directory
$ mkdir my_home
$ firejail --profile=test.profile
Reading profile test.profile
Parent pid 187773, child pid 187774
Child process initialized in 27.03 ms
[aurora@localhost ~]$ exit

```

glitsj16: @rusty-snake Sure, I understand what you mean. There has always been a difference between CLI and configuration file syntax no? That isn't limited to this particular option. I think we need to add a note to that effect in our man pages!

Issue #3952, opened 2021-02-06
New Firefox profiles break KeePassXC addon


NetSysFire:
**Bug and expected behavior**

The newest profiles break the KeePassXC Firefox addon. I verified that Firefox (and not KeePassXC) is the application causing problems by running them with and without a sandbox. It started to work again when I ran Firefox without a sandbox.

**No profile and disabling firejail**
Running it without Firejail works. I downloaded an [older package](https://archive.archlinux.org/packages/f/firejail/firejail-0.9.58-1-x86_64.pkg.tar.xz), and copied `firefox*.profile` to `~/.config/firejail/` to force it to use the old profiles. It worked again.

**Reproduce**

- Make sure to use the latest version of Firefox, Firejail and KeePassXC (also its browser addon)
- Open Firefox, the extension will complain about the fact that it can not connect

**Environment**
 - Distro: Arch Linux
 - Firejail version: 0.9.64.2-1
 I compared the checksums of the relevant profiles contained in the Arch Linux package and the ones in the git repo. I verified that these are the latest profiles.

**Additional context**

I `diff`ed the old, confirmed working profiles (although I had some U2F problems, I had to make sure my yubikey was plugged in before starting the browser), encountered [a comment](https://github.com/netblue30/firejail/blob/master/etc/inc/firefox-common-addons.inc#L70) and began working on my overrides (`.local`). I tried putting this in **`firefox-common-addons.local`**:
```
private-bin keepassxc-proxy
include whitelist-runuser-common.inc
```
but unfortunately this did not resolve this issue (with the newest profiles of course). I suspect it is either related to the newly introduced dbus-filtering or to a now inaccessible socket location, which is why I tried adding that whitelist include into my override. I could not get any useful debug output from Firefox or KeePassXC and there are not any blacklist violations in my journal.
Let me know if there is anything else I can do to debug this, but I have no clue about all the dbus filtering.

**Checklist**
 - [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
 - [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
 - [x] I have performed a short search for similar issues (to avoid opening a duplicate).
 - [ ] If it is a AppImage, `--profile=PROFILENAME` is used to set the right profile.
 - [x] Used `LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 PROGRAM` to get english error-messages.
[...]

PR #3984, opened 2021-02-13
Fix for KeePassXC plugin

nidamanx:
KeePassXC changed the socket name.
https://github.com/keepassxreboot/keepassxc/commit/a145bf91191f0a4630a7e31654aff8a8dfd09bf0
Keep also old socket name in whitelist for back compatibility

### DIFF
--- a/etc/profile-a-l/firefox.profile
+++ b/etc/profile-a-l/firefox.profile
@@ -49,7 +49,7 @@ dbus-user.own org.mpris.MediaPlayer2.firefox.*
 # implementation)
 #ignore noroot
 # Uncomment or put in your firefox.local one of the following whitelist to enable KeePassXC Plugin
-# note: run KeePassXC before Firefox to allow coomunication between them
+# NOTE: start KeePassXC before Firefox and keep it open to allow communication between them
 #whitelist ${RUNUSER}/kpxc_server
 #whitelist ${RUNUSER}/org.keepassxc.KeePassXC.BrowserServer
 ignore dbus-user none
--- a/etc/profile-a-l/firefox.profile
+++ b/etc/profile-a-l/firefox.profile
@@ -14,6 +14,11 @@ mkdir ${HOME}/.mozilla
 whitelist ${HOME}/.cache/mozilla/firefox
 whitelist ${HOME}/.mozilla
 
+# Uncomment or put in your firefox.local one of the following whitelist to enable KeePassXC Plugin
+# NOTE: start KeePassXC before Firefox and keep it open to allow communication between them
+#whitelist ${RUNUSER}/kpxc_server
+#whitelist ${RUNUSER}/org.keepassxc.KeePassXC.BrowserServer
+
 whitelist /usr/share/doc
 whitelist /usr/share/firefox
 whitelist /usr/share/gnome-shell/search-providers/firefox-search-provider.ini
@@ -48,10 +53,6 @@ dbus-user.own org.mpris.MediaPlayer2.firefox.*
 # does not work with the above lines (might depend on the portal
 # implementation)
 #ignore noroot
-# Uncomment or put in your firefox.local one of the following whitelist to enable KeePassXC Plugin
-# NOTE: start KeePassXC before Firefox and keep it open to allow communication between them
-#whitelist ${RUNUSER}/kpxc_server
-#whitelist ${RUNUSER}/org.keepassxc.KeePassXC.BrowserServer
 ignore dbus-user none
 
 # Redirect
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/firefox.profile
```diff
@@ -48,6 +48,10 @@ dbus-user.own org.mpris.MediaPlayer2.firefox.*
 # does not work with the above lines (might depend on the portal
 # implementation)
 #ignore noroot
+# Uncomment or put in your firefox.local one of the following whitelist to enable KeePassXC Plugin
+# note: run KeePassXC before Firefox to allow coomunication between them
```
rusty-snake, line 52:
```suggestion
# NOTE: start KeePassXC before Firefox and keep it open to allow communication between them
```

rusty-snake (commented):
Since it is a `whitelist`, it shoud go to the whitelist block(s) above (line 12-23).

rusty-snake:
> Since it is a whitelist, it shoud go to the whitelist block(s) above (line 12-23).

and since it is a Add-On related thing, it should go in firefox-common-addons.inc. There we conflict with `ignore include whitelist-runuser-common.inc` (which I added because of KPXC).

nidamanx:
> > Since it is a whitelist, it shoud go to the whitelist block(s) above (line 12-23).
> 
> and since it is a Add-On related thing, it should go in firefox-common-addons.inc. There we conflict with `ignore include whitelist-runuser-common.inc` (which I added because of KPXC).

Genau! :-)
Let me do a full test before
Also commenting the line `ignore include whitelist-runuser-common.inc`

nidamanx:
> > > Since it is a whitelist, it shoud go to the whitelist block(s) above (line 12-23).
> > 
> > 
> > and since it is a Add-On related thing, it should go in firefox-common-addons.inc. There we conflict with `ignore include whitelist-runuser-common.inc` (which I added because of KPXC).
> 

Very strange.
My setup (Default Debian) is ignoring firefox-common-addons.inc and firefox-common-addons.local
Starting `firejail firefox` from terminal, I cannot see parsing no one of those files.
And, it didn't work at all while setting up the lines for KPXC (and commenting the ones in firefox.local profile, 
I tried in .config/firejail and in /etc/firejail in all combinatione, but no way.

nidamanx:
> > Since it is a whitelist, it shoud go to the whitelist block(s) above (line 12-23).
> 
> and since it is a Add-On related thing, it should go in firefox-common-addons.inc. There we conflict with `ignore include whitelist-runuser-common.inc` (which I added because of KPXC).

Very strange.
My setup (Default Debian) is ignoring firefox-common-addons.inc and firefox-common-addons.local
Starting firejail firefox from terminal, I cannot see parsing no one of those files.
And, it didn't work at all while setting up the lines for KPXC (and commenting the ones in firefox.local profile,
I tried in .config/firejail and in /etc/firejail in all combinatione, but no way.

rusty-snake:
Sorry, I don't fully get you. firefox-common-addons.inc is opt-in because it has wide relaxations and is used by a minority, if this is where you hang.

nidamanx:
> Sorry, I don't fully get you. firefox-common-addons.inc is opt-in because it has wide relaxations and is used by a minority, if this is where you hang.

So, that's the reason why I cannot see the parse of firefox-common-addons.inc while starting from the terminal.
But, anyway, if I insert the lines about KPRC **only** in firefox-common-addons.inc or firefox-common-addons.local, I cannot have a fully working plugin.
It really seems I need to use this way:
```
$ ls -1 ~/.config/firejail/
firefox.local
firefox.profile
keepassxc.profile
```

P.S. firefox.profile and keepassxc.profile are there to override the default from .deb (so I'm sure to use the very last from the git)
And firefox.local contains just the few lines fos KPHC

kmk3:
(For cross-reference, this PR stems from the following issue: #3952)

nidamanx:
Tried again this morning without good results.
The only way seems really to follow [3952#issuecomment-778640094](https://github.com/netblue30/firejail/issues/3952#issuecomment-778640094)

For the test, I deleted firefox.local and used the updated version for all firefox.profiles,
No way (at least on Debian 10) to use `/etc/firejail/firefox-common-addons.local`
I tried the following lines in any combinations but really seems this .local file is ignored
```
ignore include whitelist-runuser-common.inc (commented/uncommented/ignored/included)
#private-bin keepassxc-proxy (commented/uncommented)
whitelist ${RUNUSER}/kpxc_server (never commented)
```

nidamanx:
What about this profile?
With the right socket seems going well and with no issues since 10 days

rusty-snake:
My worry is that I don't understand why firefox-common-addons.inc doesn't work.

nidamanx:
> why firefox-common-addons.inc doesn't work

Mhhh, clear.
So, better investigate more. Let me do some more tests

One question: is the new named bus from keepassxc already included somewhere? if not, why firefox-common-addons should allow it? ...maybe I'm missing something
I think the problem was only the new renamed bus. All the rest is still the same. Isn't that correct?

netblue30:
Merged, thanks!

Issue #3979, opened 2021-02-12
KMail profile in 0.9.64.4 has multiple fatal errors


creideiki:
**Bug and expected behavior**
Gentoo has been lagging a bit with updates, so I did not test any version between 0.9.64 and 0.9.64.4. In 0.9.64, KMail (and Akonadi) ran without issues using the default profile. In 0.9.64.4, the application does not run at all. It seems the refactoring of the profile done in 5532fbdb9749c5333ac03152f8c94fd364182d32 introduced several errors.

First, there are syntax errors. Several `mkdir` lines have file names containing asterisks. This gives the following error:
```
Error: "${HOME}/.cache/akonadi*" is an invalid filename: rejected character: "*"
```
I am not sure what they intend to do, but whatever it is it's not working. Especially confusing is the line
```
mkdir /tmp/akonadi-*
```
Yes, Akonadi creates a directory in `/tmp`, but its name is random and seems to have been created using `mkstemp(3)` or similar. I'm not sure how Firejail is supposed to be able to pre-create it.

Removing the asterisks makes Firejail at least accept the profile syntactically and try to run the program.

At startup, Firejail now prints the following warning:
```
***
*** Warning: cannot whitelist ${DOCUMENTS} directory
*** Any file saved in this directory will be lost when the sandbox is closed.
***
```
This seems like it would be a problem for saving e-mail attachments.

The next error comes when KMail tries to start Akonadi, which tries to start PostgreSQL. This fails with the following error:
```
org.kde.pim.akonadiserver: stdout: "waiting for server to start....2021-02-12 11:03:25.168 CET [53] FATALT:  kunde inte \xC3\xB6ppna katalog \"/usr/share/postgresql-13/timezonesets\": Filen eller katalogen finns inte\n2021-02-12 11:03:25.168 CET [53] TIPS:  Detta tyder p\xC3\xA5 en inkomplett PostgreSQL-installation alternativt att filen \"/usr/lib64/postgresql-13/bin/postgres\" har flyttats bort fr\xC3\xA5n sin korrekta plats.\n stopped waiting\n"
```
I am not sure why it is in Swedish, when I have set `LC_MESSAGES=POSIX`, but that's a different problem. The actual error is that PostgreSQL needs access to `/usr/share/postgresql-13/` in order to run. Adding the following line to `kmail.profile` fixes that:
```
whitelist /usr/share/postgresql*
```

The next problem is this message on the console:
```
[...]

PR #3985, opened 2021-02-13
Sort.py updates

rusty-snake:
0f76e53 sort.py: Better error message
19c92f2 sort.py: Print the fixed line when running in a CI
9f939b8 workflows/sort.yml: Run also if sort.py is changed

### DIFF
--- a/contrib/sort.py
+++ b/contrib/sort.py
@@ -111,6 +111,7 @@ def fix_profile(filename):
 
 def main(args):
     exit_code = 0
+    print(f"sort.py: checking {len(args)} {'profiles' if len(args) != 1 else 'profile'} ...")
     for filename in args:
         try:
             if exit_code not in (1, 101):
--- a/contrib/sort.py
+++ b/contrib/sort.py
@@ -120,8 +120,8 @@ def main(args):
         except PermissionError:
             print(f"[ Error ] Can't read/write `{filename}'")
             exit_code = 1
-        except:
-            print(f"[ Error ] An error occurred while processing `{filename}'")
+        except Exception as err:
+            print(f"[ Error ] An error occurred while processing `{filename}': {err}")
             exit_code = 1
     return exit_code
 
--- a/contrib/sort.py
+++ b/contrib/sort.py
@@ -24,6 +24,7 @@ Exit-Codes:
 
 # Requirements:
 #  python >= 3.6
+from os import getenv
 from sys import argv
 
 
@@ -80,7 +81,7 @@ def fix_profile(filename):
         lines = profile.read().split("\n")
         was_fixed = False
         fixed_profile = []
-        for line in lines:
+        for lineno, line in enumerate(lines):
             if line[:12] in ("private-bin ", "private-etc ", "private-lib "):
                 fixed_line = f"{line[:12]}{sort_alphabetical(line[12:])}"
             elif line[:13] in ("seccomp.drop ", "seccomp.keep "):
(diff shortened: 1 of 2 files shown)


contrib/sort.py
```diff
@@ -95,6 +96,8 @@ def fix_profile(filename):
                 fixed_line = line
             if fixed_line != line:
                 was_fixed = True
+                if getenv("CI"):
+                    print(f"{filename}:{lineno + 1}:{fixed_line}")
```
kmk3, line 100: Why print only on CI?

I would like to see what it does/does not when running locally.  Also, I like
the output of this print call as it is similar to what is done by both the
linter and the other script that I mentioned [here][1] (except that sort.py
prints the fixed version rather than the original one).

For example, I have been working on a PR that (among other things) adds a
makefile target for running sort.py (to then add the aforementioned linter on
it), so that it's easier to run locally and one of my gripes with it is the
lack of feedback when there are no errors.  For example, when messing with the
makefile and running the target, I wasn't sure if sort.py was even being
executed and if it was, if it did actually check the profiles that I intended
to check or if I passed an empty/wrong var.

It would be nice to see something like this at the beginning:

```
sort.py: checking 1001 profiles...
```

Example in shell (mentally replace it with the python equivalent):

```sh
$ printf '%s: checking %s profiles...\n' "$(basename "$0")" $#
```

Full output example:

```
sort.py: checking 1001 profiles...
etc/profile-a-l/ark.profile:39:private-bin 7z,aaa,ark,bash,lrzip,lsar,lz4,lzop,p7zip,rar,sh,tclsh,unar,unrar,unzip,zip,zipinfo
[ Fixed ] etc/profile-a-l/ark.profile
etc/profile-m-z/mpv.profile:75:private-bin aaa,env,mpv,python*,waf,youtube-dl
[ Fixed ] etc/profile-m-z/mpv.profile
```

Though this other print could be left for another PR of course.

Other than that, LGTM.  The commits are looking nice too.

[1]: https://github.com/netblue30/firejail/issues/3979#issuecomment-778491722

reinerh (approved):

kmk3 (approved):
Nice, I like the new diff-like output:

```
sort.py: checking 2 profiles...
etc/profile-a-l/ark.profile:39:-private-bin 7z,ark,bash,lrzip,lsar,lz4,lzop,p7zip,rar,sh,tclsh,unar,unrar,unzip,zip,zipinfo,aaa
etc/profile-a-l/ark.profile:39:+private-bin 7z,aaa,ark,bash,lrzip,lsar,lz4,lzop,p7zip,rar,sh,tclsh,unar,unrar,unzip,zip,zipinfo
[ Fixed ] etc/profile-a-l/ark.profile
etc/profile-m-z/mpv.profile:75:-private-bin env,mpv,python*,waf,youtube-dl,aaa
etc/profile-m-z/mpv.profile:75:+private-bin aaa,env,mpv,python*,waf,youtube-dl
[ Fixed ] etc/profile-m-z/mpv.profile
```

I have some more enhancement suggestions, but I'd rather finish writing them up
later / on a new thread.

**Edit:** I meant "on a new PR" rather than "on a new thread [here]".  As in,
by me this is ready for merging.

---

By the way, this is the script that I used to generate/compare the outputs:

```sh
#!/bin/sh

set -e
set -- etc/profile-a-l/ark.profile etc/profile-m-z/mpv.profile

git checkout -- "$@"
sed -i.bak -e 's/^private-bin .*/&,aaa/' -- "$@"
./contrib/sort.py "$@"
```

Then, to clean up:

```sh
find etc/profile-* -name '*.bak' -exec rm '{}' +
```

PR #3988, opened 2021-02-14
add apparmor to torbrowser-launcher

glitsj16:
At least on Arch Linux torbrowser-launcher comes with AppArmor support. Enabling apparmor in our profile sounds reasonable and shouldn't break on other distro's that do packaging differently.

### DIFF
--- a/etc/profile-m-z/torbrowser-launcher.profile
+++ b/etc/profile-m-z/torbrowser-launcher.profile
@@ -31,6 +31,7 @@ whitelist ${HOME}/.local/share/torbrowser
 include whitelist-common.inc
 include whitelist-var-common.inc
 
+apparmor
 caps.drop all
 netfilter
 nodvd


rusty-snake:
`apparmor` makes `${HOME}` `noexec`, does it not? And torbrowser-launcher installs the torbrowser under `~/.local/share/torbrowser-launcher`. I'm confused, can this work?

Vincent43:
I have the same concern as above. Was this tested?

glitsj16:
@rusty-snake @Vincent43 This PR only adds 'apparmor'. The profile already has `ignore noexec ${HOME}`:  https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/torbrowser-launcher.profile#L9. Too bad the history only goes back to the [etc reorganization](https://github.com/netblue30/firejail/commit/018d75775eab4a0f045949a9d069c57686ca2686) so I can't track things further away in time.

I have been running with apparmor in my torbrowser-launcher.local since I started to use that package on Arch Linux, which according to its 'Update History' was September 24, 2020. So yes, I tested this. That(s why I thought it was a good idea to add it here in the first place. I'll redo the testing on my side again. Would be interesting to compare observations if any of you or both can find the time to test torbrowser-launcher too with the apparmor option.

glitsj16:
@rusty-snake @Vincent43 After more extensive testing it turned out I had a AppArmor rule burried in my /etc/apparmor.d/local/firejail-default that made this work. Adding it to https://github.com/netblue30/firejail/blob/master/etc/apparmor/firejail-default should provide the missing piece of the puzzle IMO.

PR #3990, opened 2021-02-15
Follow-up fixes for torbrowser-launcher

glitsj16:
I retested enabling the apparmor option after legitimate questions posed by @rusty-snake and @Vincent43 in #3988 - thanks!. Apologies for the confusion, I forgot to add the AppArmor rule to allow access to the torbrowser-launcher executables installed under ${HOME}.

### DIFF
--- a/etc/apparmor/firejail-default
+++ b/etc/apparmor/firejail-default
@@ -82,9 +82,6 @@ owner /proc/@{PID}/{uid_map,gid_map,setgroups} w,
 owner /proc/@{PID}/oom_score_adj w,
 owner /proc/@{PID}/clear_refs w,
 
-# Needed for torbrowser-launcher
-owner @{HOME}/.local/share/torbrowser/tbb/{i686,x86_64}/tor-browser_*/Browser/** ix,
-
 ##########
 # Allow running programs only from well-known system directories. If you need
 # to run programs from your home directory, uncomment /home line.
--- a/etc/apparmor/firejail-default
+++ b/etc/apparmor/firejail-default
@@ -82,6 +82,9 @@ owner /proc/@{PID}/{uid_map,gid_map,setgroups} w,
 owner /proc/@{PID}/oom_score_adj w,
 owner /proc/@{PID}/clear_refs w,
 
+# Needed for torbrowser-launcher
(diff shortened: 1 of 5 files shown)


etc/apparmor/firejail-default
```diff
@@ -82,6 +82,9 @@ owner /proc/@{PID}/{uid_map,gid_map,setgroups} w,
 owner /proc/@{PID}/oom_score_adj w,
 owner /proc/@{PID}/clear_refs w,
 
+# Needed for torbrowser-launcher
+owner @{HOME}/.local/share/torbrowser/tbb/{i686,x86_64}/tor-browser_*/Browser/** ix,
```
Vincent43, line 86: Before such rules were recommended only as user specific overrides in `firejail-local `policy. The idea behind current `firejail-default` policy is to block executing code from user writable paths, see [comment](https://github.com/netblue30/firejail/blob/970d679d85a60792373f3b324497e91414d1b716/etc/apparmor/firejail-default#L89):

> Allow running programs only from well-known system directories

Adding such override in `firejail-default` will self-circumvent that policy as it will be bypassed by any code put on writable path. For this reason I think firejail-default shouldn't be patched such way but torrowser-launcher and any other app that execute something from home should remove appamror instead (leaving it as user opt-in if they want to tweak firejail-local policy on their own).

glitsj16: Thanks for the review. I agree it makes more sense to implement this as opt-in and adapted the PR accordingly.

etc/profile-m-z/torbrowser-launcher.profile
```diff
@@ -31,7 +31,10 @@ whitelist ${HOME}/.local/share/torbrowser
 include whitelist-common.inc
 include whitelist-var-common.inc
 
-apparmor
+# Uncomment the line below or put 'apparmor' in your torbrowser-launcher.local.
+# IMPORTANT: the relevant rule in /etc/apparmor.d/local/firejail-default will need
```
rusty-snake, line 35:
```suggestion
# IMPORTANT: the relevant rule in /etc/apparmor.d/local/firejail-local will need
```

PR #3997, opened 2021-02-18
Create nextcloud-desktop.profile

nidamanx:
It works also using the official AppImage from Nextcloud.com

### DIFF
--- a/etc/profile-m-z/nextcloud-desktop.profile
+++ b/etc/profile-m-z/nextcloud-desktop.profile
@@ -1,70 +1,10 @@
-# Firejail profile for nextcloud-desktop
-# Description: Nextcloud desktop client
+# Firejail profile alias for nextcloud
 # This file is overwritten after every install/update
 # Persistent local customizations
 include nextcloud-desktop.local
 # Persistent global definitions
-include globals.local
+# added by included profile
+#include globals.local
 
-noblacklist ${HOME}/Nextcloud
-noblacklist ${HOME}/.config/Nextcloud
-noblacklist ${HOME}/.local/share/Nextcloud
-# Uncomment or put in your nextcloud-desktop.local to allow sync with more directories.
-#noblacklist ${DOCUMENTS}
-#noblacklist ${MUSIC}
-#noblacklist ${PICTURES}
-#noblacklist ${VIDEOS}
-
-include disable-common.inc
-include disable-devel.inc
-include disable-exec.inc
-include disable-interpreters.inc
-include disable-passwdmgr.inc
-include disable-programs.inc
-include disable-shell.inc
-include disable-xdg.inc
-
-mkdir ${HOME}/Nextcloud
-mkdir ${HOME}/.config/Nextcloud
-mkdir ${HOME}/.local/share/Nextcloud
-whitelist ${HOME}/Nextcloud
-whitelist ${HOME}/.config/Nextcloud
-whitelist ${HOME}/.local/share/Nextcloud
-# Uncomment or put in your nextcloud-desktop.local to allow sync with more directories.
-#whitelist ${DOCUMENTS}
-#whitelist ${MUSIC}
-#whitelist ${PICTURES}
-#whitelist ${VIDEOS}
-include whitelist-common.inc
-include whitelist-runuser-common.inc
-include whitelist-usr-share-common.inc
-include whitelist-var-common.inc
-
-apparmor
-caps.drop all
-machine-id
-netfilter
-no3d
-nodvd
-nogroups
-nonewprivs
-noroot
-nosound
-notv
-nou2f
-novideo
-protocol unix,inet,inet6
-seccomp
-seccomp.block-secondary
-shell none
-tracelog
-
-disable-mnt
-private-bin nextcloud,nextcloud-desktop
-private-cache
-private-dev
-private-tmp
-
-dbus-user filter
-dbus-user.talk org.freedesktop.secrets
-dbus-system none
+# Redirect
+include nextcloud.profile
--- a/etc/profile-m-z/nextcloud-desktop.profile
+++ b/etc/profile-m-z/nextcloud-desktop.profile
@@ -8,9 +8,6 @@ include globals.local
 
 # Uncomment or put in your nextcloud-desktop.local to allow sync with Document directory.
 #noblacklist ${DOCUMENTS}
-noblacklist ${HOME}/Nextcloud
-noblacklist ${HOME}/.config/Nextcloud
-noblacklist ${HOME}/.local/share/Nextcloud
 
 include disable-common.inc
 include disable-devel.inc
@@ -60,6 +57,5 @@ private-dev
 private-tmp
 
 dbus-user filter
-dbus-user.own com.nextcloud.desktopclient.nextcloud
 dbus-user.talk org.freedesktop.secrets
 dbus-system none
--- a/etc/profile-m-z/nextcloud-desktop.profile
+++ b/etc/profile-m-z/nextcloud-desktop.profile
@@ -60,7 +60,7 @@ shell none
 tracelog
 
 disable-mnt
-private-bin nextcloud-desktop
+private-bin nextcloud,nextcloud-desktop
 private-cache
 private-dev
 private-tmp
--- a/etc/profile-m-z/nextcloud-desktop.profile
+++ b/etc/profile-m-z/nextcloud-desktop.profile
@@ -30,6 +30,7 @@ include whitelist-common.inc
 include whitelist-var-common.inc
 
 caps.drop all
+machine-id
 netfilter
 no3d
 nodvd
--- a/etc/profile-m-z/nextcloud-desktop.profile
+++ b/etc/profile-m-z/nextcloud-desktop.profile
@@ -13,9 +13,11 @@ noblacklist ${HOME}/.local/share/Nextcloud
 
 include disable-common.inc
 include disable-devel.inc
+include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
+include disable-shell.inc
 
 mkdir ${HOME}/Nextcloud
 mkdir ${HOME}/.config/Nextcloud
--- a/etc/profile-m-z/nextcloud-desktop.profile
+++ b/etc/profile-m-z/nextcloud-desktop.profile
@@ -6,7 +6,8 @@ include nextcloud-desktop.local
 # Persistent global definitions
 include globals.local
 
-noblacklist ${DOCUMENTS}
+# Uncomment or put in your nextcloud-desktop.local to allow sync with Document directory.
+#noblacklist ${DOCUMENTS}
 noblacklist ${HOME}/Nextcloud
 noblacklist ${HOME}/.config/Nextcloud
 noblacklist ${HOME}/.local/share/Nextcloud
@@ -18,15 +19,19 @@ include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
 include disable-shell.inc
+include disable-xdg.inc
 
 mkdir ${HOME}/Nextcloud
 mkdir ${HOME}/.config/Nextcloud
 mkdir ${HOME}/.local/share/Nextcloud
-whitelist ${DOCUMENTS}
+# Uncomment or put in your nextcloud-desktop.local to allow sync with Document directory.
+#whitelist ${DOCUMENTS}
 whitelist ${HOME}/Nextcloud
 whitelist ${HOME}/.config/Nextcloud
 whitelist ${HOME}/.local/share/Nextcloud
 include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
 caps.drop all
@@ -43,9 +48,12 @@ nou2f
 novideo
 protocol unix,inet,inet6
 seccomp
+seccomp.block-secondary
 shell none
+tracelog
 
+disable-mnt
+private-bin nextcloud-desktop
+private-cache
 private-dev
 private-tmp
-
-noexec /tmp
--- a/etc/profile-m-z/nextcloud-desktop.profile
+++ b/etc/profile-m-z/nextcloud-desktop.profile
@@ -34,6 +34,7 @@ include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
+apparmor
 caps.drop all
 machine-id
 netfilter
@@ -57,3 +58,5 @@ private-bin nextcloud-desktop
 private-cache
 private-dev
 private-tmp
+
+dbus-system none
--- a/etc/profile-m-z/nextcloud-desktop.profile
+++ b/etc/profile-m-z/nextcloud-desktop.profile
@@ -1,0 +1,48 @@
+# Firejail profile for nextcloud-desktop
+# Description: Nextcloud desktop client
+# This file is overwritten after every install/update
+# Persistent local customizations
+include nextcloud-desktop.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${DOCUMENTS}
+noblacklist ${HOME}/Nextcloud
+noblacklist ${HOME}/.config/Nextcloud
+noblacklist ${HOME}/.local/share/Nextcloud
+
+include disable-common.inc
+include disable-devel.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+
+mkdir ${HOME}/Nextcloud
+mkdir ${HOME}/.config/Nextcloud
+mkdir ${HOME}/.local/share/Nextcloud
+whitelist ${DOCUMENTS}
+whitelist ${HOME}/Nextcloud
+whitelist ${HOME}/.config/Nextcloud
+whitelist ${HOME}/.local/share/Nextcloud
+include whitelist-common.inc
+include whitelist-var-common.inc
+
+caps.drop all
+netfilter
+no3d
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+
+private-dev
+private-tmp
+
+noexec /tmp
--- a/etc/profile-m-z/nextcloud-desktop.profile
+++ b/etc/profile-m-z/nextcloud-desktop.profile
@@ -6,8 +6,14 @@ include nextcloud-desktop.local
 # Persistent global definitions
 include globals.local
 
-# Uncomment or put in your nextcloud-desktop.local to allow sync with Document directory.
+noblacklist ${HOME}/Nextcloud
+noblacklist ${HOME}/.config/Nextcloud
+noblacklist ${HOME}/.local/share/Nextcloud
+# Uncomment or put in your nextcloud-desktop.local to allow sync with more directories.
 #noblacklist ${DOCUMENTS}
+#noblacklist ${MUSIC}
+#noblacklist ${PICTURES}
+#noblacklist ${VIDEOS}
 
 include disable-common.inc
 include disable-devel.inc
@@ -21,11 +27,14 @@ include disable-xdg.inc
 mkdir ${HOME}/Nextcloud
 mkdir ${HOME}/.config/Nextcloud
 mkdir ${HOME}/.local/share/Nextcloud
-# Uncomment or put in your nextcloud-desktop.local to allow sync with Document directory.
-#whitelist ${DOCUMENTS}
 whitelist ${HOME}/Nextcloud
 whitelist ${HOME}/.config/Nextcloud
 whitelist ${HOME}/.local/share/Nextcloud
+# Uncomment or put in your nextcloud-desktop.local to allow sync with more directories.
+#whitelist ${DOCUMENTS}
+#whitelist ${MUSIC}
+#whitelist ${PICTURES}
+#whitelist ${VIDEOS}
 include whitelist-common.inc
 include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
--- a/etc/profile-m-z/nextcloud-desktop.profile
+++ b/etc/profile-m-z/nextcloud-desktop.profile
@@ -59,4 +59,7 @@ private-cache
 private-dev
 private-tmp
 
+dbus-user filter
+dbus-user.own com.nextcloud.desktopclient.nextcloud
+dbus-user.talk org.freedesktop.secrets
 dbus-system none
--- a/etc/profile-m-z/nextcloud.profile
+++ b/etc/profile-m-z/nextcloud.profile
@@ -1,0 +1,71 @@
+# Firejail profile for nextcloud
+# Description: Nextcloud desktop synchronization client
+# This file is overwritten after every install/update
+# Persistent local customizations
+include nextcloud.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/Nextcloud
+noblacklist ${HOME}/.config/Nextcloud
+noblacklist ${HOME}/.local/share/Nextcloud
+# Uncomment or put in your nextcloud.local to allow sync with more directories.
+#noblacklist ${DOCUMENTS}
+#noblacklist ${MUSIC}
+#noblacklist ${PICTURES}
+#noblacklist ${VIDEOS}
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/Nextcloud
+mkdir ${HOME}/.config/Nextcloud
+mkdir ${HOME}/.local/share/Nextcloud
+whitelist ${HOME}/Nextcloud
+whitelist ${HOME}/.config/Nextcloud
+whitelist ${HOME}/.local/share/Nextcloud
+# Uncomment or put in your nextcloud.local to allow sync with more directories.
+#whitelist ${DOCUMENTS}
+#whitelist ${MUSIC}
+#whitelist ${PICTURES}
+#whitelist ${VIDEOS}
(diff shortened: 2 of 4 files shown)


etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
...
+# This file is overwritten after every install/update
+# Persistent local customizations
+include nextcloud-desktop.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${DOCUMENTS}
+noblacklist ${HOME}/Nextcloud
+noblacklist ${HOME}/.config/Nextcloud
+noblacklist ${HOME}/.local/share/Nextcloud
```
rusty-snake, line 12: Add `blacklist`s in disable-programs.inc

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
+# Firejail profile for nextcloud-desktop
+# Description: Nextcloud desktop client
+# This file is overwritten after every install/update
+# Persistent local customizations
+include nextcloud-desktop.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${DOCUMENTS}
+noblacklist ${HOME}/Nextcloud
```
rusty-snake, line 10: I'm not sure if we want to add a blacklist for that? For what is it? If this is the place where all file are synced to, almost every program should have access to it.

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
+# Firejail profile for nextcloud-desktop
+# Description: Nextcloud desktop client
+# This file is overwritten after every install/update
+# Persistent local customizations
+include nextcloud-desktop.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${DOCUMENTS}
```
rusty-snake, line 9: Remove or add `include disable-xdg.inc`.

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
...
+
+mkdir ${HOME}/Nextcloud
+mkdir ${HOME}/.config/Nextcloud
+mkdir ${HOME}/.local/share/Nextcloud
+whitelist ${DOCUMENTS}
+whitelist ${HOME}/Nextcloud
+whitelist ${HOME}/.config/Nextcloud
+whitelist ${HOME}/.local/share/Nextcloud
+include whitelist-common.inc
+include whitelist-var-common.inc
```
rusty-snake, line 28: Can we add wruc and/or wusc?

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
...
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+
+private-dev
+private-tmp
+
+noexec /tmp
```
rusty-snake, line 48: Use `include disable-exec.inc` instead.

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
...
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
```
rusty-snake, line 43: Can we add `tracelog`?

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
...
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
```
rusty-snake, line 42: Can we add `seccomp.block-secondary`.

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
...
+shell none
+
+private-dev
+private-tmp
+
```
rusty-snake, line 47: Does it need full D-Bus access?

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
...
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+
+private-dev
```
rusty-snake, line 45: Can we add `private-cache` and other `private*` options?

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
...
+mkdir ${HOME}/.local/share/Nextcloud
+whitelist ${DOCUMENTS}
+whitelist ${HOME}/Nextcloud
+whitelist ${HOME}/.config/Nextcloud
+whitelist ${HOME}/.local/share/Nextcloud
+include whitelist-common.inc
+include whitelist-var-common.inc
+
+caps.drop all
+netfilter
```
rusty-snake, line 31: `machine-id`?

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,48 @@
...
+include disable-programs.inc
+
+mkdir ${HOME}/Nextcloud
+mkdir ${HOME}/.config/Nextcloud
+mkdir ${HOME}/.local/share/Nextcloud
+whitelist ${DOCUMENTS}
+whitelist ${HOME}/Nextcloud
+whitelist ${HOME}/.config/Nextcloud
+whitelist ${HOME}/.local/share/Nextcloud
+include whitelist-common.inc
```
rusty-snake, line 27: I'm not sure if it should be an whitelisting profile, but I also don't know what features nextcloud-desktop has.

rusty-snake (commented):
Add it to `firecfg.config` so firecfg create symlinks for it.

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: It was the same in dropbox.profile

I would eve us this full block (tested and fully working)
```
include disable-common.inc
include disable-devel.inc
include disable-exec.inc
include disable-passwdmgr.inc
include disable-programs.inc
include disable-shell.inc
```

nidamanx: It was the same in dropbox.profile

I would eve us this full block (tested and fully working)
```
include disable-common.inc
include disable-devel.inc
include disable-exec.inc
include disable-passwdmgr.inc
include disable-programs.inc
include disable-shell.inc
```

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: In QOwnNotes.profile, we have the same.
That's the default for Nextcloud.
But I really understand you point

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: Done!
That was also my choice
I skipped many params because I had a look to dropbox.profile
I think we should also have a look to it

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: I don't think so.
What do you suggest?

etc/profile-m-z/nextcloud-desktop.profile
nidamanx:
```
ls ~/.config/Nextcloud/
cookies0.db  cookies1.db  logs  nextcloud.cfg  sync-exclude.lst
```

```
ls ~/.local/share/Nextcloud/
Nextcloud_sync.log
```

etc/profile-m-z/nextcloud-desktop.profile
rusty-snake: If it works `dbus-{user,system} none`.

etc/profile-m-z/nextcloud-desktop.profile
rusty-snake: QOwnNotes has only `(no)blacklist ${HOME}/Nextcloud/Notes`. But if there are audio/video/document/... files we would need to add a noblacklist to all that viewers/editors.

etc/profile-m-z/nextcloud-desktop.profile
rusty-snake: > Client fails to login without the following 2 lines:

When you remove every `whitelist ${HOME}/...` and `include whitelist-common.inc`?

rusty-snake:
There are more templates (gui, gtk, ...) and maybe there is something like /etc/nextcloud. In addition you can run `firejail --build nextcloud-desktop | grep private-etc`.

etc/profile-m-z/nextcloud-desktop.profile
rusty-snake: > Client fails to login without the following 2 lines:

When you remove every `whitelist ${HOME}/...` and `include whitelist-common.inc`?

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: Yes, I think it could be something to think about. At least for images. Maybe only opening `whitelist ${DOWNLOADS}`.
What do you think?
BTW, I tried just now and I can load save images in QOwnNotes.
I cannot see buttons to load video or audio. So, only that whitelist should be enough

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: I was able to use only `dbus-system none`
`dbus-user filter` and `dbus-user filter` didn't allow login

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: No. No login in this way
```
#whitelist ${HOME}/Nextcloud
#whitelist ${HOME}/.config/Nextcloud
#whitelist ${HOME}/.local/share/Nextcloud
include whitelist-common.inc
```

It seems we really need those lines

nidamanx:
> firejail --build nextcloud-desktop | grep private-etc

Unfortunately, that break the AppImage nextcloud-desktop (and now, they are providing mostly only this for Linux)
I will try to search more to a better private-etc

etc/profile-m-z/nextcloud-desktop.profile
rusty-snake: The flatpak permissions look like:
```
[Session Bus Policy]
com.canonical.AppMenu.Registrar=talk
org.kde.StatusNotifierWatcher=talk
org.freedesktop.Notifications=talk
org.kde.*=own
org.freedesktop.secrets=talk
```
so you could try
```
dbus-user filter
# Maybe not necessary 
dbus-user.own com.nextcloud.desktopclient.nextcloud
# If it is required, we will add it
```

etc/profile-m-z/nextcloud-desktop.profile
rusty-snake: Because whitelisting is still enabled, you need to comment `include whitelist-common.inc` too to have a blacklisting profile.

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: Works perfectly!
Now we can even use `dbus-user filter`
`dbus-user.talk org.freedesktop.secrets` is useful in case of meeting calls.
Resolved! :-)

```
dbus-user filter
dbus-user.own com.nextcloud.desktopclient.nextcloud
dbus-system none
```

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: >you need to comment include whitelist-common.inc too 

sorry, now I did. Unfortunately same result. No login or no sync.

---
`whitelist ${HOME}/Nextcloud` is the defaults sync folder (user can change and use what he wants, of course)
`whitelist ${DOCUMENTS}` I would let as a commented option. Can be enabled in case someone wants to sync another directory in Documents
`whitelist ${HOME}/.config/Nextcloud` and `whitelist ${HOME}/.local/share/Nextcloud` are used for Nextcloud login and settings

I really don't know how to go around those whitelists

Maybe we could propose a more safe location for the Nextcloud sync directory, but what to do with config and local? We really need those

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: > I'm not sure if we want to add a blacklist for that? For what is it? If this is the place where all file are synced to, almost every program should have access to it.

now we don't need those lines anymore

```
#noblacklist ${HOME}/Nextcloud
#noblacklist ${HOME}/.config/Nextcloud
#noblacklist ${HOME}/.local/share/Nextcloud
```

I'm going to delete them from the next commit

etc/profile-m-z/nextcloud-desktop.profile
nidamanx: We can skip the mkdir but (still) not whitelist. Tell me if you think could be better.

```
#mkdir ${HOME}/Nextcloud
#mkdir ${HOME}/.config/Nextcloud
#mkdir ${HOME}/.local/share/Nextcloud
# Uncomment or put in your nextcloud-desktop.local to allow sync with Document directory.
#whitelist ${DOCUMENTS}
whitelist ${HOME}/Nextcloud
whitelist ${HOME}/.config/Nextcloud
whitelist ${HOME}/.local/share/Nextcloud
```

etc/profile-m-z/nextcloud-desktop.profile
```diff
@@ -0,0 +1,65 @@
...
+tracelog
+
+disable-mnt
+private-bin nextcloud-desktop
+private-cache
+private-dev
+private-tmp
+
+dbus-user filter
+dbus-user.own com.nextcloud.desktopclient.nextcloud
```
nidamanx, line 63: This we don't need. Belong only to client for flatpak

etc/profile-m-z/nextcloud-desktop.profile
rusty-snake: We need `noblacklist ${HOME}/.config/Nextcloud` and `noblacklist ${HOME}/.local/share/Nextcloud` after adding `blacklist`s for them in disable-programs.inc.

etc/profile-m-z/nextcloud-desktop.profile
rusty-snake: Are you sure that there is no "hidden" `whitelist`? (check with `firejail --ignore=private-bin --profile=nextcloud-desktop ls -lA $HOME`).

> We can skip the mkdir

Nope, because you can only `whitelist` file which exist, this will break if `${HOME}/.config/Nextcloud`/`${HOME}/.local/share/Nextcloud` are not present (e.g. first start).

nidamanx:
```
$ firejail --ignore=private-bin --profile=nextcloud-desktop ls -1A $HOMEReading profile /home/user/.config/firejail/nextcloud-desktop.profile
Reading profile /etc/firejail/disable-shell.inc
Reading profile /etc/firejail/whitelist-runuser-common.inc
Warning: networking feature is disabled in Firejail configuration file
Parent pid 28736, child pid 28739
Blacklist violations are logged to syslog
Child process initialized in 231.94 ms
.bashrc
.cache
.config
.local
Nextcloud
.themes

```

rusty-snake: > Reading profile /etc/firejail/whitelist-common.inc

Can you try with every whitelist commented or maybe even with `--ignore='whitelist ${HOME}'`.

nidamanx: Found!
Is this way good enough?

```
# Uncomment or put in your nextcloud-desktop.local to allow sync with Document directory.
#noblacklist ${DOCUMENTS}
noblacklist ${HOME}/Nextcloud
noblacklist ${HOME}/.config/Nextcloud
noblacklist ${HOME}/.local/share/Nextcloud

include disable-common.inc
include disable-devel.inc
include disable-exec.inc
include disable-passwdmgr.inc
include disable-programs.inc
include disable-shell.inc
include disable-xdg.inc

mkdir ${HOME}/Nextcloud
mkdir ${HOME}/.config/Nextcloud
mkdir ${HOME}/.local/share/Nextcloud
# Uncomment or put in your nextcloud-desktop.local to allow sync with Document directory.
#whitelist ${DOCUMENTS}
#whitelist ${HOME}/Nextcloud
#whitelist ${HOME}/.config/Nextcloud
#whitelist ${HOME}/.local/share/Nextcloud
#include whitelist-common.inc
#include whitelist-var-common.inc
```

rusty-snake: Maybe we could go with
```
# Uncomment (or put in your nextcloud-desktop.local) if you want to sync from these directories.
#blacklist ${DOCUMENTS}
#blacklist ${MUSIC}
#blacklist ${PICTURES}
#blacklist ${VIDEOS}
noblacklist ${HOME}/Nextcloud
noblacklist ${HOME}/.config/Nextcloud
noblacklist ${HOME}/.local/share/Nextcloud

include disable-common.inc
include disable-devel.inc
include disable-exec.inc
include disable-passwdmgr.inc
include disable-programs.inc
include disable-shell.inc
include disable-xdg.inc

# Uncomment the following (or put in your nextcloud-desktop.local) to enable extra protection.
#mkdir ${HOME}/Nextcloud
#mkdir ${HOME}/.config/Nextcloud
#mkdir ${HOME}/.local/share/Nextcloud
#whitelist ${HOME}/Nextcloud
#whitelist ${HOME}/.config/Nextcloud
#whitelist ${HOME}/.local/share/Nextcloud
#include whitelist-common.inc
```

nidamanx: **It works! **

We were on the same way in the same time :-)
I have a big delay to receive notifications
You added more nice stuff

A question: 

>Uncomment the following (or put in your nextcloud-desktop.local) to enable extra protection.

I thought it's better to have less whitelists as possible. That's the reason we were working on. Did I understood wrong?

nidamanx: > #Uncomment (or put in your nextcloud-desktop.local) if you want to sync from these directories.
> #blacklist ${DOCUMENTS}
> #blacklist ${MUSIC}
> #blacklist ${PICTURES}
> #blacklist ${VIDEOS}

Shouldn't be all 
`#noblacklist ...`

nidamanx: So, now, only this part is not really clear to me.

To enable extra protection we can enable those lines? I thought we were working to kick away the whitelists
And mkdirs are important if a dir is not yet there.
Maybe I really understood something different
```
# Uncomment the following (or put in your nextcloud-desktop.local) to enable extra protection.
#mkdir ${HOME}/Nextcloud
#mkdir ${HOME}/.config/Nextcloud
#mkdir ${HOME}/.local/share/Nextcloud
#whitelist ${HOME}/Nextcloud
#whitelist ${HOME}/.config/Nextcloud
#whitelist ${HOME}/.local/share/Nextcloud
#include whitelist-common.inc
```

rusty-snake: > I thought it's better to have less whitelists as possible. That's the reason we were working on. Did I understood wrong?

less whitelist = less permissions but no whitelist = full permissions (minus blacklisted permissions). Whitelisting is much much more safer the blacklisting. On the other hand you can only sync into/from whitelisted directories. I'm still not sure if we should even comment the disable-xdg or enable whitelisting. What's the usuall workflow? How likely is that it disturbs? If anyone else can say something too.

nidamanx: I did more tests about those 2 choices
1) If I use noblacklist, mkdir and whitelist, I cannot choose what to sync from the server and all is always downloaded 
2) if I use only noblacklist (like it is now), all /home is visible and I can sync all directories I want

I think, the first we cannot use, sync is really a pain.
The second, we need a kind of restriction
This is the second
```
# Uncomment or put in your nextcloud-desktop.local to allow sync with Document directory.
#noblacklist ${DOCUMENTS}
#noblacklist ${MUSIC}
#noblacklist ${PICTURES}
#noblacklist ${VIDEOS}
noblacklist ${HOME}/Nextcloud
noblacklist ${HOME}/.config/Nextcloud
noblacklist ${HOME}/.local/share/Nextcloud

include disable-common.inc
include disable-devel.inc
include disable-exec.inc
include disable-passwdmgr.inc
include disable-programs.inc
include disable-shell.inc
include disable-xdg.inc

# Uncomment the following (or put in your nextcloud-desktop.local) to enable extra protection.
#mkdir ${HOME}/Nextcloud
#mkdir ${HOME}/.config/Nextcloud
#mkdir ${HOME}/.local/share/Nextcloud
#whitelist ${HOME}/Nextcloud
#whitelist ${HOME}/.config/Nextcloud
#whitelist ${HOME}/.local/share/Nextcloud
#include whitelist-common.inc
```

rusty-snake:
Looks good to far. What remains:
- [ ] firecfg.config
- [ ] disable-programs.inc
- [ ] sort.py

nidamanx:
> Looks good to far. What remains:
> 
>     * [x]  firecfg.config
> 
>     * [x]  disable-programs.inc
> 
>     * [x]  sort.py

All the three steps are done.
I committed the updated file.
Thanks for all

nidamanx:
@rusty-snake 

Two topics:

### 1) Finally I found working private-etc

```
# Appimage: 
private-etc alternatives,ca-certificates,crypto-policies,fonts,host.conf,hosts,ld.so.cache,machine-id,nsswitch.conf,pki,pulse,resolv.conf,ssl
# Debian package: 
private-etc drirc,fonts,gcrypt,hosts,machine-id,Nextcloud,os-release,passwd,selinux,ssl,xdg
```

if you agree, I would mix both:
`private-etc alternatives,ca-certificates,crypto-policies,drirc,fonts,gcrypt,host.conf,hosts,ld.so.cache,machine-id,Nextcloud,nsswitch.conf,os-release,passwd,pki,pulse,resolv.conf,selinux,ssl,xdg`

### 2) We need to rename the profile
In Debian repo, the package is named `nextcloud-desktop` (and after a closer look, I found the name of the installed bin: `nextcloud`)
More than this, looking to [Repology](https://repology.org/projects/?search=nextcloud), seems way better to use the name `nextcloud`
So, if you agree, I would rename the file and re-edit  [fireconfig.config](src/firecfg/firecfg.config)

glitsj16:
> 2) We need to rename the profile
In Debian repo, the package is named nextcloud-desktop (and after a closer look, I found the name of the installed bin: nextcloud)
More than this, looking to Repology, seems way better to use the name nextcloud
So, if you agree, I would rename the file and re-edit fireconfig.config

Some distro's may still use the existing name. IMO we should keep it and add a alias (redirect) profile for the newly named binary.

nidamanx:
@rusty-snake 

Two topics:

### 1) Finally I found working private-etc

```
# Appimage: 
private-etc alternatives,ca-certificates,crypto-policies,fonts,host.conf,hosts,ld.so.cache,machine-id,nsswitch.conf,pki,pulse,resolv.conf,ssl
# Debian package: 
private-etc drirc,fonts,gcrypt,hosts,machine-id,Nextcloud,os-release,passwd,selinux,ssl,xdg
```

if you agree, I would mix both:
`private-etc alternatives,ca-certificates,crypto-policies,drirc,fonts,gcrypt,host.conf,hosts,ld.so.cache,machine-id,Nextcloud,nsswitch.conf,os-release,passwd,pki,pulse,resolv.conf,selinux,ssl,xdg`

### 2) We need to rename the profile
In Debian repo, the package is named `nextcloud-desktop` (and after a closer look, I found the name of the installed bin: `nextcloud`)
More than this, looking to [Repology](https://repology.org/projects/?search=nextcloud), seems way better to use the name `nextcloud`
So, if you agree, I would rename the file and re-edit  [fireconfig.config](src/firecfg/firecfg.config)

> > 1. We need to rename the profile
> >    In Debian repo, the package is named nextcloud-desktop (and after a closer look, I found the name of the installed bin: nextcloud)
> >    More than this, looking to Repology, seems way better to use the name nextcloud
> >    So, if you agree, I would rename the file and re-edit fireconfig.config
> 
> Some distro's may still use the existing name. IMO we should keep it and add a alias (redirect) profile for the newly named binary.

What about to create an alias for nextcloud-deskop, instead?
We could provide with the alias the backport and keep the "official" name for the main profile. So, the majority will load directly the nextcloud.profile without aliasing.

You're the expert in this. It's only an idea that came just now.
Let me know and let's do as you think is better.

glitsj16:
> What about to create an alias for nextcloud-deskop, instead?

I could live with that. But maybe we should wait for @rusty-snake to chime in on this topic.

rusty-snake:
Sound all reasonable.

nidamanx:
Ok, so, @glitsj16  and @rusty-snake 
I would like to avoid to override some procedures that I'm not aware of.
If I have confirm from both of you, I'll go on

glitsj16:
> If I will have confirm from both of you, I'll go on

Here's my confirmation. It looks good, so proceed with the redirect alias and we'll have another fine application to add to the collection. Thanks for your work.

nidamanx:
> Thanks for your work.

Thanks for YOUR work (to both of you)! Mine is just like peanuts, in comparison.

Committed all.
Let me know is something is missing or needs editing

etc/profile-m-z/nextcloud.profile
```diff
@@ -0,0 +1,71 @@
...
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+
+disable-mnt
+private-bin nextcloud
```
rusty-snake, line 63:
```suggestion
private-bin nextcloud,nextcloud-desktop
```

rusty-snake (commented):
LGTM

rusty-snake (approved):

netblue30:
merged, thanks!

Issue #3960, opened 2021-02-08
Patches from Jolla


SkewedZeppelin:
The Jolla company seems to have used Firejail in their recent OS update.
That is pretty neat.
https://forum.sailfishos.org/t/release-notes-koli-4-0-1/4542

They have some patches here which maybe should be considered for upstream inclusion:
https://github.com/sailfishos/firejail/blob/master/rpm/0002-Fix-symlinks-that-go-though-proc-self.patch
https://github.com/sailfishos/firejail/blob/master/rpm/0007-fcopy-Fix-memory-leaks.patch
https://github.com/sailfishos/firejail/blob/master/rpm/0009-sandbox-Do-not-leave-file-mounts-underneath-private-.patch

netblue30:
I think I'll grab all of them - thanks!

Tomin1:
Hi! I'm glad that there is interest in our enhancements to firejail. We would like to prepare pull requests for our patches.

netblue30:
Thnaks, that's even better. Send the patches!

Tomin1:
@netblue30 Do you think you could handle these three patches as one PR or should I create separate PRs? Additionally I think 
`0005-Add-missing-linefeeds-in-stderr-logging.patch` would be also good fit to the same PR if you agree that it's a useful fix.

I will also make PRs for some of the other patches if they have any chance of being accepted. I'm thinking I would first do those three or four already mentioned and then separately:
`0003-Add-utility-functions-for-handing-comma-separa.patch`, and
`0004-Allow-changing-protocol-list-after-initial-set.patch` together since they are part of the same improvement.

`0006-PATCH-Add-mkdir-and-mkfile-command-line-options-for-.patch` as yet another PR.

That would leave two patches that I guess are not that useful for most firejail users since they are very Sailfish OS specific.

netblue30:
> Do you think you could handle these three patches as one PR or should I create separate PRs?

Your choice, it doesn't really mater. And send patches for all the rest! Thanks a lot!

Tomin1:
> That would leave two patches that I guess are not that useful for most firejail users since they are very Sailfish OS specific.

Just to be clear, I don't intend to upstream these two. Everything else is already there.

rusty-snake:
Can we close here?

PR #3998, opened 2021-02-19
Upstreaming a set of fixes from Sailfish's packaging

Tomin1:
This is upstreaming first set of fixes from [Sailfish's firejail packaging](https://github.com/sailfishos/firejail/). These were developed as part of implementing firejail sandboxing in Sailfish OS. See also #3960 for discussion.

These are general fixes to firejail touching logging, private-etc, memory handing and symbolic link construction around /proc filesystem. Please see the commit messages for explanations. These were written by my colleague and previously reviewed by me or one of my other colleagues.

### DIFF
--- a/src/fcopy/main.c
+++ b/src/fcopy/main.c
@@ -172,6 +172,47 @@ static void mkdir_attr(const char *fname, mode_t mode, uid_t uid, gid_t gid) {
 	}
 }
 
+static char *proc_pid_to_self(const char *target)
+{
+	char *use_target = 0;
+	char *proc_pid = 0;
+
+	if (!(use_target = canonicalize_file_name(target)))
+		goto done;
+
+	// target is under /proc/<PID>?
+	static const char proc[] = "/proc/";
+	if (strncmp(use_target, proc, sizeof proc - 1))
+		goto done;
+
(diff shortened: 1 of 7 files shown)


reinerh:
One of these changes causes a regression in the test suite:
https://github.com/netblue30/firejail/pull/3998/checks?check_run_id=1933635335
```
Private /etc installed in 22.23 ms
Error mount bind: fs_etc.c:194 fs_private_dir_mount: No such file or directory
Error: proc 4439 cannot sync with peer: unexpected EOF
Peer 4441 unexpectedly exited with status 1
runner@fv-az102-314:~/work/firejail/firejail/test/profiles$ TESTING ERROR 0
```
(the other errors are mostly the same)

It might be related to this line:
```
fs_private_dir_mount("/usr/etc", RUN_USR_ETC_DIR);
```
because `/usr/etc` doesnot exist in the test environment.

Tomin1:
Nice catch from the test suite! I added the same check that fs_private_dir_copy() has to fs_private_dir_mount() which results in some duplicated work but I think one more stat() is not that bad of a tradeoff. Also added those asserts just in case.

netblue30:
all set, thanks!

PR #4001, opened 2021-02-21
signal-desktop.profile: fix typo of disable-xdg.profile

kmk3:
Added on commit f4f676745 ("Refactor electron.profile and electron based
programs (#3807)").

This appears to be the only instance of that:

    $ grep -Fnr 'include-xdg' etc
    etc/profile-m-z/signal-desktop.profile:9:ignore include-xdg.inc

Credits go to syntax highlighting on vim.

---

Also, note that the file says this:

```firejail
ignore include-xdg.inc
```

So if there was no negative feedback, should xdg dirs (other than
`${DOWNLOADS}`, that is) be allowed or should we just remove that line instead?
As all this PR would do is allow xdg dirs for the first time since at least
commit f4f676745.

### DIFF
--- a/etc/profile-m-z/signal-desktop.profile
+++ b/etc/profile-m-z/signal-desktop.profile
@@ -6,7 +6,6 @@ include signal-desktop.local
 include globals.local
 
 # Disabled until someone reported positive feedback
-ignore include-xdg.inc
 ignore include whitelist-runuser-common.inc
 ignore include whitelist-usr-share-common.inc
 ignore private-cache


rusty-snake (commented):
Since signal-dekstop is a whitelisting profile, it's safe to `include disable-xdg.inc`.

kmk3:
> Since signal-dekstop is a whitelisting profile, it's safe to `include
> disable-xdg.inc`.

Alright; removed the ignore and force-pushed.

rusty-snake (approved):

netblue30:
all set, thanks

PR #4003, opened 2021-02-23
Minor fixes for vmware

Neo00001:
These two profiles will allow `firecfg` to take care of vmware related `.desktop` files. Otherwise,manual intervention would be needed.

### DIFF
--- a/etc/profile-m-z/vmware-player.profile
+++ b/etc/profile-m-z/vmware-player.profile
@@ -1,0 +1,8 @@
+# Firejail profile for vmware-player
+# Description: The industry standard for running multiple operating systems as virtual machines on a single Linux PC.
+# This file is overwritten after every install/update
+# Persistent local customizations
+include vmware-player.local
+
+# Redirect
+include vmware.profile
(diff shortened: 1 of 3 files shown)


rusty-snake (commented):
> These two profiles will allow firecfg to take care of vmware related .desktop files. 

firecfg can't take care if there is no entry firecfg.config. Rest looks fine.

Neo00001:
> firecfg can't take care if there is no entry firecfg.config.

My understanding is if there is a profile & there exist same entry in `/usr/share/applications` as `.desktop` file then `firecfg` should create an entry in `${Home}/.local/share/application`. Anyway,I am adding those two in `firecfg.config`.

rusty-snake:
> My understanding is if there is a profile **( in `~/.config/firejail` ) or ( a profile in `/etc/firejail` and an entry in `firecfg.config` )** & there exist same entry in `/usr/share/applications` as `.desktop` file then `firecfg` should create an entry in `${Home}/.local/share/application` **if the file in `/usr/share/applications` has an absolute `Exec=` or `DBusActivatable=true`**.

So there is no dot-desktop fixup as long as the dot-desktop file has a relative `Exec` and is not `DBusActivatable` (firectl and firecfg.py behave differently). And the presents of a profile triggers firecfg only for user-profiles, if we would create symlinks for every profile in /etc/firejail, we would need to remove profile for e.g. gzip, tar, epiphany, evolution, ffmpeg, ... .

rusty-snake (approved):

netblue30:
all set, thanks!

netblue30:
all set, thanks!

PR #4004, opened 2021-02-23
add PATH_FCOPY to private-lib automatically

### LINKED ISSUE #3741: Error: failed to run /run/firejail/lib/fcopy

smitsohu:
Essentially restores 45304621a6c600d8e30e98bfbef05149caaf56c5. Removing read permission on helper executables (fcopy) broke this original fix.

Now run fldd as root in order to fix #3741 without having to compromise or give up on the new permission system. It runs as root only on binaries that are not controlled by the user, and that Firejail needs to trust anyway. Also infrastructure is put in place to add more helper binaries to private-lib, should the need arise.

This pull request might be useful also for other reasons. For example, if one day we want to run the test suite with ASan/UBsan instrumentation, private-lib should now pick up all necessary libraries automatically (again).

### DIFF
--- a/src/firejail/fs_lib.c
+++ b/src/firejail/fs_lib.c
@@ -28,6 +28,7 @@
 #define MAXBUF 4096
 
 extern void fslib_install_stdc(void);
+extern void fslib_install_firejail(void);
 extern void fslib_install_system(void);
 
 static int lib_cnt = 0;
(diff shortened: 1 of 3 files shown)


netblue30:
all set, thanks!

PR #4007, opened 2021-02-25
 fix firecfg links in restrictive sandboxes

### LINKED ISSUE #3911: firecfg breaking firejail on the command line
**Bug and expected behavior**
Running `firejail application` on a command line fails in certain circumstances. The problem occurs if a symbolic link to firejail was created in /usr/local/bin, for example by means of `firecfg`, and the sandbox is too restrictive to allow execution of a shell.

**Reproduce**
Steps to reproduce the behavior. Options usually come from a profile but are passed on the command line here; gedit is an arbitrary example and could be anything:
1. `sudo ln -s /usr/bin/firejail /usr/local/bin/gedit`
2. `firejail --noprofile --private-lib --shell=none  gedit` or
3. `firejail --profile=/etc/firejail/disable-shell.inc --shell=none gedit`
4. note that `private-bin` rescues this, because `firecfg` links in /usr/local/bin are masked

Command 2 gives me
```
Child process initialized in 780.41 ms
Warning: an existing sandbox was detected. /usr/bin/gedit will run without any additional sandboxing features
/bin/bash: error while loading shared libraries: libtinfo.so.6: cannot open shared object file: No such file or directory
```

Command 3 gives me
```
Child process initialized in 23.67 ms
Warning: an existing sandbox was detected. /usr/bin/gedit will run without any additional sandboxing features
```

**Additional context / proposed fix**
If Firejail is run inside a Firejail sandbox (run_no_sandbox), everything it does requires a shell. Would it be acceptable to remove this dependency and take the `--shell=none` code path instead?

Alternatively it should also work if the sandbox process, with `--shell=none` configured, searches the executable in $PATH, skips all these links created by `firecfg` and then runs the program from an absolute path (rather than delegating the work to `execvp`)

smitsohu:
Firejail symbolic link redirection currently depends
on a shell, but in restrictive sandboxes there
might be no execute permission, or `private-lib` might
have removed necessary libraries, or `seccomp` might
block required syscalls ...

Fix this by forcing `--shell=none`.

Closes #3911

### DIFF
--- a/src/firejail/no_sandbox.c
+++ b/src/firejail/no_sandbox.c
@@ -167,15 +167,6 @@ void run_no_sandbox(int argc, char **argv) {
 	if (setresuid(-1, getuid(), getuid()) != 0)
 		errExit("setresuid");
 
-	// use $SHELL to get shell used in sandbox, guess shell otherwise
-	cfg.shell = guess_shell();
-	if (!cfg.shell) {
-		fprintf(stderr, "Error: unable to guess your shell, please set SHELL environment variable\n");
-		exit(1);
-	}
-	else if (arg_debug)
-		printf("Selecting %s as shell\n", cfg.shell);
-
 	// process limited subset of options
 	// and find first non option arg:
 	//	- first argument not starting with --,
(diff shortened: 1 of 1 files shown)


glitsj16:
Great stuff! This should bring back some sanity to firecfg and (hopefully) diminish the number of user reported issues with it.

netblue30:
merged, thanks!

PR #4008, opened 2021-02-25
add new profile for gget

glitsj16:
See https://gget.io/ for details of the application.

### DIFF
--- a/etc/profile-a-l/gget.profile
+++ b/etc/profile-a-l/gget.profile
@@ -1,0 +1,59 @@
+# Firejail profile for gget
+# Description:  a cli. to get things. from git repos
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include gget.local
+# Persistent global definitions
+include globals.local
+
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/gget.profile
```diff
@@ -0,0 +1,59 @@
+# Firejail profile for gget
+# Description:  a cli. to get things. from git repos
```
rusty-snake, line 2:
```suggestion
# Description: a cli. to get things. from git repos
```

rusty-snake (commented):
looks good

etc/profile-a-l/gget.profile
glitsj16: Sorry, I just pushed the button when I saw this. Fixed in https://github.com/netblue30/firejail/commit/8a7b96974fd8ddae8ea4641d8e010b1e03b9e817. Thanks!

rusty-snake:
FTR: The missing firecfg.config is intention I guessed.

glitsj16:
> FTR: The missing firecfg.config is intention I guessed.

Indeed. I just stumbled onto this little app when looking for an easy/fast way to get single files from repositories. It isn't even in the AUR, nor packaged by any of the major distros we support. So I decided not to add it to firecfg. Doesn't come with a .desktop file neither. I'll try it out for a while and perhaps add it to the AUR myself. When that happens (or someone else does) I'll revisit this to add firecfg support.

PR #4009, opened 2021-02-25
Allow changing protocol list after initial set

Tomin1:
Firejail uses set-once logic for protocol list. This makes it impossible to accumulate list of allowed protocols from multiple include files. Instead of mandating that protocol is set once, allow constructing the list from multiple protocol rules.

These patches were part of [Sailfish's firejail packaging](https://github.com/sailfishos/firejail/). These were developed as part of implementing firejail sandboxing in Sailfish OS and just like the previous patches these were written by my colleague and previously reviewed by me or one of my other colleagues. See also #3960 for discussion.

### DIFF
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -451,6 +451,9 @@ int profile_check_line(char *ptr, int lineno, const char *fname);
 // add a profile entry in cfg.profile list; use str to populate the list
 void profile_add(char *str);
 void profile_add_ignore(const char *str);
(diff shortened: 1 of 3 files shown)


netblue30:
Merged, thanks!

glitsj16:
@Tomin1 Do we need bash_completion support for any of this?

rusty-snake:
There is/was no completion of the value of `--protocol`.

glitsj16:
@rusty-snake Good, that's settled thanks

PR #4015, opened 2021-02-26
dbus.c: check_object_path: Allow /StatusNotifierWatcher

rusty-snake:
Without this I get 

```
Error: Invalid dbus-user.call rule: org.kde.StatusNotifierWatcher=org.kde.StatusNotifierWatcher.*@/StatusNotifierWatcher
```
for

```
dbus-user.call org.kde.StatusNotifierWatcher=org.kde.StatusNotifierWatcher.*@/StatusNotifierWatcher
```

### DIFF
--- a/src/firejail/dbus.c
+++ b/src/firejail/dbus.c
@@ -111,7 +111,7 @@ static int check_object_path(const char *path) {
 		}
 		++p;
 	}
-	return in_segment && segments >= 2;
+	return in_segment && segments >= 1;
 }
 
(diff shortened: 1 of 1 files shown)


kris7t (approved):
Yeah, I guess even 1-segment objects paths are allowed in D-Bus.

In fact, the specification also allows the root path `/` (https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-marshaling-object-path), which is a bit odd (and not recommended at all).

PR #4016, opened 2021-02-27
Create rtv-addons.inc

pirate486743186:
rtv can be configured to launch links in external applications.
https://github.com/michael-lazar/rtv#viewing-media-links

This include is meant to be used in a .local file.
This is a proposed first draft.

didn't work: firefox
worked: w3m, eom, mpv

for w3m, you need to launch rtv with a variable like this
export RTV_BROWSER="w3m" ; /usr/bin/firejail /usr/bin/rtv

i tried to use firefox-common-addons.inc naming convention


etc/profile-m-z/rtv-addons.inc
```diff
@@ -0,0 +1,23 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+
+# You can configure rtv to open different type of links
+# in external applications. Configuration here:
+# https://github.com/michael-lazar/rtv#viewing-media-links
+# This include is meant to facilitate that configuration
+# with the use of a .local file.
+
+ignore nosound
```
glitsj16, line 10: machine-id is also known to break audio, please double-check if that needs to be ignored too

etc/profile-m-z/rtv-addons.inc
```diff
@@ -0,0 +1,23 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+
+# You can configure rtv to open different type of links
```
glitsj16, line 4: perhaps add a note too rtv.profile so users are made aware of this extra configuration

etc/profile-m-z/rtv-addons.inc
```diff
@@ -0,0 +1,23 @@
...
+# in external applications. Configuration here:
+# https://github.com/michael-lazar/rtv#viewing-media-links
+# This include is meant to facilitate that configuration
+# with the use of a .local file.
+
+ignore nosound
+ignore private-bin
+ignore dbus-system none
```
glitsj16, line 12: is granting access to the system bus really needed?

etc/profile-m-z/rtv-addons.inc
```diff
@@ -0,0 +1,23 @@
...
+ignore dbus-user none
```
glitsj16, line 13: can we use any specific dbus.user filter rules?

glitsj16 (commented):
Some minor nitpicks...

etc/profile-m-z/rtv-addons.inc
pirate486743186: lol it's not

etc/profile-m-z/rtv-addons.inc
pirate486743186: mpv seams to work, i'm on 18.04 though...
The sound is for mpv with youtube-dl, with other players, you'll need a direct link, that's rarer.

etc/profile-m-z/rtv-addons.inc
pirate486743186: i'm no expert, this is mostly my use case. eom brakes with "dbus-user none"

etc/profile-m-z/rtv-addons.inc
glitsj16: Great, let's go without touching machine-id in that case.

etc/profile-m-z/rtv-addons.inc
glitsj16: Suggestion: add something like the below to rtv.profile:
```
$ cat /etc/firejail/rtv.profile
...
include allow-python2.inc
include allow-python3.inc

# Uncomment or put in rtv.local for external application support
#include rtv-addons.inc

include disable-common.inc
```

glitsj16 (approved):

etc/profile-m-z/rtv-addons.inc
```diff
@@ -0,0 +1,22 @@
+# This file is overwritten during software install.
```
rusty-snake, line 1: `*.inc` files should be placed in `etc/inc`.

rusty-snake: https://github.com/netblue30/firejail/blob/9b56dc8e94cf2313baf19e5847a0631231497c93/Makefile.in#L135

etc/profile-m-z/rtv-addons.inc
pirate486743186: yea, that didn't work for me either

etc/profile-m-z/rtv-addons.inc
pirate486743186: ok

etc/profile-m-z/rtv-addons.inc
pirate486743186: ok

etc/inc/rtv-addons.inc
```diff
@@ -11,11 +11,13 @@ ignore nosound
 ignore private-bin
 ignore dbus-user none
 
+noblacklist ${HOME}/.cache/youtube-dl/youtube-sigfuncs/
 noblacklist ${HOME}/.config/mpv
 noblacklist ${HOME}/.mailcap
 noblacklist ${HOME}/.netrc
 noblacklist ${HOME}/.w3m
 
+whitelist ${HOME}/.cache/youtube-dl/youtube-sigfuncs/
```
rusty-snake:
```suggestion
whitelist ${HOME}/.cache/youtube-dl/youtube-sigfuncs
```

etc/inc/rtv-addons.inc
```diff
@@ -11,11 +11,13 @@ ignore nosound
 ignore private-bin
 ignore dbus-user none
 
+noblacklist ${HOME}/.cache/youtube-dl/youtube-sigfuncs/
```
rusty-snake: We have no such blacklist.

etc/inc/rtv-addons.inc
```diff
@@ -0,0 +1,23 @@
+# This file is overwritten during software install.
```
reinerh, line 1: In my opinion this should not be an `.inc` file. As `.inc` files tend to be more generic and useful for many more profiles.
Why not rename it `rtv-addons.profile`?
@rusty-snake @glitsj16 What do you think about that?

pirate486743186: i was copying the firefox include precedent

reinerh: Yes, there are a few exceptions currently (chromium, firefox, feh-network), which I don't really like (inc files also land in the main firejail package on Debian, which I try to keep minimal).
But maybe @rusty-snake or @glitsj16 have a different opinion on that.

pirate486743186: maybe the main package of firejail should take the generic ones, and the rest come with the profiles? In general the organization of the profiles is one giant mess...

glitsj16: > Why not rename it rtv-addons.profile?

@reinerh I agree that would be a much better name. The same goes for the other exceptions mentioned above. Ever since I started to use firejail I added a few mime-types with distinctive icons to visually distinguish between .inc, .local, .net and .profile files. The chromium, firefox & feh-network .inc files indeed stick out as sore thumbs.

etc/profile-m-z/rtv.profile
```diff
@@ -1,6 +1,15 @@
 # Firejail profile for rtv
 # Description: Browse Reddit from your terminal
 # This file is overwritten after every install/update
+
+# You can configure rtv to open different type of links
+# in external applications. Configuration here:
+# https://github.com/michael-lazar/rtv#viewing-media-links
+# This include is meant to facilitate that configuration
```
glitsj16, line 8: Due to the comment added below this sentence can be dropped.

etc/inc/rtv-addons.inc
glitsj16: @pirate486743186 If you need help renaming the file, there's [documentation](https://github.com/blog/1436-moving-and-renaming-files-on-github) on how to do that.

etc/profile-m-z/rtv.profile
```diff
@@ -16,6 +25,8 @@ noblacklist ${HOME}/.local/share/rtv
 include allow-python2.inc
 include allow-python3.inc
 
+# Uncomment or put in rtv.local for external application support
+#include rtv-addons.inc
```
glitsj16, line 29: #include rtv-addons.profile

glitsj16 (changes requested):
Please rename to /etc/profile-m-z/rtv-addons.profile. We can merge it when that's done. Thanks!

etc/profile-m-z/rtv.profile
pirate486743186: i moved the rest of the comment bellow too. It looks out of place otherwise.

etc/profile-m-z/rtv.profile
```diff
@@ -1,6 +1,7 @@
 # Firejail profile for rtv
 # Description: Browse Reddit from your terminal
 # This file is overwritten after every install/update
+
```
rusty-snake, line 4:
```suggestion
```

etc/profile-m-z/rtv-addons.profile
```diff
@@ -0,0 +1,23 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
```
rusty-snake, line 2:
```suggestion
# Persistent customizations should go in a .local file.
include rtv-addons.local
```

glitsj16 (approved):

rusty-snake (approved):

PR #4019, opened 2021-03-01
fix protocol list

glitsj16:


etc/profile-m-z/transmission-daemon.profile
```diff
@@ -14,6 +14,7 @@ whitelist ${HOME}/.config/transmission-daemon
 whitelist /var/lib/transmission
 
 caps.keep ipc_lock,net_bind_service,setgid,setuid,sys_chroot
+ignore protocol unix,inet,inet6
 protocol unix,inet,inet6,packet
```
rusty-snake, line 18: The lists accumulate, this `ignore` is useless because the later `protocol` just adds no protocols to the list because all are already set. We can even simplify this to `protocol packet`. A `ignore protocol` is only needed if you want to remove protocols which would be added later.

glitsj16: Leftover from my local override in which I drop inet6... Fixing it, thanks!

PR #4022, opened 2021-03-01
fix audit 'syscalls check' output

glitsj16:
Minor improvement to the --audit output.


glitsj16:
Obsolete now that https://github.com/netblue30/firejail/commit/5c95f0f9578593b5549a1c8b37693e9f419ef880 retired --audit.

PR #4025, opened 2021-03-01
add support for "play with..."

pirate486743186:
adding support for "play with..." extension. It uses a clever hack, so it doesn't need an external application in order to launch mpv.

Also trimmed ff2mpv comment block.
and added .cache/youtube-dl


etc/inc/firefox-common-addons.inc
```diff
@@ -75,16 +83,7 @@ include allow-python3.inc
...
 #include allow-lua.inc
-#include allow-python3.inc
-#mkdir ${HOME}/.config/mpv
-#mkdir ${HOME}/.config/youtube-dl
-#whitelist ${HOME}/.config/mpv
-#whitelist ${HOME}/.config/youtube-dl
-#whitelist ${HOME}/.netrc
 #whitelist /usr/share/lua
 #whitelist /usr/share/lua*
 #whitelist /usr/share/vulkan
```
rusty-snake, line 89: mpv does not work properly without this wusc additions (depending on configuration and system) IIRC. IHMO we should enable it too.

pirate486743186: you mean the /usr lines?

rusty-snake (approved):
Note on `whitelist ${HOME}/.cache/youtube-dl/youtube-sigfuncs`:
It's useless (which does not mean that it needs to be removed) without
```
nobalcklist ${HOME}/.cache/youtube-dl
ignore private-cache
```

rusty-snake:
This PR has a merge conflict, please rebase.

etc/profile-a-l/firefox-common-addons.profile
```diff
@@ -3,11 +3,15 @@
 include firefox-common-addons.local
 
 ignore include whitelist-runuser-common.inc
+ignore private-cache
 
+nobalcklist ${HOME}/.cache/youtube-dl
```
glitsj16, line 8: typo: nobalcklist --> noblacklist

PR #4035, opened 2021-03-03
Improve error messages

Tomin1:
I sent some changes recently (#3998) and now I noticed that there were some minor things that could have been better.

The first commit moves error message of missing directory while mounting after debug logging so that it's easy to see when the error happens and adds a cause message. The second commit changes %m usages introduced by the changes I made. While %m is reentrant and allows very nice and clean code, it's not consistent with the rest of firejail and also not portable across libc implementations.


reinerh:
Thanks!

Issue #4026, opened 2021-03-02
--private and $HOME


reinerh:
A Debian user [reported](https://bugs.debian.org/983746) the oddity that a `bin` directory inside a `--private` home directory is read-only.
This is caused by `read-only ${HOME}/bin` from `disable-common.inc`.

Is it actually intended that `${HOME}` applies to the private home directory as well?

rusty-snake:
If not we can close #903.

------

> I would have thought that the typical use is to create a directory,
copy a minimum number of (non-confidential) files in it, and do
all the dirty work in it via "firejail --private=". I do that for
testing actively developed 3rd party software, and this needs
something like "make install" each time the software is updated.
If bin is read-only, this doesn't work.

That's one workflow, another can be `--private=~/work` or `--private=~/kleingartenverein`. Then ${HOME} should expand to the --private home. Let's say you run ssh and firefox in ~/work, then you don't want that firefox can read the ssh keys of your company.

So I think it's better to expand ${HOME} to --private and use noblacklist/nowhitelist/read-write/ignore.

vinc17fr:
Well, if firefox cannot read the ssh configuration, then ssh won't be able to read it either (both being run in the jail)! So what's the point of storing ssh configuration in `~/work`?

rusty-snake:
> Well, if firefox cannot read the ssh configuration, then ssh won't be able to read it either (both being run in the jail)!

The workflow would look like `firejail --private=~/work firefox` and `firejail --private=~/work ssh`. Both will use their own profile (with may has `noblacklist ${HOME}/.ssh`).

> So what's the point of storing ssh configuration in ~/work?

Using different homes for work, personal stuff, association activity, ... (https://firejail.wordpress.com/documentation-2/firefox-guide/#work).

reinerh:
@vinc17fr Would it work for you to create overrides in `disable-common.local` (or your own profile if you are using one)?
Some users might already be relying on the fact that `$HOME` also applies to private home directories, so it's difficult to loosen this now.

vinc17fr:
If firefox and ssh have their own profile, I suppose that it is this profile that should blacklist/restrict the various files and directories, just like if `--private` were not used. Having to use things like `noblacklist ${HOME}/.ssh` for ssh would just be ugly and error prone.

What do you mean by overrides? Note that just unblacklisting `bin` would not be sufficient in general. There should be no restriction on files inside the private directory.

reinerh:
[...]

PR #4036, opened 2021-03-03
Add new condition ?HAS_PRIVATE:

rusty-snake:
Idea from @vinc17fr
https://github.com/netblue30/firejail/issues/4026#issuecomment-789178572

dp gets included
```
?HAS_PRIVATE: ignore include disable-programs.inc
include disable-programs.inc
```

dp gets not included
```
private
?HAS_PRIVATE: ignore include disable-programs.inc
include disable-programs.inc
```

dp gets not included
```
private ${HOME}/foo
?HAS_PRIVATE: ignore include disable-programs.inc
include disable-programs.inc
```

### DIFF
--- a/src/firejail/profile.c
+++ b/src/firejail/profile.c
@@ -157,6 +157,10 @@ static int check_nosound(void) {
 	return arg_nosound != 0;
 }
 
+static int check_private(void) {
+	return arg_private;
+}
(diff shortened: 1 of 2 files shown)


netblue30:
merged, thanks!

PR #4037, opened 2021-03-03
adding support for "play with..." extension

pirate486743186:
adding support for "play with..." extension. It uses a clever hack, so it doesn't need an external application in order to launch mpv.

Also trimmed ff2mpv comment block.
and added .cache/youtube-dl


netblue30:
merged, thanks

PR #4040, opened 2021-03-03
sandbox setup: postpone fslogger

smitsohu:
Postpone writing of log file in order to also catch filesystem modifications from x11 code


netblue30:
all in, thanks!

PR #4042, opened 2021-03-03
private-lib hardening

smitsohu:
Ensure that libraries are loaded from below a default ld.so search path

It is reasonable for firejail to expect that unprivileged users have
no write permission on these paths; lax permissions there would
mean that the system is probably screwed anyway.

A somewhat similar check exists already in `private-bin` : https://github.com/netblue30/firejail/blob/3565217343a64b71c60e376053dda6af1bfff42f/src/firejail/fs_bin.c#L166

From a functional perspective this should be no limitation, because
`private-lib` not need to bother about libraries from paths that are
not masked in the end.

### DIFF
--- a/src/firejail/fs_lib.c
+++ b/src/firejail/fs_lib.c
@@ -33,6 +33,31 @@ extern void fslib_install_system(void);
 static int lib_cnt = 0;
 static int dir_cnt = 0;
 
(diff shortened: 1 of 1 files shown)


netblue30:
merged, thanks!

PR #4043, opened 2021-03-03
private-lib: mask /usr/local/lib[,64] directories, too

smitsohu:
Currently depends on #4042 (hopefully that makes sense). 

~~I'll send another pull request to move private-lib from copying to mounting the next few days.~~ Done.

### DIFF
--- a/src/firejail/fs_lib.c
+++ b/src/firejail/fs_lib.c
@@ -33,6 +33,31 @@ extern void fslib_install_system(void);
 static int lib_cnt = 0;
 static int dir_cnt = 0;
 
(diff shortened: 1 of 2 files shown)


smitsohu:
I have lost confidence in this patch. It will create problems when there are conflicting versions of the same directory in `/usr/local/lib[,64]` and in other searched paths. This is more common than I thought.

The initial response would be to turn it upside down and not consider `/usr/local/lib[,64]` at all, which would be similar to how `private-lib` treats these paths currently.

netblue30:
merged, thanks!

PR #4045, opened 2021-03-04
private-lib: move to mount-only

### LINKED ISSUE #3980: private-lib: can we mount instead of copy?
While in general `private` options copy files in the sandbox instead of mounting them, there are two exceptions to the rule:

- `private-tmp`, which mounts a set of paths in the sandbox in order to provide essential sockets 

- `private-lib`, which currently employs a mix of copying and mounting: files are copied and directories are mounted

I would propose to move all of `private-lib` to mounts, mainly for performance reasons. For example it takes Firejail seconds to configure the sandbox `firejail --noprofile --private-lib kcalc`, even on a rather powerful notebook. On legacy hardware it can be ten seconds and more. Switching from copying to mounting accelerates sandbox setup by about an order of magnitude.

smitsohu:
Currently depends on #4043 (hopefully that makes sense)

Closes #3980

### DIFF
--- a/src/firejail/fs_lib.c
+++ b/src/firejail/fs_lib.c
@@ -23,7 +23,8 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
-#include <dirent.h>
+#include <fcntl.h>
+#include <errno.h>
 #include <glob.h>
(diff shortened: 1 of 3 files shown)


netblue30:
all set!

PR #4069, opened 2021-03-08
Add profile for checksum tools

rusty-snake:
Someone told me that md5 is insecure, so I decided it is better to sandbox it so that is can be used securely. :upside_down_face:

---
hasher-common.profile is based on archiver-common.profile.

### DIFF
--- a/etc/profile-a-l/hasher-common.profile
+++ b/etc/profile-a-l/hasher-common.profile
@@ -1,0 +1,60 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+include hasher-common.local
+
+# common profile for hasher/checksum tools
+
+blacklist ${RUNUSER}
+
+# WARNING:
+# Users can (un)restrict file access for **all** hashers by commenting/uncommenting the needed
+# include file(s) here or by putting those into hasher-common.local.
+# Another option is to do this **per hasher** in the relevant <hasher>.local.
+# Just beware that things tend to break when overtightening profiles. For example, because you only
+# need to hash/check files in ${DOWNLOADS}, other applications may need access to ${HOME}/.local/share.
+
+# Uncomment the next line (or put it into your hasher-common.local) if you don't need to hash files in disable-common.inc.
+#include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+# Uncomment the next line (or put it into your hasher-common.local) if you don't need to hash files in disable-programs.inc.
+#include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
+# Uncomment the next line (or put it into your hasher-common.local) if you don't need to hash files in disable-xdg.inc.
+#include disable-xdg.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
+net none
+no3d
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+x11 none
+
+# Uncomment the next line (or put it into your hasher-common.local) if you don't need to hash files in /tmp.
+#private-cache
+private-dev
+# Uncomment the next line (or put it into your hasher-common.local) if you don't need to hash files in /tmp.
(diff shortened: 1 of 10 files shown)


etc/profile-a-l/hasher-common.profile
```diff
@@ -0,0 +1,60 @@
...
+blacklist ${RUNUSER}
+
+# WARNING:
+# Users can (un)restrict file access for **all** hashers by commenting/uncommenting the needed
+# include file(s) here or by putting those into hasher-common.local.
+# Another option is to do this **per hasher** in the relevant <hasher>.local.
+# Just beware that things tend to break when overtightening profiles. For example, because you only
+# need to hash/check files in ${DOWNLOADS}, other applications may need access to ${HOME}/.local/share.
+
+# Uncomment the next line (or put it into your archiver-common.local) if you don't need to hash files in disable-common.inc.
```
glitsj16, line 16: hasher-common.local instead of archiver-common.local

etc/profile-a-l/hasher-common.profile
```diff
@@ -0,0 +1,60 @@
...
+#include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+# Uncomment the next line (or put it into your archiver-common.local) if you don't need to hash files in disable-programs.inc.
```
glitsj16, line 22: same here, hasher-common.local instead of archiver-common.local

etc/profile-a-l/hasher-common.profile
```diff
@@ -0,0 +1,60 @@
...
+#include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
+# Uncomment the next line (or put it into your archiver-common.local) if you don't need to hash files in disable-xdg.inc.
```
glitsj16, line 26: one more, hasher-common.local instead of archiver-common.local

etc/profile-a-l/hasher-common.profile
```diff
@@ -0,0 +1,60 @@
...
+nou2f
+novideo
+protocol unix
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+x11 none
+
+# Uncomment the next line (or put it into your archiver-common.local) if you don't need to hash files in /tmp.
```
glitsj16, line 50: almost done, hasher-common.local instead of archiver-common.local

etc/profile-a-l/hasher-common.profile
```diff
@@ -0,0 +1,60 @@
...
+#private-cache
+private-dev
+# Uncomment the next line (or put it into your archiver-common.local) if you don't need to hash files in /tmp.
```
glitsj16, line 53: last one, hasher-common.local instead of archiver-common.local

glitsj16 (approved):
Just a few occurences of archiver-common.local that need changing to hasher-common.local and this is good to go. Very nice additions!

rusty-snake:
`quiet` should be added in any way. I am unsure about firecfg, @glitsj16 what's your opinion?

netblue30:
> I am unsure about firecfg

It could be a problem when they run it from scripts. It will slow down if they do it in a loop on large number of file, for example a package manager comparing checksums - we introduce about 50 ms to start each sandbox.

glitsj16:
> I am unsure about firecfg

> It could be a problem when they run it from scripts. It will slow down if they do it in a loop on large number of files, for example a package manager comparing checksums - we introduce about 50 ms to start each sandbox.

Package management indeed poses a bit of a problem for these checksum tools, as do archivers (see #3095). I guess not all firejail users do a lot of manual/scripted package building, activities where this inherent slowing down factor quickly turns into a nuissance. Those that do are `probably` more aware of this and have the skills to work around it. By leaving them out of firecfg we tend to disfavour the first set of users. For the occasional checksumming outside of package management it would be nice to add them. What about adding a setting in firejail.config so users can have more control over their preference. That setting would have to be tied into firecfg to keep things from breaking, and in all honesty I haven't looked into the feasibility of doing that.

glitsj16:
Oops, I didn't want to close this!

netblue30:
OK, let's grab it as is, I'll put quiet in all of them. Thanks!

netblue30:
sorry, quiet was already there!

PR #4071, opened 2021-03-08
Commons of opengl-game-wrapper.sh

rusty-snake:
Add profiles for alienarena, ballbuster, colorful, gl-117, glaxium, pinball

Commons of opengl-game-wrapper.sh 

 - Add allow-opengl-game.inc
 - Add profiles for alienarena-wrapper, ballbuster-wrapper,
   colorful-wrapper, etr-wrapper, gl-117-wrapper, glaxium-wrapper,
   neverball-wrapper, neverputt-wrapper, pinball-wrapper,
   supertuxkart-wrapper
 - Use allow-opengl-game.inc in xonotic.profile and the profiles above
 - xonotic.profile: simplify private-bin by using xonotic*

### DIFF
--- a/etc/profile-m-z/xonotic.profile
+++ b/etc/profile-m-z/xonotic.profile
@@ -8,12 +8,16 @@ include globals.local
 
 noblacklist ${HOME}/.xonotic
 
+include allow-bin-sh.inc
+include allow-opengl-game.inc
+
 include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
 include disable-programs.inc
+include disable-shell.inc
 include disable-xdg.inc
 
 mkdir ${HOME}/.xonotic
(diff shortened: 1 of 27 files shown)


rusty-snake:
I'm unsure whether we should add profiles for all the GAME-wrappers or not.

Explanation: Fedora (and maybe more) installs a .desktop that uses `Exec=foo-wrapper` instead of `Exec=foo`. `/usr/bin/foo-wrapper` is a symlink to `/usr/bin/opengl-game-wrapper.sh` which checks HW-accel and then starts the game or notifies the user that there is a problem.

```bash
$ cat /usr/bin/opengl-game-wrapper.sh
#!/usr/bin/bash

. /usr/share/opengl-games-utils/opengl-game-functions.sh

GAME=`basename $0 | sed 's/-wrapper.*//'`

checkDriOK $GAME

exec $GAME "$@"
$ cat /usr/share/opengl-games-utils/opengl-game-functions.sh 
# check if DRI is available, true if it is, false otherwise

function hasDri ()
{
  if [ "`glxinfo | grep "direct rendering: " | head -n 1 | cut -d " " -f 3`" != Yes ] ||
     glxinfo | grep -qE "OpenGL renderer string: Software Rasterizer|OpenGL renderer string: Gallium .* on llvmpipe"; then
    return 1
  else
    return 0
  fi 
}

# check if DRI is available, show an error and exit if it isn't

function checkDriOK ()
{
  if ! hasDri; then
    zenity --error --text="Your system currently is not capable of hardware \
accelerated 3D. Therefore $1 cannot run.

Usually the cause of this error is that there are no Free Software drivers \
for your graphics card, please contact your graphics card manufacturer and \
kindly ask them to provide Free Software support for your card."
    exit 1;
  fi
}
```

On the one hand it makes sense to have this check inside that sandbox. On the other hand it requires a shell which can weaken the sandbox.

etc/profile-m-z/xonotic.profile
```diff
@@ -8,12 +8,16 @@ include globals.local
 
 noblacklist ${HOME}/.xonotic
 
+include allow-opengl-game.inc
```
glitsj16, line 11: nitpick: move this below include allow-bin-sh.inc for alphabetical ordering?

glitsj16 (approved):
LGTM. Only one nitpick.

glitsj16:
> I'm unsure whether we should add profiles for all the GAME-wrappers or not.
On the one hand it makes sense to have this check inside that sandbox. On the other hand it requires a shell which can weaken the sandbox.

Always difficult IMO weighing pro's and con's in such a context. How about adding a comment? I mean, keep the sandbox as tight as possible (no shell) and inform users on how to proceed if they want that? As you can probably tell, I don't have any experience with (sandboxing) games, so I can't offer anything helpful here I'm afraid.

glitsj16 (approved):
LGTM. Nice work!

PR #4079, opened 2021-03-11
Add profile for youtube-dl-gui & some other changes

Neo00001:
`private-bin` for telegram & virtualbox tested on `Arch Linux` & works pretty well. But don't know about other distro's, that's why I  commented out those lines.

### DIFF
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -48,7 +48,6 @@ private-etc alsa,alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.
 private-tmp
 
 dbus-user filter
-dbus-user.own org.telegram.desktop
 dbus-user.talk org.freedesktop.Notifications
 dbus-user.talk org.kde.StatusNotifierWatcher
 dbus-user.talk org.gnome.Mutter.IdleMonitor
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -41,7 +41,7 @@ shell none
 tracelog
 
 disable-mnt
-#private-bin telegram-desktop,telegram,Telegram,telegram-desktop
+#private-bin telegram,Telegram,telegram-desktop
 private-cache
 private-dev
 private-etc alsa,alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.cache,localtime,machine-id,os-release,passwd,pki,pulse,resolv.conf,ssl,xdg
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -41,7 +41,7 @@ shell none
 tracelog
 
 disable-mnt
-#private-bin bash,sh,telegram-desktop
+#private-bin bash,sh,telegram-desktop,telegram,Telegram,telegram-desktop
 private-cache
 private-dev
 private-etc alsa,alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.cache,localtime,machine-id,os-release,passwd,pki,pulse,resolv.conf,ssl,xdg
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -41,7 +41,7 @@ shell none
 tracelog
 
 disable-mnt
-#private-bin bash,sh,telegram-desktop,telegram,Telegram,telegram-desktop
+#private-bin telegram-desktop,telegram,Telegram,telegram-desktop
 private-cache
 private-dev
 private-etc alsa,alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.cache,localtime,machine-id,os-release,passwd,pki,pulse,resolv.conf,ssl,xdg
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -47,4 +47,10 @@ private-dev
 private-etc alsa,alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.cache,localtime,machine-id,os-release,passwd,pki,pulse,resolv.conf,ssl,xdg
 private-tmp
 
+dbus-user filter
+dbus-user.own org.telegram.desktop
+dbus-user.talk org.freedesktop.Notifications
+dbus-user.talk org.kde.StatusNotifierWatcher
+dbus-user.talk org.gnome.Mutter.IdleMonitor
+dbus-user.talk org.freedesktop.ScreenSaver
 dbus-system none
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -36,10 +36,15 @@ noroot
 notv
 protocol unix,inet,inet6,netlink
 seccomp
+seccomp.block-secondary
 shell none
+tracelog
 
 disable-mnt
+#private-bin bash,sh,telegram-desktop
 private-cache
 private-dev
 private-etc alsa,alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.cache,localtime,machine-id,os-release,passwd,pki,pulse,resolv.conf,ssl,xdg
 private-tmp
+
+dbus-system none
--- a/etc/profile-m-z/virtualbox.profile
+++ b/etc/profile-m-z/virtualbox.profile
@@ -44,7 +44,7 @@ shell none
 tracelog
 
 #disable-mnt
-#private-bin basename,bash,env,gawk,grep,ps,readlink,sh,virtualbox,VirtualBox,VBox,VBox*,vbox*,vboximg-mount,whoami
+#private-bin basename,bash,env,gawk,grep,ps,readlink,sh,virtualbox,VirtualBox,VBox*,vbox*,whoami
 private-cache
 private-etc alsa,asound.conf,ca-certificates,conf.d,crypto-policies,dconf,fonts,hostname,hosts,ld.so.cache,localtime,machine-id,pki,pulse,resolv.conf,ssl
 private-tmp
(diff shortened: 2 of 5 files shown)


etc/profile-m-z/telegram.profile
```diff
@@ -36,10 +36,15 @@ noroot
 notv
 protocol unix,inet,inet6,netlink
 seccomp
+seccomp.block-secondary
 shell none
+tracelog
 
 disable-mnt
+#private-bin bash,sh,telegram-desktop
```
rusty-snake, line 44: There seem to be at least three names for the binary:

```
$ ls /etc/firejail | grep -i telegram
telegram-desktop.profile
telegram.profile
Telegram.profile
```

```suggestion
#private-bin bash,sh,telegram,Telegram,telegram-desktop
```

and bash,sh is blacklisted by `include disable-shell.inc` so a (commented) `include allow-bin-sh.inc` would make sense.

etc/profile-m-z/telegram.profile
```diff
@@ -36,10 +36,15 @@ noroot
...
 private-cache
 private-dev
 private-etc alsa,alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.cache,localtime,machine-id,os-release,passwd,pki,pulse,resolv.conf,ssl,xdg
 private-tmp
+
+dbus-system none
```
rusty-snake, line 50: I already thought about a `dbus-user filter` policy https://github.com/netblue30/firejail/issues/3638#issuecomment-695988562. Maybe you can try if the following works (still very weak).

```
dbus-user filter
dbus-user.own org.telegram.desktop
dbus-user.talk org.kde.StatusNotifierWatcher
dbus-user.talk org.gnome.Mutter.IdleMonitor
```

etc/profile-m-z/youtube-dl-gui.profile
```diff
@@ -0,0 +1,58 @@
...
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/youtube-dlg
+whitelist ${HOME}/.config/youtube-dlg
+whitelist ${VIDEOS}
```
rusty-snake, line 25: Just thinking: `whitelist ${DOWNLOADS}`, yes / no? IDK

etc/profile-m-z/youtube-dl-gui.profile
```diff
@@ -0,0 +1,58 @@
...
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+machine-id
+netfilter
+no3d
```
rusty-snake, line 35: This maybe decreases performances of ffmpeg while converting.

etc/profile-m-z/youtube-dl-gui.profile
Neo00001: let's go for `whitelist ${DOWNLOADS}` instead of `whitelist ${VIDEOS}`

etc/profile-m-z/youtube-dl-gui.profile
Neo00001: I haven't noticed any performance degradation. Perhaps, on big files performance degradation can be noticeable.I will report you back on this.

etc/profile-m-z/telegram.profile
Neo00001: on kde I need just
```
dbus-user filter
dbus-user.talk org.kde.StatusNotifierWatcher
```
btw, is `dbus-user.talk org.freedesktop.ScreenSaver` needed?

etc/profile-m-z/telegram.profile
Neo00001: I'm running with `include disable-shell.inc` & without `include allow-bin-sh.inc`. Wait a minute, telegram was not running without `bash` or `sh`. So with `include disable-shell.inc` telegram should not run anyway if I'm not wrong. But,I didn't face any issue whatsoever.

etc/profile-m-z/telegram.profile
rusty-snake: IDK, they are just picked from it's flatpak, which has a lot of talk rules. It has even a talk rules for a `org.freedesktop.portal.foobar`, though flatpak always allows `org.freedesktop.portal.*`. If it is needed, it's used to inhibit the screensaver in e.g. video-calls.

etc/profile-m-z/telegram.profile
Neo00001: well got the reason. it was getting called by `dbus`. though with `private-bin` I don't need `include allow-bin-sh.inc` or `dbus`. A bit confused here.What's your interpretation?

etc/profile-m-z/telegram.profile
Neo00001: then we should allow bare minimum. & rest should be commented out. what are necessary on gnome for basic functioning.

etc/profile-m-z/telegram.profile
Neo00001: I was running `telegram` from terminal without referring  absolute path. probably that's why `bash` was needed. telegram can run without shell it seems. Sorry for all the non-sense. need a good sleep.

etc/profile-m-z/youtube-dl-gui.profile
Neo00001: yep. it degrades performance.

etc/profile-m-z/telegram.profile
```diff
@@ -36,10 +36,15 @@ noroot
 notv
 protocol unix,inet,inet6,netlink
 seccomp
+seccomp.block-secondary
 shell none
+tracelog
 
 disable-mnt
+#private-bin telegram-desktop,telegram,Telegram,telegram-desktop
```
rusty-snake, line 44:
```suggestion
#private-bin telegram,Telegram,telegram-desktop
```

etc/profile-m-z/virtualbox.profile
```diff
@@ -44,6 +44,7 @@ shell none
 tracelog
 
 #disable-mnt
+#private-bin basename,bash,env,gawk,grep,ps,readlink,sh,virtualbox,VirtualBox,VBox,VBox*,vbox*,vboximg-mount,whoami
```
rusty-snake, line 47:
```suggestion
#private-bin basename,bash,env,gawk,grep,ps,readlink,sh,virtualbox,VirtualBox,VBox*,vbox*,whoami
```

etc/profile-m-z/telegram.profile
Neo00001: wow... what happens to my eyes! my eyes!

etc/profile-m-z/telegram.profile
rusty-snake: In general I agree with you to only allow a bare minimum. However there is always a difference between adding restrictions and weakening. Here we start introducing restrictions at all and I don't want break to much that was working before. Especial if you look at https://github.com/netblue30/firejail/wiki/Restrict-D-Bus. Allowing `org.freedesktop.ScreenSaver` if `org.freedesktop.Notifications` is already allowed adds virtually no new capabilities.

etc/profile-m-z/youtube-dl-gui.profile
rusty-snake: IMHO we should remove it then.

Neo00001: > IMHO we should remove it then.

or rather keep it commented out. which one you prefer?

rusty-snake: TBH I don't care. It's not that extra protections that it's worth to be used if you only download slow files IMHO.

etc/profile-m-z/telegram.profile
```diff
@@ -47,4 +47,10 @@ private-dev
 private-etc alsa,alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.cache,localtime,machine-id,os-release,passwd,pki,pulse,resolv.conf,ssl,xdg
 private-tmp
 
+dbus-user filter
+dbus-user.own org.telegram.desktop
```
rusty-snake, line 51: If `busctl --user list | grep org.telegram.desktop` does not show it while you run telegram, we can remove it.

etc/profile-m-z/youtube-dl-gui.profile
Neo00001: download speed remains same,it's the post-download processing which is affected.I would like to keep it commented out.

etc/profile-m-z/telegram.profile
Neo00001: yea. I thought gnome or any other de has it.

etc/profile-m-z/youtube-dl-gui.profile
```diff
@@ -0,0 +1,58 @@
...
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+# Following degrades performances
+#caps.drop all
+#machine-id
+#netfilter
+#no3d
```
rusty-snake, line 35:
```suggestion
caps.drop all
machine-id
netfilter
# Following degrades performances
#no3d
```

Only `no3d` degrades performances.
`no3d` `blacklist`s [`/dev/dri`](https://en.wikipedia.org/wiki//dev/dri), therefore ffmpeg falls back to software en/decoding instead of hardware accelerated en/decoding.

Neo00001: oh.. then no need for no3d.

rusty-snake (approved):

netblue30:
all in, thanks!

PR #4085, opened 2021-03-11
Document enabling debugging for Node.js

tredondo:


etc/profile-m-z/nodejs-common.profile
```diff
@@ -1,5 +1,6 @@
 # Firejail profile for Node.js
-# Description: Common profile for npm/yarn
+# Description: Asynchronous event-driven JavaScript runtime
```
rusty-snake, line 2: It's nodejs-common.profile not nodejs.profile. The description is right.

```
$ grep nodejs-common.profile /etc/firejail/*
/etc/firejail/npm.profile:include nodejs-common.profile
/etc/firejail/yarn.profile:include nodejs-common.profile
```

etc/profile-m-z/nodejs-common.profile
```diff
@@ -1,5 +1,6 @@
...
+# See comments for enabling debugging with certain IDEs
```
rusty-snake, line 3: I don't think this is necessary, we have comments like (un)comment for edge-case bla in a lot profiles.

rusty-snake (commented):
IMHO the descriptions should be changed, rest LGTM.

etc/profile-m-z/nodejs-common.profile
tredondo: You're right now that I looked again at the profiles. So there's no specific profile for Node.js then. Should the description be "Common profile for npm, yarn and Node.js"? And maybe have a `node.profile` that includes it? I don't know why there's no node.profile.

glitsj16: > So there's no specific profile for Node.js then. Should the description be "Common profile for npm, yarn and Node.js"? And maybe have a node.profile that includes it? I don't know why there's no node.profile.

Correct, we don't have node.profile (yet). A _very_ untested example:
```
# Firejail profile for node
# Description: Evented I/O for V8 javascript
quiet
include node.local
include globals.local

# Redirect
include nodejs-common.profile
```

_If_ we introduce such `node.profile` you could change the description of `nodejs-common.profile`. My Node.js experience is way too limited to judge whether that's a good idea or not.

netblue30:
merged, thanks!

PR #4095, opened 2021-03-13
make appimage mounts private to sandbox

smitsohu:
Mount AppImage on `/run/firejail/appimage` in the sandbox process, rather than on a subdirectory in the main process. This way the mount is destroyed automatically when the sandbox terminates, and it is private to the sandbox and not a global resource (except for root, who can always peek into the loop device).


netblue30:
merged, thanks!

PR #4098, opened 2021-03-14
Create bcompare.profile

tredondo:
My first profile, please review and squash the commits :grin:

In particular, I don't really understand what `ipc-namespace` is for, but the template says it's for CLI-only apps, so I've left it commented out.

Also unclear on the private-* and dbus stuff.

### DIFF
--- a/etc/profile-a-l/bcompare.profile
+++ b/etc/profile-a-l/bcompare.profile
@@ -7,6 +7,7 @@ include bcompare.local
 # Persistent global definitions
 include globals.local
 
+noblacklist ${HOME}/.config/bcompare
 # In case the user decides to include disable-programs.inc, still allow
 # KDE's Gwenview to view images via right click -> Open With -> Associated Application
 noblacklist ${HOME}/.config/gwenviewrc
--- a/etc/profile-a-l/bcompare.profile
+++ b/etc/profile-a-l/bcompare.profile
@@ -33,7 +33,6 @@ include disable-write-mnt.inc
 
 apparmor
 caps.drop all
-# ipc-namespace
 machine-id
 net none
 no3d
--- a/etc/profile-a-l/bcompare.profile
+++ b/etc/profile-a-l/bcompare.profile
@@ -8,7 +8,8 @@ include bcompare.local
 # Persistent global definitions
 include globals.local
 
-# In case the user decides to include disable-programs.inc, still allow KDE's Gwenview to view images via right click -> Open With -> Associated Application
+# In case the user decides to include disable-programs.inc, still allow
+# KDE's Gwenview to view images via right click -> Open With -> Associated Application
 noblacklist ${HOME}/.config/gwenviewrc
 
 include disable-common.inc
--- a/etc/profile-a-l/bcompare.profile
+++ b/etc/profile-a-l/bcompare.profile
@@ -1,0 +1,65 @@
+# Firejail profile for Scootersoft Beyond Compare
+# Description: directory compare and file compare utility
+# Disables the network, which only impacts checking for updates.
+# This file is overwritten after every install/update
+
+# Persistent local customizations
+include bcompare.local
+# Persistent global definitions
+include globals.local
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+# Allow comparing e.g. Firefox profile files
+# include disable-programs.inc
+# Uncommenting this breaks launch
+# include disable-shell.inc
+include disable-write-mnt.inc
+include disable-xdg.inc
+
+# include whitelist-common.inc
+# include whitelist-runuser-common.inc
+# include whitelist-usr-share-common.inc
+# include whitelist-var-common.inc
+
+#allusers
+apparmor
+caps.drop all
+# ipc-namespace
+machine-id
+net none
+no3d
+# nodvd
+nogroups
+nonewprivs
+noroot
+# nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+# shell none
+# tracelog
+
+# disable-mnt
+# private
+# private-bin program
+private-cache
+private-dev
+# see /usr/share/doc/firejail/profile.template for more common private-etc paths.
+# private-etc alternatives,fonts,machine-id
+# private-lib
+# private-opt none
+# Necessary because of the `include disable-exec.inc` line. Prevents error "Error fstat: fs.c:504 fs_remount_simple: Transport endpoint is not connected ... cannot sync with peer: unexpected EOF Peer [...] unexpectedly exited with status 1"
+private-tmp
+
+# dbus-user none
+# dbus-system none
+
+# memory-deny-write-execute
+noexec ${HOME}
+# read-only ${HOME}
--- a/etc/profile-a-l/bcompare.profile
+++ b/etc/profile-a-l/bcompare.profile
@@ -29,7 +29,6 @@ include disable-write-mnt.inc
 # include whitelist-usr-share-common.inc
 # include whitelist-var-common.inc
 
-#allusers
 apparmor
 caps.drop all
 # ipc-namespace
--- a/etc/profile-a-l/bcompare.profile
+++ b/etc/profile-a-l/bcompare.profile
@@ -8,12 +8,15 @@ include bcompare.local
 # Persistent global definitions
 include globals.local
 
+# In case the user decides to include disable-programs.inc, still allow KDE's Gwenview to view images via right click -> Open With -> Associated Application
+noblacklist ${HOME}/.config/gwenviewrc
+
 include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
-# Allow comparing e.g. Firefox profile files
+# Allow comparing e.g. Firefox profile files, so don't disable access to programs
 # include disable-programs.inc
 # Uncommenting this breaks launch
 # include disable-shell.inc
--- a/etc/profile-a-l/bcompare.profile
+++ b/etc/profile-a-l/bcompare.profile
@@ -18,8 +18,8 @@ include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
 include disable-passwdmgr.inc
-# Allow comparing e.g. Firefox profile files, so don't disable access to programs
-# include disable-programs.inc
+# Uncomment the next line (or put it into your bcompare.local) if you don't need to compare files in disable-programs.inc
+#include disable-programs.inc
 # Uncommenting this breaks launch
 # include disable-shell.inc
 include disable-write-mnt.inc
@@ -37,7 +37,7 @@ caps.drop all
 #machine-id
 net none
 no3d
-#nodvd
+nodvd
 nogroups
 nonewprivs
 noroot
--- a/etc/profile-a-l/bcompare.profile
+++ b/etc/profile-a-l/bcompare.profile
@@ -2,7 +2,6 @@
 # Description: directory and file compare utility
 # Disables the network, which only impacts checking for updates.
 # This file is overwritten after every install/update
-
 # Persistent local customizations
 include bcompare.local
 # Persistent global definitions
--- a/etc/profile-a-l/bcompare.profile
+++ b/etc/profile-a-l/bcompare.profile
@@ -33,14 +33,16 @@ include disable-write-mnt.inc
 
 apparmor
 caps.drop all
-machine-id
+# Uncommenting might break Pulse Audio
+#machine-id
 net none
 no3d
-# nodvd
+#nodvd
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
+# Firejail profile for Beyond Compare by Scooter Software
+# Description: directory and file compare utility
+# Disables the network, which only impacts checking for updates.
+# This file is overwritten after every install/update
+
```
rusty-snake, line 5:
```suggestion
```

No blank line here

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+# Persistent local customizations
+include bcompare.local
+# Persistent global definitions
+include globals.local
+
+# In case the user decides to include disable-programs.inc, still allow KDE's Gwenview to view images via right click -> Open With -> Associated Application
```
rusty-snake, line 11:
```suggestion
# In case the user decides to include disable-programs.inc, still allow
# KDE's Gwenview to view images via right click -> Open With -> Associated Application
```

break long lines

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+# Allow comparing e.g. Firefox profile files, so don't disable access to programs
+# include disable-programs.inc
+# Uncommenting this breaks launch
+# include disable-shell.inc
```
rusty-snake, line 22:
```suggestion
```

no need to keep such lines (in general it's often enough to `include allow-bin-sh.inc`)

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+include disable-write-mnt.inc
+# Don't disable ${DOCUMENTS}, ${MUSIC}, ${PICTURES}, ${VIDEOS}
+# include disable-xdg.inc
```
rusty-snake, line 25:
```suggestion
```

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+nonewprivs
+noroot
+# nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+# shell none
+# tracelog
```
rusty-snake, line 50: Can we uncomment them?

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+# include whitelist-common.inc
+# include whitelist-runuser-common.inc
+# include whitelist-usr-share-common.inc
+# include whitelist-var-common.inc
+
+#allusers
+apparmor
+caps.drop all
+# ipc-namespace
+machine-id
```
rusty-snake, line 36: Either comment this too or uncomment `nosound` (`machine-id` breaks PA).

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+
+# include whitelist-common.inc
+# include whitelist-runuser-common.inc
+# include whitelist-usr-share-common.inc
+# include whitelist-var-common.inc
+
+#allusers
+apparmor
+caps.drop all
+# ipc-namespace
```
rusty-snake, line 35:
```suggestion
```

FYI: Causes issues with X11 depending on {{nobody knows when exactly}}

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+include disable-write-mnt.inc
+# Don't disable ${DOCUMENTS}, ${MUSIC}, ${PICTURES}, ${VIDEOS}
+# include disable-xdg.inc
+
+# include whitelist-common.inc
+# include whitelist-runuser-common.inc
+# include whitelist-usr-share-common.inc
+# include whitelist-var-common.inc
+
+#allusers
```
rusty-snake, line 32:
```suggestion
```

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+private-dev
+# see /usr/share/doc/firejail/profile.template for more common private-etc paths.
+# private-etc alternatives,fonts,machine-id
+# private-lib
+# private-opt none
+# Necessary because of the `include disable-exec.inc` line. Prevents error "Error fstat: fs.c:504 fs_remount_simple: Transport endpoint is not connected ... cannot sync with peer: unexpected EOF Peer [...] unexpectedly exited with status 1"
+private-tmp
+
+# dbus-user none
+# dbus-system none
```
rusty-snake, line 65: Can we enabled them?

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+
+# memory-deny-write-execute
+noexec ${HOME}
+# read-only ${HOME}
```
rusty-snake, line 69:
```suggestion
```

`noexec ${HOME}` is in disable-exec.inc

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+
+# disable-mnt
+# private
+# private-bin program
+private-cache
+private-dev
+# see /usr/share/doc/firejail/profile.template for more common private-etc paths.
+# private-etc alternatives,fonts,machine-id
+# private-lib
+# private-opt none
```
rusty-snake, line 60:
```suggestion
```

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+nou2f
+novideo
+protocol unix
+seccomp
+# shell none
+# tracelog
+
+# disable-mnt
+# private
+# private-bin program
```
rusty-snake, line 54:
```suggestion
```

etc/profile-a-l/bcompare.profile
```diff
@@ -0,0 +1,69 @@
...
+
+# Persistent local customizations
+include bcompare.local
+# Persistent global definitions
+include globals.local
+
+# In case the user decides to include disable-programs.inc, still allow KDE's Gwenview to view images via right click -> Open With -> Associated Application
+noblacklist ${HOME}/.config/gwenviewrc
+
+include disable-common.inc
```
rusty-snake, line 14: I would suggest that we use the same comment format as in meld.profile and kdiff3.profile: `Uncomment the next line (or put it into your meld.local) if you don't need to compare files in FOO`

```suggestion
# Uncomment the next line (or put it into your meld.local) if you don't need to compare files in disable-common.inc
#include disable-common.inc
```

rusty-snake (commented):
Looks good, some nitpicks to go.

- add it to src/firecfg/firecfg.config
- if it has it's own configuration files/dirs, add them to disable-programs.inc

etc/profile-a-l/bcompare.profile
dandv: Yeah that's because there's no blank line in the template, but I think there *should* be a blank line in the template, to allow for separation between a user-friendly header, and the nitty-gritty of the profile.

tredondo: Yeah, that's become there's no blank line in the template, but I think there *should* be a blank line in the template, to separate the user-friendly header from the nitty-gritty of the profile.

rusty-snake: No profile has a blank line there, so if you think so you should open an issues IMHO. So we all can discuss this and change every profile in a separate PR.

etc/profile-a-l/bcompare.profile
tredondo: Beyond Compare still seems to run fine.

tredondo:
>     * add it to src/firecfg/firecfg.config
>     * if it has it's own configuration files/dirs, add them to disable-programs.inc

Added those instructions [to the wiki](https://github.com/netblue30/firejail/wiki/Creating-Profiles/_compare/52dca3086f2ff8429212c67243cd660d01ecaa85...ca2c4887be1bf5abe81b2b1e8ce97cc8d23cf08c).

etc/profile-a-l/bcompare.profile
tredondo: Yes, runs fine.

tredondo:
Should I add `noblacklist ${HOME}/.config/bcompare`, given that I've added bcompare to `disable-programs.inc`?

etc/profile-a-l/bcompare.profile
tredondo: Committed with `meld` replaced by `bcompare`.

etc/profile-a-l/bcompare.profile
tredondo: Launching VLC from the context menu of an MP3 does play it just fine with `machine-id` enabled, on my KDE neon 5.21. But I've commented `machine-id` anyway.

On a related note, despite `novideo`, VLC launched from the context menu of an `mp4`, does play it.

Also, should `nodvd` be uncommented? Is that a multimedia API of some sort, so if the user wants to compare files from a DVD, `nodvd` wouldn't have an effect because the DVD would be mounted via `/media/` ?

etc/profile-a-l/bcompare.profile
tredondo: I've kept them because to me it was surprising to realize that `xdg` was related to `${DOCUMENTS}`, and I thought others who base the profile on this one might benefit from the comment. For the same reason, I added potentially superfluous comments like

```
# Might break Pulse Audio
#machine-id
```

or

```
# Allow associated applications to play sound files
#nosound
```

etc/profile-a-l/bcompare.profile
tredondo: I've kept such lines because to me it was surprising to realize that `xdg` was related to `${DOCUMENTS}`, and I thought others who base the profile on this one might benefit from the comment. For the same reason, I added potentially superfluous comments like

```
# Might break Pulse Audio
#machine-id
```

or

```
# Allow associated applications to play sound files
#nosound
```

etc/profile-a-l/bcompare.profile
rusty-snake: `nodvd`: `/dev/sr0`, `/dev/sr1`, ...
`novideo`: `/dev/video0`, `/dev/video1`, ...

rusty-snake: > Launching VLC from the context menu of an MP3 does play it just fine with machine-id enabled, on my KDE neon 5.21. But I've commented machine-id anyway.

Second time I hear that PA works with `machine-id`. Anyway it was broken for years and until someone can confirm this for Debian/OpenSUSE Leap over Fedora to Arch/OpenSUSE Tumbleweed we should should comment it.

> On a related note, despite novideo, VLC launched from the context menu of an mp4, does play it.

Because `novideo` is about your webcam

> Also, should nodvd be uncommented? Is that a multimedia API of some sort, so if the user wants to compare files from a DVD, nodvd wouldn't have an effect because the DVD would be mounted via /media/ ?

DVD are usually accessed via gvfs/gio/kio/... which escape the sandbox via D-Bus.

etc/profile-a-l/bcompare.profile
rusty-snake: I meant to use this message for the others (disable-programs.inc, wusc, ..) too.

etc/profile-a-l/bcompare.profile
tredondo: That seems a bit arbitrary... disable all shells, but then enable back `bash`, `dash`, and `sh`, but not the much more popular `zsh`. Why bother disabling shells then?

etc/profile-a-l/bcompare.profile
tredondo: Sent a [PR re. `novideo` become about webcams](https://github.com/netblue30/firejail/pull/4100).

tredondo:
@rusty-snake / others, can we have a look-over and see what *really* needs to be done to merge this PR?

I'm somewhat partial to the comments I left above disable-xdg and disable-shell, but if I must, I can remove them.

netblue30:
Merged, thanks!

etc/profile-a-l/bcompare.profile
rusty-snake: zsh is popular as interactive shell yes, but not to write script in it as you can not assume it's installed on system. Usually you write you scripts for `/bin/sh` which should be present on any POSIX compatible system or `/bin/bash` which should be present on the most GNU/Linux systems. `allow-bin-sh.inc` does exactly what it names says, it allows `/bin/sh`. `/bin/sh` is a symlink to `/bin/bash` on the most system resp. `/bin/dash` on Debian/Ubuntu. Therefore they need a `noblacklist` too.

PR #4116, opened 2021-03-19
Update vmware.profile & dbus-policy for amarok

Neo00001:
should I create `mkdir` entry for `/var/lib/vmware`?

### DIFF
--- a/etc/profile-a-l/amarok.profile
+++ b/etc/profile-a-l/amarok.profile
@@ -34,10 +34,16 @@ private-dev
 # private-etc alternatives,asound.conf,ca-certificates,crypto-policies,machine-id,pki,pulse,ssl
 private-tmp
 
+# If you ain't on kde-plasma you need to uncomment the following
 dbus-user filter
 dbus-user.own org.kde.amarok
+#dbus-user.own org.kde.kded
+#dbus-user.own org.kde.klauncher
+# The following one is needed for keyboard shortcuts even on kde. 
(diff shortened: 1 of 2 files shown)


glitsj16:
> should I create mkdir entry for /var/lib/vmware?

The mkdir and mkfile options only apply to paths under ${HOME} AFAIK.

rusty-snake:
manpage:

> Create  a  directory in user home, under /tmp, or under /run/user/<UID> before the sandbox is started.

rusty-snake (commented):
amarok: Did you test the tray-icon? If not it maybe needs `dbus-user.own org.kde.*`.
vmware: agree to @glitsj16

Neo00001:
> It might need `writable-var` though

should I add `writable-var`?



> amarok: Did you test the tray-icon? If not it maybe needs `dbus-user.own org.kde.*`

yep.tray-icon was working smoothly.no need for `dbus-user.own org.kde.*`. btw tested on Arch kde, if you think other de may need this then I will add it.

Neo00001:
> should I add `writable-var`?

after going through there their documentation it seems `write` permission is needed.So adding it.

Neo00001:
amarok's dbus-policy checked on gnome, will check on xfce later.Afaik, xfce's implementation is similar to gnome if not the same.

netblue30:
> Should I merge this?

Absolutely! When you are done, you just merge it.

Issue #4158, opened 2021-04-02
Dunst notifications with Signal-Desktop not working


vwheezy22:
**Bug and expected behavior**
- After running `sudo firecfg` to create profiles for my applications and running signal-desktop, notifications through dunst no longer work.
- Expected a notification to pop up. 

**No profile and disabling firejail**
- Running signal-desktop under `--noprofile` and through `/usr/bin/signal-desktop` exhibits the correct behavior.

**Reproduce**
1. Run in bash `firejail signal-desktop`
2. Receive a notification through Signal
3. Observe lack of notification by dunst

**Environment**
 - 5.11.11-arch1-1
 - 0.9.64.4

**Additional context**
Tried overriding the profile with some dbus options knowing that dunst uses it (I think).
```
ignore private-tmp
ignore nodbus
dbus-system.talk org.freedesktop.Notifications
include /etc/firejail/signal-desktop.profile
```
I have a feeling I just don't know enough about DBus or firejail itself to resolve this myself. Sorry if this belongs in questions and thank you!

<details><summary> debug output </summary>

```
Reading profile /home/vwheezy/.config/firejail/signal-desktop.profile
Autoselecting /bin/bash as shell
Building quoted command line: 'signal-desktop'
Command name #signal-desktop#
Found signal-desktop.profile profile in /home/vwheezy/.config/firejail directory
Reading profile /etc/firejail/signal-desktop.profile
Reading profile /etc/firejail/electron.profile
Found electron.profile profile in /etc/firejail directory
Reading profile /etc/firejail/disable-common.inc
Found disable-common.inc profile in /etc/firejail directory
Reading profile /etc/firejail/disable-devel.inc
Found disable-devel.inc profile in /etc/firejail directory
Reading profile /etc/firejail/disable-exec.inc
Found disable-exec.inc profile in /etc/firejail directory
Reading profile /etc/firejail/disable-interpreters.inc
Found disable-interpreters.inc profile in /etc/firejail directory
Reading profile /etc/firejail/disable-passwdmgr.inc
Found disable-passwdmgr.inc profile in /etc/firejail directory
Reading profile /etc/firejail/disable-programs.inc
Found disable-programs.inc profile in /etc/firejail directory
Reading profile /etc/firejail/disable-xdg.inc
Found disable-xdg.inc profile in /etc/firejail directory
Reading profile /etc/firejail/whitelist-common.inc
Found whitelist-common.inc profile in /etc/firejail directory
[...]

PR #4161, opened 2021-04-04
allow notifications + comment fixes

glitsj16:
Inspired by #4158.

Regarding the comment:
```
#ignore private-cache --> signal-desktop doesn't use ~/.cache
```
IMO we can drop these. Both `whitelist-runuser-common.inc` and `whitelist-usr-share-common.inc` are already getting included via electron.profile and that seems to work as expected. On my Arch Linux box signal-desktop doesn't use ~/.cache so it should be safe to keep `private-cache` as well.

### DIFF
--- a/etc/profile-m-z/signal-desktop.profile
+++ b/etc/profile-m-z/signal-desktop.profile
@@ -19,10 +19,9 @@ read-only ${HOME}/.mozilla/firefox/profiles.ini
 mkdir ${HOME}/.config/Signal
 whitelist ${HOME}/.config/Signal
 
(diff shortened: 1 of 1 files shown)


rusty-snake:
Is `ignore private-tmp` required for notifications?

> Regarding the comment:
> ```
> # Disabled until someone reported positive feedback
> #include whitelist-runuser-common.inc
> #include whitelist-usr-share-common.inc
> ```

The comment is `ignore include …`.

rusty-snake:
Is `private-etc groups,passwd` required for anything? `Error getpwuid: main.c:236 init_cfg: Success` is in @vwheezy22 debug output because of the firecfg symlink (`firejail --debug signal-desktop` vs. `firejail --debug /usr/bin/signal-desktop`).

glitsj16:
> Is ignore private-tmp required for notifications?

Not sure. I don't have a smartphone to fully test the app to begin with :D...

PR #4164, opened 2021-04-04
Update Librewolf profile and Add Sway profile

vnepogodin:
Parent is shutting down then including firefox-common-addons.profile
[error.log](https://github.com/netblue30/firejail/files/6255116/error.log)

### DIFF
--- a/etc/profile-a-l/librewolf.profile
+++ b/etc/profile-a-l/librewolf.profile
@@ -31,7 +31,7 @@ include whitelist-usr-share-common.inc
 
 # librewolf requires a shell to launch on Arch. We can possibly remove sh though.
 # Add the next line to your librewolf.local to enable private-bin.
-#private-bin bash,dbus-launch,dbus-send,env,librewolf,python*,sh,which
+#private-bin dbus-launch,dbus-send,librewolf,sh
 # Add the next line to your librewolf.local to enable private-etc. Note
 # that private-etc must first be enabled in firefox-common.local.
 #private-etc librewolf
--- a/etc/profile-a-l/librewolf.profile
+++ b/etc/profile-a-l/librewolf.profile
@@ -31,7 +31,7 @@ include whitelist-usr-share-common.inc
 
 # librewolf requires a shell to launch on Arch. We can possibly remove sh though.
 # Add the next line to your librewolf.local to enable private-bin.
-#private-bin dbus-launch,dbus-send,librewolf,sh
+#private-bin bash,dbus-launch,dbus-send,env,librewolf,python*,sh,which
 # Add the next line to your librewolf.local to enable private-etc. Note
 # that private-etc must first be enabled in firefox-common.local.
 #private-etc librewolf
--- a/etc/profile-a-l/librewolf.profile
+++ b/etc/profile-a-l/librewolf.profile
@@ -14,8 +14,7 @@ mkdir ${HOME}/.librewolf
 whitelist ${HOME}/.cache/librewolf
 whitelist ${HOME}/.librewolf
 
-# Uncomment (or add to librewolf.local) the following lines if you want to
-# use the migration wizard.
+# Add the next lines to your librewolf.local if you want to use the migration wizard.
 #noblacklist ${HOME}/.mozilla
 #whitelist ${HOME}/.mozilla
 
@@ -30,11 +29,11 @@ whitelist /usr/share/mozilla
 whitelist /usr/share/webext
 include whitelist-usr-share-common.inc
 
-# librewolf requires a shell to launch on Arch.
-#private-bin bash,dbus-launch,dbus-send,env,librewolf,sh,which
-# Fedora use shell scripts to launch librewolf, at least this is required
-#private-bin basename,bash,cat,dirname,expr,false,librewolf,getenforce,ln,mkdir,pidof,restorecon,rm,rmdir,sed,sh,tclsh,true,uname
-# private-etc must first be enabled in firefox-common.profile
+# librewolf requires a shell to launch on Arch. We can possibly remove sh though.
+# Add the next line to your librewolf.local to enable private-bin.
+#private-bin bash,dbus-launch,dbus-send,env,librewolf,python*,sh,which
+# Add the next line to your librewolf.local to enable private-etc. Note
+# that private-etc must first be enabled in firefox-common.local.
 #private-etc librewolf
 
 dbus-user filter
--- a/etc/profile-a-l/librewolf.profile
+++ b/etc/profile-a-l/librewolf.profile
@@ -19,16 +19,40 @@ whitelist ${HOME}/.librewolf
 #noblacklist ${HOME}/.mozilla
 #whitelist ${HOME}/.mozilla
 
+# Uncomment or put in your librewolf.local one of the following whitelist to enable KeePassXC Plugin
+# NOTE: start KeePassXC before Librewolf and keep it open to allow communication between them
+#whitelist ${RUNUSER}/kpxc_server
+#whitelist ${RUNUSER}/org.keepassxc.KeePassXC.BrowserServer
+
 whitelist /usr/share/doc
-whitelist /usr/share/gnome-shell/search-providers/firefox-search-provider.ini
 whitelist /usr/share/gtk-doc/html
 whitelist /usr/share/mozilla
 whitelist /usr/share/webext
 
-# librewolf requires a shell to launch on Arch. We can possibly remove sh though.
-#private-bin bash,dbus-launch,dbus-send,env,librewolf,python*,sh,which
+# librewolf requires a shell to launch on Arch.
+#private-bin bash,dbus-launch,dbus-send,env,librewolf,sh,which
+# Fedora use shell scripts to launch librewolf, at least this is required
+#private-bin basename,bash,cat,dirname,expr,false,librewolf,getenforce,ln,mkdir,pidof,restorecon,rm,rmdir,sed,sh,tclsh,true,uname
 # private-etc must first be enabled in firefox-common.profile
 #private-etc librewolf
+
+dbus-user filter
+# Uncomment or put in your librewolf.local to enable native notifications.
+#dbus-user.talk org.freedesktop.Notifications
+# Uncomment or put in your librewolf.local to allow to inhibit screensavers
+#dbus-user.talk org.freedesktop.ScreenSaver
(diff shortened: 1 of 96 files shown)


etc/profile-a-l/librewolf.profile
```diff
@@ -19,6 +19,8 @@ whitelist ${HOME}/.librewolf
 #noblacklist ${HOME}/.mozilla
 #whitelist ${HOME}/.mozilla
 
+include whitelist-usr-share-common.inc
```
rusty-snake, line 22: That's firefox' block, should we add some of them too?
https://github.com/netblue30/firejail/blob/ba2d8a8e9bebd274fa5aa554e792164a7981e110/etc/profile-a-l/firefox.profile#L22-L28

etc/profile-a-l/sway.profile
```diff
@@ -0,0 +1,17 @@
+# Firejail profile for Sway
+# Description: i3-compatible Wayland compositor 
+# This file is overwritten after every install/update
+# Persistent local customizations
+include sway.local
+# Persistent global definitions
+include globals.local
+
+# all applications started in sway will run in this profile
+noblacklist ${HOME}/.config/sway
```
rusty-snake, line 10: Add a `blacklist` for it in disable-common.inc.

IIRC sway uses `~/.config/i3` as fallback if there is no `~/.config/sway`. We should add `nobalcklist ${HOME}/.config/i3`.

rusty-snake (commented):
LGTM, some suggestions below.

Should we make sway redirect to i3?

vnepogodin:
No it's should have own folder `$HOME/.config/sway`

And what about `$HOME/.config/i3`?
Sway doesn't work completely well with i3 config

etc/profile-a-l/librewolf.profile
vnepogodin: no it's not needed

rusty-snake: It's not needed to start librewolf. But what is with other features. Does it support `/usr/share/webext`? Or do we want `/usr/share/doc`? (The questions are for anybody who has a opinion).

etc/profile-a-l/sway.profile
vnepogodin: done

BL4CKH47H4CK3R:
> LGTM, some suggestions below.
> 
> Should we make sway redirect to i3?

Sway is almost i3 the .config are also same the main diif is wayland and x11 but why need to redirect to i3. I think sway profile will also be cool. if not working then we can redirect to i3 for sure :+1:

rusty-snake:
> Sway is almost i3 

That's the reason.

etc/profile-a-l/librewolf.profile
glitsj16: I think we should add the relevant ones. AFAICT librewolf supports webext. So the only path NOT to whitelist would be /usr/share/firefox IMO.

glitsj16:
@rusty-snake I agree it would make sense to implement sway as a redirect profile. Perhaps the term 'redirect' caused confusion in this context for other posters.

BL4CKH47H4CK3R (commented):
@glitsj16, I know that desktop environment or windows manager doesn't matter when I am using firejail. But I see people are creating profile like for i3 which is redundant so I told. Even `archiso` is enough to run firejail :smile:

rusty-snake (dismissed):
Re-add `include whitelist-usr-share-common.inc`.

vnepogodin:
> Re-add `include whitelist-usr-share-common.inc`.

Why ?

rusty-snake:
TL;DR: It whitelist lots of **important** `/usr/share` folders.
**Always use `include {allow,whitelist}-common.inc` in PRs.**

The role of `{allow,whitelist}-common.inc` is to make maintaining profiles easier and cause less breakage. *Always* use them in PRs unless you can 100% guarantee that they are not required. The paths to some resources can differ between distros, others depend on the HW where it runs or the configuration. Or the version you run.  Have you tested that it works on every major distro, with any configuration, any DE, any HW? It's not enough that it can start there. Can it play audio? Is HTTPS working? What about spellchecking? Just the first line  (`whitelist /usr/share/alsa`) can you guarantee that not `whitelist`ing `/usr/share/alase`  does not break sound? With raw alsa?

vnepogodin:
> TL;DR: It whitelist lots of **important** `/usr/share` folders.
> 
> **Always use `include {allow,whitelist}-common.inc` in PRs.**
> 
> 
> 
> The role of `{allow,whitelist}-common.inc` is to make maintaining profiles easier and cause less breakage. *Always* use them in PRs unless you can 100% guarantee that they are not required. The paths to some resources can differ between distros, others depend on the HW where it runs or the configuration. Or the version you run.  Have you tested that it works on every major distro, with any configuration, any DE, any HW? It's not enough that it can start there. Can it play audio? Is HTTPS working? What about spellchecking? Just the first line  (`whitelist /usr/share/alsa`) can you guarantee that not `whitelist`ing `/usr/share/alase`  does not break sound? With raw alsa?

I agree but how about ?

https://github.com/netblue30/firejail/pull/4164#discussion_r606858628

rusty-snake:
(Privacy is out of the scope of firejail, it's for security).

> its not needed as it reveals lots of important /usr/share folders

It's not needed because it reveals important folders???!

> like /usr/share/fonts which can used for font fingerprinting 

This does not prevent font fingerprinting. Actually it makes the situation worse. If you `whitelist /usr/share/fonts` and don't install to much other fonts you have the same font FP as any other user of your distro. If you omit it and put on a gas mask you're even more unique. 

> and OS detection

Your OS can be detected anyway.
- Linux has an other TCP/IP-Stack then Window => Passive detection possible.
- Some math operations on floating point number has other results on Linux then Window. => Active detection via JavaScript
- Your scrollbar-width depends on your GTK-Theme and version => Active detection via JS or CSS possible.
- Your HTTP-UA contains your OS.
- Your JS-UA contains your OS, even with RFP.
- ...

> Linux and windows common font are not same so its a problem.

Masking your fonts still does not make them the same.

> Librewolf can launch

yes

> work perfectly without this options

no (I'm 90% sure it can not open HTTPS on some system without wusc)

rusty-snake:
BTW I don't think `whitelist /usr/share/gnome-shell/search-providers/firefox-search-provider.ini` (*firefox*-search-provider) is used.

etc/profile-a-l/librewolf.profile
```diff
@@ -19,16 +19,40 @@ whitelist ${HOME}/.librewolf
...
 whitelist /usr/share/doc
-whitelist /usr/share/gnome-shell/search-providers/firefox-search-provider.ini
 whitelist /usr/share/gtk-doc/html
 whitelist /usr/share/mozilla
 whitelist /usr/share/webext
 
-# librewolf requires a shell to launch on Arch. We can possibly remove sh though.
-#private-bin bash,dbus-launch,dbus-send,env,librewolf,python*,sh,which
+# librewolf requires a shell to launch on Arch.
+#private-bin bash,dbus-launch,dbus-send,env,librewolf,sh,which
```
rusty-snake, line 34: Based on https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=librewolf#n325 it's just `private-bin librewolf,sh` if anyone want to test ...

etc/profile-a-l/librewolf.profile
```diff
@@ -19,16 +19,40 @@ whitelist ${HOME}/.librewolf
...
+# Fedora use shell scripts to launch librewolf, at least this is required
+#private-bin basename,bash,cat,dirname,expr,false,librewolf,getenforce,ln,mkdir,pidof,restorecon,rm,rmdir,sed,sh,tclsh,true,uname
```
rusty-snake, line 36: I don't see a Fedora package of librewolf (https://librewolf-community.gitlab.io/install/) ... AppImage and Flatpak seem to be the only install option for (non hacker) fedora users ... remove this.

rusty-snake:
This PR has a a merge conflict, please rebase.

etc/profile-a-l/librewolf.profile
```diff
@@ -14,15 +14,45 @@ mkdir ${HOME}/.librewolf
...
+
+whitelist /usr/share/doc
+whitelist /usr/share/gtk-doc/html
+whitelist /usr/share/mozilla
+whitelist /usr/share/webext
+include whitelist-usr-share-common.inc
+
 # librewolf requires a shell to launch on Arch. We can possibly remove sh though.
+# Add the next line to your librewolf.local to enable private-bin.
 #private-bin bash,dbus-launch,dbus-send,env,librewolf,python*,sh,which
```
rusty-snake, line 34:
```suggestion
#private-bin bash,dbus-launch,dbus-send,env,librewolf,sh,which
```

rusty-snake (approved):

etc/profile-a-l/librewolf.profile
vnepogodin: I've had added before but it can't merge changes

rusty-snake: After a rebase (unlike a merge of master) it should work w/o conflicting. Rebasing and squashing some of the commits makes sense anyway if you look at the history. So if you can do an interactive rebase.

vnepogodin: done :smiley:

etc/profile-a-l/librewolf.profile
```diff
@@ -18,12 +18,41 @@ whitelist ${HOME}/.librewolf
...
+#whitelist ${RUNUSER}/org.keepassxc.KeePassXC.BrowserServer
+
+whitelist /usr/share/doc
+whitelist /usr/share/gtk-doc/html
+whitelist /usr/share/mozilla
+whitelist /usr/share/webext
+include whitelist-usr-share-common.inc
+
 # librewolf requires a shell to launch on Arch. We can possibly remove sh though.
 # Add the next line to your librewolf.local to enable private-bin.
```
rusty-snake, line 33: Maybe something like this (feel free to improve)?
```suggestion
# Add the next line to your librewolf.local to enable private-bin (Arch Linux).
```

BTW: For what does it need `dbus-launch`?

glitsj16 (approved):

rusty-snake (approved):

rusty-snake:
merged, thanks!

PR #4165, opened 2021-04-04
Encourage making overrides in *.local files

glitsj16:
We should try to inform users that the best place for persistent local customizations is in .local files. This PR refactors/streamlines comments to make that a bit more obvious.

### DIFF
--- a/etc/profile-a-l/archiver-common.profile
+++ b/etc/profile-a-l/archiver-common.profile
@@ -6,24 +6,19 @@ include archiver-common.local
 
 blacklist ${RUNUSER}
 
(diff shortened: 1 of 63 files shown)


rusty-snake:
Didn't read all rewrite but looks great. If this catched all `grep -i uncomment /etc/firejail/*` we can squash and merged this.

glitsj16:
> Didn't read all rewrite but looks great. If this catched all grep -i uncomment /etc/firejail/* we can squash and merged this.

@rusty-snake
There's around 40 other profiles I need to go through, these were the easier bit to track down and edit. I should have made that clear. Let's convert to draft until all files are properly dealt with. Thanks for having a look!

PR #4188, opened 2021-04-13
Create tmux.profile

rusty-snake:
requested in #1139 by @vatonbero


etc/profile-m-z/tmux.profile
```diff
@@ -0,0 +1,45 @@
...
+# Persistent global definitions
+include globals.local
+
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}
+
+noblacklist /tmp/tmux-*
+
+# include disable-common.inc
+include disable-devel.inc
```
reinerh, line 16: i think tmux is also often used by developers.

PR #4193, opened 2021-04-14
add passwd to private-etc

glitsj16:
Now we have the new quodlibet.profile I've tested its functionality on a setup where users only have exfalso, but not quodlibet. Both profiles are looking good, but I noticed the below error when firejailing exfalso:

```
...
/usr/lib/python3.9/site-packages/quodlibet/qltk/filesel.py:130: Warning: getpwuid_r(): failed due to unknown user id (1001)
...
```

Adding `passwd` to private-etc fixes this for uid's > 1000.

### DIFF
--- a/etc/profile-m-z/quodlibet.profile
+++ b/etc/profile-m-z/quodlibet.profile
@@ -60,7 +60,7 @@ tracelog
 private-bin exfalso,operon,python*,quodlibet,sh
 private-cache
 private-dev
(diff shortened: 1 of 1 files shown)


rusty-snake:
> Adding passwd to private-etc fixes this for uid's > 1000.

Has this anything to do with the UID? All `getpw*` can not work without `/etc/passwd` because they are the API for it.

glitsj16:
> Has this anything to do with the UID? All getpw* can not work without /etc/passwd because they are the API for it.

I stand corrected. Happy to realize that by adding passwd to private-etc the PR at least didn't break anything :-D

Issue #3991, opened 2021-02-15
private-etc doesn't allow subdirs


g-v-egidy:
**Bug and expected behavior**
I want to allow a program access to the directory /etc/pki/ca-trust/ and below, but not to /etc/pki/tls/private.
As the private keys are really sensitvie, I want not just rely on blacklisting, but remove the private keys fully from the mount with the private-etc option.

When I try to use this in a profile file:
```
private-etc alternatives,crypto-policies,hostname,hosts,ld.so.cache,resolv.conf,localtime,ssl,pki/ca-trust
```
I get the error:
```
Error: "pki/ca-trust" is an invalid filename
```

The manpage suggests that copying subdirs of /etc is allowed as the example given there:
```
Example:
              $ firejail --private-etc=group,hostname,localtime, \
              nsswitch.conf,passwd,resolv.conf,default/motd-news
```
uses default/motd-news.

**Environment**
 - CentOS 8.3 on x86_64.
 - firejail 0.9.64.4

smitsohu:
A simple workaround for now would be to switch from `private-etc` to `whitelist`. It is about as secure as long as you don't add a `writable-etc` option.
```
whitelist /etc/alternatives
whitelist /etc/crypto-policies
whitelist /etc/hostname
whitelist /etc/hosts
whitelist /etc/ld.so.cache
whitelist /etc/resolv.conf
whitelist /etc/localtime
whitelist /etc/ssl
whitelist /etc/pki/ca-trust
```

smitsohu:
Alternatively you can also layer `noblacklist` / `blacklist` on top of the private-etc:
```
noblacklist /etc/pki/ca-trust
blacklist /etc/pki/*
private-etc alternatives,crypto-policies,hostname,hosts,ld.so.cache,resolv.conf,localtime,ssl,pki
```

g-v-egidy:
Thanks for your quick reply.

I'm already using the blacklist solution for now. But it would be nice if private-etc could be improved to also allow the subdirs, because I think this is more secure.

brisad:
Hi! I recently found Firejail, and thought I'd start to help out a little. So I just took a look a this problem and wanted to see if I could fix it.

I am guessing here that the expected behavior would be that
```
private-etc pki/ca-trust
```
would yield `/etc/pki/ca-trust`.

`private-lib` on the other hand, which does support sub-directories, maps 
```
private-lib a/b/c
```
[...]

PR #4204, opened 2021-04-20
man: corrections regarding --private-FOO options

jsquyres:
Commit 0.9.60-1070-g40d3604f (40d3604f) updated the man pages with respect to
--private-opt, --private-etc, and --private-srv.  It was made after
testing firejail 0.9.52 (from Ubuntu 18.04).  However, it
unfortunately did not accurately reflect the the behavior of the
current HEAD at the time, because commit 0.9.56-rc1-14-ga9242301 (a9242301) had
previously slightly changed the behavior of these three options (after
0.9.52), and was released in 0.9.56.  The man pages changes made in
commit 40d3604f were therefore not entirely correct.

This commit updates the man pages to describe the behavior as
implemented in a9242301 (and is still the behavior as of the current
HEAD: 0.9.64-737-g937815ba (937815ba)).

Signed-off-by: Jeff Squyres <jsquyres@cisco.com>

### DIFF
--- a/src/man/firejail-profile.txt
+++ b/src/man/firejail-profile.txt
@@ -295,7 +295,9 @@ Use the options no3d, nodvd, nosound, notv, nou2f and novideo for additional res
 Build a new /etc in a temporary
 filesystem, and copy the files and directories in the list.
 The files and directories in the list must be expressed as relative to
-the /etc directory.
+the /etc directory, and must not contain the / character
+(e.g., /etc/foo must be expressed as foo, but /etc/foo/bar --
+expressed as foo/bar -- is disallowed).
 All modifications are discarded when the sandbox is closed.
 #ifdef HAVE_PRIVATE_HOME
 .TP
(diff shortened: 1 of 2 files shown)


src/man/firejail-profile.txt
```diff
@@ -295,7 +295,9 @@ Use the options no3d, nodvd, nosound, notv, nou2f and novideo for additional res
 Build a new /etc in a temporary
 filesystem, and copy the files and directories in the list.
 The files and directories in the list must be expressed as relative to
-the /etc directory.
+the /etc directory, and must not contain the / character
+(e.g., /etc/foo must be expressed as foo, but /etc/foo/bar --
+expressed as foo/bar -- is disallowed).
```
rusty-snake, line 300: For reference: #3991

rusty-snake (approved):
BTW: There are even more disallowed chars. However they are disallowed for all file-system paths.

https://github.com/netblue30/firejail/blob/2609e5cf0b82222d83637fa1bd2538645999f04d/src/firejail/macros.c#L318

src/man/firejail-profile.txt
jsquyres: Would you like me to amend the text to include the full list of disallowed chars and/or reference #3991?

jsquyres: FWIW: I specifically mentioned `/` because it seemed the easiest way to say "you can only have a file or dir that's directly in `/etc` / `/opt` / `/srv` -- you can't have anything that's in a subdir of `/etc` / `/opt` / `/srv`.  If there's a better way to express that, I'm open to alternate wording.

rusty-snake: > Would you like me to amend the text to include the full list of disallowed chars and/or reference #3991?

No. When #3991 gets implemented we can update the manpage. And `Error: "foo;bar" is an invalid filename: rejected character: ";"` is one of the most helpful messages we have.

glitsj16:
@rusty-snake Can we merge this? Or do you have any outstanding requests for @jsquyres?

rusty-snake:
Can be merged.

rusty-snake:
thanks

glitsj16:
@rusty-snake Great, I'll take care of it. Thanks!

rusty-snake:
#4215 and we can revert the for private-etc ...

PR #3688, opened 2020-10-23
rework chromium

rusty-snake:
+ 516d0811 has removed fundamental security features.
   (remove caps.drop=all, nonewprivs, noroot, seccomp, protocol; add
caps.keep)  Though this is only necessary if running under a kernel which
disallow   unprivileged userns clones. Arch's linux-hardened and debian kernel
are   patched accordingly. Arch's linux and linux-lts kernels support this
   restriction via sysctk (kernel.unprivileged_userns_clone=0) as users
opt-in.   Other kernels such as mainline or fedora/redhat always support
unprivileged   userns clone and have no sysctl parameter to disable it. Debian and
Arch   users can enable it with 'sysctl kernel.unprivileged_userns_clone=1'.
   This commit adds a chromium-common-hardened.inc which can be included
in   chromium-common to enhance security of chromium-based programs.

 + chromium-common.profile: add private-cache

 + chromium-common.profile: add wruc and wusc, but disable it for the   following   profiles until tested. tests welcome.

    - [ ] bnox, dnox, enox, inox, snox
    - [ ] brave
    - [ ] flashpeak-slimjet
    - [ ] google-chrome, google-chrome-beta, google-chrome-unstable
    - [ ] iridium
    - [ ] min
    - [ ] opera, opera-beta

 + move vivaldi-snapshot paths from vivaldi-snapshot.profile to vivaldi.
   /usr/bin/vivaldi is a symlink to /etc/alternatives/vivaldi which can be  vivaldi-stable, vivaldi-beta or vivaldi-snapshot. vivaldi-snapshot.profile  missed also some features from vivaldi.profile, solve this by making it  redirect to vivaldi.profile. TODO: exist new paths such as .local/lib/vivaldi  also for vivaldi-snapshot?

 + ~Move drm-relaktions in vivaldi.profile behind BROWSER_ALLOW_DRM.~

 + create chromium-browser-privacy.profile (closes #3633)

PR #4207, opened 2021-04-22
[WIP] refactor vivaldi profiles

glitsj16:
Minor changes to avoid creating unneeded directories and files depending on the specific vivaldi setup.

IMO we can harden the vivaldi profiles a bit more:
- keep apparmor intact and add an opt-in rule to /etc/apparmor.d/local/firejail-default (cfr. brave and torbrowser-launcher);
- put some of the DRM-related options (see snippet below) behind a `?BROWSER_ALLOW_DRM:` conditional (like we do in other web browser profiles).

I've only recently started to experiment with Vivaldi, so I might overlook something. Any opinions/suggestions on this are very welcome.

```
# Allow HTML5 Proprietary Media & DRM/EME (Widevine)
ignore apparmor
ignore noexec /var
noblacklist /var/opt
whitelist /var/opt/vivaldi
writable-var
```

### DIFF
--- a/etc/profile-m-z/vivaldi-snapshot.profile
+++ b/etc/profile-m-z/vivaldi-snapshot.profile
@@ -3,5 +3,17 @@
 # Persistent local customizations
 include vivaldi-snapshot.local
 
+noblacklist ${HOME}/.cache/vivaldi-snapshot
+noblacklist ${HOME}/.config/vivaldi-snapshot
+
+mkdir ${HOME}/.cache/vivaldi-snapshot
+mkdir ${HOME}/.config/vivaldi-snapshot
+whitelist ${HOME}/.cache/vivaldi-snapshot
+whitelist ${HOME}/.config/vivaldi-snapshot
+
+# Add the next line to your vivaldi-snapshot.local to enable private-bin.
+# Note: this first needs to be enabled in vivaldi.local too.
+#private-bin vivaldi-snapshot
+
 # Redirect
 include vivaldi.profile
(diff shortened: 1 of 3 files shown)


rusty-snake:
> put some of the DRM-related options (see snippet below) behind a ?BROWSER_ALLOW_DRM: conditional (like we do in other web browser profiles).

I already consider this in #3688. Because vivaldi shows a error pop-up direct after starting if we block widevine (instead of when a website wants to use widevine), I decided to not do so. Likely we would get extra issues because of this (error-)message, even if people does not even use DRM

> Minor changes to avoid creating unneeded directories and files depending on the specific vivaldi setup
> so I might overlook something.

I added them in #3688. Read and undo.

> vivaldi sync needs full D-Bus access

Strictly speak: It can work with filter but nobody has figured out the names which needs talk rules yet.

glitsj16:
@rusty-snake Thanks for refering to #3688, that clears up some of the gripes I encountered when actually using Vivaldi! Looking at this again after your suggestions made me realize the only thing I'd keep is the AppArmor stuff. All the rest is more suited for a personal vivaldi.local. I only need a chromium-based browser for occasional DRM content play - which I prefer to completely disable in my main FF browser setup. On Arch the Widevine files are installed under /usr and don't need any special treatment of stuff in ${HOME}, like the `read-write ${HOME}/.local/lib/vivaldi`. For now I prefer to close this PR and come back to it later, if ever - I don't like these chromium forks much anyway.

PR #4215, opened 2021-04-25
Add support for subdirs in private-etc

### LINKED ISSUE #3991: private-etc doesn't allow subdirs

brisad:
Gosh it's been a while since I wrote C, and this solution turned out to use more pointer-manipulation that I'd hoped to. I wanted to avoid allocations, but this solution... I don't like it.

But since I already wrote it, and it seems to work, I thought I'd add a PR and you can take a look to see if it's anything to keep.

Fixes #3991

### DIFF
--- a/src/firejail/fs_etc.c
+++ b/src/firejail/fs_etc.c
@@ -76,6 +76,44 @@ void fs_machineid(void) {
 	}
 }
 
+// Duplicate directory structure from src to dst by creating empty directories.
+// The paths _must_ be identical after their respective prefixes.
+// When finished, dst will point to the target directory. That is, if
+// it starts out pointing to a file, it will instead be truncated so
+// that it contains the parent directory instead.
+static void build_dirs(char *src, char *dst, size_t src_prefix_len, size_t dst_prefix_len) {
+	char *p = src + src_prefix_len + 1;
+	char *q = dst + dst_prefix_len + 1;
+	char *r = dst + dst_prefix_len;
+	struct stat s;
+	bool last = false;
+	*r = '\0';
+	for (; !last; p++, q++) {
+		if (*p == '\0') {
+			last = true;
+		}
+		if (*p == '\0' || (*p == '/' && *(p - 1) != '/')) {
+			// We found a new component of our src path.
+			// Null-terminate it temporarily here so that we can work
+			// with it.
+			*p = '\0';
+			if (stat(src, &s) == 0 && S_ISDIR(s.st_mode)) {
+				// Null-terminate the dst path and undo its previous
+				// termination.
+				*q = '\0';
+				*r = '/';
+				r = q;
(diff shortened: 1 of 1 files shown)


src/firejail/fs_etc.c
```diff
@@ -76,6 +76,44 @@ void fs_machineid(void) {
 	}
 }
 
+// Duplicate directory structure from src to dst by creating empty directories.
+// The paths _must_ be identical after their respective prefixes.
+// When finished, dst will point to the target directory. That is, if
+// it starts out pointing to a file, it will instead be truncated so
+// that it contains the parent directory instead.
+static void build_dirs(char *src, char *dst, size_t src_prefix_len, size_t dst_prefix_len) {
+    char *p = src + src_prefix_len + 1;
```
kmk3, line 85: This file uses tabs for indentation.

brisad: Whoops, completely missed that. Thanks for the head up.

kmk3: > Whoops, completely missed that. Thanks for the heads up.

No problem.  So could you fix the indentation, amend the commit and force-push?

I'd suggest a command to do it automatically, but the proposed `build_dirs`
function is indented using 4 spaces and the changes in `duplicate` are indented
with a mix of 4 and 8 spaces.

kmk3: > Sure! I just pushed the update.
>
> Ah, yes, I even mixed tab length with my spaces, so I should've noticed that
> inconsistency :)

Thanks.

> Thanks for reviewing. As I wrote above, I'm not really happy with this one,
> so I'm all ears for suggestions for improvements. Don't have anything against
> rewriting it, perhaps using sprintf's for building up paths instead.

No problem.  I'm not really familiar with firejail's C code base, so based on
the output of `tig --no-merges master -- '*.h' '*.c'`, I think someone like
@smitsohu would be of more help in that regard.

netblue30:
Merged in, thanks!

PR #4219, opened 2021-04-26
Some minor changes

Neo00001:


etc/profile-a-l/kcalc.profile
```diff
@@ -41,13 +47,19 @@ nou2f
...
+private-cache
 private-dev
+private-etc alternatives,fonts,ld.so.cache,locale,locale.conf
 # private-lib - problems on Arch
 private-tmp
 
 dbus-user none
 dbus-system none
+
+#memory-deny-write-execute
```
rusty-snake, line 65: If mdwe breaks it does not need to be added.

rusty-snake (commented):
LGTM

etc/profile-a-l/kcalc.profile
Neo00001: on arch it works smoothly, have to test on other distros.

PR #4170, opened 2021-04-06
steam: some more games added

matthew-cline:
Games added:
* Don't Starve
* Dungeons of Dredmor
* Epic
* Loop Hero
* Pillars of Eternity I
* Rogue Legacy I
* Slay the Spire modding
* Steam World Dig I & II

PR #4225, opened 2021-04-30
steam.profile: fix rogue legacy paths and syntax

kmk3:
Due to using globbing on mkdir, the current version causes this:

@davidebeatrici commented on 2021-04-23[1]:

> ```
> Error: "${HOME}/.local/share/RogueLegacy*" is an invalid filename: rejected character: "*"
> ```

Added on commit a603d4d ("steam: some more games added") / PR #4170.

The wildcard was used because Rogue Legacy apparently looks up multiple
different paths for the config and also for the data[1][2][3]:

1. ~/.config/RogueLegacy
2. ~/.config/RogueLegacyStorageContainer
3. ~/.local/share/RogueLegacy
4. ~/.local/share/RogueLegacyStorageContainer

The ones containing "RogueLegacyStorageContainer" appear to be legacy
paths (i.e.: paths which are only created by older versions of Rogue
Legacy)[2].

So replace all globs with the full paths because:

* The paths are known a priori (unlike, say, `/var/lib/libpcre*`)
* There aren't too many of them

And use only the non-legacy paths on mkdir.  Besides mirroring what the
current version of Rogue Legacy does (and avoiding the creation of
unnecessary dirs), this is also done because _if_ the following applies
(i.e.: this was not tested):

* legacy paths take precedence over non-legacy paths
* the first path clobbers the other ones (i.e.: rather than "merge")
* save data exists in a non-legacy path (i.e.: path 3 in this case)
* firejail creates all 4 paths

Then it would make the newly-created and empty path 4 clobber the
non-legacy path 3 and thus make it seem like no save files exist.  This
would persist even if steam is run without firejail afterwards, as the
empty directory would still be there.  Losing (or appearing to lose)
game saves can be very unfortunate, so create just the non-legacy paths
to avoid confusion.

[1] #4170 (comment)
[2] https://steamcommunity.com/app/241600/discussions/1/846957366713233279/
[3] https://www.pcgamingwiki.com/wiki/Rogue_Legacy#Game_data

---

Original message below (for the thread to make sense):

---

steam.profile: remove rogue legacy to fix syntax

Remove the mkdir with globbing, which causes this:

@davidebeatrici commented on 2021-04-23[1]:

> ```
> Error: "${HOME}/.local/share/RogueLegacy*" is an invalid filename: rejected character: "*"
> ```

Added on commit bd7ad371e ("steam: also added paths to
disable-programs.inc") / PR #4170.

The wildcard was used because Rogue Legacy apparently looks up multiple
different paths for the config and also for the data[1][2][3]:

1. ~/.config/RogueLegacy
2. ~/.config/RogueLegacyStorageContainer
3. ~/.local/share/RogueLegacy
4. ~/.local/share/RogueLegacyStorageContainer

The ones containing "RogueLegacyStorageContainer" appear to be legacy
paths (i.e.: paths which are only created by older Rogue Legacy
versions)[2].  So _if_ the following applies (i.e.: it was not tested):

* legacy paths take precedence over non-legacy paths
* the first path clobbers the other ones (i.e.: rather than "merge")
* save data exists in a non-legacy path (i.e.: path 3 in this case)
* firejail creates all 4 paths (as whitelisting requires existing paths)

Then it would make the newly-created and empty path 4 clobber the
non-legacy path 3 and thus make it seem like no save files exist.  This
would persist even if steam is run without firejail afterwards, as the
empty directory would still be there.  Losing (or appearing to lose)
game saves can be very unfortunate, so just ignore all the Rogue Legacy
paths for now to avoid confusion.

Note: This is a problem whenever a program supports multiple clobbering
paths and mkdir/mkfile + whitelisting is involved.

[1] https://github.com/netblue30/firejail/pull/4170#issuecomment-825405930
[2] https://steamcommunity.com/app/241600/discussions/1/846957366713233279/
[3] https://www.pcgamingwiki.com/wiki/Rogue_Legacy#Game_data

### DIFF
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -138,6 +138,7 @@ blacklist ${HOME}/.config/Rambox
 blacklist ${HOME}/.config/Riot
 blacklist ${HOME}/.config/Rocket.Chat
 blacklist ${HOME}/.config/RogueLegacy
+blacklist ${HOME}/.config/RogueLegacyStorageContainer
 blacklist ${HOME}/.config/Signal
 blacklist ${HOME}/.config/Sinew Software Systems
 blacklist ${HOME}/.config/Slack
@@ -612,7 +613,8 @@ blacklist ${HOME}/.local/share/QGIS
 blacklist ${HOME}/.local/share/QMediathekView
 blacklist ${HOME}/.local/share/QuiteRss
 blacklist ${HOME}/.local/share/Ricochet
-blacklist ${HOME}/.local/share/RogueLegacy*
+blacklist ${HOME}/.local/share/RogueLegacy
+blacklist ${HOME}/.local/share/RogueLegacyStorageContainer
 blacklist ${HOME}/.local/share/Shortwave
 blacklist ${HOME}/.local/share/Steam
 blacklist ${HOME}/.local/share/SteamWorldDig
--- a/etc/profile-m-z/steam.profile
+++ b/etc/profile-m-z/steam.profile
@@ -10,6 +10,7 @@ noblacklist ${HOME}/.config/Epic
 noblacklist ${HOME}/.config/Loop_Hero
 noblacklist ${HOME}/.config/ModTheSpire
 noblacklist ${HOME}/.config/RogueLegacy
+noblacklist ${HOME}/.config/RogueLegacyStorageContainer
 noblacklist ${HOME}/.killingfloor
 noblacklist ${HOME}/.klei
 noblacklist ${HOME}/.local/share/3909/PapersPlease
@@ -22,7 +23,8 @@ noblacklist ${HOME}/.local/share/feral-interactive
 noblacklist ${HOME}/.local/share/IntoTheBreach
 noblacklist ${HOME}/.local/share/Paradox Interactive
 noblacklist ${HOME}/.local/share/PillarsOfEternity
-noblacklist ${HOME}/.local/share/RogueLegacy*
+noblacklist ${HOME}/.local/share/RogueLegacy
+noblacklist ${HOME}/.local/share/RogueLegacyStorageContainer
 noblacklist ${HOME}/.local/share/Steam
 noblacklist ${HOME}/.local/share/SteamWorldDig
 noblacklist ${HOME}/.local/share/SteamWorld Dig 2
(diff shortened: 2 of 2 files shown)


rusty-snake:
1. I would keep the `(no)blacklist` in any case.
2. We can `whitelist` w/o `mkdir` (ideally with a comment to prevent future "fixes"). This `whitelist` the folder if it exist and ignores it otherwise.

kmk3:
> 1. I would keep the `(no)blacklist` in any case.

Good catch.

> 2. We can `whitelist` w/o `mkdir` (ideally with a comment to prevent future
>    "fixes"). This `whitelist` the folder if it exist and ignores it
>    otherwise.

So it is possible to `whitelist` a non-existing directory now?  When was that
implemented?

I remember that not too long ago using e.g.: `whitelist ${HOME}/doesnotexist`
would cause firejail to simply die on parsing.

rusty-snake:
> So it is possible to `whitelist` a non-existing directory now? When was that
> implemented?
> 
> I remember that not too long ago using e.g.: `whitelist ${HOME}/doesnotexist`
> would cause firejail to simply die on parsing.

wc has `whitelist` w/o `mkdir` of paths that not exist on any system by default for ever.
https://github.com/netblue30/firejail/commit/347dbc8cff7f0e17445cd547268455a09cba1f16#diff-a3e4440e096ece98571dc3c76b3ae6f71f6f42a126b4c02a27de5aaac9ef0d23

> So it is possible to `whitelist` a non-existing directory now?

`whitelist` a non-existing directory in the meaning of ignore this `whitelist`.

kmk3:
> > So it is possible to `whitelist` a non-existing directory now? When was
> > that implemented?
> >
> > I remember that not too long ago using e.g.: `whitelist
> > ${HOME}/doesnotexist` would cause firejail to simply die on parsing.
> 
> wc has `whitelist` w/o `mkdir` of paths that not exist on any system by
> default for ever.
> [347dbc8#diff-a3e4440e096ece98571dc3c76b3ae6f71f6f42a126b4c02a27de5aaac9ef0d23](https://github.com/netblue30/firejail/commit/347dbc8cff7f0e17445cd547268455a09cba1f16#diff-a3e4440e096ece98571dc3c76b3ae6f71f6f42a126b4c02a27de5aaac9ef0d23)
> 
> > So it is possible to `whitelist` a non-existing directory now?
> 
> `whitelist` a non-existing directory in the meaning of ignore this
> `whitelist`.

Damn.  I was going to say that maybe the error that I was thinking about
happened because I whitelisted a path that is not supported (e.g.:
`whitelist /foobar`), but even that does not error out.

I guess I confused this with another whitelist-related issue then.

But it's good to know that non-existing paths are supported; this will be very
helpful on a branch that I was working on a while ago.

---

Force-pushed.  Changes:

* mkdir the non-legacy paths
* Use the 4 paths everywhere else

kmk3:
> 2. We can `whitelist` w/o `mkdir` (ideally with a comment to prevent future "fixes").

Now I see what you mean by this.  I'll add a comment on the profile.

rusty-snake (approved):

Issue #3236, opened 2020-02-22
private-lib + python is broken


rusty-snake:
`firejail --noprofile '--private-lib=*' python3` does not work because python need files from /usr/lib/python3.7 and /usr/lib64/python3.7 but private-lib simply mounts the private-lib over both /usr/lib and /usr/lib64 with the python3.7 directory from /usr/lib.

Issue #3189, opened 2020-01-27
Sharing data through /run subdirs


jonleivent:
The fact that --private-tmp does not create a private /run should be documented better.  It seems that unless one blacklists or read-only's subdirs of /run (careful not to do so for /run/firejail), then there are always ways that a firejailed app can share data through items in /run with other apps outside the jail, regardless of the other firejail options used. I think many would expect --private-tmp to cover /run, since /run is just a better organized alternative to /tmp.  

Here's a scenario: A user sets up two firejails for two apps.  One jailed app given read-only access to sensitive data but is cut off from all network access.  The other is given network access but is cut off from reading sensitive data.  Those two apps can communicate through the /run/user/1000/pulse dir, for instance, and tunnel the sensitive data out to the network.

Perhaps there should be a --private-run option?

smitsohu:
As `/run/user/$UID` is the only user-writable place below `/run` usually, you can solve this as follows since 0.9.60

```
whitelist ${RUNUSER}/bus
whitelist ${RUNUSER}/gnupg
whitelist ${RUNUSER}/pulse
read-only ${RUNUSER}/gnupg
read-only ${RUNUSER}/pulse
```

Assuming that `/run/shm`, if it exists, is not a directory but a symbolic link to `/dev/shm` (which is the most common setup).

smitsohu:
> Perhaps there should be a --private-run option?

Good question! Over time there is likely going to be whitelist support for /run, so we could have a whitelist-run-common.inc at some point, similar to what we are doing now in whitelist-var-common.inc or whitelist-usr-share-common.inc.

rusty-snake:
> As `/run/user/$UID` is the only user-writable place below `/run` usually, you can solve this as follows since 0.9.60
> 
> ```
> whitelist ${RUNUSER}/bus
> whitelist ${RUNUSER}/gnupg
> whitelist ${RUNUSER}/pulse
> read-only ${RUNUSER}/gnupg
> read-only ${RUNUSER}/pulse
> ```

For Wayland: `whitelist ${RUNUSER}/wayland-?`

smitsohu:
> As /run/user/$UID is the only user-writable place below /run usually, you can solve this as follows since 0.9.60

We have also /run/lock, but Firejail mounts a tmpfs on it as long as there is a symbolic link /var/lock -> /run/lock

jonleivent:
Thanks for the help!   I also found that it was possible to blacklist many of the items within /run.

[...]

PR #4229, opened 2021-05-01
Whitelist2

### LINKED ISSUE #2041: [enhancement] Allow whitelisting arbitrary directories
What do people think of adding the ability to whitelist files/directories within `/usr` (not _just_ files/directories within `/usr/share/`)?

smitsohu:
Somewhat experimental whitelist implementation. Fixes #2041.

All top level directories are allowed except /proc, /sys and /run. As an exception from the exception, /sys/module and /run/user/$UID are allowed. This way all profiles will continue to work. Another special case is /usr, where the subdirectories (like /usr/share) are top level directories for the purpose of whitelisting.

For now all restrictions regarding symbolic links are gone (and `follow-symlink-as-user` from firejail.config is without effect). I'm not entirely sure if that is sustainable, but it can always be added back.

Otherwise this implementation should be very close to the current one.

Maybe it would also make sense to reimplement private-lib as whitelist then, in order to prevent name collisions as in #3236

### DIFF
--- a/etc/firejail.config
+++ b/etc/firejail.config
@@ -116,6 +116,10 @@
 # Enable or disable whitelisting support, default enabled.
 # whitelist yes
 
+# Disable whitelist top level directories, in addition to those
+# that are disabled out of the box. None by default; this is an example.
+# whitelist-disable-topdir /etc,/usr/etc
+
 # Enable or disable X11 sandboxing support, default enabled.
 # x11 yes
 
--- a/src/firejail/checkcfg.c
+++ b/src/firejail/checkcfg.c
@@ -35,6 +35,7 @@ char *xvfb_extra_params = "";
 char *netfilter_default = NULL;
 unsigned long join_timeout = 5000000; // microseconds
 char *config_seccomp_error_action_str = "EPERM";
+char **whitelist_reject_topdirs = NULL;
 
 int checkcfg(int val) {
 	assert(val < CFG_MAX);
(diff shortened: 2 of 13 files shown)


rusty-snake:
> except /run

No way to target #3189 here too?

smitsohu:
> No way to target #3189 here too?

There was not really a _strong_ reason to deny /run. I'll add /run top level directory support.

smitsohu:
Ok, I rewrote the messed up commit history and added support for /run.

Two nested whitelist top level directories (/run, /run/user/$UID and user home) and all the automatisms are a bit difficult to get right, but I think I'm now close to it.

glitsj16:
@smitsohu Another great piece of coding! Whenever you think this is ready for testing, feel free to ping.

smitsohu:
@glitsj16 If you want you can give it a try.

There are still two warnings from the gcc static analyzer, I need to look at them more closely but think they are false positives.

netblue30:
Let's try it out, thanks!

mardy:
What is the reason for not allowing /proc? With a blacklist, there's the risk that some kernel modules expose some sensitive information in nonstandard /proc/ files, so I'd rather whitelist the files that I know the application needs, rather than risking to give access to unforeseen files.

smitsohu:
> What is the reason for not allowing /proc? With a blacklist, there's the risk that some kernel modules expose some sensitive information in nonstandard /proc/ files, so I'd rather whitelist the files that I know the application needs, rather than risking to give access to unforeseen files.

Yes, it would be desirable to include /proc and /sys.

I excluded /proc mainly because Firejail itself still needs to access it after the whitelist has been set up, though one could argue that the same applies to blacklisting.

Probably there is no really good reason to exclude /sys. The time I wrote this there was a bit too much respect/scare about subtle effects that missing files in /proc and /sys could have on other sandboxed setuid tools.

smitsohu:
> What is the reason for not allowing /proc? With a blacklist, there's the risk that some kernel modules expose some sensitive information in nonstandard /proc/ files, so I'd rather whitelist the files that I know the application needs, rather than risking to give access to unforeseen files.

The procfs is not an easy target for a sandboxing tools anyways, because per-process subdirectories are created after the sandbox is up already. In many cases AppArmor, Tomoyo, SELinux will serve you better.

Do you have a particular kernel module in mind?

Perhaps something can be done using the `subset` or `hidepid` procfs mount options.

rusty-snake:
To further complicate this, there are programs that check whether /proc/PATH is somehow tampered with: https://docs.rs/rustix/latest/rustix/procfs/index.html

Issue #1793, opened 2018-03-02
ktorrent won't save configuration when used with default profile for it


zorates:
firejail version 0.9.52
ktorrent 5.1.0

Launching `firejail ktorrent` with the default profile in /etc/firejail/ktorrent.profile, changing some settings in ktorrent and clicking apply produces the following error on the console:

> Qt Warning: Couldn't write "/home/myuser/.config/ktorrentrc" . Disk full?

No error when launching with the --noprofile argument however.

I checked ktorrent.profile and it contains the following lines:

> noblacklist ${HOME}/.config/ktorrentrc
> whitelist ${HOME}/.config/ktorrentrc

so it appears the file is properly whitelisted, but it still doesn't work.

Fred-Barclay:
G'day @zorates 
Do you know what those settings are you changed in ktorrent? 
Thanks!
Fred

zorates:
For example the 'Default save location' or the the 'Maximum upload speed'.

I first ran ktorrent without firejail if that makes any difference, so the `ktorrentrc` file already existed and contained my initial setup. I then started running it in firejail, without any apparent issues, until I tried to change to above settings and notice that they don't survive a restart. This lead me to investigate the cause, which turned out to be firejail.

smitsohu:
I see the same with all KDE apps when I try to add whitelisting to their profiles, and Firefox has similar problems when it attempts to update the default browser in mimeapps.list.

This might have to do with KDE and Firefox trying to run rename() on whitelisted files, which fails because whitelisted files are mount points inside the sandbox and can't be renamed. It could be a lack of fantasy on my side, but I think this is a hard limitation in the interaction between KDE apps (or Firefox) and Firejail. 

Also means we can drop all `mkfile` instructions in KDE app profiles with whitelist (i.e. akregator and ktorrent).

rusty-snake:
still an issue?

smitsohu:
Basically it is a limitation in the kernel. It can be addressed either there or in the relevant libraries, there is not much that Firejail can do (apart from dirty LD_PRELOAD hacks).

rusty-snake:
See https://github.com/netblue30/firejail/issues/3697#issuecomment-716468502 for a workaround.

PR #4230, opened 2021-05-02
New profile for neochat

Kishore96in:
This is a firejail profile for Neochat, the KDE Matrix Client ( https://apps.kde.org/id/neochat/ ).
I have tested this only on Arch Linux (and everything seems to work).
I've also added the neochat config files to `disable-programs.inc`.

### DIFF
--- a/etc/inc/whitelist-1793-workaround.inc
+++ b/etc/inc/whitelist-1793-workaround.inc
@@ -1,0 +1,26 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+include allow-nodejs.local
+# This works around bug 1793, and allows whitelisting to be used for some KDE applications.
+
+whitelist ${HOME}/.config/ibus
+whitelist ${HOME}/.config/mimeapps.list
+whitelist ${HOME}/.config/pkcs11
+read-only ${HOME}/.config/pkcs11
+whitelist ${HOME}/.config/user-dirs.dirs
+read-only ${HOME}/.config/user-dirs.dirs
+whitelist ${HOME}/.config/user-dirs.locale
+read-only ${HOME}/.config/user-dirs.locale
+whitelist ${HOME}/.config/fontconfig
+whitelist ${HOME}/.config/Kvantum
+whitelist ${HOME}/.config/Trolltech.conf
+whitelist ${HOME}/.config/QtProject.conf
+whitelist ${HOME}/.config/kdeglobals
+whitelist ${HOME}/.config/kio_httprc
+whitelist ${HOME}/.config/kioslaverc
+whitelist ${HOME}/.config/ksslcablacklist
+whitelist ${HOME}/.config/qt5ct
+whitelist ${HOME}/.config/qtcurve
+
+blacklist ${HOME}/.config/*
+whitelist ${HOME}/.config
--- a/etc/inc/whitelist-1793-workaround.inc
+++ b/etc/inc/whitelist-1793-workaround.inc
@@ -3,24 +3,27 @@
 include allow-nodejs.local
 # This works around bug 1793, and allows whitelisting to be used for some KDE applications.
 
-whitelist ${HOME}/.config/ibus
-whitelist ${HOME}/.config/mimeapps.list
-whitelist ${HOME}/.config/pkcs11
-read-only ${HOME}/.config/pkcs11
-whitelist ${HOME}/.config/user-dirs.dirs
-read-only ${HOME}/.config/user-dirs.dirs
-whitelist ${HOME}/.config/user-dirs.locale
-read-only ${HOME}/.config/user-dirs.locale
-whitelist ${HOME}/.config/fontconfig
-whitelist ${HOME}/.config/Kvantum
-whitelist ${HOME}/.config/Trolltech.conf
-whitelist ${HOME}/.config/QtProject.conf
-whitelist ${HOME}/.config/kdeglobals
-whitelist ${HOME}/.config/kio_httprc
-whitelist ${HOME}/.config/kioslaverc
-whitelist ${HOME}/.config/ksslcablacklist
-whitelist ${HOME}/.config/qt5ct
-whitelist ${HOME}/.config/qtcurve
+noblacklist ${HOME}/.config/ibus
+noblacklist ${HOME}/.config/mimeapps.list
+noblacklist ${HOME}/.config/pkcs11
+noblacklist ${HOME}/.config/user-dirs.dirs
+noblacklist ${HOME}/.config/user-dirs.locale
+noblacklist ${HOME}/.config/dconf
+noblacklist ${HOME}/.config/fontconfig
+noblacklist ${HOME}/.config/gtk-2.0
+noblacklist ${HOME}/.config/gtk-3.0
+noblacklist ${HOME}/.config/gtk-4.0
+noblacklist ${HOME}/.config/gtkrc
+noblacklist ${HOME}/.config/gtkrc-2.0
+noblacklist ${HOME}/.config/Kvantum
+noblacklist ${HOME}/.config/Trolltech.conf
+noblacklist ${HOME}/.config/QtProject.conf
+noblacklist ${HOME}/.config/kdeglobals
+noblacklist ${HOME}/.config/kio_httprc
+noblacklist ${HOME}/.config/kioslaverc
+noblacklist ${HOME}/.config/ksslcablacklist
+noblacklist ${HOME}/.config/qt5ct
+noblacklist ${HOME}/.config/qtcurve
 
 blacklist ${HOME}/.config/*
 whitelist ${HOME}/.config
--- a/etc/inc/whitelist-1793-workaround.inc
+++ b/etc/inc/whitelist-1793-workaround.inc
@@ -1,6 +1,6 @@
 # This file is overwritten during software install.
 # Persistent customizations should go in a .local file.
-include allow-nodejs.local
+include whitelist-1793-workaround.local
 # This works around bug 1793, and allows whitelisting to be used for some KDE applications.
 
 noblacklist ${HOME}/.config/ibus
--- a/etc/profile-m-z/neochat.profile
+++ b/etc/profile-m-z/neochat.profile
@@ -63,7 +63,6 @@ dbus-user filter
 dbus-user.own org.kde.neochat
 dbus-user.talk org.freedesktop.Notifications
 dbus-user.talk org.kde.StatusNotifierWatcher
-dbus-user.talk org.kde.kwalletd5
 dbus-system none
 
 join-or-start neochat
--- a/etc/profile-m-z/neochat.profile
+++ b/etc/profile-m-z/neochat.profile
@@ -61,7 +61,7 @@ disable-mnt
 private-bin neochat
 private-dev
 ##private-etc Common,Networking,GUI,Qt,KDE,D-Bus
-private-etc alternatives,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,mime.types,xdg,ca-certificates,ssl,pki,crypto-policies,nsswitch.conf,resolv.conf,hosts,host.conf,hostname,protocols,services,rpc,fonts,pango,X11,Trolltech.conf,kde4rc,kde5rc,dbus-1,machine-id
+private-etc alternatives,ca-certificates,crypto-policies,dbus-1,fonts,host.conf,hostname,hosts,kde4rc,kde5rc,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,machine-id,mime.types,nsswitch.conf,pango,pki,protocols,resolv.conf,rpc,services,ssl,Trolltech.conf,X11,xdg
 private-tmp
 
 dbus-user filter
--- a/etc/profile-m-z/neochat.profile
+++ b/etc/profile-m-z/neochat.profile
@@ -62,11 +62,8 @@ private-bin neochat
 private-dev
 ##private-etc Common,Networking,GUI,Qt,KDE,D-Bus
 private-etc alternatives,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,mime.types,xdg,ca-certificates,ssl,pki,crypto-policies,nsswitch.conf,resolv.conf,hosts,host.conf,hostname,protocols,services,rpc,fonts,pango,X11,Trolltech.conf,kde4rc,kde5rc,dbus-1,machine-id
-##private-lib LIBS
-##private-opt NAME
 private-tmp
 
-## dbus-user permissions copied from the flatpak package
 dbus-user filter
 dbus-user.own org.kde.neochat
 dbus-user.talk org.freedesktop.Notifications
--- a/etc/profile-m-z/neochat.profile
+++ b/etc/profile-m-z/neochat.profile
@@ -62,7 +62,6 @@ private-tmp
 dbus-user filter
 dbus-user.own org.kde.neochat
 dbus-user.talk org.freedesktop.Notifications
-dbus-user.talk com.canonical.AppMenu.Registrar
 dbus-user.talk org.kde.StatusNotifierWatcher
 dbus-user.talk org.kde.kwalletd5
 dbus-system none
--- a/etc/profile-m-z/neochat.profile
+++ b/etc/profile-m-z/neochat.profile
@@ -11,7 +11,6 @@ noblacklist ${HOME}/.config/KDE/neochat
 noblacklist ${HOME}/.config/neochatrc
 noblacklist ${HOME}/.config/neochat.notifyrc
 noblacklist ${HOME}/.local/share/KDE/neochat
-noblacklist ${DOWNLOADS}
 
 include disable-common.inc
 include disable-devel.inc
--- a/etc/profile-m-z/neochat.profile
+++ b/etc/profile-m-z/neochat.profile
@@ -66,5 +66,3 @@ dbus-user.own org.kde.neochat
 dbus-user.talk org.freedesktop.Notifications
 dbus-user.talk org.kde.StatusNotifierWatcher
 dbus-system none
-
-join-or-start neochat
--- a/etc/profile-m-z/neochat.profile
+++ b/etc/profile-m-z/neochat.profile
@@ -39,6 +39,7 @@ include whitelist-var-common.inc
 
 apparmor
 caps.drop all
+machine-id
 netfilter
 nodvd
 nogroups
--- a/etc/profile-m-z/neochat.profile
+++ b/etc/profile-m-z/neochat.profile
@@ -7,6 +7,7 @@ include neochat.local
 include globals.local
 
 noblacklist ${HOME}/.cache/KDE/neochat
+noblacklist ${HOME}/.config/KDE
 noblacklist ${HOME}/.config/KDE/neochat
 noblacklist ${HOME}/.config/neochatrc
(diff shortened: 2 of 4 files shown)


etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
+# Firejail profile for neochat
+# Description: Matrix Client
+# This file is overwritten after every install/update
+# Persistent local customizations
+include neochat.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.cache/KDE/neochat
```
glitsj16, line 9: Add this as blacklist to disable-programs.inc too. You'll find the section # ${HOME}/.cache directory further down in that file.

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+noblacklist ${HOME}/.config/KDE/neochat
+noblacklist ${HOME}/.config/neochatrc
+noblacklist ${HOME}/.config/neochat.notifyrc
+noblacklist ${HOME}/.local/share/KDE/neochat
+noblacklist ${DOWNLOADS}
```
glitsj16, line 14: ${DESKTOP} and ${DOWNLOADS} are special, they're never blacklisted in disable-xdg.inc. So you can remove this line.

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+include disable-shell.inc
+include disable-write-mnt.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.cache/KDE/neochat
+mkdir ${HOME}/.config/KDE/neochat.conf
+mkdir ${HOME}/.local/share/KDE/neochat
+mkfile ${HOME}/.config/neochatrc
+mkfile ${HOME}/.config/neochat.notifyrc
+
```
glitsj16, line 31: No need for a blank line here.

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+whitelist ${HOME}/.cache/KDE/neochat
+whitelist ${HOME}/.config/KDE/neochat.conf
+whitelist ${HOME}/.config/neochatrc
+whitelist ${HOME}/.local/share/KDE/neochat
+whitelist ${HOME}/.config/neochat.notifyrc
+whitelist ${DOWNLOADS}
+include whitelist-common.inc
+
```
glitsj16, line 39: No need for a blank line here.

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+tracelog
+
+disable-mnt
+private-bin neochat
+private-dev
+##private-etc Common,Networking,GUI,Qt,KDE,D-Bus
```
glitsj16, line 63: Only relevant in the template, can be removed here.

glitsj16 (approved):
Besides some aesthetic nitpicking this is ready for merge. Nice work!

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+noblacklist ${DOWNLOADS}
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
```
rusty-snake, line 23:
```suggestion
```

Already `blacklist`ed by `disable-mnt`.

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+nodvd
+nogroups
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
```
rusty-snake, line 56: Can we add `seccomp.block-secondary`?

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+whitelist ${DOWNLOADS}
+include whitelist-common.inc
+
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+netfilter
```
rusty-snake, line 46: `machine-id`?

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+tracelog
+
+disable-mnt
+private-bin neochat
```
rusty-snake, line 61: `private-cache`?

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+private-dev
+##private-etc Common,Networking,GUI,Qt,KDE,D-Bus
+private-etc alternatives,ca-certificates,crypto-policies,dbus-1,fonts,host.conf,hostname,hosts,kde4rc,kde5rc,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,machine-id,mime.types,nsswitch.conf,pango,pki,protocols,resolv.conf,rpc,services,ssl,Trolltech.conf,X11,xdg
+private-tmp
+
+dbus-user filter
+dbus-user.own org.kde.neochat
+dbus-user.talk org.freedesktop.Notifications
+dbus-user.talk com.canonical.AppMenu.Registrar
```
rusty-snake, line 70:
```suggestion
```

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+dbus-user.talk org.kde.StatusNotifierWatcher
```
rusty-snake, line 71: No need for `dbus-user.own org.kde.*`?

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+dbus-user.talk org.kde.kwalletd5
+dbus-system none
+
+join-or-start neochat
```
rusty-snake, line 75: Add a comment why this is necessary.

etc/profile-m-z/neochat.profile
```diff
@@ -0,0 +1,75 @@
...
+
+mkdir ${HOME}/.cache/KDE/neochat
+mkdir ${HOME}/.config/KDE/neochat.conf
+mkdir ${HOME}/.local/share/KDE/neochat
+mkfile ${HOME}/.config/neochatrc
+mkfile ${HOME}/.config/neochat.notifyrc
+
+whitelist ${HOME}/.cache/KDE/neochat
+whitelist ${HOME}/.config/KDE/neochat.conf
+whitelist ${HOME}/.config/neochatrc
```
rusty-snake, line 34: Can neochat save changed settings?

rusty-snake (commented):
Some further hardening suggestions.

etc/profile-m-z/neochat.profile
Kishore96in: `private-cache` makes the application loading visibly slow. When I was testing the profile, I thought it wasn't working (due to the very slow loading), which is why I did not add that option. Should I add it?

etc/profile-m-z/neochat.profile
Kishore96in: All features seem to work with the current configuration, so I guess that is not required.

etc/profile-m-z/neochat.profile
rusty-snake: then, no

etc/profile-m-z/neochat.profile
rusty-snake: I just asked because the kde (or qt?) libs need this for tray-icons to work.

etc/profile-m-z/neochat.profile
Kishore96in: Actually, no. I thought I'd checked that. :-( 
It works if I don't whitelist anything, but these are all the files which I can see created when I run it with `firejail --private`.

etc/profile-m-z/neochat.profile
Kishore96in: Removed.

etc/profile-m-z/neochat.profile
Kishore96in: Done. I copied the dbus stuff from the flatpack package, but it works without this one.

etc/profile-m-z/neochat.profile
Kishore96in: I've commented it out since it seems to work on my machine without it, but this permission was there in the flatpak as well, so I am not sure if it may be needed, perhaps for some different sign-on method which I haven't tried.

etc/profile-m-z/neochat.profile
Kishore96in: The tray icon is shown (and I can interact with it) on my machine without `dbus-user.own org.kde.*`. Is this some distro-dependent thing?

etc/profile-m-z/neochat.profile
Kishore96in: Should this be separate from `seccomp`, or should it replace it?

Kishore96in: Added

etc/profile-m-z/neochat.profile
Kishore96in: Added

etc/profile-m-z/neochat.profile
Kishore96in: It works if I whitelist `~/.config`. Not sure how to find out which file/subfolder I'm missing.

rusty-snake: I asked because that's an known issue with KDE programs (#1793).

I would say comment the `whitelist`ing (not wruc,wusc,wcv) and add a comment like "Makes settings immutable, add to your neochat.local after you configured neochat."

etc/profile-m-z/neochat.profile
rusty-snake: >  Is this some distro-dependent thing?

No. It depends on the program/library that is used to add a tray icon. Maybe it also depended on DE but the most users will use KDE apps in KDE and GTK app in GNOME/Xfce.

etc/profile-m-z/neochat.profile
rusty-snake: My comment refers to the `join-or-start neochat`. GH always adds the three lines above if you comment on a single line.

etc/profile-m-z/neochat.profile
Kishore96in: Would it make sense to whitelist `~/.config`? At least that way, the rest of the user's home directory will be protected.

rusty-snake: Can you test if this works.
```
noblacklist ${HOME}/.config/KDE
noblacklist ${HOME}/.config/KDE/neochat
noblacklist ${HOME}/.config/neochatrc
noblacklist ${HOME}/.config/neochat.notifyrc
blacklist ${HOME}/.config/*
whitelist ${HOME}/.config
whitelist ${HOME}/.local/share/KDE/neochat
whitelist ${HOME}/.cache/KDE/neochat
include whitelist-common.inc
include disable-common.inc
include disable-programs.inc
read-only ${HOME}/.config/*
read-write  ${HOME}/.config/neochatrc
read-write  ${HOME}/.config/KDE/neochat
read-write  ${HOME}/.config/neochat.notifyrc
```

Kishore96in: It doesn't even remember the login if I `blacklist ${HOME}/.config/*`. If I remove just that command, it works, but still doesn't remember any changed settings.

Kishore96in: If I add this but comment out `read-only ${HOME}/.config/*`, I can change settings.

Kishore96in: So if I do
```
noblacklist ${HOME}/.cache/KDE/neochat
noblacklist ${HOME}/.config/KDE
noblacklist ${HOME}/.config/KDE/neochat
noblacklist ${HOME}/.config/neochatrc
noblacklist ${HOME}/.config/neochat.notifyrc
noblacklist ${HOME}/.local/share/KDE/neochat

blacklist ${HOME}/.config/*

mkdir ${HOME}/.config/neochat
mkdir ${HOME}/.cache/KDE/neochat
mkdir ${HOME}/.config/KDE/neochat.conf
mkdir ${HOME}/.local/share/KDE/neochat
mkfile ${HOME}/.config/neochatrc
mkfile ${HOME}/.config/neochat.notifyrc

whitelist ${HOME}/.config
whitelist ${HOME}/.cache/KDE/neochat
whitelist ${HOME}/.local/share/KDE/neochat
```
I can change settings in the program. The program can see other files in `.config`, but can't open them. However, it can access all files in `.config/KDE/`. Should I go ahead and do this?

etc/profile-m-z/neochat.profile
Kishore96in: When is `join-or-start` typically used? I checked now, and even without that, it just brings the currently running instance to the foreground and exits.

etc/profile-m-z/neochat.profile
rusty-snake: > mkdir ${HOME}/.config/…

Can be skipped

> all files in .config/KDE/

We can `blacklist .config/KDE/*` too

>  Should I go ahead and do this?

1. We will likely need to noblacklist some of them `grep '\.config' /etc/firejail/whitelist-common.inc`
2. @glitsj16 what do you think?

glitsj16: I'm not at all familiar with KDE I'm afraid. Currently there's no mention of .config/KDE in any of our files AFAICT. @Kishore96in What do you have under ~/.config/KDE on your system? If this is a stupid question, disregard. Like I said, I never touched KDE and I'm not even sure I get the question :-)

rusty-snake: The main questions is: Should we this #1793 workaround?
```
noblacklist ${HOME}/.config/a
noblacklist ${HOME}/.config/b
noblacklist ${HOME}/.config/c
blacklist ${HOME}/.config/*
whitelist ${HOME}/.config
```

Pros:
- whitelisting enabled by default

Cons:
- ugly (we should add a comment that explains what we are doing)
- maintenance to be in sync with whitelist-common

glitsj16: @rusty-snake Thanks for providing context.
> maintenance to be in sync with whitelist-common
Personally I would go for the strongest hardening. So bring in the #1793 workaround. We already have quite a complex maintenance situation, one more shouldn't hurt that much as long as we're aware of it (like with another comment).

rusty-snake: > Should I go ahead and do this?

@Kishore96in yes but add `noblacklist`s for `grep '\.config' /etc/firejail/whitelist-common.inc` (the kde/qt stuff is enough IMHO) and a comment what we are doing here (above the `blacklist ~/.config/*`?).

Kishore96in: > @Kishore96in yes but add `noblacklist`s for `grep '\.config' /etc/firejail/whitelist-common.inc` (the kde/qt stuff is enough IMHO) and a comment what we are doing here (above the `blacklist ~/.config/*`?).

Would it be better to do the blacklisting in a separate file (say `blacklist-common-kde.inc`) and include it here, so that this fix can be easily applied to other profiles if necessary?

rusty-snake: A `whitelist-1793-workaround.inc` (or kde-workaround) or what evername (I would use a `whitelist-` prefix as it is used to enable whitelisting. If not `allow-` prefix) like
```
# HEADER
# COMMENT
noblacklist ${HOME}/.config/Trolltech.conf
noblacklist ${HOME}/.config/QtProject.conf
noblacklist …
blacklist ${HOME}/.config/*
whitelist ${HOME}/.config
```
? Sound good to me.

etc/profile-m-z/neochat.profile
Kishore96in: I've removed the join-or-start thing. It already detects the currently running instance (presumably over dbus) and exits.

etc/inc/whitelist-1793-workaround.inc
```diff
@@ -0,0 +1,26 @@
...
+whitelist ${HOME}/.config/fontconfig
+whitelist ${HOME}/.config/Kvantum
+whitelist ${HOME}/.config/Trolltech.conf
+whitelist ${HOME}/.config/QtProject.conf
+whitelist ${HOME}/.config/kdeglobals
+whitelist ${HOME}/.config/kio_httprc
+whitelist ${HOME}/.config/kioslaverc
+whitelist ${HOME}/.config/ksslcablacklist
+whitelist ${HOME}/.config/qt5ct
+whitelist ${HOME}/.config/qtcurve
```
rusty-snake, line 23: whitelist -> noblacklist

If we use a own .inc file for it, we should include the hole list IMHO.

`fjp cat whitelist-common.inc | grep "\.config" | grep whitelist | sed s/whitelist/noblacklist/`:

```suggestion
noblacklist ${HOME}/.config/ibus
noblacklist ${HOME}/.config/mimeapps.list
noblacklist ${HOME}/.config/pkcs11
noblacklist ${HOME}/.config/user-dirs.dirs
noblacklist ${HOME}/.config/user-dirs.locale
noblacklist ${HOME}/.config/dconf
noblacklist ${HOME}/.config/fontconfig
noblacklist ${HOME}/.config/gtk-2.0
noblacklist ${HOME}/.config/gtk-3.0
noblacklist ${HOME}/.config/gtk-4.0
noblacklist ${HOME}/.config/gtkrc
noblacklist ${HOME}/.config/gtkrc-2.0
noblacklist ${HOME}/.config/Kvantum
noblacklist ${HOME}/.config/Trolltech.conf
noblacklist ${HOME}/.config/QtProject.conf
noblacklist ${HOME}/.config/kdeglobals
noblacklist ${HOME}/.config/kio_httprc
noblacklist ${HOME}/.config/kioslaverc
noblacklist ${HOME}/.config/ksslcablacklist
noblacklist ${HOME}/.config/qt5ct
noblacklist ${HOME}/.config/qtcurve
```

etc/profile-m-z/neochat.profile
Kishore96in: > blacklist .config/KDE/*

Doing this seems to make `.config/KDE/neochat` inaccessible as well (even with the explicit noblacklist earlier), so I have not done it. As of the current version of the PR, the files in `.config/KDE` are accessible to the program (this folder contains config files of some other KDE apps as well), but the program cannot access other unnecessary folders or files in .config.

Otherwise, the program works now, and settings etc. can be changed and persist.

rusty-snake (approved):
form my side

Kishore96in:
It looks like clicking on links to open in browser doesn't work inside the sandbox. Is this something that can be done in a browser- and desktop-independent way, or should we leave it to the users to allow this?

Kishore96in:
Also, I get warnings:
```
```
when launching it. Should I blacklist these directories in the profile?

rusty-snake:
> It looks like clicking on links to open in browser doesn't work inside the sandbox. Is this something that can be done in a browser- and desktop-independent way, or should we leave it to the users to allow this?

Outside of the scope here

> Also, I get warnings:
>
> ```
> Warning: /sbin directory link was not blacklisted
> Warning: /usr/sbin directory link was not blacklisted
> ```
>
> when launching it. Should I blacklist these directories in the profile?

https://github.com/netblue30/firejail/blob/master/etc/inc/disable-common.inc#L422-L425

netblue30:
Merging in, thanks!

PR #4234, opened 2021-05-02
support older gstreamer setups in xfce4-mixer

glitsj16:


etc/profile-m-z/xfce4-mixer.profile
```diff
@@ -19,7 +19,7 @@ include disable-xdg.inc
 
 mkfile ${HOME}/.config/xfce4/xfconf/xfce-perchannel-xml/xfce4-mixer.xml
 whitelist ${HOME}/.config/xfce4/xfconf/xfce-perchannel-xml/xfce4-mixer.xml
-whitelist /usr/share/gstreamer
+whitelist /usr/share/gstreamer-*
```
rusty-snake, line 22: This removes the `whitelist` for `/usr/share/gstreamer` (w/o `-VERSION` extension).

For Fedora no package seems to provide `/usr/share/gstreamer`, maybe that fine.

glitsj16: Yeah, there's nothing I could find on Arch Linux and Debian/Ubuntu either that installs to that path. That's why I needed a 'whitelist /usr/share/gstreamer-0.10' in xfce4-mixer.local for example. So I think we're ok here.

kmk3: @rusty-snake yesterday:

> This removes the `whitelist` for `/usr/share/gstreamer` (w/o `-VERSION`
> extension).
> 
> For Fedora no package seems to provide `/usr/share/gstreamer`, maybe that
> fine.

I saw this on master and was about to ask the same thing.

@glitsj16 yesterday:

> Yeah, there's nothing I could find on Arch Linux and Debian/Ubuntu either
> that installs to that path. That's why I needed a 'whitelist
> /usr/share/gstreamer-0.10' in xfce4-mixer.local for example. So I think we're
> ok here.

Please consider mentioning things like that in the commit message.  It makes it
clear why it was removed and that it was an intentional change and not an
accidental removal.  Doesn't have to be anything fancy; your comment as it is
would have sufficed.

Also, since you already went through the trouble of checking on multiple
distros, it's extra helpful to know the results of the investigation.

glitsj16: > Please consider mentioning things like that in the commit message in the
future.

@kmk3 I'll try to be more specific in the future. I understand your reasoning. FTR I did only find /usr/share/gstreamer-{0.10,1.0} out there.

PR #4239, opened 2021-05-04
Add noinput to all profiles with private-dev

rusty-snake:


SkewedZeppelin:
This just maintains parity with with previous behavior correct?

Maybe these should be commented: barrier, kodi, xournal

SkewedZeppelin (approved):

reinerh (approved):

PR #4248, opened 2021-05-05
[WIP] joplin cli profile

sak96:
- [x] The ordering of options follow the rules described in [/usr/share/doc/firejail/profile.template](https://github.com/netblue30/firejail/blob/master/etc/templates/profile.template).  
   > Hint: The profile-template is very new. If you install firejail with your package manager, it may be missing. In order to follow the latest rules, it is recommended to use the template from the repository.

- [x]  Order the arguments of options alphabetically. You can easily do this with [sort.py](https://github.com/netblue30/firejail/tree/master/contrib/sort.py).  

   | Distro | Path |
   | ------ | ---- |

   Note also that the sort.py script exists only since firejail `0.9.61`
   
- [x] Testing
  - [x] able to open application
  - [x] able to sync with dropbox
  - [x] able to open vim 
  - [x] **IMPORTANT** add noinput before submmiting
  - [x] understand noblacklist vs whitelist

### DIFF
--- a/etc/profile-a-l/joplin.profile
+++ b/etc/profile-a-l/joplin.profile
@@ -1,0 +1,80 @@
+# Firejail profile for joplin
+# Description: A note taking and to-do application
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include joplin.local
+# Persistent global definitions
+include globals.local
+
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}
+
+noblacklist ${HOME}/.config/joplin
+noblacklist ${HOME}/.config/nano
+noblacklist ${HOME}/.emacs
+noblacklist ${HOME}/.emacs.d
+noblacklist ${HOME}/.nanorc
+noblacklist ${HOME}/.vim
+noblacklist ${HOME}/.vimrc
+
+include allow-bin-sh.inc
+include allow-nodejs.inc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/joplin
+whitelist ${HOME}/.config/joplin
+whitelist ${HOME}/.config/nano
+whitelist ${HOME}/.emacs
+whitelist ${HOME}/.emacs.d
+whitelist ${HOME}/.nanorc
+whitelist ${HOME}/.vim
+whitelist ${HOME}/.vimrc
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+machine-id
+netfilter
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+
+disable-mnt
+private-cache
+private-dev
+private-etc alternatives,ca-certificates,crypto-policies,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,mime.types,nsswitch.conf,pki,protocols,resolv.conf,rpc,services,ssl,xdg
+private-tmp
+
+read-only ${HOME}/.config/nano
+read-only ${HOME}/.emacs
+read-only ${HOME}/.emacs.d
+read-only ${HOME}/.nanorc
+read-only ${HOME}/.vim
+read-only ${HOME}/.viminfo
+read-only ${HOME}/.vimrc
+
+dbus-user none
+dbus-system none
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -279,6 +279,7 @@ blacklist ${HOME}/.config/iridium
 blacklist ${HOME}/.config/itch
 blacklist ${HOME}/.config/jami
 blacklist ${HOME}/.config/jd-gui.cfg
+blacklist ${HOME}/.config/joplin
 blacklist ${HOME}/.config/k3brc
 blacklist ${HOME}/.config/kaffeinerc
 blacklist ${HOME}/.config/kalgebrarc
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -398,6 +398,7 @@ jdownloader
 jerry
 jitsi
 jitsi-meet-desktop
+joplin
 jumpnbump
 jumpnbump-menu
 k3b


sak96:
@glitsj16 

sorry for clouding the issue with my questions. 
i hope you can give some suggestion here. i am still familiarizing with firejail.

Q: what is difference between noblacklist vs whitelist ?

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,54 @@
...
+# This file is overwritten after every install/update
+quiet
+
+# Persistent local customizations
+include joplin.local
+# Persistent global definitions
+include globals.local
+
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}/wayland-*
```
rusty-snake, line 12:
```suggestion
```

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,54 @@
+# Firejail profile for joplin
+# Description: A note taking and to-do application
+# This file is overwritten after every install/update
+quiet
+
```
rusty-snake, line 5:
```suggestion
```

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,54 @@
...
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-shell.inc
+include disable-write-mnt.inc
+include disable-xdg.inc
+
+noblacklist /usr/share/joplin
```
rusty-snake, line 27: No such `balcklist`

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,54 @@
...
+mkdir ${HOME}/.config/joplin
+whitelist ${HOME}/.config/joplin
```
rusty-snake, line 29: Add `blacklist` for it in disable-programs.inc and `noblacklist` in this file.

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,54 @@
...
+whitelist ${HOME}/.config/nano
+whitelist ${HOME}/.emacs
+whitelist ${HOME}/.emacs.d
+whitelist ${HOME}/.nanorc
+whitelist ${HOME}/.vim
+whitelist ${HOME}/.viminfo
+whitelist ${HOME}/.vimrc
```
rusty-snake, line 36: Add `include whitelist-common.inc`

rusty-snake, line 36: Can we `read-only` them?

rusty-snake (commented):
What was the reason to drop the most stuff of https://github.com/netblue30/firejail/issues/1139#issuecomment-831384616 resp. https://github.com/netblue30/firejail/issues/1139#issuecomment-831917967?

Possible further hardening options:
- [ ] `private-cache`
- [ ] `private-bin`
- [ ] `private-etc`
- [ ] `disable-mnt`
- [ ] `tracelog`
- [ ] `private-tmp`
- [ ] wruc, wusc, wvc
- [ ] `dbus-{user,system} none`
- [ ] `seccomp.block-secondary`
- [ ] `noinput` <- very new in firejail git. Just add it, it is the current behaviour of firejail 0.9.64.4.
- [ ] `netfilter`
- [ ] ~mdwe~

etc/profile-a-l/joplin.profile
sak96: this was added to avoid following output. 
```
...
Reading profile /etc/firejail/allow-bin-sh.inc
Reading profile /etc/firejail/allow-nodejs.inc
Reading profile /etc/firejail/disable-exec.in
...
```

rusty-snake: I forget that we it's a cli program. Right. Leave the `quiet`, but remove the blank line.

etc/profile-a-l/joplin.profile
sak96: is this needed.
whitelist-common.inc seems looks to be more oriented for ui.

rusty-snake: yes

sak96:
@rusty-snake 

off-topic: 
how did you put the pr in draft ?
can you tell where is the button for that ?

rusty-snake:
If you create a PR: https://github.blog/2019-02-14-introducing-draft-pull-requests/
On a existing PR: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-stage-of-a-pull-request

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,71 @@
...
+
+dbus-user none
+dbus-system none
+
+private-dev
+private-cache
+private-etc none
+disable-mnt
+private-tmp
+seccomp.block-secondary
```
rusty-snake, line 71: Move direct under `seccomp`

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,71 @@
...
+shell none
+
+dbus-user none
+dbus-system none
+
+private-dev
+private-cache
+private-etc none
+disable-mnt
+private-tmp
```
rusty-snake, line 70: order

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,71 @@
...
+protocol unix,inet,inet6
+seccomp
+shell none
+
+dbus-user none
+dbus-system none
+
+private-dev
+private-cache
+private-etc none
```
rusty-snake, line 68: Does not work

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,71 @@
...
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+shell none
+
+dbus-user none
+dbus-system none
```
rusty-snake, line 64: Move at the very end of this file (in a own block).

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,71 @@
...
+include allow-nodejs.inc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
```
rusty-snake, line 23: Remove (protected by `disable-mnt`)

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,71 @@
...
+include disable-xdg.inc
+
+include whitelist-common.inc
```
rusty-snake, line 26: Move direct at the end of the whitelisting block

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,71 @@
...
+read-only ${HOME}/.emacs
+whitelist ${HOME}/.emacs.d
+read-only ${HOME}/.emacs.d
+whitelist ${HOME}/.nanorc
+read-only ${HOME}/.nanorc
+whitelist ${HOME}/.vim
+read-only ${HOME}/.vim
+whitelist ${HOME}/.viminfo
+read-only ${HOME}/.viminfo
+whitelist ${HOME}/.vimrc
```
rusty-snake, line 43: Move the whitelist in the whitelist block and the read-only between the private and the dbus block.

rusty-snake (commented):
some moving.

Can you please squash and force push only at the end. squash and force push is good before merging but horrible to review (you don't see the changes).

etc/profile-a-l/joplin.profile
sak96: it is working. i used it from the [here](https://github.com/netblue30/firejail/blob/9adeaf0c6f302c3091496848bc6e3b732064e1e3/etc/profile-m-z/notify-send.profile#L53).

sak96: ok i don't know what to add for starters i will add these thing you mentioned.

sak96:
> What was the reason to drop the most stuff of [#1139 (comment)](https://github.com/netblue30/firejail/issues/1139#issuecomment-831384616) resp. [#1139 (comment)](https://github.com/netblue30/firejail/issues/1139#issuecomment-831917967)?

looks like there are three application cli, desktop and appimage.
the once in this profile are specifically for cli.
the profile there was just a draft.

#  Possible further hardening options: 
* [x]  `private-cache`
* [ ]  `private-bin`
> private-bin would required to allow shells (bash/zsh/..) + editor(vim/emacs/..) and so on.
* [x]  `private-etc`
* [x]  `disable-mnt`
* [ ]  `tracelog`
> this is for audit ?? can you explain ?
* [x]  `private-tmp`
* [ ]  wruc, wusc, wvc
> some examples. i don't get you. 
* [x]  `dbus-{user,system} none`
* [x]  `seccomp.block-secondary`
> is this required as i have added full seccomp.
* [ ]  `noinput` <- very new in firejail git. Just add it, it is the current behaviour of firejail 0.9.64.4.
> will add at the end
* [x]  `netfilter` 
* [ ]  ~mdwe~
> this doesn't need to be added. i guess ?

rusty-snake:
> wruc, wusc, wvc

whitelist-usr-share-common, whitelist-runuser-common, whitelist-var-common

>  mdwe

Will break with node I guess. If don't add it. (mdwe=memory-deny-write-execute)

> is this required as i have added full seccomp.

seccomp: block all syscalls in @defualt-nodebuggers 
seccomp.block-secondary block all non-native syscalls

> tracelog

> This option enables auditing blacklisted files and  directories.
> A message is sent to syslog in case the file or the directory is accessed.

kmk3:
@rusty-snake reviewed 19 hours ago:

> some moving.
> 
> Can you please squash and force push only at the end. squash and force push
> is good before merging but horrible to review (you don't see the changes).

I agree that it's annoying when in happens while reviewing and personally, I
try to comment a summary of the changes after force-pushing.  But FWIW, there
is a (kind of manual) way to see the changes before/after the force-push on
GitHub:

> sak96 force-pushed the sak96:master branch from bbe0160 to 8f7087d 19 hours ago

Append `$fromcommit..$tocommit` to this url:

* <https://github.com/netblue30/firejail/compare/>

Like so:

* <https://github.com/netblue30/firejail/compare/bbe0160..8f7087d>

Ideally, GitHub would put a link to that in the force-push event message.

Note: This won't work if the branch is also rebased onto a different base
(e.g.: rebase onto a more recent master).  In that case, I'd say that it indeed
likely becomes "horrible to review", as I suppose you'd have to check out both
versions of the branch locally and rebase the old one onto the same base to see
the exact changes.

If there were more force-pushes in the meantime, you could also compare only
the last version you reviewed against the latest version.  Example (considering
bbe0160 as the last reviewed version):

> sak96 force-pushed the sak96:master branch from bbe0160 to 8f7087d 19 hours ago
> sak96 force-pushed the sak96:master branch from 974258a to 4be87e7 19 hours ago

Would yield:

* <https://github.com/netblue30/firejail/compare/bbe0160..4be87e7>

kmk3:
> Ideally, GitHub would put a link to that in the force-push event message.

And here I was, trying to write some js to do that, but the link was there all
along, on the "force-pushed" text:

> sak96 force-pushed the sak96:master branch from bbe0160 to 8f7087d 20 hours ago

> https://github.com/netblue30/firejail/compare/bbe016065e297829f09f269b3e875341077d29fd..8f7087d7b270b3fca33ed32f018b9b520ba1304a

Hopefully, the part I said about multiple force-pushes is still useful.

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,82 @@
...
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/joplin
+
```
rusty-snake, line 34:
```suggestion
```

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,82 @@
...
+whitelist ${HOME}/.config/joplin
+whitelist ${HOME}/.config/nano
+whitelist ${HOME}/.emacs
+whitelist ${HOME}/.emacs.d
+whitelist ${HOME}/.nanorc
+whitelist ${HOME}/.vim
+whitelist ${HOME}/.vimrc
+
```
rusty-snake, line 42:
```suggestion
```

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,82 @@
...
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-var-common.inc
```
rusty-snake, line 45: whitelist-usr-share-common?

rusty-snake (commented):
LGTM (now you can squash and force push)

etc/profile-a-l/joplin.profile
sak96: that is because you need to whitelist /usr/share/joplin /usr/share/vim and so on

sak96:
https://github.com/netblue30/firejail/compare/602ff74..955bb15
yes force pushed.

rusty-snake (approved):

etc/profile-a-l/joplin.profile
rusty-snake: yes, then whitelist them

sak96:
@rusty-snake 
for joplin-desktop i will be raising another request in weeks time.
the profile request should complete with that one.

sak96:
@rusty-snake can you please point out on what is left ?
@glitsj16  can you elaborate on connecting vim via it's own sandbox.
- currently the app requires editor to edit nodes, which might be vim or emacs or such and such.
- is it possible to mention the same in profiles ?

glitsj16:
> @glitsj16 can you elaborate on connecting vim via it's own sandbox.
    - currently the app requires editor to edit nodes, which might be vim or emacs or such and such.
    - is it possible to mention the same in profiles ?

Originally I was thinking it would be possible to just edit these joplin notes directly in vim (or any other firejail-supported editor for that matter) in its own sandbox. But the way our CLI editor profiles are designed aren't suited for that anyway I now realize. They all include `disable-programs.inc` for example, which is a pretty big collection of blacklists, and having to noblacklist the files you'd like to access directly thru an editor's own sandbox would be rather unmanageable/impractical). You've already added emacs, nano and vim support I notice, so please ignore my remark. It's my own lack of understanding of the design of these editor profiles that caused the confusion I'm afraid.

On an entirely different note, according to [this page](https://aur.archlinux.org/packages/joplin-desktop/) joplin-desktop is electron-based. That begs the question why is there no separate `joplin-desktop.profile` in this PR? You could implement it as a redirect to our electron.profile for example (have a look at one of our +20 other profiles that are done that way for electron apps). But you're much better placed to call this, as I assume you use this app regularly.

glitsj16:
> I still working on joplin-desktop.profile

@sak96 I understand. Take your time. It's just something we have to bare in mind.

> If vim/emacs/nano is opened by something like joplin edit todos-of-today this discussion is relevant. And btw vim/emacs/nano aren't in firecfg.config.

@rusty-snake Good point! My braincell is being flooded with corona-lockdown relaxing rules going into effect here tonight... I hope we can all enjoy a less cramped future soonish. Take care! Over & out :D

sak96:
looks like the app doesn't have access to network. i tried removing netfilter as well. 
this is required for sync with dropbox/nextcloud/... but seems like it is not able to connect to internet. 
can you suggest the changes. removing netfilter does nothing.

rusty-snake (changes requested):
sort.py CI is failing because of latest commit. Rest it still LGTM.

glitsj16:
> looks like the app doesn't have access to network. i tried removing netfilter as well.
this is required for sync with dropbox/nextcloud/... but seems like it is not able to connect to internet.
can you suggest the changes. removing netfilter does nothing.

Try integrating what's in the `profile.template` for [networking](https://github.com/netblue30/firejail/blob/master/etc/templates/profile.template#L190).

rusty-snake:
We definitely need something like `private-etc @network(ing)` (#1734 or #3424/#3412 as generic workaround). I use [`include private-etc:net.inc`](https://gitlab.com/rusty-snake/kyst/-/blob/f4d93bd61bdf9c044e36e73c398bb841864d6641/firejail/private-etc:net.inc) in my locals as workaround for now.

sak96:
@netblue30 i think [kyst](https://gitlab.com/rusty-snake/kyst/) is cool.
i am stealing some config for my local :smirk:

for now i am using networking and common from template profile.

etc/profile-a-l/joplin.profile
```diff
@@ -0,0 +1,81 @@
...
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+
+disable-mnt
+private-cache
+private-dev
+private-etc alternatives,ca-certificates,crypto-policies,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,mime.types,nsswitch.conf,pki,protocols,resolv.conf,rpc,services,ssl,xdg
+
```
rusty-snake, line 69:
```suggestion
```

sak96:
closing this as i am no longer using joplin. in case any one wants to continue diff is atttached as a patch.
[joplin_cli.patch.zip](https://github.com/netblue30/firejail/files/7141302/joplin_cli.patch.zip)

PR #4251, opened 2021-05-06
whitelist /var/lib/aspell in whitelist-var-common.inc

pirate486743186:
In claws-mail you can't switch the language in the spellchecker without this line. I'm assuming this should be generic.

### DIFF
--- a/etc/inc/whitelist-var-common.inc
+++ b/etc/inc/whitelist-var-common.inc
@@ -4,6 +4,7 @@ include whitelist-var-common.local
 
 # common /var whitelist for all profiles
 
+whitelist /var/lib/aspell
 whitelist /var/lib/ca-certificates
 whitelist /var/lib/dbus
 whitelist /var/lib/menu-xdg


glitsj16:
Thanks for reporting this. On Arch Linux claws-mail uses [enchant](https://archlinux.org/packages/extra/x86_64/enchant/) for spell checking. That being a generic app users can opt for `aspell`, `hspell`, `hunspell`, `libvoikko` or `nuspell` according to the enchant dependencies. Do you happen to know any of these other engines besides aspell have their own, similar counterpart in /var/lib too that we'd need to support?

glitsj16 (commented):
I understand the added whitelist fixes this for claws-mail, no problem there. I'm just wondering if whitelist-var-common.inc is the best place to add it - _if_ claws-mail is the only affected app we could add it to its profile instead. I'm not familiar with spell checkers to judge whether they all use a similar path under /var/lib that we'd need to support. Let's try to find more info on that before we open up paths that are not widely used.

rusty-snake:
Seems to be an Arch specific path.
```
$ en_US-locale dnf provides /var/lib/aspell 
Error: No Matches found
```

What's in it?

pirate486743186:
i'm on ubuntu 18.04 32bit
i don't know about arch

/var/lib/aspell is installed by aspell packages
for example aspell-fr

the same behavior happens with pluma too
there's an unrelated bug there, you need 
`--ignore=memory-deny-write-execute`
to run it.

to be precise, switching english variants works fine, if you try to switch to an other language, it fails.

it's not arch specific, i'm on ubuntu
it seams that english is treated differently, because it's the default
try aspell-fr
it seams to be files generated during install
/var/lib/aspell has a README that says

> This directory will contain files related to aspell hash autobuild.

it has .rws files for the installed languages, plus some special files.

my guess, it's that it's some sort of compilation

glitsj16:
Arch doesn't use /var/lib for non-english aspell dictionaries. It places those under /usr/lib/aspell-0.60 (cfr. the [PKGBUILD](https://github.com/archlinux/svntogit-packages/blob/packages/aspell-de/trunk/PKGBUILD#L33)).

pirate486743186:
i meant, that because english is the default, it's not affected by this problem. In ubuntu also, files of all languages are placed in the same folders as english. These folders are different from arch though.

in ubuntu, they are /usr/lib/aspell/, /usr/share/aspell and /var/lib/aspell

rusty-snake:
`/usr/lib64/aspell-0.60` for Fedora (64bit).

Since these aspell files are available on other distros then ubu, we can merge this IMHO.

PR #4254, opened 2021-05-07
pluma broken with memory-deny-write-execute

pirate486743186:
with memory-deny-write-execute you just get a messed up window.

The same error was in engrampa, so most likely, they are other gtk3 application profiles with the same error.

### DIFF
--- a/etc/profile-m-z/pluma.profile
+++ b/etc/profile-m-z/pluma.profile
@@ -50,6 +50,4 @@ private-tmp
 # dbus-user none
 # dbus-system none
 
-memory-deny-write-execute
-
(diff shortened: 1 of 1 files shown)


glitsj16:
> The same error was in engrampa, so most likely, they are other gtk3 application profiles with the same error.

engrampa doesn't have mdwe...

rusty-snake:
https://github.com/netblue30/firejail/commit/936cb321ecec13e2d39dd7b4d6e0bd92a8be10fc#diff-319702f732d484f09647854e99289835f3713f8f9ae7e96564e12f406058ebcc

PR #4255, opened 2021-05-07
Node.js stack refactoring

glitsj16:
An attempt to bring some sanity to our node.js related profiles. Based on info gathered from:
- our existing profiles
- https://wiki.archlinux.org/title/Node.js
- https://github.com/nvm-sh/nvm/blob/master/README.md

~~Not being a Node developer I'm marking this as WIP for now.~~ I've been testing this, but the node stack still has many dark spots for me to unravel. Would be great if this could be more thoroughly tested. Open to all suggestions, input etcetera.

IMPORTANT NOTE: none of the Node.js stack profiles are in firecfg. It is up to the user to create the necessary scripts/symlinks to sandbox the following commands:
- /usr/bin/node
- /usr/bin/npm
- /usr/bin/nvm
- /usr/bin/yarn


etc/profile-m-z/nodejs-common.profile
```diff
@@ -7,9 +7,26 @@ include nodejs-common.local
 # added by caller profile
 #include globals.local
 
+# Add the next line to your nodejs-common.local if you prefer to disable gatsby telemetry.
+#env GATSBY_TELEMETRY_DISABLED=1
```
rusty-snake, line 11: `env` should be placed in the mdwe,ro,rw,join-or-start block at the end.

etc/profile-m-z/nodejs-common.profile
```diff
@@ -7,9 +7,26 @@ include nodejs-common.local
...
+
 blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}
 
+ignore read-only ${HOME}/.npm-packages
+ignore read-only ${HOME}/.npmrc
+ignore read-only ${HOME}/.nvm
+ignore read-only ${HOME}/.yarnrc
```
rusty-snake, line 19: Maybe prefer `read-write`? IDK

etc/profile-m-z/semver.profile
```diff
@@ -0,0 +1,10 @@
+# Firejail profile for semver
```
rusty-snake, line 1: IDK what it is, but sound like a cli tool. Can we add `private-bin semver`?

etc/profile-m-z/nodejs-common.profile
glitsj16: Done.

etc/profile-m-z/nodejs-common.profile
glitsj16: > Maybe prefer `read-write`? IDK

I agree that `read-write` sounds a lot more logical than the `ignore blah` construct. Just checked and I used it in git-cola, gnome-{logs,system-log} and yelp too apparently. Can't remember why OTOH but I'll change it after testing it gives the same result. That'll be over the weekend though, I'm clocking off for today. Thanks for the suggestions, much appreciated. And enjoy the weekend!

etc/profile-m-z/semver.profile
glitsj16: It's a semantic versioning tool apparently. And it has the `#!/usr/bin/env node` shebang. So technically it doesn't need to have a separate profile. But I wanted to try a more tightly defined private-bin later on where it could prove useful. Just haven't gotten around to that yet :-)

etc/profile-m-z/nodejs-common.profile
rusty-snake: FWIW: The only case where `ignore read-only` should be used instead of `read-write` are redirect profiles IMHO.

PR #4257, opened 2021-05-07
read-write fixes

glitsj16:
Impacts comments only, no functional changes.


rusty-snake (changes requested):
`read-only`/`read-write` have a last-win logic. If you add `read-write` to a .local it has no-effect because the it must come after `read-only`.

Corrected:
> FWIW: The only case where `ignore read-only` should be used instead of `read-write` are redirect profiles [and .locals] IMHO.

PR #4209, opened 2021-04-23
Map /dev/input with "--private-dev", add "--no-input" option to disable it

davidebeatrici:
Fixes #2203.

---

By default only joystick devices (`/dev/input/js*`) can be accessed.
At least, that's the case on Debian: the other entries have more restrictive permissions.
The original owner and group are `root` and `input`, respectively.
However, until we have granular input control options, allowing access to joysticks only is better than nothing.

```bash
$ ls -l /dev
total 0
lrwxrwxrwx  1 nobody nogroup      8 23 apr 07.22 cdrom -> /dev/sr0
lrwxrwxrwx  1 nobody nogroup      8 23 apr 07.22 cdrw -> /dev/sr0
drwxr-xr-x  3 nobody nogroup    100 22 apr 19.18 dri
lrwxrwxrwx  1 nobody nogroup      8 23 apr 07.22 dvd -> /dev/sr0
lrwxrwxrwx  1 nobody nogroup      8 23 apr 07.22 dvdrw -> /dev/sr0
lrwxrwxrwx  1 nobody nogroup     13 23 apr 07.22 fd -> /proc/self/fd
crw-rw-rw-  1 nobody nogroup   1, 7 23 apr 07.22 full
crw-rw----+ 1 nobody nogroup 244, 0 22 apr 19.18 hidraw0
crw-rw----+ 1 nobody nogroup 244, 1 22 apr 19.18 hidraw1
crw-rw----+ 1 nobody nogroup 244, 2 22 apr 19.18 hidraw2
crw-rw----+ 1 nobody nogroup 244, 3 22 apr 19.18 hidraw3
crw-rw----+ 1 nobody nogroup 244, 4 22 apr 19.18 hidraw4
crw-rw----+ 1 nobody nogroup 244, 5 22 apr 19.18 hidraw5
drwxr-xr-x  4 nobody nogroup    760 23 apr 07.22 input
srw-rw-rw-  1 nobody nogroup      0 22 apr 19.18 log
crw-rw-rw-  1 nobody nogroup   1, 3 23 apr 07.22 null
lrwxrwxrwx  1 nobody nogroup     13 23 apr 07.22 ptmx -> /dev/pts/ptmx
drwxr-xr-x  2 nobody nogroup      0 23 apr 07.22 pts
crw-rw-rw-  1 nobody nogroup   1, 8 23 apr 07.22 random
drwxrwxrwt  2 nobody nogroup     40 23 apr 07.22 shm
[...]

PR #4259, opened 2021-05-08
contrib/vim: add missing noinput command to syn match

kmk3:
Added on commit a90386d77 ("Map /dev/input with "--private-dev", add
"--no-input" option to disable it") / PR #4209.  See also commit
0cee0ba5a ("Add noinput to all profiles with private-dev").

Misc: I noticed that it was missing due to the lack of syntax
highlighting on etc/profile-m-z/webstorm.profile.

Cc: @davidebeatrici (as the author of #4209)


davidebeatrici (approved):

rusty-snake (approved):
FWIW: `dbus-*` has no highlighting too.

kmk3:
Sorry for the late reply, I've been AFK.

> FWIW: `dbus-*` has no highlighting too.

Good catch.  This one is more complicated as it has 3 different forms and
arguments (compared to `noinput` having a single form with no arguments):

```
dbus-(user|system) (filter|none)
dbus-(user|system).(own|talk|see) name
dbus-(user|system).(call|broadcast) callrule
```

The commands/subcommands can be matched basically as is above, but validating
the arguments is not so easy.  Looking at profile.c and dbus.c, I managed to
turn [`check_bus_or_interface_name`][1] into a regex for `name`:

```sh
grep -E '^(\*|[A-Za-z_-][A-Za-z0-9_-]*(\.[A-Za-z_-][A-Za-z0-9_-]*)*((\.[A-Za-z_-][A-Za-z0-9_-]*)+|\.\*))$'
```

It seems to match the behavior of the function in all the tests that I came up
with, but it looks quite gigantic and the [`dbus_check_call_rule`][2] function
(for `callrule`) is even more complex than the former (though the regex above
can probably be reused in a `nextgroup=`).

I could also just make the arguments match `.*`; not sure if half-highlighting
better or worse than no highlighting.

I think I'll try it in a new PR, but feel free to discuss it here in the
meantime.

[1]: https://github.com/netblue30/firejail/blob/master/src/firejail/dbus.c#L52
[2]: https://github.com/netblue30/firejail/blob/master/src/firejail/dbus.c#L121

rusty-snake:
All newly added commands/conditions since https://github.com/netblue30/firejail/commit/a1cc4a5568675ab036ba942410eac5aa9b0eec7c (I found): `dbus-*`, `seccomp-error-action`, `seccomp.32*`, `allow-debuggers`, `?HAS_NOSOUND`, `?HAS_X11`, `?HAS_NET`, `private-cwd`

kmk3:
> All newly added commands/conditions since
> [a1cc4a5](https://github.com/netblue30/firejail/commit/a1cc4a5568675ab036ba942410eac5aa9b0eec7c)
> (I found): `dbus-*`, `seccomp-error-action`, `seccomp.32*`,
> `allow-debuggers`, `?HAS_NOSOUND`, `?HAS_X11`, `?HAS_NET`, `private-cwd`

Good catch.  Would you mind creating an issue/task for that (and ideally
another just for `dbus-*`, as it seems much more complex)?  This would help
make it more visible and ensure that it does not get lost over time.  And I'd
really like to archive finished PRs/issues as "done", in order to unclutter the
inbox.

In a similar vein, I plan on moving some scattered ideas from my comments to
issues for the same reasons.

Additionally, I have some other branches that I'm working on in the meantime,
so if you want to do the changes, feel free to do so.

PR #4273, opened 2021-05-13
Try to fix #2310 -- Can't create run directory without suid-root

### LINKED ISSUE #2310: Can't create run directory without suid-root
I compiled firejail with `--disable-suid` and tried running it, but it fails to start:
```
Error mkdir: util.c:1014 create_empty_dir_as_root: Permission denied
```
It tries to create its runtime hierarchy in `/run/firejail`, which is not possible as unprivileged user.
If this is not supported, it should check the uid and print a more friendly error message if the user is not root.
If an unprivileged user should be able to run it (without suid-root), maybe it can create required directories in the user's home directory.

rusty-snake:
(untested)


src/firejail/main.c
```diff
@@ -982,6 +982,14 @@ int main(int argc, char **argv, char **envp) {
 	int arg_caps_cmdline = 0; 	// caps requested on command line (used to break out of --chroot)
 	char **ptr;
 
+#ifndef HAVE_SUID
+	if (geteuid() != 0) {
+		fprintf(stderr, "Error: Firejail need to be SUID.");
```
glitsj16, line 987: nitpick/typo: Error: Firejail needs to be SUID.

glitsj16 (approved):
LGTM

src/firejail/main.c
rusty-snake: fixed, thanks.

src/firejail/main.c
```diff
@@ -982,6 +982,14 @@ int main(int argc, char **argv, char **envp) {
...
+		fprintf(stderr, "Assuming firejail is istalled in /usr/bin, execute the following command as root:");
```
reinerh, line 988: typo: installed

rusty-snake: fixed, thx.

src/firejail/main.c
```diff
@@ -982,6 +982,14 @@ int main(int argc, char **argv, char **envp) {
 	int arg_caps_cmdline = 0; 	// caps requested on command line (used to break out of --chroot)
 	char **ptr;
 
+#ifndef HAVE_SUID
+	if (geteuid() != 0) {
+		fprintf(stderr, "Error: Firejail needs to be SUID.");
```
reinerh, line 987: You also want to add newlines (`\n`), as otherwise the error message looks a bit weird (everything in one line).

rusty-snake: Absolutely right.

reinerh (approved):

netblue30:
all in, thanks!

PR #4278, opened 2021-05-14
rename noautopulse to keep-config-pulse

kmk3:
Changes:

* add the keep-config-pulse option
* make noautopulse an alias for keep-config-pulse
* deprecate the noautopulse option
* misc: fix indentation of --keep-dev-shm on src/firejail/usage.c

Even though noautopulse is not intended for hardening, it looks like it
is, because it starts with "no", just like no3d, noroot, etc).  In fact,
it is the only "no" option that differs in such a way.

And it has been accidentally misused as such before; see PR #4269 and
commit e4beaeaa8 ("drop noautopulse from agetpkg").

So effectively rename it to keep-config-pulse in order to avoid
confusion.  This is similar to the keep-var-tmp and keep-dev-shm
options, which are used to "leave a path alone", just like noautopulse.

Note: The changes on this patch are based on the ones from commit
617ff40c9 ("add --noautopulse arg for complex pulse setups") / PR #1854.

See #4269 for the discussion.

Cc: @vermeeren (as the author of #1854)


rusty-snake (commented):
:+1: 

OT: `_firejail.in` and `usage.c` are not aware of `noinput`. I'll add them.

src/firejail/profile.c
```diff
@@ -423,7 +423,7 @@ int profile_check_line(char *ptr, int lineno, const char *fname) {
 		return 0;
 	}
 	else if (strcmp(ptr, "noautopulse") == 0) {
```
glitsj16, line 425: Does this need changing too?

glitsj16 (approved):
LGTM

src/firejail/profile.c
rusty-snake: Nope, changing this would mean to hard-fail (with invalid/unknow-command) if a user has `noautopulse` in their .locals.

glitsj16: Thanks for clearing things up. Looks ready for merge!

vermeeren (approved):

Fred-Barclay (approved):

PR #4283, opened 2021-05-16
Fix Lutris profile

RandomVoid:
Recently I updated my OS to Ubuntu 21.04 and I discovered that profile for Lutris has been added. Unfortunately Lutris v0.5.8.3 didn't start when I tried to use it. After spending some time on debugging I found two problems (both are fixed in this pull request):
1. Lutris is using ldconfig, so it requires access to /sbin and /usr/sbin. I've added noblacklist option for both paths.
2. Games with enabled Feral GameMode won't run if D-Bus is disabled. I've disabled options dbus-user and dbus-system.

### DIFF
--- a/etc/profile-a-l/lutris.profile
+++ b/etc/profile-a-l/lutris.profile
@@ -14,6 +14,10 @@ noblacklist ${HOME}/.config/lutris
 noblacklist ${HOME}/.local/share/lutris
 # noblacklist ${HOME}/.wine
 noblacklist /tmp/.wine-*
+# Don't block access to /sbin and /usr/sbin to allow using ldconfig. Otherwise
+# Lutris won't even start.
+noblacklist /sbin
+noblacklist /usr/sbin
 
 ignore noexec ${HOME}
 
--- a/etc/profile-a-l/lutris.profile
+++ b/etc/profile-a-l/lutris.profile
@@ -74,5 +74,7 @@ shell none
 #private-dev
 private-tmp
 
-dbus-user none
+dbus-user filter
+dbus-user.own net.lutris.Lutris
+dbus-user.talk com.feralinteractive.GameMode
 dbus-system none


etc/profile-a-l/lutris.profile
```diff
@@ -70,5 +74,6 @@ shell none
 #private-dev
 private-tmp
 
-dbus-user none
-dbus-system none
+# Games with enabled Feral GameMode won't run if D-Bus system is disabled.
+#dbus-user none
+#dbus-system none
```
rusty-snake, line 79: Does `dbus-system none` really break gamemode?

etc/profile-a-l/lutris.profile
```diff
@@ -70,5 +74,6 @@ shell none
 #private-dev
 private-tmp
 
-dbus-user none
-dbus-system none
+# Games with enabled Feral GameMode won't run if D-Bus system is disabled.
+#dbus-user none
```
rusty-snake, line 78: Does this work?
```suggestion
dbus-user filter
dbus-user.talk com.feralinteractive.GameMode
```

rusty-snake (commented):
LGTM, but I think we can keep some of the dbus restrictions.

etc/profile-a-l/lutris.profile
RandomVoid: You're right. `dbus-system none` doesn't break GameMode.

rusty-snake:
FWIW: I did some playing with `gamemoderun` in #4201.

etc/profile-a-l/lutris.profile
RandomVoid: Unfortunately Lutris won't start with this options. Following message is logged:
```
(lutris:5): dbind-WARNING **: 18:17:44.334: Couldn't connect to accessibility bus: Failed to connect to socket /tmp/dbus-Gky7oRM3c7: No such file or directory
```

rusty-snake: Try this, if lutris still don't work we need to go with `#dbus-user none`.
```
dbus-user filter
dbus-user.own net.lutris.Lutris
dbus-user.talk com.feralinteractive.GameMode
#dbus-user.talk com.deepin.WMSwitcher
#dbus-user.talk org.gnome.Mutter.DisplayConfig
#dbus-user.talk org.kde.KWin
#dbus-user.talk org.freedesktop.ScreenSaver
#dbus-user.talk org.mate.ScreenSaver
#dbus-user.talk org.xfce.ScreenSaver
```

etc/profile-a-l/lutris.profile
RandomVoid: I've updated commit.

etc/profile-a-l/lutris.profile
RandomVoid: Thanks for suggestion. Following configuration works correctly:
```
dbus-user filter
dbus-user.own net.lutris.Lutris
dbus-user.talk com.feralinteractive.GameMode
```
I've updated commit.

etc/profile-a-l/lutris.profile
```diff
@@ -70,5 +74,8 @@ shell none
 #private-dev
 private-tmp
 
-dbus-user none
+# Games with enabled Feral GameMode won't run if D-Bus user is none.
```
rusty-snake, line 77: IMHO no need
```suggestion
```

rusty-snake (approved):

etc/profile-a-l/lutris.profile
RandomVoid: OK, commit updated.

netblue30:
all in, thanks!

PR #4284, opened 2021-05-16
Add cargo.profile

rusty-snake:
@reinerh here is one :wink: (#4281)

### DIFF
--- a/etc/profile-a-l/cargo.profile
+++ b/etc/profile-a-l/cargo.profile
@@ -1,0 +1,80 @@
+# Firejail profile for cargo
+# Description: The Rust package manager
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include cargo.local
+# Persistent global definitions
+include globals.local
+
+ignore noexec ${HOME}
+ignore noexec /tmp
+
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}
+
+noblacklist ${HOME}/.cargo/credentials
+noblacklist ${HOME}/.cargo/credentials.toml
+
+# Allows files commonly used by IDEs
+include allow-common-devel.inc
+
+# Allow lua (blacklisted by disable-interpreters.inc)
+#include allow-lua.inc
+
+# Allow perl (blacklisted by disable-interpreters.inc)
+#include allow-perl.inc
+
+# Allow python (blacklisted by disable-interpreters.inc)
+#include allow-python2.inc
+#include allow-python3.inc
+
+# Allow ssh (blacklisted by disable-common.inc)
+#include allow-ssh.inc
+
+include disable-common.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+#mkdir ${HOME}/.cargo
+#whitelist ${HOME}/YOUR_CARGO_PROJECTS
+#whitelist ${HOME}/.cargo
+#whitelist ${HOME}/.rustup
+#include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+caps.drop all
+ipc-namespace
+machine-id
+netfilter
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+
+disable-mnt
+#private-bin cargo,rustc
+private-cache
+private-dev
+private-tmp
+
+dbus-user none
+dbus-system none
+
+read-write ${HOME}/.cargo/bin
--- a/etc/profile-a-l/cargo.profile
+++ b/etc/profile-a-l/cargo.profile
@@ -19,16 +19,6 @@ noblacklist ${HOME}/.cargo/credentials.toml
 # Allows files commonly used by IDEs
 include allow-common-devel.inc
 
-# Allow lua (blacklisted by disable-interpreters.inc)
-#include allow-lua.inc
-
-# Allow perl (blacklisted by disable-interpreters.inc)
-#include allow-perl.inc
-
-# Allow python (blacklisted by disable-interpreters.inc)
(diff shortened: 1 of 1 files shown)


reinerh (approved):

reinerh:
Just tested your profile with a normal/harmless rust project, and once with the bad POC with malicious macros.
Works fine and it prevented the POC from reading my SSH key. :+1:

etc/profile-a-l/cargo.profile
```diff
@@ -0,0 +1,80 @@
...
+protocol unix,inet,inet6
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+
+disable-mnt
+#private-bin cargo,rustc
+private-cache
+private-dev
```
glitsj16, line 74: no private-etc?

etc/profile-a-l/cargo.profile
```diff
@@ -0,0 +1,80 @@
...
+private-tmp
+
+dbus-user none
+dbus-system none
+
```
glitsj16, line 79: can we add mdwe?

etc/profile-a-l/cargo.profile
```diff
@@ -0,0 +1,80 @@
...
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}
+
+noblacklist ${HOME}/.cargo/credentials
+noblacklist ${HOME}/.cargo/credentials.toml
+
+# Allows files commonly used by IDEs
+include allow-common-devel.inc
+
+# Allow lua (blacklisted by disable-interpreters.inc)
```
glitsj16, line 22: I can see the additional allow-foo.inc's are commented. Does cargo _ever_ need anything from lua, perl and/or python?

etc/profile-a-l/cargo.profile
rusty-snake: WFM:
```
private-etc alternatives,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,group,magic,magic.mgc,passwd,ca-certificates,crypto-policies,host.conf,hostname,hosts,nsswitch.conf,pki,protocols,resolv.conf,services,rpc,ssl
```

etc/profile-a-l/cargo.profile
rusty-snake: WFM too

etc/profile-a-l/cargo.profile
rusty-snake: lua,perl,python: if you project needs them
ssh: dependencies in private git repositories like

```
[dependencies.foo]
git = "ssh://git@github.com/bar/foo.git"
```

etc/profile-a-l/cargo.profile
```diff
@@ -0,0 +1,80 @@
...
+ignore noexec /tmp
+
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}
+
+noblacklist ${HOME}/.cargo/credentials
+noblacklist ${HOME}/.cargo/credentials.toml
+
+# Allows files commonly used by IDEs
+include allow-common-devel.inc
```
glitsj16, line 20: There's this note in the [cargo book](https://doc.rust-lang.org/cargo/reference/config.html):
```
Note: Cargo also reads config files without the .toml extension, such as .cargo/config. Support for the .toml
extension was added in version 1.39 and is the preferred form. If both files exist, Cargo will use the file
without the extension.
```
Currently we have mixed support for ${HOME}/.cargo/{config,crates.toml} in allow-common-devel.inc and disable-programs.inc. Perhaps we could cover both formats? I mean, add ${HOME}/.cargo/{config.toml,crates} to both files...

glitsj16 (approved):
LGTM. Just a few questions on things I'm not sure off...

etc/profile-a-l/cargo.profile
glitsj16: Same here. I think we can add it.

etc/profile-a-l/cargo.profile
glitsj16: WFM too. Always nice to have mdwe IMO.

etc/profile-a-l/cargo.profile
glitsj16: Thanks for clearing that up. I'm happy with this now I understand.

etc/profile-a-l/cargo.profile
```diff
@@ -0,0 +1,80 @@
...
+
+#mkdir ${HOME}/.cargo
+#whitelist ${HOME}/YOUR_CARGO_PROJECTS
+#whitelist ${HOME}/.cargo
+#whitelist ${HOME}/.rustup
+#include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
```
glitsj16, line 50: FWIW I tested apparmor support. All it takes is a comment here and a few additional lines in /etc/apparmor.d/local/firejail-default. Maybe I'm overdoing it, what do you think?

Suggestion
```
# NOTE: to enable apparmor you will need to uncomment and edit
# the 'cargo + rustup' rules in /etc/apparmor.d/local/firejail-default.
# Then you can add 'apparmor' to your cargo.local.
#apparmor
```

For [firejail-local](https://github.com/netblue30/firejail/blob/master/etc/apparmor/firejail-local):
```
## Uncomment to opt-in to apparmor for 'cargo + rustup'.
#+ If you use whitelisting change ${HOME}/YOUR_CARGO_PROJECTS below to your Rust projects directory.
#+ Otherwise you can keep that line commented.
#owner @{HOME}/.cargo/bin/*/** ix,
#owner @{HOME}/.cargo/env ix,
#owner @{HOME}/.rustup/toolchains/*/** ix,
#owner @{HOME}/YOUR_CARGO_PROJECTS/** ix,
```

rusty-snake: `cargo install` needs `/tmp` mounted `exec` IIRC. Just saying, I don't even know if `apparmor` makes it `noexec`.

glitsj16: More testing confirms what you're saying. AA support is out.

netblue30:
all in, thanks!

PR #4287, opened 2021-05-19
Restrict /usr/libexec

rusty-snake:
WIP

### DIFF
--- a/etc/profile-a-l/0ad.profile
+++ b/etc/profile-a-l/0ad.profile
@@ -10,6 +10,8 @@ noblacklist ${HOME}/.cache/0ad
 noblacklist ${HOME}/.config/0ad
 noblacklist ${HOME}/.local/share/0ad
 
+blacklist /usr/libexec
+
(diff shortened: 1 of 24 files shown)


glitsj16 (approved):
FWIW [Arch Linux packaging guidelines](https://wiki.archlinux.org/title/Arch_package_guidelines#Package_etiquette) strongly discourage placing anything in /usr/libexec. On that OS users should already be covered. Nice addition nonetheless!

netblue30:
merged!

reinerh:
> > FWIW Arch Linux packaging guidelines strongly discourage placing anything in /usr/libexec.
> 
> Same for Debian. `/usr/libexec` is used by Fedora. And maybe OpenSUSE? IDK.

It seems like Debian is moving towards placing binaries into `/usr/libexec`, because they want to follow FHS 3.0. At least the main linting tool is now informing about that:
https://lintian.debian.org/tags/executable-in-usr-lib

PR #4288, opened 2021-05-19
Create whitelist-run-common.inc

rusty-snake:
**This needs testing. Add it to your globals.local. Does everything still work?**

If someone who use systemd-resolved can say more which resolv.conf is necessary on such system.

### DIFF
--- a/etc/inc/whitelist-run-common.inc
+++ b/etc/inc/whitelist-run-common.inc
@@ -3,6 +3,7 @@
 include whitelist-run-common.local
 
 whitelist /run/NetworkManager/resolv.conf
+whitelist /run/cups/cups.sock
 whitelist /run/dbus/system_bus_socket
 whitelist /run/systemd/resolve/resolv.conf
 whitelist /run/systemd/resolve/stub-resolv.conf
--- a/etc/inc/whitelist-run-common.inc
+++ b/etc/inc/whitelist-run-common.inc
@@ -1,0 +1,8 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+include whitelist-run-common.local
+
+whitelist /run/NetworkManager/resolv.conf
+whitelist /run/dbus/system_bus_socket
+whitelist /run/systemd/resolve/resolv.conf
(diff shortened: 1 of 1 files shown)


pirate486743186:
lol, nothing works anymore
`Error: invalid whitelist path /run/NetworkManager/resolv.conf
Error: proc 4224 cannot sync with peer: unexpected EOF
Peer 4225 unexpectedly exited with status 1
`
ubuntu 18.04 32bit, latest firejail from ppa
The path exist, actually firejail rejects all lines, not sure what's wrong.

rusty-snake:
I'm not sure which version is in the ppa but you need #4302.

rusty-snake:
I have `whitelist /run/dbus/system_bus_socket` in globals.local for over one month now and had no trouble (except for printing which is now solved by adding `whitelist /run/cups/cups.sock`). Nerveless, my resolv.conf is a file (not a symlink) for example and I don't things like kerberos, avahi, VPNs, ... so it might don't work for all system configurations. 

If nobody reports testing here, I go forward, merged this PR and start adding it to some profiles.

pirate486743186:
in order to test this, you need to compile firejail from source....
So, not a lot of potential testers...

rusty-snake:
> in order to test this, you need to compile firejail from source....

No, you can it install with your package manager in Arch Linux and Fedora Rawhide. And other distros like Debian unstable, Debian stable+backports, OpenSUSE tumbleweed, Manjaro will follow soon.

> So, not a lot of potential testers...

The most people (frequently) hanging here and reading this compile firejail from source as it looks for me. Anyway we don't need much testers a few with different distros and dns setups are enough. There are 5 whitelists ATM so there aren't to much possibilities to differ. The only thing that I worry about are resolv.conf symlinks.

glitsj16:
> This needs testing. Add it to your globals.local. Does everything still work?

FWIW, I've been running with this without any issues on Arch Linux. But my resolv.conf also is a file instead of a symlink (never used systemd-resolved) so on that part I cannot add anything relevant.

glitsj16 (approved):
LGTM

Issue #2359, opened 2019-01-21
Support subpaths in macros (like `${PICTURES}/Screenshots`)


ghost:
Hello for flameshot I need to add `mkdir ${PICTURES}/Screenshots` but unfortunately subfolders are only supported for `${HOME}`.
So with a whitelist profile the program is returning an error because there is no Screenshots folder and it looks like it can't create it, you need to create it manually (assuming that you have `noblacklist`ed and `whitelist`ed the whole `${PICTURES}`).
Also for privacy it would be better to allow only a subfolder of Pictures instead of giving access to everything.

It would be very useful, thanks.

PR #4290, opened 2021-05-20
Update Zoom profile for meeting recordings

squatched:
By default, Zoom records meetings to ${HOME}/Documents/Zoom. Add that
folder to the whitelist so that future users don't lose their meeting
recordings upon shutting Zoom down.

That was a painful loss for me today =(.

### DIFF
--- a/etc/profile-m-z/zoom.profile
+++ b/etc/profile-m-z/zoom.profile
@@ -18,15 +18,18 @@ ignore dbus-system none
 
 noblacklist ${HOME}/.config/zoomus.conf
 noblacklist ${HOME}/.zoom
+noblacklist ${HOME}/Documents
 
 nowhitelist ${DOWNLOADS}
 
 mkdir ${HOME}/.cache/zoom
 mkfile ${HOME}/.config/zoomus.conf
 mkdir ${HOME}/.zoom
+mkdir ${HOME}/Documents/Zoom
 whitelist ${HOME}/.cache/zoom
 whitelist ${HOME}/.config/zoomus.conf
 whitelist ${HOME}/.zoom
+whitelist ${HOME}/Documents/Zoom
 
 # Disable for now, see https://github.com/netblue30/firejail/issues/3726
 #private-etc alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,machine-id,nsswitch.conf,pki,resolv.conf,ssl


rusty-snake (changes requested):
#4006

> The problem is that "Documents" can also be "Dokumente", "Dokumenty", "Dokumentuak", "文档", "Документы", ... so `whitelist ${DOCUMENTS}` must be used instead [As you already discovered]. This has the drawback that `${DOCUMENTS}/Zoom` is treated literally, so zoom has full access to `~/Documents`.

I'm not sure if we want to give zoom this wide rights by default. Maybe we should add a comment that says what needs to be added to zoom.local to make meeting-recodring/notes-saving work. 

cc @VorpalBlade

squatched:
Thank you for the feedback, I feel dumb for not thinking of localization but, c'est la vie.

Regardless, if there's any way I can help bottom-out on this issue, I'd love to do it because this REALLY hurt when it happened and I'd like to help others avoid the pain-point. I'm also fairly new to firejail and am only recently really digging into what it's capable of and tweaking it to my needs so, if you're willing to help me understand, I'm super happy to help out.

Also, if this would be better suited to a bug report, I'll close this pull request then open one.

The first issue I'm running into is `Error: "${DOCUMENTS}/Zoom" is an invalid filename: rejected character: "{"` when I try and use `${DOCUMENTS}` instead of `${HOME}\Documents` (even with the `DOCUMENTS` environment variable exported explicitly and the given folder already created, I'm running Arch and my firejail version is 0.9.64.4 from the [community repo](https://archlinux.org/packages/community/x86_64/firejail/).

The second issue is... well... I'm just not sure how to accomplish the appropriate whitelisting here. I completely agree that we don't want to give Zoom access to the documents folder at large by default so... Advice is greatly appreciated.

rusty-snake:
> The first issue I'm running into is `Error: "${DOCUMENTS}/Zoom" is an invalid filename: rejected character: "{"` when I try and use `${DOCUMENTS}`

That's what I mean with "This has the drawback that `${DOCUMENTS}/Zoom` is treated literally".
Issues for it: #2359 (already linked in #4006).

> even with the `DOCUMENTS` environment variable exporte

The `${DOCUMENTS}` macro expands to `XDG_DOCUMENTS_DIR` in `~/.config/user-dirs.dirs` or hardcoded defaults.

> Also, if this would be better suited to a bug report, I'll close this pull request then open one.

We have issues for both `${DOCUMENTS}/foobar` and `Documents/Zoom`. No need.

> The second issue is... well... I'm just not sure how to accomplish the appropriate whitelisting here. I completely agree that we don't want to give Zoom access to the documents folder at large by default so... Advice is greatly appreciated.

Unless someone adds support for `${DOCUMENTS}/Zoom`, we need to use `whitelist ${DOCUMENTS}` if we want to have it working by default.

squatched:
OK, thank you for the clarification. Super appreciate it.

I'll adjust this pull request to be just comments in the config that indicate what can happen and how to fix it.

kmk3 (approved):

etc/profile-m-z/zoom.profile
```diff
@@ -18,15 +18,18 @@ ignore dbus-system none
 
 noblacklist ${HOME}/.config/zoomus.conf
 noblacklist ${HOME}/.zoom
+noblacklist ${HOME}/Documents
```
rusty-snake, line 21:
```suggestion
noblacklist ${DOCUMENTS}
```

rusty-snake (commented):
Looks fine, but it still works only with english locale.

PR #4293, opened 2021-05-22
configure*: fix build with non-bash /bin/sh

kmk3:
The configure script happens to work if /bin/sh supports the non-POSIX
"+=" operator (e.g.: bash) and fails otherwise (e.g.: dash).

This usage first appeared on configure.ac on commit 66a476419 ("gcov
support"), which is from 2016.

If the --enable-apparmor flag is passed to ./configure (which is the
default on Arch Linux), running `make` fails due to the missing
-lapparmor LDFLAG.  Thus, building firejail-git from the AUR does not
work if /bin/sh is e.g.: dash.

Errors when running the build commands below from makepkg:

    $ ./configure --prefix=/usr --enable-apparmor >/dev/null
    ./configure: 3174: EXTRA_CFLAGS+= -mindirect-branch=thunk: not found
    ./configure: 3246: EXTRA_CFLAGS+= -fstack-clash-protection: not found
    ./configure: 3282: EXTRA_CFLAGS+= -fstack-protector-strong: not found
    ./configure: 3518: EXTRA_CFLAGS+= : not found
    $ make >/dev/null
    /usr/bin/ld: apparmor.o: in function `apparmor_test':
    /tmp/firejail-git/src/firejail-git/src/jailcheck/apparmor.c:28: undefined reference to `aa_gettaskcon'
    collect2: error: ld returned 1 exit status
    make[1]: *** [Makefile:10: jailcheck] Error 1
    make: *** [Makefile:42: src/jailcheck/jailcheck] Error 2
    make: *** Waiting for unfinished jobs....
    /usr/bin/ld: apparmor.o: in function `print_apparmor':
    /tmp/firejail-git/src/firejail-git/src/firemon/apparmor.c:28: undefined reference to `aa_gettaskcon'
    collect2: error: ld returned 1 exit status
    make[1]: *** [Makefile:10: firemon] Error 1
    make: *** [Makefile:42: src/firemon/firemon] Error 2
    /usr/bin/ld: join.o: in function `extract_apparmor':
    /tmp/firejail-git/src/firejail-git/src/firejail/join.c:65: undefined reference to `aa_is_enabled'
    /usr/bin/ld: sandbox.o: in function `set_apparmor':
    /tmp/firejail-git/src/firejail-git/src/firejail/sandbox.c:133: undefined reference to `aa_change_onexec'
    collect2: error: ld returned 1 exit status
    make[1]: *** [Makefile:10: firejail] Error 1
    make: *** [Makefile:42: src/firejail/firejail] Error 2

Without the apparmor flag, the CFLAGS related to HAVE_SPECTRE do not get
applied either, but `make` does not error out, so the problem is harder
to detect in this case.

Diff comparing the output of `./configure 2>&1` when running without and
then with this patch:

    $ git --no-pager diff --no-index configure_current.log configure_patch.log
    diff --git a/configure_current.log b/configure_patch.log
    index f5e814f..099d836 100644
    --- a/configure_current.log
    +++ b/configure_patch.log
    @@ -10,12 +10,9 @@ checking for gcc option to accept ISO C89... none needed
     checking for a BSD-compatible install... /usr/bin/install -c
     checking for ranlib... ranlib
     checking whether C compiler accepts -mindirect-branch=thunk... yes
    -./configure: 3174: EXTRA_CFLAGS+= -mindirect-branch=thunk: not found
     checking whether C compiler accepts -mretpoline... no
     checking whether C compiler accepts -fstack-clash-protection... yes
    -./configure: 3246: EXTRA_CFLAGS+= -fstack-clash-protection: not found
     checking whether C compiler accepts -fstack-protector-strong... yes
    -./configure: 3282: EXTRA_CFLAGS+= -fstack-protector-strong: not found
     checking for pkg-config... /usr/bin/pkg-config
     checking pkg-config is at least version 0.9.0... yes
     checking for gawk... yes
    @@ -88,7 +85,7 @@ Configuration options:
        busybox workaround: no
        Spectre compiler patch: yes
        EXTRA_LDFLAGS:
    -   EXTRA_CFLAGS:
    +   EXTRA_CFLAGS:  -mindirect-branch=thunk -fstack-clash-protection -fstack-protector-strong
        fatal warnings:
        Gcov instrumentation:
        Install contrib scripts: yes

### DIFF
--- a/configure
+++ b/configure
@@ -3171,7 +3171,7 @@ fi
 { $as_echo "$as_me:${as_lineno-$LINENO}: result: $ax_cv_check_cflags___mindirect_branch_thunk" >&5
 $as_echo "$ax_cv_check_cflags___mindirect_branch_thunk" >&6; }
 if test "x$ax_cv_check_cflags___mindirect_branch_thunk" = xyes; then :
-  HAVE_SPECTRE="yes" && EXTRA_CFLAGS+=" -mindirect-branch=thunk"
+  HAVE_SPECTRE="yes" && EXTRA_CFLAGS="$EXTRA_CFLAGS -mindirect-branch=thunk"
 
 else
   :
@@ -3207,7 +3207,7 @@ fi
 { $as_echo "$as_me:${as_lineno-$LINENO}: result: $ax_cv_check_cflags___mretpoline" >&5
 $as_echo "$ax_cv_check_cflags___mretpoline" >&6; }
 if test "x$ax_cv_check_cflags___mretpoline" = xyes; then :
-  HAVE_SPECTRE="yes" && EXTRA_CFLAGS+=" -mretpoline"
+  HAVE_SPECTRE="yes" && EXTRA_CFLAGS="$EXTRA_CFLAGS -mretpoline"
 
 else
   :
@@ -3243,7 +3243,7 @@ fi
 { $as_echo "$as_me:${as_lineno-$LINENO}: result: $ax_cv_check_cflags___fstack_clash_protection" >&5
 $as_echo "$ax_cv_check_cflags___fstack_clash_protection" >&6; }
 if test "x$ax_cv_check_cflags___fstack_clash_protection" = xyes; then :
-  HAVE_SPECTRE="yes" && EXTRA_CFLAGS+=" -fstack-clash-protection"
+  HAVE_SPECTRE="yes" && EXTRA_CFLAGS="$EXTRA_CFLAGS -fstack-clash-protection"
 
 else
   :
@@ -3279,7 +3279,7 @@ fi
 { $as_echo "$as_me:${as_lineno-$LINENO}: result: $ax_cv_check_cflags___fstack_protector_strong" >&5
 $as_echo "$ax_cv_check_cflags___fstack_protector_strong" >&6; }
 if test "x$ax_cv_check_cflags___fstack_protector_strong" = xyes; then :
-  HAVE_SPECTRE="yes" && EXTRA_CFLAGS+=" -fstack-protector-strong"
+  HAVE_SPECTRE="yes" && EXTRA_CFLAGS="$EXTRA_CFLAGS -fstack-protector-strong"
 
 else
   :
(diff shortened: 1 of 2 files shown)


kmk3:
@reinerh By the way, how does the build work on Debian if it uses dash as
/bin/sh by default?  Is the build done with `bash ./configure` or something
like that?

I don't know much about Debian packaging, but I tried looking at [1] from [2],
and I found one such invocation, but only on a test.  From tests/network-test:

```bash
#!/bin/bash

LOGFILE="$ADTTMP/test-network.log"

# copy tests to temporary directory, as current one might be read-only
cp -a test "$ADTTMP"
cd "$ADTTMP/test"

pushd network
bash -x ./configure
runuser -u nobody -g nogroup -- bash -x ./network.sh | tee "$LOGFILE"
popd

echo "======================================"
grep "TESTING" "$LOGFILE"
echo "======================================"

[ $(grep -c "TESTING ERROR" "$LOGFILE") -gt 0 ] && exit 1

exit 0
```

While the main build looks rather default[3].  From `head -n 15 rules`:

```makefile
#!/usr/bin/make -f

#export DH_VERBOSE=1

export DEB_BUILD_MAINT_OPTIONS = hardening=+all

%:
        dh $@

override_dh_auto_configure:
        dh_auto_configure -- --enable-apparmor --enable-contrib-install=no

override_dh_fixperms-arch:
        dh_fixperms
        chmod 4755 debian/firejail/usr/bin/firejail
```

Does `dh_auto_configure` use bash internally?

[1] https://deb.debian.org/debian/pool/main/f/firejail/firejail_0.9.58.2-2+deb10u2.debian.tar.xz
[2] https://packages.debian.org/buster/firejail
[3] https://www.debian.org/doc/manuals/maint-guide/dreq.en.html#defaultrules

reinerh:
Hi @kmk3,

`dh_auto_configure` will call `configure` with certain parameters, but it does not enforce a specific shell.
Debian uses `/bin/dash` by default as `/bin/sh`, so I wondered why it's not affected by this problem.

It looks like on Debian `./configure` calls itself via bash automatically. I tried even to start `./configure` from an open dash shell, and in the process list I could still see `/bin/bash ./configure`. So I guess it is somehow wrapping itself and invoke itself with bash (if it finds that bash is available).

I'm not exactly sure where this is coming from. Maybe from autotools. But I'm quite sure this does not come from Debian tool (debhelper), I didn't find anything in the code there.

But your changes look good to me. Getting rid of bashisms is worth it, thanks!

reinerh (approved):

Fred-Barclay (approved):

Issue #4297, opened 2021-05-23
CI is broken (yet again)


kmk3:
It looks like it was broken by #4229.  Workflow run:

* <https://github.com/netblue30/firejail/runs/2534847161>

<details>
<summary>Log excerpt</summary>
<p>

```
gcc-10 -g -O2 -ggdb -W -Wall -Werror -O2 -DVERSION='"0.9.65"'  -DPREFIX='"/usr"' -DSYSCONFDIR='"/etc/firejail"' -DLIBDIR='"/usr/lib"' -DBINDIR='"/usr/bin"'  -DHAVE_OUTPUT -DHAVE_X11 -DHAVE_PRIVATE_HOME -DHAVE_APPARMOR  -DHAVE_USERTMPFS -DHAVE_DBUSPROXY -DHAVE_FIRETUNNEL -DHAVE_GLOBALCFG -DHAVE_CHROOT -DHAVE_NETWORK -DHAVE_USERNS -DHAVE_FILE_TRANSFER -DHAVE_WHITELIST -DHAVE_SELINUX  -fstack-protector-all -D_FORTIFY_SOURCE=2 -fPIE -Wformat -Wformat-security -fstack-clash-protection -fstack-protector-strong -fanalyzer   -c fs_whitelist.c -o fs_whitelist.o
In function ‘tmpfs_topdirs’:
fs_whitelist.c:398:1: error: leak of ‘topdirs’ [CWE-401] [-Werror=analyzer-malloc-leak]
  398 | }
      | ^
  ‘fs_whitelist’: events 1-4
    |
    |  523 | void fs_whitelist(void) {
    |      |      ^~~~~~~~~~~~
    |      |      |
    |      |      (1) entry to ‘fs_whitelist’
    |  524 |  ProfileEntry *entry = cfg.profile;
    |  525 |  if (!entry)
    |      |     ~ 
    |      |     |
    |      |     (2) following ‘false’ branch (when ‘entry’ is non-NULL)...
    |......
    |  528 |  if (asprintf(&runuser, "/run/user/%u", getuid()) == -1)
    |      |  ~~ ~ 
    |      |  |  |
    |      |  |  (4) following ‘false’ branch...
    |      |  (3) ...to here
    |
  ‘fs_whitelist’: event 5
    |
    |../include/common.h:39:164:
[...]

PR #4300, opened 2021-05-23
CI: build with GCC 11

reinerh:
this fixes a new false positive memory leak (#4297), but unfortunately
opens a few new false positives (#4274).
therefore let it ignore memleak checks for now, until the detection
is a bit more stable in GCC.

Fixes: #4274, #4297

ping @smitsohu @rusty-snake @glitsj16 @kmk3

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,9 +24,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: install dependencies
-      run: sudo apt-get install gcc-10 libapparmor-dev libselinux1-dev expect xzdec
+      run: sudo apt-get install gcc-11 libapparmor-dev libselinux1-dev expect xzdec
     - name: configure
-      run: CC=gcc-10 ./configure --enable-fatal-warnings --enable-analyzer --enable-apparmor --enable-selinux --prefix=/usr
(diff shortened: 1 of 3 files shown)


reinerh:
@smitsohu whitelist.exp test is now failing:
```
$ firejail --whitelist=~/fjtest-file-lnk --whitelist=~/fjtest-dir-lnk

Parent pid 19111, child pid 19112
Child process initialized in 70.08 ms
user@host:~$ TESTING ERROR 30
```

the test is doing the following:
```
# symlinks outside home to a file we don't own
send -- "rm ~/fjtest-file-lnk\r"
after 200
send -- "ln -s /etc/passwd ~/fjtest-file-lnk\r"
after 200
send -- "firejail --whitelist=~/fjtest-file-lnk --whitelist=~/fjtest-dir-lnk\r"
expect {
        timeout {puts "TESTING ERROR 30\n";exit}
        "invalid whitelist path"
}
expect {
        timeout {puts "TESTING ERROR 31\n";exit}
        "cannot sync with peer"
}
sleep 1
```
it expects that firejail throws an error, when a file is whitelisted that symlinks to outside home.
is this now considered allowed? or a bug in the whitelist?

smitsohu:
@reinerh These symlink restrictions are gone, so we will need to disable the tests.

smitsohu:
I'll try to fix the broken tests in a separate pull request. Thanks!

PR #4325, opened 2021-06-02
jail_prober: enable absolut include directives

florianbegusch:
contrib/jail_prober.py strips include directives which is what is most often the culprit. This PR remedies that.

In addition I decided to set the firejail profiles directory to '/etc/firejail'. If this is not desired please share your suggestions.

### DIFF
--- a/contrib/jail_prober.py
+++ b/contrib/jail_prober.py
@@ -70,6 +70,19 @@ def get_args(profile_path):
     return profile
 
 
+def absolute_include(word):
+    home = os.environ['HOME']
+    path = home + '/.config/firejail/'
+
+    option, filename = word.split('=')
+    absolute_filename = path + filename
+
+    if not os.path.isfile(absolute_filename):
+        absolute_filename = '${CFG}/' + filename
+
+    return option + '=' + absolute_filename
+
+
 def arg_converter(arg_list, style):
     """
     Convert between firejail command-line arguments (--example=something) and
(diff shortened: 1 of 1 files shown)


rusty-snake (commented):
:+1: but we should change the `'include' not in word` IMO. I also added two nice to have comments.

As list comprehension fan I suggest the following, but fell free to ignore it :sunglasses:  .
```python3
new_args = [
    absolute_include(word) if word.startswith('--include') else word
    for word in new_args
]
```

contrib/jail_prober.py
```diff
@@ -94,9 +94,15 @@ def arg_converter(arg_list, style):
...
-    # Remove invalid '--include' args if converting to command-line form
     elif style == 'to_commandline':
-        new_args = [word for word in new_args if 'include' not in word]
+        temp_args = []
+        for word in new_args:
+            if 'include' not in word:
+                temp_args.append(word)
+            else:
+                option, filename = word.split('=')
+                temp_args.append(option + '=' + '/etc/firejail/' + filename)
```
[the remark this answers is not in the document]
florianbegusch: thx for clarifying that one

contrib/jail_prober.py
```diff
@@ -94,11 +107,14 @@ def arg_converter(arg_list, style):
...
     elif style == 'to_commandline':
-        new_args = [word for word in new_args if 'include' not in word]
+        new_args = [
+            absolute_include(word) if word.startswith('--include')
+            else word
+            for word in new_args
+        ]
 
-    return new_args
+        return new_args
```
rusty-snake, line 117: LGTM.com alert:
```suggestion
    return new_args
```

rusty-snake (approved):

contrib/jail_prober.py
florianbegusch: glad it caught that one

rusty-snake:
Merged, thanks.

Issue #1817, opened 2018-03-16
Error running ssh commands with arguments


chtaylo2:
This might have been on the radar a couple years back: https://github.com/netblue30/firejail/issues/719

I'm trying to allow ssh login w/ a firejail shell and run commands with arguments. 
ie.:  mkdir -p /var/crash

Both of these are not working:
ssh test@0 "mkdir -p /var/crash"
firejail -c "mkdir -p /var/crash"

```
[root@centos users]# firejail --version
firejail version 0.9.44.10
[root@netdump-poc-01 users]# firejail -c "mkdir -p /var/crash"
Reading profile /etc/firejail/server.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-passwdmgr.inc

** Note: you can use --noprofile to disable server.profile **

/bin/bash: mkdir -p /var/crash: No such file or directory
```

Is there any suggested work arounds for this?

chtaylo2:
I have the same setup with /usr/bin/firejail as the shell in /etc/passwd. However, on ssh login, I get:

"/bin/bash: mkdir -p ~/crash: No such file or directory"

chtaylo2:
You're correct on the quotes with -c. It works once you remove the quotes.

```
[root@netdump-poc-01 ~]# firejail --profile=/etc/firejail/default.profile -c touch ~/LOCAL_FILE
[root@netdump-poc-01 ~]# ls -al ~/LOCAL_FILE
-rw-r--r-- 1 root root 0 Mar 30 17:53 /root/LOCAL_FILE
```

However, doing this through an SSH session it does fail. Below is a snippet: 
```
[root@vm-client ~]# ssh -qv netdump@netdump-poc-01 -i /etc/kdump_id_rsa touch /tmp/FROM_REMOTE
OpenSSH_7.4p1, OpenSSL 1.0.2k-fips  26 Jan 2017
.....
debug1: Sending environment.
debug1: Sending env LANG = en_US.UTF-8
debug1: Sending command: touch /tmp/FROM_REMOTE
/bin/bash: touch /tmp/FROM_REMOTE: No such file or directory
debug1: client_input_channel_req: channel 0 rtype exit-status reply 0
debug1: client_input_channel_req: channel 0 rtype eow@openssh.com reply 0
debug1: channel 0: free: client-session, nchannels 1
[...]

Issue #1644, opened 2017-11-13
strange nested loop in firejail 0.9.50 when trying to use it as login-shell


lowshoe:
trying to use firejail as login-shell. that worked perfectly in  0.9.44.10 but now breaks with an error. possibly related to #1326.

```sh
root@myhostname:~$ cat /etc/os-release
NAME="Oracle Linux Server"
VERSION="7.4

root@myhostname:~$ uname -a
Linux myhostname 4.1.12-103.9.2.el7uek.x86_64 #2 SMP Tue Oct 31 16:43:46 PDT 2017 x86_64 x86_64 x86_64 GNU/Linux

root@myhostname:~$ grep firejail /etc/passwd
myuser:x:1002:1002::/home/myuser:/bin/firejail

root@myhostname:~$ su -l myuser
Letzte Anmeldung: Montag, den 13. November 2017, 14:30:19 CET auf pts/2
Reading profile /etc/firejail/default.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc

** Note: you can use --noprofile to disable default.profile **

Parent pid 5181, child pid 5182
Child process initialized in 35.22 ms
Warning: an existing sandbox was detected. -l will run without any additional sandboxing features
Warning: an existing sandbox was detected. '-l'  will run without any additional sandboxing features
[..]
'"' '"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"'"' '"'"'"'"'"'"'"'"' '"'"' '  will run without any additional sandboxing features
Error cmdline_length: cmdline.c:145 build_cmdline: Argument list too long

Parent is shutting down, bye...
```

the same happens when i try to access the host through ssh with the same user. I also tried with firejail-0.9.48 but the error is the same.

netblue30:
We've seen this one coming and going., definitely a bug!

lowshoe:
just tested with ssh. unfortunately yes.
```
ssh user@host
user@host's password: 
Last login: Wed Aug  8 13:43:15 2018 from myworkstation
Warning: an existing sandbox was detected. -l will run without any additional sandboxing features
Warning: an existing sandbox was detected. '-l'  will run without any additional sandboxing features
Warning: an existing sandbox was detected. "'"'-l'"'"' '  will run without any additional sandboxing features
[...]

Issue #887, opened 2016-11-01
restricted shell bug


dmio:
It seems, that after 0.9.34 version, firejail unable to execute commands with args thru restricted shell.
Examples:
```
# cat /etc/passwd|grep firejail
test:x:1001:1001::/home/test:/usr/bin/firejail
# su -l test -c "ls -l"
Autoselecting /bin/bash as shell
Command name #ls#
...
Running 'ls -l'  command through /bin/bash
execvp argument 0: /bin/bash
execvp argument 1: -c
execvp argument 2: 'ls -l' 
/bin/bash: ls -l: command not found
# firejail -c "date && date"
/bin/bash: date && date: command not found
```
Same with ssh:
```
% ssh test@localhost ls -l
/bin/bash: ls -l: command not found
```
Git over ssh:
```
% LANG=c git push origin
/bin/bash: git-receive-pack '/home/git/my/xyz.git': No such file or directory
fatal: Could not read from remote repository.
```

chtaylo2:
+1 for this issue.  I'd like to use firejail for a remote kdump server, so being able to execute remote commands would be great. Thanks for the great work on this tool.

jsquyres:
See #4326 for a possible solution.

rusty-snake:
This issue is fixed by #4326 as I understand.

PR #4326, opened 2021-06-02
cmdline.c: optionally quote the resulting command line

jsquyres:
The commit message contains an abbreviated version of this explanation.  This PR is a possible solution for #1644, #887, and #1817.

-----

# Short version

`sshd` invokes firejail slightly differently than how firejail is normally invoked on the command line.  One possible fix is to recognize that firejail was launched by sshd and slightly modify its quoting behavior.

# More details

There's a difference between invoking firejail with a multi-token command:

```sh
# Using --quiet for brevity
$ firejail --quiet echo hello world
hello world
```

and ssh'ing in a non-interactive multi-token command with firejail as the login shell (assume a trivial `username: --shell=/bin/bash` line in `/etc/firejail/login.users`):

```
$ ssh username@othernode echo hello world
/bin/bash: echo hello world: command not found
```

The difference is that in the `ssh` case, the local `sshd` on `somenode` will invoke the following:

* `argv[0]`: `/usr/bin/firejail`
* `argv[1]`: `-c`
* `argv[2]`: `echo hello world`

Notice how `echo hello world` is a single token (`argv[2]`). 

For simplicity of explanation, let's take `ssh`/`sshd` out of the situation and just explain two different command line cases.

## Case 1: a single token

Consider:

```sh
# Using --quiet for brevity
$ firejail --quiet -c "echo hello world"
/bin/bash: echo hello world: command not found
```

This is equivalent to what `sshd` does: `echo hello world` is a single token (`argv[3]`, in this example).

Firejail ultimately invokes `src/firejail/cmdline.c:quote_cmdline()` at https://github.com/netblue30/firejail/blob/4522ccb4ef529fe9cafa60178027be139ce0e592/src/firejail/cmdline.c#L67 

`quote_cmdline()` adds an additional set of single quotes around tokens.  The user's shell -- let's assume it's Bash -- is ultimately executed as:

* `argv[0]`: `/bin/bash`
* `argv[1]`: `-c`
* `argv[2]`: `'echo hello world' `

With the extra quotes, Bash won't split up `argv[2]`, and therefore won't find an intrinsic or executable named `echo hello world `.  Bash then rightfully generates an error.

## Case 2: multiple tokens

Consider:

```sh
# Using --quiet for brevity
$ firejail --quiet -c echo hello world
hello world
```

Note that `echo` `hello` `world` are all their own `argv` tokens (`argv[3]` through `argv[5]` in this example).  Firejail's `quote_cmdline()` behavior is therefore different; bash is ultimately executed as:

* `argv[0]`: `/bin/bash`
* `argv[1]`: `-c`
* `argv[2]`: `'echo' 'hello' 'world' `

Bash therefore is able to split up the resulting `argv[2]`, find an `echo` intrinsic or executable, and things proceed as expected (i.e., `hello world` is emitted and we get an exit status of 0).

## Proposed solution

One possible solution is to have `quote_cmdline()` behave differently depending on whether Firejail was invoked by `sshd` or not.

This PR uses the already-set `parent_sshd` flag to tell `build_cmdline()` (which, in turn, invokes `quote_cmdline()`) to *not* add extra quotes around the all-the-tokens-are-in-a-single-argv[x] token.  Bash is therefore ultimately invoked as:

* `argv[0]`: `/bin/bash`
* `argv[1]`: `-c`
* `argv[2]`: `echo hello world `

Which, while this is slightly different than case 2, above (i.e., there's no extra quotes around each sub-token), seems to be sufficient.

### DIFF
--- a/src/firejail/cmdline.c
+++ b/src/firejail/cmdline.c
@@ -26,7 +26,7 @@
 #include <assert.h>
 #include <errno.h>
 
-static int cmdline_length(int argc, char **argv, int index) {
+static int cmdline_length(int argc, char **argv, int index, bool want_extra_quotes) {
 	assert(index != -1);
 
 	unsigned i,j;
@@ -46,10 +46,11 @@ static int cmdline_length(int argc, char **argv, int index) {
 					len += 3;
 				in_quotes = false;
 			} else {
-				if (!in_quotes)
+				if (!in_quotes && want_extra_quotes)
 					len++;
 				len++;
-				in_quotes = true;
+				if (want_extra_quotes)
+					in_quotes = true;
 			}
 		}
 		if (in_quotes) {
@@ -64,7 +65,7 @@ static int cmdline_length(int argc, char **argv, int index) {
 	return len;
 }
 
-static void quote_cmdline(char *command_line, char *window_title, int len, int argc, char **argv, int index) {
+static void quote_cmdline(char *command_line, char *window_title, int len, int argc, char **argv, int index, bool want_extra_quotes) {
 	assert(index != -1);
 
 	unsigned i,j;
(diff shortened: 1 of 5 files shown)


glitsj16 (commented):
LGTM. Squashing the referenced bugs/issues looks awesome!

netblue30:
Thanks!

rusty-snake:
Can we #1644, #887 and #1817?

jsquyres:
Should I back-port this to the `LTSbase` branch?

Issue #4294, opened 2021-05-22
Firejail broke latest Bitwarden by blocking network access 


iandstanley:
Not sure if this is the fault of firejail's default profiles or whether it is the packaging made for Debian 10 of the firejail-profiles

firejail/stable,stable 0.9.58.2-2+deb10u2 amd64

firejail-profiles/stable,stable 0.9.58.2-2+deb10u2 a

**Bug and expected behavior**
- Describe the bug.
Firejail fails to launch Bitwarden 1.24.6

Firejail complains that networking feature is disabled (which breaks the vault safe sync)

```
Warning: networking feature is disabled in Firejail configuration file
Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
```

then complains about it's own config and then aborts.

```
Warning: not remounting /run/user/1000/gvfs
Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
Child process initialized in 1095.02 ms
[17:0522/205528.592544:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/Bitwarden/chrome-sandbox is owned by root and has mode 4755.
```

I suspect that if u2f/yubikey/fido2 is enabled on the client this will break as well 

- What did you expect to happen?
Bitwarden to just load 

** ERROR LOG**
```
Warning: not remounting /run/user/1000/gvfs
Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
Child process initialized in 1095.02 ms
[17:0522/205528.592544:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/Bitwarden/chrome-sandbox is owned by root and has mode 4755.
```

the chrome-sandbox is owned by root and is 4755 

I checked the chrome-sandbox 

**No profile and disabling firejail**
- What changed calling `firejail --noprofile /path/to/program` in a terminal?
bitwarden loads and runs normally and can sync to online vault.

- What changed calling the program by path (e.g. `/usr/bin/vlc`)?
/Bitwarden-1.24.6-amd64.deb

**Reproduce**

sudo apt install firejail firejail-profiles
sudo firecfg 

run bitwarden from either launch icon or command line 

**Environment**
[...]

PR #4332, opened 2021-06-04
Refactor bitwarden as electron redirect

rusty-snake:
See #4294.

@glitsj16 @iandstanley can you say something about the `ignore` block.

```
ignore include whitelist-runuser-common.inc
ignore include whitelist-usr-share-common.inc
ignore disable-mnt
ignore dbus-user none
ignore dbus-system none
```

### DIFF
--- a/etc/profile-a-l/bitwarden.profile
+++ b/etc/profile-a-l/bitwarden.profile
@@ -6,11 +6,8 @@ include bitwarden.local
 # Persistent global definitions
 include globals.local
 
-ignore include whitelist-runuser-common.inc
+# Disabled until someone reported positive feedback
 ignore include whitelist-usr-share-common.inc
-ignore disable-mnt
-ignore dbus-user none
-ignore dbus-system none
 
 ignore noexec /tmp
 
--- a/etc/profile-a-l/bitwarden.profile
+++ b/etc/profile-a-l/bitwarden.profile
@@ -6,54 +6,28 @@ include bitwarden.local
 # Persistent global definitions
 include globals.local
 
+ignore include whitelist-runuser-common.inc
+ignore include whitelist-usr-share-common.inc
+ignore disable-mnt
+ignore dbus-user none
+ignore dbus-system none
+
 ignore noexec /tmp
 
 noblacklist ${HOME}/.config/Bitwarden
 
-include disable-common.inc
-include disable-devel.inc
-include disable-exec.inc
-include disable-interpreters.inc
-include disable-passwdmgr.inc
-include disable-programs.inc
 include disable-shell.inc
-include disable-xdg.inc
 
 mkdir ${HOME}/.config/Bitwarden
 whitelist ${HOME}/.config/Bitwarden
-whitelist ${DOWNLOADS}
-include whitelist-common.inc
-include whitelist-var-common.inc
 
-apparmor
-caps.drop all
 machine-id
-netfilter
 no3d
-nodvd
-nogroups
-noinput
-nonewprivs
-noroot
 nosound
-notv
-nou2f
-novideo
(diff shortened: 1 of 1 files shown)


glitsj16:
@rusty-snake I haven't used the desktop app in a long time and don't have it installed anymore. But I downloaded the AppImage (version 1.26.5) to test this refactored profile. Here are my observations of running it for a while and testing functionality.

(1) I cannot reproduce #4294. The app works fine here with the existing bitwarden.profile, vault syncing included. I've confirmed this with both `network no/yes` in firejail.config. Whether that's due to having version 1.26.5 while the OP references 1.24.6 is not something I can judge. Upstream just mentions 'Bug Fixes' on their [release](https://github.com/bitwarden/desktop/releases) page.

> Latest Bitwarden .DEB downloaded from Bitwarden's own website Bitwarden-1.24.6-amd64.deb

Could be a typo, but in any case 1.24.6 was released  on Jan 26 and there have been updates since then.

Regarding the `ignore` block:
```
disable-mnt
dbus-user none
dbus-system none
```
These can be used, I couldn't detect anything breaking. That includes tray functionality.

Additionally I could use private-dev with the AppImage as well, so the conditional
#?HAS_APPIMAGE: ignore private-dev
doesn't seem needed...

One other thing I noticed is that the current bitwarden.profile has
caps.drop all
while electron.profile uses
caps.keep sys_admin,sys_chroot
For me, again, using the AppImage, it works fine with caps.drop all...

Hopefully @iandstanley can clear up some of all this.

rusty-snake:
Thanks. Here are some notes on that
1. This PR does not try to fix the network issue from #4294, just the it's a elcetron program issue.
2. Good to here that we can remove the `ignore` block. However if you only tested the AI I will keep to `ignore` wusc.
3. Of course a AI can not contains a privileged (e.g suid) program. Therefore it work with `caps.drop all` (and `--appimage` enforces it anyway).

iandstanley:
Hic @rusty-snake 

I'm in the same position as I was forced to uninstall firejail as the debian packages broke my daily drivers (without Bitwarden I couldn't do anything).

rusty-snake:
> I'm in the same position as I was forced to uninstall firejail as the debian packages broke my daily drivers (without Bitwarden I couldn't do anything).

:confused: Why uninstalling firejail if one program fails? Why not just use bitwarden w/o firejail and keep it for everything else?

iandstanley:
> > I'm in the same position as I was forced to uninstall firejail as the debian packages broke my daily drivers (without Bitwarden I couldn't do anything).
> 
>  Why uninstalling firejail if one program fails? Why not just use bitwarden w/o firejail and keep it for everything else?

Just way to busy to spend extensive time debugging. as it stopped me dead in my tracks.

The combination of bitwarden app breaking and firefox not allowing me to use fido2 second factors meant I had no access to any of my passwords and couldn't log into virtually every online account I have. 

I'd already lost a couple of days trying to debug what was going on and deadlines were looming.

Normally I would persevere through issues to fix them (like the issue I had trying to use a GPG key in Yubikey on a second machine - after nearly a month I managed to debug  the issue down to an issue with pinentry and /dev/pts/? permissions )
But this one hitting all access to my passwords reduced productivity to 0% which can only last so long before the world starts shouting.

Once the pressure is off again I will come back to it ... it's a great technology

netblue30:
@rusty-snake: is it done? are you merging? I'm asking because I have a release coming in a few days!

rusty-snake:
Yes, the current state can be merged. Enabled wusc can be done later, there are more electron redirects with a bigger "Disabled until someone reported positive feedback" block.

netblue30:
cool!

Issue #3219, opened 2020-02-11
Crashes with an AMD GPU with Mesa >= 19.3.4 and seccomp


creideiki:
Since not long ago (I unfortunately don't have exact dates or versions for when it happened, but I think it started with Firefox 72.0.2), Firefox hangs at startup under Firejail. This happens on two machines with AMD GPUs, but not on three others with Intel GPUs. All five systems are running up-to-date Gentoo Linux unstable.

Trying on a completely empty profile directory, Firefox gets a little bit through its startup:
```
Reading profile /etc/firejail/firefox.profile
Reading profile /etc/firejail/whitelist-usr-share-common.inc
Reading profile /etc/firejail/firefox-common.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-exec.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-common.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Warning: noroot option is not available
Parent pid 12102, child pid 12103
Warning: An abstract unix socket for session D-BUS might still be available. Use --net or remove unix from --protocol set.
Post-exec seccomp protector enabled
Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
Child process initialized in 117.95 ms
1581450106377   addons.webextension.doh-rollout@mozilla.org     WARN    Loading extension 'doh-rollout@mozilla.org': Reading manifest: Invalid extension permission: networkStatus
```
And then hangs. The process tree in the sandbox looks like this:
```
 ~ $ ps -A --forest -o pid,comm
  PID COMMAND
    1 firejail
    9 firefox
   57  \_ GPU Process
```
And all threads in the GPU process are hung:
```
 ~ # strace -f -p 12160
strace: Process 12160 attached with 3 threads
[pid 12163] futex(0x7f54e41feb78, FUTEX_WAIT_PRIVATE, 0, NULL <unfinished ...>
[pid 12162] epoll_wait(6,  <unfinished ...>
[pid 12160] restart_syscall(<... resuming interrupted read ...>^Cstrace: Process 12160 detached
```
As are the ones in the main Firefox process:
```
 ~ # strace -f -p 12112
strace: Process 12112 attached with 41 threads
[...]

PR #4340, opened 2021-06-05
augment seccomp lists in firejail.config

smitsohu:
Contributes to #3219.

I didn't know where to put kcmp, so still no full solution.

### DIFF
--- a/src/firejail/checkcfg.c
+++ b/src/firejail/checkcfg.c
@@ -35,6 +35,7 @@ char *xvfb_extra_params = "";
 char *netfilter_default = NULL;
 unsigned long join_timeout = 5000000; // microseconds
 char *config_seccomp_error_action_str = "EPERM";
+char *config_seccomp_filter_add = NULL;
 char **whitelist_reject_topdirs = NULL;
 
 int checkcfg(int val) {
@@ -221,6 +222,10 @@ int checkcfg(int val) {
 			// timeout for join option
 			else if (strncmp(ptr, "join-timeout ", 13) == 0)
 				join_timeout = strtoul(ptr + 13, NULL, 10) * 1000000; // seconds to microseconds
+
+			// add rules to default seccomp filter
+			else if (strncmp(ptr, "seccomp-filter-add ", 19) == 0)
+				config_seccomp_filter_add = seccomp_check_list(ptr + 19);
 
 			// seccomp error action
 			else if (strncmp(ptr, "seccomp-error-action ", 21) == 0) {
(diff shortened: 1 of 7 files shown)


smitsohu:
There is still a bug, where removing syscalls leads to empty seccomp lists (and these are not allowed). I'll add a minimal fix.

topimiettinen (commented):
Nice idea! This allows the local admin or distro to make changes which may not make sense for others.

src/firejail/checkcfg.c
```diff
@@ -222,6 +223,10 @@ int checkcfg(int val) {
 			else if (strncmp(ptr, "join-timeout ", 13) == 0)
 				join_timeout = strtoul(ptr + 13, NULL, 10) * 1000000; // seconds to microseconds
 
+			// modify default seccomp list
+			else if (strncmp(ptr, "seccomp-list-modify ", 20) == 0)
+				config_seccomp_list_modify = seccomp_check_list(ptr + 20);
```
topimiettinen, line 228: If `seccomp-list-modify` is specified more than once, the last entry wins. It would be better if all entries would be honored. Also the memory is not free()d.

topimiettinen (commented):
Command line version would be nice and of course documentation is a must. Otherwise looks great.

topimiettinen:
> There is still a bug, where removing syscalls leads to empty seccomp lists (and these are not allowed). I'll add a minimal fix.

Perhaps the empty seccomp lists could be forbidden when user specifies them but allowed for internal use (or when the architecture specific version is empty etc).

src/firejail/checkcfg.c
smitsohu: Hm, it seems that firejail.config in general doesn't expect settings to be specified more than once currently. I can add that if wanted, but would prefer to collect the changes in a separate pull request.

smitsohu:
> Command line version would be nice

Command line version exists already: `--seccomp=` 

> and of course documentation is a must

Done.

Unlrelated to the changed code, I moved some parts of `--seccomp` documentation to `--seccomp=` because I had the impression it fits better there.

> Perhaps the empty seccomp lists could be forbidden when user specifies them but allowed for internal use (or when the architecture specific version is empty etc).

Done (for default+drop list).

src/firejail/checkcfg.c
topimiettinen: Sorry, I was thinking about the command line switch. Yes, it's indeed material for another PR.

PR #4345, opened 2021-06-06
Add allow-pandoc.local

florianbegusch:


etc/inc/allow-pandoc.inc
```diff
@@ -0,0 +1,9 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+include allow-pandoc.local
+
+noblacklist ${PATH}/pandoc
+noblacklist /usr/bin/pandoc
+noblacklist /usr/lib/ghc-*
+noblacklist /usr/lib/*pandoc*
```
rusty-snake, line 8: Never `blacklist`ed. None of them.

rusty-snake (changes requested):
Where would it be used?

etc/inc/allow-pandoc.inc
```diff
@@ -0,0 +1,9 @@
...
+whitelist /usr/share/pandoc*
```
rusty-snake, line 9: Remove.

None of the `allow-*.inc` includes have `whitelist`ing options as they break profile which do not `whitelist`ing in this location (exception is `allow-opengl-game.inc` because all profiles using it have wusc+`private-bin`).

rusty-snake (changes requested):

etc/inc/allow-pandoc.inc
florianbegusch: Very true. The whitelist directive is enough.

I use pandoc to display html email in neomutt, I thought I'd share it. Might share the neomutt.profile as well.

If one does not whitelist `/usr/share/pandoc` the following happens:
```
$ neomutt
Could not find data file /usr/share/pandoc/data/abbreviations
Press any key to continue...
```

PR #4348, opened 2021-06-08
Update weechat.profile

NetSysFire:
weechat needs access to `/usr/share/weechat` if you have any global scripts installed. The directory is empty by default, so there is no additional risk here.

### DIFF
--- a/etc/profile-m-z/weechat.profile
+++ b/etc/profile-m-z/weechat.profile
@@ -11,6 +11,8 @@ noblacklist ${HOME}/.weechat
 include disable-common.inc
 include disable-programs.inc
 
+whitelist /usr/share/weechat
+
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/weechat.profile
```diff
@@ -11,6 +11,8 @@ noblacklist ${HOME}/.weechat
 include disable-common.inc
 include disable-programs.inc
 
+whitelist /usr/share/weechat
+
```
rusty-snake, line 15:
```suggestion
```

rusty-snake (approved):

etc/profile-m-z/weechat.profile
NetSysFire: imho the whitespace helps separate the logical blocks here, but this is not the right place to discuss this

reinerh (approved):

reinerh:
Thanks!

Issue #2747, opened 2019-06-04
alpine mail client terminates if run using firejail


jose1711:
To reproduce:
- install alpine (https://repo.or.cz/alpine.git/bundles)
- run `firejail --noprofile alpine`
- do not do anything, just wait
- alpine will terminate with `Parent is shutting down, bye...` message

I've seen this with Raspbian and Arch Linux

chiraag-nataraj:
Here's a debug log running `alpine` with debugging level 60.
[alpine-debug.txt](https://github.com/netblue30/firejail/files/3254043/alpine-debug.txt)

luca-rei:
I know it is an old thread but I can confirm this happen even on gentoo
from what I understand alpine need to be spawn from a subterminal, so as a workaround you could use a launcher and call that .sh from firejail. 
Another issue with alpine is that firejail do not permit to delete or rename over a file in whitelist like configuration file (.pinerc) so here the workaround is to save configuration in a file .pinerc-bck, whitelist that one and in the launcher copy that file in .pinerc and on exit copy .pinerc over .pine-bck (*copy* not rename or delete)
At the end the launcher should be something like
#!/bin/bash
cp .pinerc-bck .pinerc
/usr/bin/alpine -passfile ~/.alpinepass
cp .pinerc .pinerc-bck

and firejail profile:
noblacklist ~/.alpine-launcher
#noblacklist ~/.pinerc
noblacklist ~/.pinerc-bck
noblacklist ~/mail
noblacklist ~/mail/saved-messages
noblacklist ~/mail/sent-mail
noblacklist ~/.pine-interrupted-mail
noblacklist ~/.pine-debug1
noblacklist ~/.bashrc

noblacklist ~/.alpinepass
noblacklist ~/.alpine-smime
noblacklist ~/.alpine-smime/.pwd
noblacklist ~/.alpine-smime/.pwd/MasterPassword.crt
noblacklist ~/.alpine-smime/.pwd/MasterPassword.key
noblacklist ~/.alpine-smime/private
noblacklist ~/.alpine-smime/ca
noblacklist ~/.alpine-smime/public


include /etc/firejail/disable-common.inc
include /etc/firejail/disable-devel.inc
include /etc/firejail/disable-interpreters.inc
include /etc/firejail/disable-programs.inc
include /etc/firejail/disable-exec.inc
include /etc/firejail/disable-passwdmgr.inc
include /etc/firejail/disable-xdg.inc

[...]

PR #4350, opened 2021-06-08
creating alpine.profile

pirate486743186:
terminal email/newsgroups client.
It dates back from the early 90s,  has a lot of complex features similar to mutt, i'm only using a very small subset of them. The profile is good for simple use.

It also has some terrible folder management.

alpinef is actually a symlink to alpine

beh, there's a bug
#2747


rusty-snake (approved):

etc/profile-a-l/alpine.profile
```diff
@@ -7,6 +7,9 @@ include alpine.local
 # Persistent global definitions
 include globals.local
 
+# Workaround for bug https://github.com/netblue30/firejail/issues/2747
+# firejail --ignore="shell none" --private-bin=bash --noblacklist=/bin/bash --profile=alpine bash -c '(alpine)'
```
rusty-snake, line 11: Is `--ignore="shell none"` necessary?

pirate486743186: you need bash to launch alpine in a subshell

rusty-snake: > you need bash to launch alpine in a subshell

Yes you need a shell (bash or sh IDC) in the sandbox that launches alpine (IDK why but if this is the workaround ...). But why do you need to ignore `shell none` if you `bash -c`?

rusty-snake: `shell none` does not disable/blacklist (all) shells inside the sandbox (for that we have `disable-shell.inc`). It changes the way firejail starts the sandbox.
`firejail --noprofile --shell=none /bin/bash -c firefox` is mostly the same as `firejail --noprofile firefox` (note that there are edge cases and the /bin/bash vs. $SHELL bug).

pirate486743186: If i remove it, i get this (/usr/local/bin/bash doesn't exist)

```
Error: execute permission denied for /usr/local/bin/bash
Error: no suitable bash executable found
```

rusty-snake: Why is your bash at `/usr/local/bin/bash`? What does `firejail --ignore=private-bin --profile=alpine which -a bash` and  `firejail --ignore=private-bin --profile=alpine ls -l ADD_BASH_PATH_HERE` show?

pirate486743186: there's no bash executable there

```
$ firejail --ignore=private-bin --profile=alpine which -a bash
```

```
$ firejail --ignore=private-bin --profile=alpine ls -l /bin/bash
-r-------- 1 nobody nogroup 0 Jun 10 12:14 /bin/bash
```


```
$ firejail --ignore=private-bin --profile=alpine ls -l /usr/local/bin/bash
ls: cannot access '/usr/local/bin/bash': No such file or directory
```







firejail from the ppa, 32 bit ubuntu 18.04
```
$ firejail --version
firejail version 0.9.64.4

Compile time support:
	- AppArmor support is enabled
	- AppImage support is enabled
	- chroot support is enabled
	- SELinux support is enabled
```

rusty-snake: This isn't caused by `shell=none` (try `firejail --noprofile --shell=none ls -l /bin/bash`). `ingore=shell=none` makes it work as a (not protable?) side effect I guess.

```suggestion
# firejail --private-bin=sh --include='${CFG}/allow-bin-sh.inc' --profile=alpine sh -c '(alpine)'
```

pirate486743186:
alpine first saves a file with a random number in the name, then renames it to the correct name.
This is why i deactivated the whitelists in ${HOME}

rusty-snake:
merged, thanks.

PR #4365, opened 2021-06-19
Fix sndio support

lxeiqr:
This patch allows sandboxed programs to access a sndiod server and play audio on systems where sndio handles audio.

### DIFF
--- a/etc/inc/whitelist-common.inc
+++ b/etc/inc/whitelist-common.inc
@@ -23,8 +23,8 @@ read-only ${HOME}/.local/share/applications
 whitelist ${HOME}/.local/share/icons
 whitelist ${HOME}/.local/share/mime
 whitelist ${HOME}/.mime.types
-whitelist ${HOME}/.uim.d
 whitelist ${HOME}/.sndio/cookie
+whitelist ${HOME}/.uim.d
 
 # dconf
 mkdir ${HOME}/.config/dconf
--- a/etc/inc/whitelist-common.inc
+++ b/etc/inc/whitelist-common.inc
@@ -24,6 +24,7 @@ whitelist ${HOME}/.local/share/icons
 whitelist ${HOME}/.local/share/mime
 whitelist ${HOME}/.mime.types
 whitelist ${HOME}/.uim.d
+whitelist ${HOME}/.sndio/cookie
 
 # dconf
 mkdir ${HOME}/.config/dconf
@@ -82,7 +83,3 @@ whitelist ${HOME}/.kde4/share/config/ksslcablacklist
 whitelist ${HOME}/.kde4/share/config/oxygenrc
 whitelist ${HOME}/.kde4/share/icons
 whitelist ${HOME}/.local/share/qt5ct
-
-# sndio
-whitelist ${HOME}/.sndio/cookie
-whitelist /tmp/sndio/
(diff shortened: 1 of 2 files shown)


SkewedZeppelin (changes requested):
-1, don't add /tmp to this, will effectively apply private-tmp to all profiles using whitelist-common

SkewedZeppelin:
This is likely something you should add to your locals.

rusty-snake (changes requested):
- [ ] Move it into the common block (the first one)
- [x] Only `whitelist` in `${HOME}` (if `whitelist /tmp/sndio` is necessary, we should add it to `private-tmp`)

lxeiqr:
>     * [ ]  Move it into the common block (the first one)
> 
>     * [ ]  Only `whitelist` in `${HOME}` (if `whitelist /tmp/sndio` is necessary, we should add it to `private-tmp`)
I've removed that last line and moved the cookie whitelist into the common block. However, I'm not quite sure how to handle the /tmp/sndio directory. Usually it contains UNIX domain sockets (e.g. /tmp/sndio/sndio0) used to connect to the server.

etc/inc/whitelist-common.inc
```diff
@@ -24,6 +24,7 @@ whitelist ${HOME}/.local/share/icons
 whitelist ${HOME}/.local/share/mime
 whitelist ${HOME}/.mime.types
 whitelist ${HOME}/.uim.d
+whitelist ${HOME}/.sndio/cookie
```
rusty-snake, line 27:
```suggestion
whitelist ${HOME}/.sndio/cookie
whitelist ${HOME}/.uim.d
```

rusty-snake:
> However, I'm not quite sure how to handle the /tmp/sndio directory. Usually it contains UNIX domain sockets (e.g. /tmp/sndio/sndio0) used to connect to the server.

I think a `profile_add("whitelist /tmp/sndio");` in `fs_private_tmp` works

https://github.com/netblue30/firejail/blob/447e30e2c9b24f4ec6af5520a37a77068e66383b/src/firejail/fs.c#L1251-L1299

lxeiqr:
> > However, I'm not quite sure how to handle the /tmp/sndio directory. Usually it contains UNIX domain sockets (e.g. /tmp/sndio/sndio0) used to connect to the server.
> 
> I think a `profile_add("whitelist /tmp/sndio");` in `fs_private_tmp` works
> 
> https://github.com/netblue30/firejail/blob/447e30e2c9b24f4ec6af5520a37a77068e66383b/src/firejail/fs.c#L1251-L1299

Done, now it works properly without any extra tweaking.

lxeiqr:
Done, now it works properly without any extra tweaking.

src/firejail/fs.c
```diff
@@ -1274,6 +1274,9 @@ void fs_private_tmp(void) {
 	// read-only x11 directory
 	profile_add("read-only /tmp/.X11-unix");
 
+        // whitelist sndio directory
+        profile_add("whitelist /tmp/sndio");
```
rusty-snake, line 1278: 8 spaces -> 1 tab

rusty-snake (commented):
LGTM

rusty-snake (approved):

netblue30:
Thanks for the patch!

PR #4369, opened 2021-06-22
rtv profile correction

pirate486743186:
less is used to read comments
other minor corrections


etc/profile-m-z/rtv-addons.profile
```diff
@@ -9,6 +9,7 @@ include rtv-addons.local
 
 ignore nosound
 ignore private-bin
+#private-bin w3m,mpv,youtube-dl
```
rusty-snake, line 12: move down as per template.

rusty-snake (approved):

netblue30:
all in, thanks!

PR #4370, opened 2021-06-23
allow access to pkgconfig

glitsj16:
Building (specific) packages on Arch Linux (from AUR) needs access to /usr/share/pkgconfig. I've only seen it happen with cargo, so I added the whitelist to cargo.profile. We might want to add it to wusc instead. Suggestions welcome!

### DIFF
--- a/etc/profile-a-l/cargo.profile
+++ b/etc/profile-a-l/cargo.profile
@@ -34,6 +34,7 @@ include disable-xdg.inc
 #whitelist ${HOME}/.cargo
 #whitelist ${HOME}/.rustup
 #include whitelist-common.inc
+whitelist /usr/share/pkgconfig
 include whitelist-runuser-common.inc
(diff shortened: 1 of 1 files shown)


rusty-snake (commented):
LGTM

I wouldn't add it to wusc. Only IDEs and co will need it. However:

```
$ grep usr $(grep allow-common-devel /etc/firejail/* -l)
/etc/firejail/atom.profile:ignore include whitelist-usr-share-common.inc
/etc/firejail/cargo.profile:include whitelist-usr-share-common.inc
```

glitsj16:
> $ grep usr $(grep allow-common-devel /etc/firejail/* -l)
/etc/firejail/atom.profile:ignore include whitelist-usr-share-common.inc
/etc/firejail/cargo.profile:include whitelist-usr-share-common.inc

@rusty-snake  Thanks, we're okay as-is in that case.

PR #4371, opened 2021-06-24
Correct amule.profile for upnp

chrpinedo:
In order UPnP to work netlink protocol must be added

### DIFF
--- a/etc/profile-a-l/amule.profile
+++ b/etc/profile-a-l/amule.profile
@@ -33,6 +33,7 @@ nosound
 notv
 nou2f
 novideo
+# Add netlink protocol to use UPnP
 protocol unix,inet,inet6
 seccomp
 shell none


rusty-snake (commented):
Since `netlink` opens a wide attack surface (to the kernel) it may be better to add it commented if UPnP is a rarely used feature of aMule.

chrpinedo:
I agree with you that one commented line could be interesting to other users.

smitsohu:
@chrpinedo Are you interested in adding a comment to the profile?

Otherwise I would suggest to merge as-is, in order to get the bug fixed.

chrpinedo:
Since netlink protocol should be avoided when it is not required, I modified the commit to add one comment to the profile. If you agree, you could merge it.

rusty-snake (approved):

smitsohu:
Merged!

PR #4360, opened 2021-06-18
gcov: add missing gcov.h includes

kmk3:
Fixes the following "implicit declaration" warning (13 occurrences in
total) when building with gcov support:

    $ pacman -Q gcc10
    gcc10 1:10.2.0-3
    $ CC=gcc-10 && export CC
    $ ./configure --prefix=/usr --enable-apparmor --enable-gcov >/dev/null
    $ make >/dev/null
    appimage.c: In function ‘appimage_set’:
    appimage.c:140:2: warning: implicit declaration of function ‘__gcov_flush’ [-Wimplicit-function-declaration]
      140 |  __gcov_flush();
          |  ^~~~~~~~~~~~
    interface.c: In function ‘print_sandbox’:
    interface.c:149:3: warning: implicit declaration of function ‘__gcov_flush’ [-Wimplicit-function-declaration]
      149 |   __gcov_flush();
          |   ^~~~~~~~~~~~
    netstats.c: In function ‘netstats’:
    netstats.c:246:4: warning: implicit declaration of function ‘__gcov_flush’ [-Wimplicit-function-declaration]
      246 |    __gcov_flush();
          |    ^~~~~~~~~~~~
    [...]

Note: The commands above were executed from makepkg, while building
firejail-git from the AUR.

Note2: gcc-10 was used because the build fails with the current gcc
version (11.1.0) on Artix Linux.  The failure happens because
__gcov_flush was removed on gcc 11.1.0[1]; this will be addressed later.

Note3: The following command helped find the affected files:

    $ git grep -Fl __gcov -- src

[1] https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=811b7636cb8c10f1a550a76242b5666c7ae36da2

PR #4373, opened 2021-06-25
gcov: fix build failure with gcc 11.1.0

kmk3:
The build currently fails if gcov support is enabled:

    $ pacman -Q gcc
    gcc 11.1.0-1
    $ ./configure --prefix=/usr --enable-apparmor --enable-gcov >/dev/null
    $ make >/dev/null
    [...]
    netstats.c: In function ‘netstats’:
    netstats.c:250:25: warning: implicit declaration of function ‘__gcov_flush’; did you mean ‘__gcov_dump’? [-Wimplicit-function-declaration]
      250 |                         __gcov_flush();
          |                         ^~~~~~~~~~~~
          |                         __gcov_dump
    [...]
    /usr/bin/ld: netstats.o: in function `netstats':
    /tmp/firejail-git/src/firejail-git/src/firemon/netstats.c:250: undefined reference to `__gcov_flush'
    [...]
    collect2: error: ld returned 1 exit status
    make[1]: *** [Makefile:10: firemon] Error 1
    make: *** [Makefile:42: src/firemon/firemon] Error 2
    [...]

This happens because __gcov_flush was removed on gcc 11.1.0[1] [2] [3].
See the following gcc commits:

* d39f7dc8d5 ("Do locking for __gcov_dump and __gcov_reset as well.")
* c0532db47d ("Use __gcov_dump and __gcov_reset in execv and fork context.")
* 811b7636cb ("Remove __gcov_flush.")

Its implementation did the following[4]:

      __gcov_lock ();
      __gcov_dump_int ();
      __gcov_reset_int ();
      __gcov_unlock ();

As hinted in the commit messages above, the function is no longer needed
because locking is now done inside each of __gcov_dump and __gcov_reset.

So add an implementation of __gcov_flush (on a new gcov_wrapper.h file)
for gcc >= 11.1.0, which just calls __gcov_dump and then __gcov_reset.

    $ git grep -Flz '#include <gcov.h>' -- '*.c' |
      xargs -0 -I '{}' sh -c \
      "printf '%s\n' \"\`sed 's|<gcov\\.h>|\"../include/gcov_wrapper.h\"|' '{}'\`\" >'{}'"

Note: This is the continuation of commit 31557e9c7 ("gcov: add missing
gcov.h includes") / PR #4360.

[1] https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=d39f7dc8d558ca31a661b02d08ff090ce65e6652
[2] https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=c0532db47d092430f8e8f497b2dc53343527bb13
[3] https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=811b7636cb8c10f1a550a76242b5666c7ae36da2
[4] https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libgcc/libgcov-interface.c;h=855e8612018d1c9caf90396a3271337aaefdb9b3#l86

### DIFF
--- a/src/firejail/appimage.c
+++ b/src/firejail/appimage.c
@@ -29,7 +29,7 @@
 #include <errno.h>
 
 #ifdef HAVE_GCOV
-#include <gcov.h>
+#include "../include/gcov_wrapper.h"
 #endif
 
 static char *devloop = NULL;	// device file
--- a/src/firejail/chroot.c
+++ b/src/firejail/chroot.c
@@ -30,7 +30,7 @@
 #endif
 
 #ifdef HAVE_GCOV
-#include <gcov.h>
+#include "../include/gcov_wrapper.h"
 #endif
 
 // exit if error
--- a/src/firejail/fs.c
+++ b/src/firejail/fs.c
@@ -34,7 +34,7 @@
 #endif
 
 #ifdef HAVE_GCOV
-#include <gcov.h>
+#include "../include/gcov_wrapper.h"
 #endif
 
 #define MAX_BUF 4096
--- a/src/firejail/fs_mkdir.c
+++ b/src/firejail/fs_mkdir.c
@@ -26,7 +26,7 @@
 #include <string.h>
 
 #ifdef HAVE_GCOV
-#include <gcov.h>
+#include "../include/gcov_wrapper.h"
 #endif
 
 static void check(const char *fname) {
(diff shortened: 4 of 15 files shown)


kmk3:
Note: I trimmed many lines but the message is still rather long.  Let me know
if it's too long.

Note2: I haven't used gcov and I don't know if supporting gcc 10 is
desired/makes sense; I was just working on `configure*` when I noticed this
issue.

Note3: This seems to be the easier way to fix the build, but I could also
replace all occurrences of `__gcov_flush` with `__gcov_dump` / `__gcov_reset`
instead if that's better.

netblue30:
all in, thanks

PR #4376, opened 2021-06-27
gcov: use no-op functions if not enabled

kmk3:
Instead of wrapping every gcov function call in an ifdef.

Note: The usage of `((void)0)` is based on section 7.2 of the C99
standard (N1256)[1] [2]:

> 7.2 Diagnostics <assert.h>
>
> 1 The header <assert.h> defines the assert macro and refers to another
> macro,
>
>     NDEBUG
>
> which is not defined by <assert.h>. If NDEBUG is defined as a macro
> name at the point in the source file where <assert.h> is included, the
> assert macro is defined simply as
>
>     #define assert(ignore) ((void)0)

See also assert.h(0p) from POSIX.1-2017[3].

Note: This is a continuation of commit b408b20c7 ("gcov: fix build
failure with gcc 11.1.0") / PR #4373.

[1] http://www.open-std.org/JTC1/SC22/WG14/www/docs/n1256.pdf
[2] https://port70.net/~nsz/c/c99/n1256.html#7.2
[3] https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/assert.h.html

### DIFF
--- a/src/firejail/appimage.c
+++ b/src/firejail/appimage.c
@@ -21,16 +21,13 @@
 // sudo mount -o loop krita-3.0-x86_64.appimage mnt
 
 #include "firejail.h"
+#include "../include/gcov_wrapper.h"
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/mount.h>
 #include <fcntl.h>
 #include <linux/loop.h>
 #include <errno.h>
-
-#ifdef HAVE_GCOV
-#include "../include/gcov_wrapper.h"
-#endif
 
 static char *devloop = NULL;	// device file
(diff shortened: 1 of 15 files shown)


kmk3:
Note: There have been no other changes to gcov-related code on master recently,
so all my recent gcov PRs were made on the same branch to make it easier to see
a diff of everything (especially since some of the earlier changes are kind of
undone on the latest commit):

CLI:

```sh
git diff e770ab6d8..5106b2ec4
```

GitHub: <https://github.com/netblue30/firejail/compare/e770ab6d8..5106b2ec4>

Anyway, this is probably the last PR specific to gcov.

netblue30:
all set!

PR #4388, opened 2021-07-07
allow/deny in zsh completion

rusty-snake:


reinerh (approved):

netblue30:
somehow I missed this one

rusty-snake:
There are multiple places except of the source-code which needs to be updated when new commands are added (and usually we forget at least one). That's why I want to add a check-list to CONTRIUTING.md. I'm currently at this. Anything forgotten?
 
```markdown
- [ ] Update manpages: firejail(1) and firejail-profile(5)
- [ ] Update zsh-completions
- [ ] Update vim syntax files
```

reinerh:
(there is also a bash completion (which still contains whitelist) ;) )

Issue #4379, opened 2021-06-30
Replace "whitelist" and "blacklist" terms by "allowlist" and "blocklist"/"denylist"


ned64:
There has been a debate in the IT sec community for a while now about the terms "whitelist" and "blacklist", see e.g. here

  [dice.com article: ‘Whitelist,’ ‘Blacklist’: The New Debate Over Security Terminology](https://insights.dice.com/2020/07/17/whitelist-blacklist-the-new-debate-over-security-terminology/)

 While certainly not meant by firejail or others in any racial or other negative sense there is at least a potential problem of connotation.  "Black" items are to excluded/blocked, while "white" ones will be included/accepted.  In a world where there are racism problems in most societies - many of which are about dark and light skin colour - this can be experienced by some as just one more implicit bias of white/black being good/bad.

I therefore suggest that `--blacklist` options and `--whitelist` options (including all derivatives) be renamed `--blocklist` and `--allowlist` (etc.) as some companies have already started doing.

Alternatively one could use derivatives of `allow` and `deny` - terms known from Apache (`httpd.conf` is where I first came across them decades ago).

In whatever way the change is implemented, the old options would need to be kept (but deprecated a.s.a.p.) as alternative names for existing user configurations not to break, perhaps for quite some time.  In all own releases the replacement should be complete, though, including contributed profiles and documentation.

glitsj16:
Looks like a duplicate of https://github.com/netblue30/firejail/issues/3447.

ned64:
@glitsj16 Well spotted. I _did_ search before posting but it didn't turn up.

glitsj16:
> I did search before posting but it didn't turn up.

@ned64 That does happen sometimes. Especially with 'keywords' like _whitelist_ and _blacklist_ being all over our code base it can help to run a search through the web search engine of your choosing instead of through the GitHub search.

netblue30:
Tentative move to allow/deny under way. I put a message on the main README.md, and redirected the old #3447 issue here.

netblue30:
I'm done for now. Everything should be working as before. blacklist/whitelist are set as aliases, nothing should break.

In src/tools/profcleaner.c I have a small tool you can use to convert your local profiles - details in the .c file. Any suggestions in man pages just put them in git. Probably there will be several iterations until we get them right. Hold on on changes throughout the wiki until we release it.

netblue30:
Phase 2

"deny" seems to be working, but "allow" sucks!

[...]

PR #4389, opened 2021-07-07
Create profcleaner.sh

rusty-snake:
profcleaner.c is just sed, I was wondering why we need C for that.

------------

I must say here again that I do not like the blacklist/whitelist -> allow/deny change.
1. Having a command named `allow` that
   - does not necessary allow access to a file is confusing (for new users)
   - have site effects on other files is confusing (for new users)
2. The term "allow" is already used for `allow-*.inc` include, maybe we should renamed them to `nodisable-*.inc`/`nodeny-*.inc`
3. We still have `whitelist-*.inc`
4. We need some standardized terminology (commands vs. options vs. ...; does `deny` drop,block,disable a file / access to a file; ...)
5. We need to migrate comments in profiles as well.

### DIFF
--- a/src/tools/profcleaner.sh
+++ b/src/tools/profcleaner.sh
@@ -22,14 +22,14 @@ if [[ $1 == --help ]]; then
 	cat <<-EOM
 	USAGE:
 	  profcleaner.sh --help        Show this help message and exit
-	  profcleaner.sh --all         Clean all profiles in /etc/firejail
+	  profcleaner.sh --system      Clean all profiles in /etc/firejail
 	  profcleaner.sh --user        Clean all profiles in ~/.config/firejail
 	  profcleaner.sh /path/to/profile1 /path/to/profile2 ...
 	EOM
 	exit 0
 fi
 
-if [[ $1 == --all ]]; then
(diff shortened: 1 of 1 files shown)


src/tools/profcleaner.sh
```diff
@@ -0,0 +1,45 @@
...
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+if [[ $1 == --help ]]; then
+	cat <<-EOM
+	USAGE:
+	  profcleaner.sh --help        Show this help message and exit
+	  profcleaner.sh --all         Clean all profiles in /etc/firejail
```
reinerh, line 25: maybe `--global` or `--system` instead of `--all`?

reinerh (approved):

netblue30:
Yes, .sh is better than .c. I went for .c because of lines such as "include whitelist-common.inc". Will do another round of changes and will fix the .sh. So back to https://github.com/netblue30/firejail/issues/4379

PR #4399, opened 2021-07-13
rkhunter fix

glitsj16:
rkhunter has `/etc/rkhunter.d` and `/etc/rkhunter.conf` (at least on Arch Linux). Cover both paths via globbing.

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -5,63 +5,63 @@ include disable-common.local
 # The following block breaks trash functionality in file managers
 #read-only ${HOME}/.local
 #read-write ${HOME}/.local/share
-deny  ${HOME}/.local/share/Trash
+deny ${HOME}/.local/share/Trash
 
 # History files in $HOME and clipboard managers
-deny-nolog  ${HOME}/.*_history
-deny-nolog  ${HOME}/.adobe
-deny-nolog  ${HOME}/.cache/greenclip*
-deny-nolog  ${HOME}/.histfile
-deny-nolog  ${HOME}/.history
-deny-nolog  ${HOME}/.kde/share/apps/klipper
-deny-nolog  ${HOME}/.kde4/share/apps/klipper
-deny-nolog  ${HOME}/.local/share/fish/fish_history
-deny-nolog  ${HOME}/.local/share/klipper
-deny-nolog  ${HOME}/.macromedia
-deny-nolog  ${HOME}/.mupdf.history
-deny-nolog  ${HOME}/.python-history
-deny-nolog  ${HOME}/.python_history
-deny-nolog  ${HOME}/.pythonhist
-deny-nolog  ${HOME}/.lesshst
-deny-nolog  ${HOME}/.viminfo
-deny-nolog  /tmp/clipmenu*
+deny-nolog ${HOME}/.*_history
+deny-nolog ${HOME}/.adobe
+deny-nolog ${HOME}/.cache/greenclip*
+deny-nolog ${HOME}/.histfile
+deny-nolog ${HOME}/.history
+deny-nolog ${HOME}/.kde/share/apps/klipper
+deny-nolog ${HOME}/.kde4/share/apps/klipper
+deny-nolog ${HOME}/.local/share/fish/fish_history
+deny-nolog ${HOME}/.local/share/klipper
+deny-nolog ${HOME}/.macromedia
+deny-nolog ${HOME}/.mupdf.history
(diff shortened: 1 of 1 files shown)


rusty-snake:
If I `dnf rq -l` for the program in 110a74f094abcb4f2763d76e204fb3c9743fa9a1, Fedora has a at least `/etc/aide.conf` and `/etc/sysconfig/rkhunter` in addition.

glitsj16:
@rusty-snake Thanks, added those Fedora paths.

rusty-snake:
While we are at it, what about `deny  /etc/sysconfig` and `deny  /etc/default`?

glitsj16:
> While we are at it, what about deny /etc/sysconfig and deny /etc/default?

Added, thanks for the suggestions!

rusty-snake:
> fix doubled spacing

I'd suggest to undo this here and do it for all profile in  separate PR. (Which we then can exclude from `git blame` too).

glitsj16:
> I'd suggest to undo this here and do it for all profile in separate PR. (Which we then can exclude from git blame too).

Makes sense, done!

rusty-snake:
FTR: This must be megred by the "Squash and Merge" option.

glitsj16:
> FTR: This must be megred by the "Squash and Merge" option.

Indeed. Let's hold off merging until we can fix another CI failure... as suggested by @kmk3.

kmk3:
@rusty-snake commented 5 hours ago:

> > fix doubled spacing
> 
> I'd suggest to undo this here and do it for all profile in separate PR.
> (Which we then can exclude from `git blame` too).

I agree that it would be better to change all at once rather than otherwise,
but for now I think that this should also be avoided.

Even though it can be ignored for `git blame`, it makes diffing files across
these changes harder (and I think that it still shows on `git log`).

So please don't make more big changes like that until we have settled on the
commands to be used on profiles.  See also:

<https://github.com/netblue30/firejail/issues/4379#issuecomment-878840569>

kmk3:
@rusty-snake commented 4 hours ago:

> FTR: This must be megred by the "Squash and Merge" option.

@glitsj16 To ensure that the sorting commits aren't accidentally merged, I can
rebase this branch if you want.

Or if you want to try it yourself:

```sh
git stash push # optional: save anything important
git checkout rkhunter
git tag bkp # optional: in case things go bad; use any name you want
git rebase -i "$(git merge-base master rkhunter)"
```

This is the default text/"script" that comes up:

```
pick 647866e6f rkhunter fix
pick d233b85e9 additional deny's for Fedora and some sorting changes
pick 331a53c67 more deny's
pick e0ac12ded fix doubled spacing
pick e33c56f93 revert fix double spacing
```

Mark the commits to be squashed as **squash**.  This makes each of them be
squashed into the commit above it on the list.

Mark the commits to be removed as **drop** (or just delete its entire line).

Note: The only parts that git cares about is `<action> <hash>`; the rest is
ignored (so don't bother trying to change the title in here).

```
pick 647866e6f rkhunter fix
pick d233b85e9 additional deny's for Fedora and some sorting changes
squash 331a53c67 more deny's
drop e0ac12ded fix doubled spacing
drop e33c56f93 revert fix double spacing
```

A file appears with the messages of both commits 2 and 3.  Its contents will be
the commit message of commit 2.

Delete the parts of the message which are not relevant anymore ("more deny's"
in this case).  Then just save and quit.

Look at the result:

```sh
git log -p --reverse "$(git merge-base master HEAD)..HEAD"

# alternatively:
tig "$(git merge-base master HEAD)..HEAD"
```

If the branch looks good, force-push:

```sh
git push --force-with-lease origin
```

If not, reset to the backup:

```sh
git reset --hard bkp
```

To cancel an ongoing rebase and discard all changes:

```sh
git rebase --abort
```

---

Note: I have some comments on the changes, but this is more important at the
moment.

netblue30:
HI guys, we merge? we don't merge? I'll bring in some more files. I'm looking after directories in etc used only by servers.

kmk3 (changes requested):
@netblue30 commented [1 hour ago](https://github.com/netblue30/firejail/pull/4399#issuecomment-880788076):

> HI guys, we merge? we don't merge? I'll bring in some more files. I'm looking
> after directories in etc used only by servers.

(Adding this review just to clarify that changes were requested)

As per [this comment][1], these commits should be squashed (or just dropped):

* e0ac12ded fix doubled spacing
* e33c56f93 revert fix double spacing

[1]: https://github.com/netblue30/firejail/pull/4399#issuecomment-879390346

PR #4407, opened 2021-07-17
Fix Firefox 'Profile not found' - whitelist /run/user/xxx/firefox

0x6a61:
After updating ArchLinux Firefox quits with error message "Profile not found".
Apparently Firefox now uses /run/user/pid/ to store user profiles during runtime.


rusty-snake (commented):
/run/usr/UID is a tmpfs, that means it does not persist restarts. If you `whitelist ${RUNUSER}/*firefox*`, the directory you want to whitelist must already exist. How is it created? Firefox can't do this. Is it done by a init-script?

> /run/user/pid/ 

It's your uid not not the pid of firefox.

SkewedZeppelin:
I tested up-to-date Arch yesterday and had no issue with Firefox.

0x6a61:
As it turned out the directory is created by profile-sync-daemon (https://wiki.archlinux.org/title/Profile-sync-daemon).
Strange that the error occurs now, profile-sync-daemon is installed since 2016.

How do you suggest to proceed? Will the line cause warnings/errors if the folder does not exist? If so, maybe we can add the line as a comment + explanation?

glitsj16:
> Strange that the error occurs now, profile-sync-daemon is installed since 2016.

Probaby due to changes in the Firefox-related profiles between firejail releases. For example adding wruc in https://github.com/netblue30/firejail/commit/096d0de5f8bb253d0c1035796464bc5982f06f81.

> How do you suggest to proceed? Will the line cause warnings/errors if the folder does not exist? If so, maybe we can add the line as a comment + explanation?

Suggestion:
```
...
include whitelist-common.inc
# If you use profile-sync-daemon, add 'blacklist ${RUNUSER}/*firefox*' to your globals.local
# and add 'noblacklist ${RUNUSER}/*firefox*' and 'whitelist ${RUNUSER}/*firefox*' to your
# firefox-common.local.
...
```

rusty-snake:
> Suggestion:
> 
> ```
> ...
> include whitelist-common.inc
> # If you use profile-sync-daemon, add 'blacklist ${RUNUSER}/*firefox*' to your globals.local
> # and add 'noblacklist ${RUNUSER}/*firefox*' and 'whitelist ${RUNUSER}/*firefox*' to your
> # firefox-common.local.
> include whitelist-runuser-common.inc
> ...
> ```

Why not making it default? Add `(blacklist|deny) ${RUNUSER}/*firefox*` to `disable-programs.inc`and `(noblacklist|nodeny ${RUNUSER}/*firefox*`+`(allow|whitelist) ${RUNUSER}/*firefox*` to firefox.profile. And a note in firefox.profile what it is like ` profile-sync-daemon puts profiles here`.

> As it turned out the directory is created by profile-sync-daemon

So it's created at login or so, in any case before firefox is started. Good to hear.

> Will the line cause warnings/errors if the folder does not exis

No.

etc/profile-a-l/firefox-common.profile
```diff
@@ -26,6 +27,7 @@ mkdir ${HOME}/.local/share/pki
 allow  ${DOWNLOADS}
 allow  ${HOME}/.pki
 allow  ${HOME}/.local/share/pki
+allow ${RUNUSER}/*firefox*
```
rusty-snake, line 30: Since this path is only for firefox, we should add it to firefox.profile IMHO.

rusty-snake (approved):

netblue30:
@0x6a61 - all merged in. Thanks for the patch!

PR #4390, opened 2021-07-07
Exclude allow/deny move in profile from git blame

rusty-snake:
fe0f975f447d59977d90c3226cc8c623b31b20b3 (move whitelist/blacklist to
allow/deny) is just a huge rename¹ w/o effects to the profile. Ignoreing
it in git blame to see which commits actually added/changed a
allow/nodeny command.

Configure git to use .git-blame-ignore-revs:

    git config blame.ignoreRevsFile .git-blame-ignore-revs

EDIT: ¹ but it also https://github.com/netblue30/firejail/commit/fe0f975f447d59977d90c3226cc8c623b31b20b3#diff-7c733ed074575806d2755a48dd858c22c318d0190737f64299fa6dda414ba29b and https://github.com/netblue30/firejail/commit/fe0f975f447d59977d90c3226cc8c623b31b20b3#diff-9f33b84e2dc73f7861f9d367b48fe0e3cf23832184311457b6ebef69714c358c

PR #4410, opened 2021-07-18
Revert "move whitelist/blacklist to allow/deny"

kmk3:
This reverts commit fe0f975f447d59977d90c3226cc8c623b31b20b3.

Note: This only reverts the changes from etc.

The 4 aliases introduced on commit 45f2ba544 are mere, well, aliases.
That is, they fail to address the different usability problems discussed
on [#3447][3447] and in fact only make things more confusing (as has
already been mentioned on [this][4379] and later comments).  The main
reason is that the aliases do not meaningfully map to the original
commands.  For example, the commands from each pair below seem like they
would do the exact same thing:

* `allow` and `nodeny`
* `deny` and `noallow`

Additionally, if these aliases are not the final commands, but only a
test/work-in-progress, then keeping the wide-scale search/replace
changes made on commit fe0f975f4 would only serve to cause confusion, as
users of firejail-git, contributors and downstream projects might start
changing the commands used on their profiles, only to later have to
change them again, potentially to completely different commands.

The sooner this is undone the better, as (besides the above reasons) the
more profile changes there are between the original commit and the
revert, the harder it is to e.g.: `git diff` versions of files across
the following revision ranges: before the commit, after the commit but
before the revert and after the revert.  Note: This is still the case
even if a commit is [ignored by `git blame`][4390].

So let us revert fe0f975f4 and only reapply similar large-scale changes
once we have discussed and settled on better commands.

How the revert was applied: Despite using the auto-generated message
from `git revert`, to ensure correctness and to avoid conflicts the
changes were reverted in different steps: Firstly, revert the files
which can be safely reverted directly ("filestorevert"):

    # Find out which files have been changed on fe0f975f44, but have not
    # been changed afterwards and list them on "filestorevert"
    git show --pretty='' --name-only fe0f975f44 -- etc | LC_ALL=C sort >allfiles
    git diff --name-only fe0f975f44..master -- etc | LC_ALL=C sort >filestoignore
    comm -2 -3 allfiles filestoignore >filestorevert

    # Note: There are 3 extra files on filestoignore because they were
    # added after commit fe0f975f44
    wc -l allfiles filestoignore filestorevert | head -n 3
    #   797 allfiles
    #     8 filestoignore
    #   792 filestorevert

    # Automatically revert files in "filestorevert"
    # See https://stackoverflow.com/a/23401018/10095231
    tr '\n' '\000' <filestorevert | xargs -0 git show fe0f975f44 -- |
    git apply --reverse

    printf 'Total files reverted:\n'
    git diff --name-only | wc -l
    # 792

Secondly, do some search/replace on the rest:

    tr '\n' '\000' <filestoignore | xargs -0 sed -i.bak \
      -e 's/allow  /whitelist /' -e 's/noallow  /nowhitelist /' \
      -e 's/deny  /blacklist /' -e 's/nodeny  /noblacklist /' \
      -e 's/deny-nolog  /blacklist-nolog /'

    find etc -name '*.bak' -print0 | xargs -0 rm

Thirdly, verify the result.  The following command shows the difference
between all the changes in etc from before fe0f975f44 and this commit
(inclusive):

    git diff fe0f975f44~1 -- etc

From the output, it looks like all alias changes are fully reverted and
that the other changes to etc (from after fe0f975f44) remain, so the
revert seems to be done correctly.

[3447]: https://github.com/netblue30/firejail/issues/3447
[4379]: https://github.com/netblue30/firejail/issues/4379#issuecomment-876460222
[4390]: https://github.com/netblue30/firejail/issues/4390


kmk3:
> Thirdly, verify the result.  The following command shows the difference
> between all the changes in etc from before fe0f975f44 and this commit
> (inclusive):
> 
>     git diff fe0f975f44~1 -- etc

GitHub version:

<https://github.com/netblue30/firejail/compare/fe0f975f44~1..f43382f1e9707b4fd5e63c7bfe881912aa4ee994>

Note: There is no way to diff by path on GitHub, so the above also shows
changes in unrelated dirs.

Note2: I plan to add the revert commit to `.git-blame-ignore-revs`, like in
[#4390][4390], but I'd rather do it after this PR (i.e.: on a follow-up PR), to
be sure of the final commit hash.

Note3: While another large-scale search/replace may very well be done on most
profiles, it is also possible that it would be done individually or in batches
instead, for two reasons:

1. In order to fix the "whitelisting requires 3 commands" annoyance, a simple
   search/replace won't do; whitelisting profiles would require at least some
   other additional changes (like deleting the lines involving the other 2
   commands), which is not as simple to get right by scripting (without
   mangling the profiles, that is).

2. I have another WIP proposal from months ago (which will likely be posted a
   while after the first one, on a separate issue), regarding the ordering of
   some whitelist-related paths on profile.template (to fix some
   inconsistencies which have come up during some PR reviews), which would
   require splitting and sorting the relevant lines.  If this ends up being
   agreed upon and if most profiles are going to be changed anyway, it could be
   helpful to apply these changes at the same time, in order to avoid rework
   and to avoid more noise in the history.  Note: This sorting issue is
   probably less widely known and not as crucial as the first reason, but I
   thought I'd leave it here for completeness.

[4390]: https://github.com/netblue30/firejail/issues/4390

netblue30:
merged, sorry for delay!

kmk3:
> merged, sorry for delay!

That's alright; I'm just glad it's finally done now.

PR #4413, opened 2021-07-20
Add new command checklist to CONTRIBUTING.md

rusty-snake:


kmk3 (approved):
Nice; I've wondered about this before.

I think it would also be helpful to add an example of a PR that does all that.
If I may suggest one: #4278 (:p)

reinerh (approved):

Fred-Barclay (approved):

PR #4420, opened 2021-07-26
ordering and additions

glitsj16:
Redo https://github.com/netblue30/firejail/pull/4399.

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -169,9 +169,9 @@ blacklist /var/lib/systemd
 #blacklist /var/run/systemd
 
 # openrc
-blacklist /etc/init.d/
+blacklist /etc/init.d
 blacklist /etc/rc.conf
-blacklist /etc/runlevels/
+blacklist /etc/runlevels
 
 # VirtualBox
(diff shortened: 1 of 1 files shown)


etc/inc/disable-common.inc
```diff
@@ -159,23 +159,23 @@ blacklist ${RUNUSER}/gsconnect
...
+blacklist /etc/systemd/system
+blacklist /var/lib/systemd
 # creates problems on Arch where /etc/resolv.conf is a symlink to /var/run/systemd/resolve/resolv.conf
 #blacklist /var/run/systemd
 
 # openrc
-blacklist /etc/runlevels/
 blacklist /etc/init.d/
 blacklist /etc/rc.conf
+blacklist /etc/runlevels/
```
Fred-Barclay, line 174: I feel like this is a separate question... but do we need the training `/`? It seems to have been inconsistently applied in this file

glitsj16: Good point. Removed trailing slashes, thanks @Fred-Barclay.

netblue30:
merged!

PR #4462, opened 2021-08-09
Create disable-X11.inc

rusty-snake:
Move https://gitlab.com/rusty-snake/kyst/-/blob/122cda1b1492213467d1d583671502a549d912bb/firejail/deny-X11.inc upstream.

### DIFF
--- a/etc/inc/disable-X11.inc
+++ b/etc/inc/disable-X11.inc
@@ -1,0 +1,15 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+include disable-X11.local
+
+blacklist /tmp/.X11-unix
+blacklist ${HOME}/.Xauthority
+blacklist ${RUNUSER}/gdm/Xauthority
+blacklist ${RUNUSER}/.mutter-Xwaylandauth*
+blacklist ${RUNUSER}/xauth_*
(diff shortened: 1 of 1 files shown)


etc/inc/disable-X11.inc
```diff
@@ -0,0 +1,15 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+include disable-X11.local
+
+blacklist /tmp/.X11-unix
+blacklist ${HOME}/.Xauthority
+blacklist ${RUNUSER}/gdm/Xauthority
+blacklist ${RUNUSER}/.mutter-Xwaylandauth*
+blacklist ${RUNUSER}/xauth_*
+#blacklist ${RUNUSER}/[[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]]-[[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]]-[[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]]-[[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]]-[[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]][[:xdigit:]]
```
reinerh, line 10: Out of curiosity I just tried to find where support for these regex classes is implemented, but I didn't succeed.
The only other place using them seems to be `whitelist-runuser-common.inc`.
Would this actually work? If yes, do you know where this is implemented?

reinerh: Interesting, `glob(7)` is documenting this. I wasn't aware that `glob()` supports character classes.

reinerh (approved):

etc/inc/disable-X11.inc
rusty-snake: Yes it is `glob()`. If it would be a regex, I could write `[[:xdigit:]]{8}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{12}` to match a UUID which would be shorter and much more readable. Too bad that you can not do it.

And, FTR, this line is commented because there are maybe other programs out there with the stupid idea of using a UUID without any prefix/suffix that indicates the program/usage.

kmk3:
Related discussion:

* #4854

PR #4476, opened 2021-08-22
Fix hanging arp_check

### LINKED ISSUE #3958: firejail hangs with net parameter
Hi,

**Bug and expected behavior**

some recent arch update must have (partially) broken firejails --net parameter. It used to work fine but now the program just hangs when executed with --net on a bridge device.
For some reason, one of my systemd startup service has no issues starting with --net.
What I have tried so far: downgrade kernel, downgrade networkmanager, downgrade firejail with no success.
I have also encountered #3948 which could be "solved" by downgrading the kernel from 5.10.13 to 5.10.12
I used networkmanager to create the bridge.

**No profile and disabling firejail**
- What changed calling `firejail --noprofile /path/to/program` in a terminal?
nothing changed
- What changed calling the program by path (check `which <program>` or `firejail --list` while the sandbox is running)?
nothing changed

**Reproduce**
1. Run in bash `firejail --noprofile --net=br1 --defaultgw=192.168.0.1 --dns=192.168.0.1 --ip=192.168.0.254 --debug sh`
2. See error
```
Autoselecting /bin/bash as shell
Building quoted command line: 'sh' 
Command name #sh#
get interface br1 configuration
MTU of br1 is 1500.
Bridge device br1 at 192.168.0.2/24
Trying 192.168.0.254 ...
```
3. It just hangs forever.

**Environment**
 - Linux distribution and version (ie output of `lsb_release -a`, `screenfetch` or `cat /etc/os-release`)

```
cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://www.archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux
```

 - Firejail version (output of `firejail --version`) exclusive or used git commit (`git rev-parse HEAD`) 

```
firejail version 0.9.64.2

Compile time support:
	- AppArmor support is enabled
	- AppImage support is enabled
	- chroot support is enabled
	- SELinux support is disabled
```

<details><summary> debug output </summary>

```
Autoselecting /bin/bash as shell
Building quoted command line: 'sh' 
Command name #sh#
get interface br1 configuration
MTU of br1 is 1500.
Bridge device br1 at 192.168.0.2/24
Trying 192.168.0.254 ...
```

</details>

minus7:
arp_check relied on select(2) decreasing the timeout. This doesn't seem
to be the case on Linux anymore, thus arp_check tends to hang when the
interface sees a lot of traffic. Calculating the timeout explicitly
solves the problem.

Fixes #3958

### DIFF
--- a/src/firejail/arp.c
+++ b/src/firejail/arp.c
@@ -20,6 +20,7 @@
 #include "firejail.h"
 #include <sys/socket.h>
 #include <sys/ioctl.h>
+#include <sys/time.h>
 #include <linux/if_ether.h>			  //TCP/IP Protocol Suite for Linux
 #include <net/if.h>
 #include <netinet/in.h>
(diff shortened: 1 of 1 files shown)


netblue30:
Thanks for the fix!

Issue #4421, opened 2021-07-26
Fix new profiles after [no]deny / [no]allow revert


adgross:
Files from the commit a25e4a101129471f3221df5471e2fc8fa6a1c4e6 still uses deny / nodeny and allow / noallow after merging the revert 8b50039a1fad123b90172fadc85bc232e97eb6d1.
The revert commit f43382f1e9707b4fd5e63c7bfe881912aa4ee994 is older than these profiles, while the merge itself happened later, so the revert missed the new files.

Affected files:
- io.github.lainsce.Notejot.profile [[L9-L10](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/io.github.lainsce.Notejot.profile#L9-L10), [L23-L25](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/io.github.lainsce.Notejot.profile#L23-L25)]
- lifeograph.profile [[L9-L11](https://github.com/netblue30/firejail/blob/master/etc/profile-a-l/lifeograph.profile#L9-L11), [L22-L23](https://github.com/netblue30/firejail/blob/master/etc/profile-a-l/lifeograph.profile#L22-L23)]
- rednotebook.profile [[L9-L10](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/rednotebook.profile#L9-L10), [L25-L33](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/rednotebook.profile#L25-L33)]
- zim.profile [[L9-L10](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/zim.profile#L9-L10), [L16](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/zim.profile#L16), [L29-L38](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/zim.profile#L29-L38)]

rusty-snake:
And #4407.

rusty-snake:
2e4d52ec60a677586decc62b0d1bc26757a2856a

Fred-Barclay:
Ha I ran across the files by accident in that commit - thanks to @adgross for catching these as well!

PR #4487, opened 2021-08-27
.git-blame-ignore-revs: add revert of allow/deny move

kmk3:
Add commit f43382f1e ("Revert "move whitelist/blacklist to allow/deny"")
from PR #4410.

As mentioned on commit b023b9a6f ("Exclude allow/deny move in profile
from git blame") / PR #4390, commit fe0f975f4 ("move whitelist/blacklist
to allow/deny") "is just a huge rename", and so is the revert of it.

Note that there is a follow-up to f43382f1e: commit 2e4d52ec6 ("Revert
allow/deny additional files") (sort of related to #4421).  It renames a
bit too much, which is later fixed by commit 3836131f3 ("Fix zim and
rednotebook").  Since these are small changes and since they involve
regressions, neither commit is added.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,4 @@
 # move whitelist/blacklist to allow/deny
 fe0f975f447d59977d90c3226cc8c623b31b20b3
+# Revert "move whitelist/blacklist to allow/deny"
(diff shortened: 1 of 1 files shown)


netblue30:
merged, thanks!

PR #4493, opened 2021-08-30
fix duplicate globals

pirate486743186:
i'm sure netblue was just testing us :P

### DIFF
--- a/etc/profile-a-l/gallery-dl.profile
+++ b/etc/profile-a-l/gallery-dl.profile
@@ -5,7 +5,8 @@ quiet
 # Persistent local customizations
 include gallery-dl.local
 # Persistent global definitions
-include globals.local
+# added by included profile
+#include globals.local
 
 noblacklist ${HOME}/.config/gallery-dl
 noblacklist ${HOME}/.gallery-dl.conf
--- a/etc/profile-m-z/yt-dlp.profile
+++ b/etc/profile-m-z/yt-dlp.profile
@@ -5,7 +5,8 @@ quiet
 # Persistent local customizations
 include yt-dlp.local
 # Persistent global definitions
-include globals.local
+# added by included profile
+#include globals.local
 
 noblacklist ${HOME}/.cache/yt-dlp
 noblacklist ${HOME}/.config/yt-dlp


reinerh:
Hm, I don't understand why this should be commented.
But looking at profiles in `/etc/firejail` there seems to be a wild mix of profiles with comment and without.

rusty-snake:
@reinerh we don't want to include globals.local twice.
https://github.com/netblue30/firejail/blob/a44fa49cb6cb6a00badd25b8f56e60e701fe8cc1/etc/templates/redirect_alias-profile.template#L9

rusty-snake (approved):

kmk3:
@pirate486743186 commented on [Aug 30](https://github.com/netblue30/firejail/pull/4493#issue-722741903):

> i'm sure netblue was just testing us :P

On one hand, there were pending suggested changes, on the other, there were
only approving reviews on the PRs (i.e.: no review was marked as a
comment/request for changes).

I'd say that the confusion here arises from people having different
expectations and thus it could potentially be alleviated by defining some basic
review/merge workflow guidelines.  Examples:

* If the PR contains things which should be changed anyway (as opposed to e.g.:
  nice-to-have further improvements), avoid marking the review as approved
* If there are any pending suggested changes, don't merge.  If you really want
  to merge it, ask about it wait for responses.

I think that it would be really nice to have something like that, as it would
help ensure that everyone is on the same page.

@netblue30 @rusty-snake What do you think?

---

Links for reference: #4484 #4486

rusty-snake:
If you know how [gerrit](https://www.gerritcodereview.com/) handles code-review approvals/blocks, you never stop to think why GitHub/GitLab doesn't have a system like this.

> * +2 Looks good to me, approved
> * +1 Looks good to me, but someone else must approve
> * 0 No score
> * -1 I would prefer that you didn’t submit this
> * -2 Do not submit

https://gerrit-review.googlesource.com/Documentation/intro-gerrit-walkthrough.html#_reviewing_the_change

kmk3:
@rusty-snake commented [on Aug 31](https://github.com/netblue30/firejail/pull/4493#issuecomment-909355360):

> If you know how [gerrit](https://www.gerritcodereview.com/) handles
> code-review approvals/blocks, you never stop to think why GitHub/GitLab
> doesn't have a system like this.
> 
> > * +2 Looks good to me, approved
> > * +1 Looks good to me, but someone else must approve
> > * 0 No score
> > * -1 I would prefer that you didn’t submit this
> > * -2 Do not submit
> 
> <https://gerrit-review.googlesource.com/Documentation/intro-gerrit-walkthrough.html#_reviewing_the_change>

Interesting; I don't know much about gerrit, but when there is nothing like the
above defined on a given GitHub project and there are multiple reviewers on
PRs, I sometimes assume that things work in a similar fashion.  I have also
worked near teams that did something resembling that as well.

rusty-snake:
thx

PR #4502, opened 2021-09-02
Revert "allow/deny help and man pages"

kmk3:
This reverts commit a11707ea273e5665047f8a7d9387ba07f08d72f6.

The man pages currently direct users to use the aliases instead of the
commands, which some users of firejail-git may end up doing.  Example:

https://github.com/netblue30/firejail/discussions/4496

So revert the man page changes as well to avoid confusion.

Note: This is not a full revert.  The commit in question also contains
some string formatting fixes on src/firejail/usage.c (related to dbus
and netmask), which are left intact.

Relates to #4410.


kmk3:
> Note: This is not a full revert. The commit in question also contains some
> string formatting fixes on src/firejail/usage.c (related to dbus and
> netmask), which are left intact.

Since there is a lot of code that is moved around on commit a11707ea2 (which
makes sense given that it's about renaming things), it's hard to tell tell if
there are more changes that are not directly related to the renaming.  Let me
know if I missed anything.

rusty-snake:
IMHO we do not need to revert blacklist-to-deny for options like seccomp or caps.drop.

Using --blacklist/--whitelist in the manpage for now (until this gets more progress) sounds reasonable.

kmk3:
@rusty-snake commented [on Sep 3](https://github.com/netblue30/firejail/pull/4502#issuecomment-912736445):

> IMHO we do not need to revert blacklist-to-deny for options like seccomp or
> caps.drop.

I assume that you're talking about these hunks:

> ```diff
> -	"    --caps.drop=capability,capability - drop capabilities.\n"
> -	"    --caps.keep=capability,capability - allow capabilities.\n"
> +	"    --caps.drop=capability,capability - blacklist capabilities filter.\n"
> +	"    --caps.keep=capability,capability - whitelist capabilities filter.\n"
> ```

> ```diff
> -	"    --seccomp - enable seccomp filter and drop the default syscalls.\n"
> -	"    --seccomp=syscall,syscall,syscall - enable seccomp filter, drop the\n"
> +	"    --seccomp - enable seccomp filter and apply the default blacklist.\n"
> +	"    --seccomp=syscall,syscall,syscall - enable seccomp filter, blacklist the\n"
>  	"\tdefault syscall list and the syscalls specified by the command.\n"
>  	"    --seccomp.block-secondary - build only the native architecture filters.\n"
>  	"    --seccomp.drop=syscall,syscall,syscall - enable seccomp filter, and\n"
> -	"\tdrop the syscalls specified by the command.\n"
> +	"\tblacklist the syscalls specified by the command.\n"
>  	"    --seccomp.keep=syscall,syscall,syscall - enable seccomp filter, and\n"
> -	"\tallow the syscalls specified by the command.\n"
> +	"\twhitelist the syscalls specified by the command.\n"
> ```

While I like the short naming of .drop/.keep (considering that they are
subcommands), I'd say that by themselves (i.e.: without further elaboration)
they may be confusing for users unfamiliar with firejail/seccomp.  Thus, I
think that making it explicit that e.g.: "drop open,close" means "blacklist
open,close" and that "keep open,close" means "whitelist open,close" may help
clarify things.  I mean, personally, knowing about the concept of "privdrop", I
could likely infer that "drop" means "blacklist", but I don't find "keep" to be
synonymous with "whitelist".  It could also have meant "nodrop", in the same
sense that `noblacklist` is the opposite of `blacklist`.  Considering that
"allow capabilities" is used above to describe `caps.keep`, this is similar to
the question of whether an `allow` command would make more sense as meaning
`noblacklist`, `whitelist`, neither, both or something else entirely, but I
digress.

Anyway, if you really want I can remove the above, as the other changes are the
main ones.  Either way, I'll leave this open for a few more days for any more
feedback, and then make all changes at once (if any), to avoid spamming #4410
with timeline events.
 
> Using --blacklist/--whitelist in the manpage for now (until this gets more
> progress) sounds reasonable.

:+1:

kmk3:
@rusty-snake ping

Thoughts on the above?

rusty-snake:
No, if you think reverting them is better then do it.

And I agreed that drop and blacklist are mostly interchangeable while keep can be a bit confusing 

```
$ firejail --quiet --noprofile --caps.keep=sys_admin --caps.drop=sys_admin --caps.keep=sys_admin cat /proc/self/status | grep CapBnd 
CapBnd:	0000000000000000
```

Issue #4509, opened 2021-09-04
Nextcloud profile broken - needs 3D and system tray access


zarandya:
The nextcloud client profile is broken: nextcloud crashes at startup as it is trying to use OpenGL for its gui. Also, it has no access to the system tray. 

I added this to `nextcloud.local` to fix it:
```
ignore no3d
dbus-user.talk org.freedesktop.StatusNotifierItem
dbus-user.talk org.kde.StatusNotifierWatcher
```

rusty-snake:
Is enough to add this?
```
ignore no3d
dbus-user.talk org.kde.StatusNotifierWatcher
```

zarandya:
> Is enough to add this?
> 
> ```
> ignore no3d
> dbus-user.talk org.kde.StatusNotifierWatcher
> ```

It indeed is.

PR #4510, opened 2021-09-04
Add new condition ALLOW_TRAY

rusty-snake:


reinerh (approved):

kmk3:
> ```diff
> diff --git a/etc/firejail.config b/etc/firejail.config
> index 2e355586b..5111bb769 100644
> --- a/etc/firejail.config
> +++ b/etc/firejail.config
> @@ -2,6 +2,9 @@
>  # keyword-argument pairs, one per line. Most features are enabled by default.
>  # Use 'yes' or 'no' as configuration values.
>  
> +# Allow programs to display a tray icon
> +# allow-tray no
> +
>  # Enable AppArmor functionality, default enabled.
>  # apparmor yes
> ```

From the PR date, I'd guess that this is related to issues like #4509:

@zarandya commented [on Sep 4](https://github.com/netblue30/firejail/issues/4509#issue-988302836):

> The nextcloud client profile is broken: nextcloud crashes at startup as it is
> trying to use OpenGL for its gui. Also, it has no access to the system tray.
> 
> I added this to `nextcloud.local` to fix it:
> 
> ```
> ignore no3d
> dbus-user.talk org.freedesktop.StatusNotifierItem
> dbus-user.talk org.kde.StatusNotifierWatcher
> ```

@zarandya commented [on Sep 4](https://github.com/netblue30/firejail/issues/4509#issuecomment-912991632):

> > Is enough to add this?
> > ```
> > ignore no3d
> > dbus-user.talk org.kde.StatusNotifierWatcher
> > ```
> 
> It indeed is.

So is this intended to just allow interacting with the system tray through
dbus?  If so, wouldn't it suffice to add some dbus filters like the above in a
new allow-system-tray.inc file?

rusty-snake:
> So is this intended to just allow interacting with the system tray through
dbus?

Yes

> If so, wouldn't it suffice to add some dbus filters like the above in a
new allow-system-tray.inc file?

It's a different idea. For tray¹ icons to work you need `dbus-user.talk org.kde.StatusNotifierWatcher` and, depending on the implementation, `dbus-user.own org.kde.*`. Both are unsafe and you don't want them allowed by default. On the other hand some users might heavily use tray icons. Therefore the Idea is to have a option in firejail.config which enables tray icons globally (profiles require `?ALLOW_TRAY: dbus-user.talk org.kde.StatusNotifierWatcher`) instead of creating a lot .locals. This is much easier for novice users, especially if it is integrated into firejail-welcome.sh.

¹ There are many tray icon standards but this is the most common one. 

General: I have already argued against allow-*.inc includes for D-Bus.

kmk3:
@rusty-snake commented [on Sep 10](https://github.com/netblue30/firejail/pull/4510#issuecomment-916657799):

> > So is this intended to just allow interacting with the system tray through
> > dbus?
> 
> Yes
> 
> > If so, wouldn't it suffice to add some dbus filters like the above in a new
> > allow-system-tray.inc file?
> 
> It's a different idea. For tray¹ icons to work you need `dbus-user.talk
> org.kde.StatusNotifierWatcher` and, depending on the implementation,
> `dbus-user.own org.kde.*`. Both are unsafe and you don't want them allowed by
> default. On the other hand some users might heavily use tray icons. Therefore
> the Idea is to have a option in firejail.config which enables tray icons
> globally (profiles require `?ALLOW_TRAY: dbus-user.talk
> org.kde.StatusNotifierWatcher`) instead of creating a lot .locals. This is
> much easier for novice users, especially if it is integrated into
> firejail-welcome.sh.

> `?ALLOW_TRAY: dbus-user.talk org.kde.StatusNotifierWatcher`

That makes sense; +1 for the condition.

It wasn't clear to me how this was intended to be implemented, so my first
thought was that the filters were going to be hardcoded in firejail.

> ¹ There are many tray icon standards but this is the most common one.

Indeed; link for reference: #4053.

kmk3:
@rusty-snake commented [on Sep 10](https://github.com/netblue30/firejail/pull/4510#issuecomment-916657799):

> General: I have already argued against allow-\*.inc includes for D-Bus.

(Continued on #4524)

PR #4521, opened 2021-09-09
Create disable-proc.inc

rusty-snake:
Any suggestions **and testing** welcome.


topimiettinen:
Nice idea. I did something similar with systemd for all user apps (using `InaccessiblePaths=` in `display-manager.service` and `user@.service`) and found that some apps want access to specific `/proc` files or they don't work otherwise. Here are my notes:
```
qps: /proc/loadavg /proc/stat
ps: /proc/meminfo /proc/uptime 
kscreenlocker: /proc/modules /proc/driver 
libreoffice: /proc/version 
gnome-system-monitor: /proc/cpuinfo /proc/vmstat 
```

But the baseline could (should) be tighter and it can then be relaxed for specific apps.

rusty-snake:
Thanks for the input. 

I use it for firefox for a while now (no issues) and added it to my globals.local yesterday.
My list so far (only reports by `tracelog`, no breakage)
- meminfo & zoneinfo: accessed very frequently (~1s) (maybe in a loop until success) by newsflash
- filesystems: WebKit*, gsettings, mpv, meld, gucharmap, ...
- version: flameshot, keepassxc

rusty-snake:
Worked fine for 1week. Let's move forward to `/proc/sys`. Current draft:
```
blacklist /proc/sys/abi
blacklist /proc/sys/crypto
blacklist /proc/sys/debug
blacklist /proc/sys/fs
blacklist /proc/sys/net
blacklist /proc/sys/user
blacklist /proc/sys/vm

noblacklist /proc/sys/kernel/osrelease
noblacklist /proc/sys/kernel/yama
blacklist /proc/sys/*/*
```

topimiettinen:
What about `/proc/sys/dev` and `/proc/sys/kernel`?

rusty-snake:
> /proc/sys/dev

accidentally deleted

> /proc/sys/kernel

- libreoffice wants to fopen `/proc/sys/kernel/osrelease`. I didn't noticed anything broken. 
- chrom* (tested with brave) does not start if it can not read `/proc/sys/kernel/yama/ptrace_scope`.

Therefore I used `blacklist /proc/sys/*/*` except for `/proc/sys/kernel/osrelease` and `/proc/sys/kernel/yama`.

topimiettinen:
I tried adding an audit rule to check for user's `/proc/sys` accesses like this
``` bash
$ cat /etc/audit/rules.d/50-local-proc-sys.rules 
-a always,exit -S all -F dir=/proc/sys -F perm=rwxa -F uid=1000 -k proc-sys
```
but it does not seem to work.

smitsohu:
Can this be merged?

rusty-snake:
I've nothing to add from my side.

smitsohu:
Let's merge.

PR #4559, opened 2021-09-18
Add ld.so.preload to all private-etc lines

rusty-snake:
Command:

```bash
sed -i -E "s/^private-etc /private-etc ld.so.preload,/" \
    $(grep -LE "^private-etc .*ld.so.preload" etc/profile-*/*) \
    && python3 contrib/sort.py etc/profile-*/*
```


smitsohu (approved):

netblue30:
merged!

Issue #2758, opened 2019-06-10
Firejail does not work with a custom hosts file


rmrgh:
When running firejail with a custom hosts file:
`$ firejail --noprofile --private --hosts-file=hosts firefox -no-remote`
it reports the following errors:
```
Parent pid 32480, child pid 32481
Error: invalid /etc/hosts file
Error: proc 32480 cannot sync with peer: unexpected EOF
Peer 32481 unexpectedly exited with status 1
```

My hosts file contains only:
```
127.0.0.1 sample.localhost
```

Even with an empty file, it report the same errors.

**Firejail Version:** 0.9.58
**Linux Distribution:** NixOS 19.03

rusty-snake:
Can't reproduce with firejail 0.9.61 under Fedora 29.
```
$ firejail --private --noprofile --hosts-file=hfile firefox-wayland -no-remote
… No Error about hosts
$ cat hfile
127.0.0.1 some.localhost
```

chiraag-nataraj:
I even just tried `firejail --hosts-file=/dev/null` and it worked.

netblue30:
There could be two reasons it to fail: if the host file is a symbolic link, or if the user doesn't have read access to the file. Also, try to put a full path for the host file, something like "--host-file=/home/username/hostfile". Works fine on 0.9.58, that code didn't change in ages.

rmrgh:
I've tried with a VoidLinux installation with firejail version 0.9.60 and it worked just fine. I think it's something specific for NixOS since the `/etc/hosts` present on it is a symbolic link and can't be directly manipulated (readonly).

rmrgh:
I've tested on my VoidLinux installation by moving the `/etc/hosts` file to a different location and symblinking it back to `/etc/hosts` and the same error showed up.

Nudin:
I got the same issue on Arch with firejail 0.9.62. The file is no symlink and readable by the user.
```
$ touch /home/michi/empty
$ chmod 777 /home/michi/empty
$ firejail --noprofile --hosts-file=/home/michi/empty
Parent pid 87979, child pid 87980
Error: invalid /etc/hosts file
Error: proc 87979 cannot sync with peer: unexpected EOF
Peer 87980 unexpectedly exited with status 1
```

Enteee:
[...]

PR #4560, opened 2021-09-19
Remove /etc/hosts is_link check

onny:
This patch fixes https://github.com/netblue30/firejail/issues/2758#issuecomment-805174951 . Especially for NixOS users, the file `/etc/hosts` is a symlink and firejail fails with exiting after `is_link()` check.
Removing this check fixes the issue.


netblue30:
Thanks!

PR #4515, opened 2021-09-07
Rework issue templates

rusty-snake:

PR #4479, opened 2021-08-25
Issue template improvements

kmk3:
Make sections easier to see and edit.

By the way, I've intended to do this for a while, as I always thought that the
template looked rather crooked.  I had assumed that it was just created from
scratch by hand like any other file, but recently I learned that GitHub is what
generates issue templates using such unusual markup for some reason.  Now I see
that it's even mentioned in the commit that adds the template:

```console
$ git show --pretty='%h %ai %s%n%n%b' --name-status bd29bf7202
bd29bf720 2020-04-07 16:40:07 -0500 Add bug report template

(Mostly) auto-generated with GitHub, will need tweaking over time

A       .github/ISSUE_TEMPLATE/bug_report.md
```

Cc: @Fred-Barclay (from the commit) @reinerh @rusty-snake (from #4468)

PR #4561, opened 2021-09-19
Issue template improvements2

kmk3:
Use `LC_ALL` and add a checked item example.

Relates to #4479 and #4515.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,6 +7,13 @@ assignees: ''
 
 ---
 
+<!--
+See the following links for help with formatting:
+
+https://guides.github.com/features/mastering-markdown/
+https://docs.github.com/en/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax
+-->
+
 ### Description
 
 _Describe the bug_
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -30,7 +30,7 @@ _What actually happened_
 
 ### Behavior without a profile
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -44,6 +44,7 @@ _Any other detail that may help to understand/debug the problem_
 
 ### Checklist
 
+- [x] This is a checked item.
```
rusty-snake, line 47: :+1: for this.

I am not sure if this is the prettiest solution (having an own ugly example checkitem). Maybe something like

```markdown
<!--
- [x] This is a checked item.
-->
```

```markdown
### Checklist (check them like this `- [x]`)
```

Maybe it would make sense to link to https://guides.github.com/features/mastering-markdown/ too. WDYT?

rusty-snake (approved):
Approved but see my comment below.

.github/ISSUE_TEMPLATE/bug_report.md
kmk3: @rusty-snake [on Sep 19](https://github.com/netblue30/firejail/pull/4561#discussion_r711730650):

> +1 for this.
> 
> I am not sure if this is the prettiest solution (having an own ugly example
> checkitem).

Certainly not a pretty solution, though it's the most dummy-proof one that I
could think of.  And considering the formatting of some issues, dummy-proofing
seems like a helpful quality...

> Maybe something like
> 
> ```gfm
> 
> ```

I had thought about doing this, but it seemed a bit less obvious and it takes
more space.  If it works, it's good enough though.  Added it.

> Maybe it would make sense to link to
> https://guides.github.com/features/mastering-markdown/ too. WDYT?

Done.

rusty-snake (approved):

reinerh (approved):

PR #4572, opened 2021-09-22
Add missing final newlines

a1346054:
Split from #4497

Usually it is my editor warning about the missing newlines, but you can also run a command such as:
```sh
git ls-files | while IFS= read -r f; do [[ $(tail -c1 "$f" | xxd -p) == '0a' ]] || echo "$f"; done
```
which will give you a list of files in the repository that don't contain the final newline. You might need to tweak it a bit, since it also reports binary and empty files.


reinerh (approved):

netblue30:
all in, thanks!

kmk3 (approved):
@a1346054 commented [on Sep 22](https://github.com/netblue30/firejail/pull/4572#issue-1004375562):

> Split from #4497

I'll copy and paste part of [my review from there](https://github.com/netblue30/firejail/pull/4497#pullrequestreview-746590774):

> > * fix missing newlines at end of files
> 
> LGTM.
> 
> I'd personally avoid doing this until there is a commit hook/CI check to detect
> such things, but it does not change too many files, so sure.

kmk3:
@a1346054 commented [on Sep 22](https://github.com/netblue30/firejail/pull/4572#issue-1004375562):

> Usually it is my editor warning about the missing newlines, but you can also
> run a command such as:
> 
> ```shell
> git ls-files | while IFS= read -r f; do [[ $(tail -c1 "$f" | xxd -p) == '0a' ]] || echo "$f"; done
> ```
> 
> which will give you a list of files in the repository that don't contain the
> final newline. You might need to tweak it a bit, since it also reports binary
> and empty files.

Here's an improved version:

```sh
git ls-files --eol | grep -v -e '^i/-text ' -e '^i/none ' -e '^i/ ' |
cut -f 2 | while IFS= read -r f; do test "`tail -c1 "$f" | xxd -p`" = '0a' ||
printf '%s\n' "$f"; done
```

It ignores binary files, empty files and symlinks.  Also, it should work on
POSIX shells.

a1346054:
A small nitpick, that command will miss out text files that only contain 1 line and are not terminated by newline.

```shell
printf '%s' "this is a line with no ending newline" > newfile
```

newfile will be missed out.

kmk3:
@a1346054 commented [on Sep 24](https://github.com/netblue30/firejail/pull/4572#issuecomment-926670490):

> A small nitpick, that command will miss out text files that only contain 1
> line and are not terminated by newline.
> 
> ```shell
> printf '%s' "this is a line with no ending newline" > newfile
> ```
> 
> newfile will be missed out.

Good catch.  I tried to take some shortcuts and misunderstood the output of
`--eol`.  This should hopefully be good enough (but I didn't test it too much):

```sh
git ls-files --eol | grep -v -e '^i/-text ' | cut -f 2 | while IFS= read -r f
do test -s "$f" && test ! -h "$f" &&
test "`tail -c 1 "$f" | od -A n -t xC`" != ' 0a' && printf '%s\n' "$f"; done
```

This tests for empty files and symlinks directly and uses `od` for portability
(so it should only require POSIX tools and git).

I've written something similar to this before, but probably using `xargs -0`
instead of `read`.  I think that `read` should be faster (as it does not
require a subshell), but it may have issues with tricky filenames (I'm not
sure).

Nowadays, I would just use `git-stripspace` on the same set of files, as it
fixes other whitespace issues as well.  See commit 00437760d ("makefiles: fix
whitespace") / PR #4024 for an example.  But that would change way more files
on this repo and it would be better to only have to fix it once, so that's why
I'd rather put a commit hook/CI check first.

Cc: @Fred-Barclay (as another whitespace fixer)

a1346054:
We can build more and more complex commands to try to catch every possible edge case, and it is fun :)

I never used the -z option to git-ls-files because thankfully none of my filenames ever contain embedded newlines or other things that would warrant it, but it can be a good idea to include that nonetheless.

kmk3:
@a1346054 commented [on Sep 24](https://github.com/netblue30/firejail/pull/4572#issuecomment-926831591):

> We can build more and more complex commands to try to catch every possible
> edge case, and it is fun :)

Indeed.  One of the best ways to learn these things IMHO.

> I never used the -z option to git-ls-files because thankfully none of my
> filenames ever contain embedded newlines or other things that would warrant
> it, but it can be a good idea to include that nonetheless.

Besides having to worry less about filename shenanigans, I like to use null
characters because sometimes I get tripped up by IFS and nulls always just work
for me, as long as all commands involved support them.  At least for  commands
like `find` and `xargs`, the most common implementations (GNU, BSDs, busybox)
seem to support them.  Though I wouldn't use it for `git ls-files` in this case
because busybox doesn't support `grep -z`, for example.  I'd use it afterwards
to avoid dealing with IFS and because `xargs` seems to be more "composable" (as
in, you can just keep on piping things), so to speak.  Like in this example
from commit 7c2ef659c ("makefiles: replace character class with plain char") /
PR #4139:

>     $ find . -name .git -prune -o -type f \
>       \( -name Makefile -o -name Makefile.in \
>       -o -name '*.mk' -o -name '*.mk.in' \) -print0 |
>       xargs -0 grep -Fl '$(wildcard *.[h])' | tr '\n' '\000' |
>       xargs -0 sed -i.bak -e \
>       's/\$(wildcard \*.\[h\])/$(wildcard *.h)/'
>
>     $ find . -name .git -prune -o -type f \
>       -name '*.bak' -exec rm '{}' +

PR #4575, opened 2021-09-22
Trim excess whitespace

a1346054:
Split from #4497 for further discussion.

For trailing whitespace a commit hook/CI check could be implemented, but most of the excess whitespace is within other text which makes it harder for automated tools to deal with. I mainly spotted it when reading through the files.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,3 @@
-
 If your PR isn't about profiles or you have no idea how to do one of these, skip the following and go ahead with this PR.
 
 If you submit a PR for new profiles or changing profiles, please do the following:
--- a/.github/workflows/sort.yml
+++ b/.github/workflows/sort.yml
@@ -19,4 +19,3 @@ jobs:
     - uses: actions/checkout@v2
     - name: check profiles
       run: ./contrib/sort.py etc/*/{*.inc,*.profile}
-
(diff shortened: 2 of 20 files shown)


reinerh (approved):

kmk3 (changes requested):
>      src/man/firejail-profile.txt                  | 32 ++++++++---------
>      src/man/firejail.txt                          | 36 +++++++++----------

Please leave existing double-spacing (".  ") as is.

It makes text much more readable in monospace fonts IMO.  It also makes editing
less error-prone, as it allows e.g.: vim to detect when a sentence ends vs when
you're just using an abbreviation (e.g.: "Mr.") and act accordingly on
commands that operate on sentences.

For a more extensive write-up, I found the following blog post:

* <http://stevelosh.com/blog/2012/10/why-i-two-space/>

It's kind of a rant, but it explains the main points.

Other than that, LGTM.

a1346054:
The arguments are sound, I'll rework the PR to not remove those two spaces after a sentence.

The original reason why I removed them was to make it consistent with all the other text in firejail, where two spaces are not used.

a1346054:
Rebased the PR to implement your review comments.

kmk3 (approved):
@a1346054 commented [on Sep 24](https://github.com/netblue30/firejail/pull/4575#issuecomment-926651114):

> The arguments are sound, I'll rework the PR to not remove those two spaces
> after a sentence.
> 
> The original reason why I removed them was to make it consistent

That makes sense.

> with all the other text in firejail, where two spaces are not used.

To be fair, there are a few more places where it's used (see with e.g.:
`git grep -I -n '\.  '`).

@a1346054 commented [on Sep 25](https://github.com/netblue30/firejail/pull/4575#issuecomment-927170464):

> Rebased the PR to implement your review comments.

Thanks for that and for the whitespace fixes.

Issue #4578, opened 2021-09-27
`PATH_MAX` missed on musl libc


dm9pZCAq:
### Description
build failed due to `PATH_MAX` missed on musl libc

### Steps to Reproduce

build on musl libc (`voidlinux`, `alpinelinux`, `kisslinux`)

### Environment

- Gentoo linux with musl profile
- trying to build `firejail-0.9.64.4`

### Log

```
main.c:1003:26: error: use of undeclared identifier 'PATH_MAX'
                if (strlen(argv[i]) >= MAX_ARG_LEN) {
                                       ^
./firejail.h:364:22: note: expanded from macro 'MAX_ARG_LEN'
#define MAX_ARG_LEN (PATH_MAX + 32) // --foobar=PATH
                     ^
main.c:1015:23: error: use of undeclared identifier 'PATH_MAX'
                if (strlen(*ptr) >= MAX_ENV_LEN) {
                                    ^
./firejail.h:658:22: note: expanded from macro 'MAX_ENV_LEN'
#define MAX_ENV_LEN (PATH_MAX + 32)     // FOOBAR=SOME_PATH
                     ^
```

### How to fix
here is patch from `voidlinux`
https://github.com/void-linux/void-packages/blob/master/srcpkgs/firejail/patches/musl.patch

with this patch everything builds fine for me

kmk3:
@dm9pZCAq commented [on Sep 27](https://github.com/netblue30/firejail/issues/4578#issue-1008298442):

> ### Description
> 
> build failed due to `PATH_MAX` missed on musl libc
> 
> ### Steps to Reproduce
> 
> build on musl libc (`voidlinux`, `alpinelinux`, `kisslinux`)
> 
> ### Environment
> 
> * Gentoo linux with musl profile
> * trying to build `firejail-0.9.64.4`
> 
> ### Log
> 
> ```
> main.c:1003:26: error: use of undeclared identifier 'PATH_MAX'
>                 if (strlen(argv[i]) >= MAX_ARG_LEN) {
>                                        ^
> ./firejail.h:364:22: note: expanded from macro 'MAX_ARG_LEN'
> #define MAX_ARG_LEN (PATH_MAX + 32) // --foobar=PATH
>                      ^
> main.c:1015:23: error: use of undeclared identifier 'PATH_MAX'
>                 if (strlen(*ptr) >= MAX_ENV_LEN) {
>                                     ^
[...]

PR #4579, opened 2021-09-27
fix compilation on musl (#4578)

dm9pZCAq:


kmk3:
Upon closer inspection, there appear to be more issues related to limits.h, so
I created #4583.  Sorry for the duplicated effort; let's continue this over
there.

Misc: Link to the musl issue for cross-reference: #4578.

Issue #4328, opened 2021-06-03
Some firejail output goes to stdout instead of stderr


slowpeek:
**Bug and expected behavior**
- Some firejail output goes to stdout instead of stderr.

**Reproduce**
1. Run and see
    ```
    firejail --seccomp=\!name_to_handle_at pwd 2>/dev/null
    Seccomp list in: !name_to_handle_at, check list: @default-keep, prelist: unknown,
    Seccomp list in: !name_to_handle_at, check list: @default-keep, prelist: unknown,
    /etc/firejail
    ```
   It is not an artificial example btw. The seccomp filter comes from keepassxc profile inherited by keepassxc-cli hence the above lines are prepended to keepassxc-cli exported xml. `FIREJAIL_QUIET=yes` does suppress the lines but those are definitely debug information, it should not pollute stdout by default.

**Environment**
 - ubuntu focal
 - 0.9.64.4

**Additional context**
Source reference:
https://github.com/netblue30/firejail/blob/c50dd74f615aaa9d0ebd6a1c459c5b3268ba1b9d/src/lib/syscall.c#L1675-L1695

**Checklist**
 - [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
 - [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
 - [x] I have performed a short search for similar issues (to avoid opening a duplicate).
 - [x] If it is a AppImage, `--profile=PROFILENAME` is used to set the right profile.
 - [x] Used `LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 PROGRAM` to get english error-messages.
 - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
 - [x] This is not a question. Questions should be asked in https://github.com/netblue30/firejail/discussions.

rusty-snake:
1. IMHO we should print these messages only with `arg_debug`. Normal users likely doesn't even know what all those seccomp is.
2. For stdout vs. stderr we have #3749/#4275

netblue30:
I'm sending them to stderr for now, we'll have to rework it for #3749 and #4275.

Issue #4275, opened 2021-05-13
Rework quiet, debug and normal console output


rusty-snake:
We have multiple issues about how we print something to the console, which all go to the same code. Therefore I make this meta-issue.

#4168: `Error: invalid --debug command line option` if quiet-by-default is set in firejail.config
#3125: no way to selectively disable quiet-by-default in firejail.config
#3503: "ignore quiet" does not work in override (.local) files
#3354: Option --quiet suppresses important warnings
#3749: Where to print debug messages (stdout or stderr)?
#2322: [minor feature] add visual cue for firejail outputs

There are some more which aren't handled here like:

#2743: [enhancement] Print better error in the case of ownership issues

-----

Good to read: https://github.com/netblue30/firejail/issues/4168#issuecomment-815068523

IMHO we should only print warnings and errors to stderr by default and be quiet for everything else. The current `quiet` would be then a hard-queit if a program really need full ownership of stderr. In addition a `--info`/`--verbose`/`--debug=1`/`--debug=little`/whatever will show the current default output.

All the warnings/errors should make clear that they are printed by firejail (e.g. `firejail:Error:`). For the debug output it would be enough to print something like `--- end of firejail output ---`.

glitsj16:
> All the warnings/errors should make clear that they are printed by firejail (e.g. firejail:Error:).

This reminds me of #3325, which never really got of the ground. @topimiettinen [mentioned](https://github.com/netblue30/firejail/pull/3325#issuecomment-610337451) a possibility to improve errExit() and use it everywhere with a prefix. IMO that would solve this part of the issue. Just adding it here for future reference.

slowpeek:
I'm not sure is there a single reason to use stdout in firejail. It is a wrapper, hence it should not touch neither stdin nor stdout.

Issue #3749, opened 2020-11-13
Where to print debug messages (stdout or stderr)?


smitsohu:
While the vast majority of Firejail's debug messages currently go to `stdout`, occasionally they are printed to `stderr` as well. I'm curious if there is desire to have a uniform behavior (one way or the other).

Background: When attempting a fix for #2301 I ran into the problem that I cannot print messages to `stdout` for a short period of time (otherwise Firejail would mix debug info and file content).

glitsj16:
No personal desire in this context, besides avoiding such a mix obviously...

rusty-snake:
Related: #3354

smitsohu:
Personally no strong opinion either.

But having two different output destinations is somewhat inconvenient and error prone, as users might capture debug output by redirecting it to a file, potentially missing the important messages that are still going to `stderr`.

glitsj16:
@smitsohu That's indeed a very good point. I think I saw that kind of behaviour in the past and redirected both stdout and stderr to a file to avoid missing important info. Not sure if we can improve things in that regard.

PR #4585, opened 2021-09-29
add more EUID improvements

smitsohu:
Run more whitelist code with EUID of the user.

### DIFF
--- a/src/fcopy/main.c
+++ b/src/fcopy/main.c
@@ -88,7 +88,8 @@ static void selinux_relabel_path(const char *path, const char *inside_path) {
 		if (arg_debug)
 			printf("Relabeling %s as %s (%s)\n", path, inside_path, fcon);
 
-		setfilecon_raw(procfs_path, fcon);
+		if (setfilecon_raw(procfs_path, fcon) != 0 && arg_debug)
+			printf("Cannot relabel %s: %s\n", path, strerror(errno));
 	}
 	freecon(fcon);
  close:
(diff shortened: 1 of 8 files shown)


src/fcopy/main.c
```diff
@@ -88,7 +88,8 @@ static void selinux_relabel_path(const char *path, const char *inside_path) {
 		if (arg_debug)
 			printf("Relabeling %s as %s (%s)\n", path, inside_path, fcon);
 
-		setfilecon_raw(procfs_path, fcon);
+		if (setfilecon_raw(procfs_path, fcon) != 0 && arg_debug)
+			printf("Cannot relabel %s: %s\n", path, strerror(errno));
```
kmk3, line 92: I don't know much about the context, but shouldn't this go to stderr?

src/firejail/selinux.c
```diff
@@ -68,8 +79,16 @@ void selinux_relabel_path(const char *path, const char *inside_path)
 		if (arg_debug)
 			printf("Relabeling %s as %s (%s)\n", path, inside_path, fcon);
 
-		setfilecon_raw(procfs_path, fcon);
+		if (!called_as_root)
+			EUID_ROOT();
+
+		if (setfilecon_raw(procfs_path, fcon) != 0 && arg_debug)
+			printf("Cannot relabel %s: %s\n", path, strerror(errno));
```
kmk3, line 86: Same question here.

src/fcopy/main.c
rusty-snake: For reference: #4328, #3749, #4275.

Issue #4558, opened 2021-09-18
`--tracelog` and `--trace` override `/etc/ld.so.preload` inside the sandbox


rusty-snake:
### Description

`--tracelog` and `--trace` override `/etc/ld.so.preload` inside the sandbox.

### Steps to Reproduce

```console
$ cat /etc/ld.so.preload
/usr/lib64/libhardened_malloc.so 
$ firejail --noprofile --tracelog cat /etc/ld.so.preload
[…]
/run/firejail/lib/libtracelog.so
```

### Expected behavior

`--tracelog` and `--trace` append to an existing `/etc/ld.so.preload`.

```console
$ firejail --noprofile --tracelog cat /etc/ld.so.preload
[…]
/usr/lib64/libhardened_malloc.so /run/firejail/lib/libtracelog.so
```

### Actual behavior

`--tracelog` and `--trace` override an existing `/etc/ld.so.preload`.

### Additional context

Workaround is to `env LD_PRELOAD=…` in globals.local.

### Environment

- Fedora Workstation 34
- Firejail 0.9.67 from 7aec067-dirty

### Checklist

OT: We need to split-up between regressions with programs/profiles and bug in firejail itself.

- [x] I have performed a short search for similar issues (to avoid opening a duplicate).

### Log

<details>
<summary>Output of <code>firejail --debug /path/to/program</code></summary>
<p>

```
Autoselecting /bin/bash as shell
Building quoted command line: 'cat' '/etc/ld.so.preload' 
Command name #cat#
DISPLAY=:0 parsed as 0
Using the local network stack
Parent pid 407949, child pid 407950
Initializing child process
Host network configured
PID namespace installed
Mounting tmpfs on /run/firejail/mnt directory
Creating empty /run/firejail/mnt/seccomp directory
Creating empty /run/firejail/mnt/seccomp/seccomp.protocol file
Creating empty /run/firejail/mnt/seccomp/seccomp.postexec file
Creating empty /run/firejail/mnt/seccomp/seccomp.postexec32 file

** Warning: dropping all Linux capabilities and setting NO_NEW_PRIVS prctl **

Mounting /proc filesystem representing the PID namespace
Basic read-only filesystem:
Mounting read-only /etc
2095 2032 253:1 /etc /etc ro,relatime master:1 - xfs /dev/mapper/fedora_localhost--live-root rw,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
[...]

PR #4586, opened 2021-09-29
trace, tracelog: don't truncate /etc/ld.so.preload in sandbox

smitsohu:
Add support for preloaded libraries (like for example [libaslrmalloc](https://github.com/topimiettinen/libaslrmalloc)) to `--trace` and `--tracelog` options.

#4558


netblue30:
merged!

PR #4587, opened 2021-10-05
Fix vscodium

### LINKED ISSUE #3871: Firejail profile for VsCodium doesn't work: vscodium was renamed to codium
I had a problem of not getting updates for the way to get Visual Studio Code on Debian via a repository instead of the GitHub releases: https://gitlab.com/paulcarroty/vscodium-deb-rpm-repo/-/issues/50

I did not get any error or alike and "solved" it by going through that repo's guide in the readme again which I installed VsCodium another time, now not called vscodium but "codium" and the latest version (now I have two VSCodiums installed).

However, when I run `firejail --profile=/etc/firejail/vscodium.profile codium` I now get this error message:

```
Unable to write program user data.

Please make sure the following directories are writeable:

/home/username/.config/VSCodium
/home/username/.vscode-oss/extensions
/run/user/1000
```

I already created a /home/username/.config/firejail/code.local with:

```
noblacklist ${HOME}/.config/VSCodium
noblacklist ${HOME}/.vscode-oss/extensions
```

At least two issues remain:

* The vscodium profile (that's /etc/firejail/code.profile and /etc/firejail/vscodium.profile) seems to be broken/outdated
* How to solve the problem with /run/user/1000?

`echo $DBUS_SESSION_BUS_ADDRESS` returns `unix:path=/run/user/1000/bus`

Also from the default vscodium profile it looks like it's not possible to install extensions. Is that correct? If so that should be changed too but it would be a separate issue.

System: Debian10/KDE
Firejail: 0.9.64 (latest from backports)

kmk3:
Fixes #3871.

Cc: @henrythebuilder @mYnDstrEAm @rusty-snake (from #3871)

### DIFF
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -142,6 +142,7 @@ blacklist ${HOME}/.config/SubDownloader
 blacklist ${HOME}/.config/Thunar
 blacklist ${HOME}/.config/Twitch
 blacklist ${HOME}/.config/Unknown Organization
+blacklist ${HOME}/.config/VSCodium
 blacklist ${HOME}/.config/VirtualBox
 blacklist ${HOME}/.config/Whalebird
 blacklist ${HOME}/.config/Wire
--- a/etc/profile-a-l/codium.profile
+++ b/etc/profile-a-l/codium.profile
@@ -1,0 +1,10 @@
+# Firejail profile alias for VSCodium
+# This file is overwritten after every install/update
+# Persistent local customizations
+include codium.local
+# Persistent global definitions
+# added by included profile
+#include globals.local
+
+# Redirect
+include vscodium.profile
--- a/etc/profile-m-z/vscodium.profile
+++ b/etc/profile-m-z/vscodium.profile
@@ -7,6 +7,8 @@ include vscodium.local
 #include globals.local
 
 noblacklist ${HOME}/.VSCodium
+noblacklist ${HOME}/.config/VSCodium
+noblacklist ${HOME}/.vscode-oss
 
 # Redirect
 include code.profile
--- a/etc/profile-m-z/vscodium.profile
+++ b/etc/profile-m-z/vscodium.profile
@@ -1,4 +1,4 @@
-# Firejail profile alias for Visual Studio Code
+# Firejail profile alias for VSCodium
 # This file is overwritten after every install/update
 # Persistent local customizations
 include vscodium.local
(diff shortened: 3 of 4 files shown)


rusty-snake (approved):

henrythebuilder:
Thanks @kmk3 
I tried the changes (only for profiles) and they seem to work.
Two additional comments:
- about #3871,  and in general for VSCODE-related profiles (eg VSCODIUM type derivative applications), the profiles to use are those that include '_changes for Electron_', in particular  the commit [Refactor code.profile as electron redirect](https://github.com/netblue30/firejail/commit/8f867d029a6ae7b757190d9f273886d4bbc1344b), in this way with the latest profiles all seems to work without any local modification.
- For those like me use VSSCODE self-compiled and/or installed locally in _portable mode_ it is necessary to enable access to the local installation directory, for example with `read-write ${HOME}/bin/vscode` in the local profile.

:+1:

kmk3:
@henrythebuilder commented [on Oct 5](https://github.com/netblue30/firejail/pull/4587#issuecomment-934211647):

> Thanks @kmk3 I tried the changes (only for profiles) and they seem to work. Two additional comments:
> 
> * about [Firejail profile for VsCodium doesn't work: vscodium was renamed to
>   codium #3871](https://github.com/netblue30/firejail/issues/3871),  and in
>   general for VSCODE-related profiles (eg VSCODIUM type derivative
>   applications), the profiles to use are those that include '_changes for
>   Electron_', in particular  the commit [Refactor code.profile as electron
>   redirect](https://github.com/netblue30/firejail/commit/8f867d029a6ae7b757190d9f273886d4bbc1344b),
>   in this way with the latest profiles all seems to work without any local
>   modification.

Thanks for testing!

> * For those like me use VSCODE self-compiled and/or installed locally in
>   _portable mode_ it is necessary to enable access to the local installation
>   directory, for example with `read-write ${HOME}/bin/vscode` in the local
>   profile.
> 
> +1

Interesting, I didn't know that it had a portable mode on Linux.

henrythebuilder:
Portable mode is interesting because it is easy to manage without installation in the system (Windows, Linux or macOS), especially in the case of _self compilation_:

- [Portable Mode instructions for Visual Studio Code](https://code.visualstudio.com/docs/editor/portable)
- [How to run VSCodium in portable mode](https://github.com/VSCodium/vscodium/blob/master/DOCS.md#portable)

:+1:

netblue30:
merged!

PR #4590, opened 2021-10-05
Read mount id also on legacy kernels

smitsohu:
Recursive remounting (`noexec`, `read-only`, `read-write`) currently needs a mount id. So far this id could not be obtained on kernels before 3.15.

Add support for recursive remounting also on legacy kernels, like frequently found on mobile devices. Should help SailfishOS community ports.


Tomin1:
The oldest kernels we use in Sailfish OS are 3.10 and 3.18.

PR #4591, opened 2021-10-05
s/S_IWRITE/S_IWUSR/

kmk3:
They are equivalent, but only the latter is POSIX.  See sys_stat.h(0p)
of POSIX.1-2017[1].  From Section 14.9.5, The Mode Bits for Access
Permission of the glibc manual[2]:

> S_IWUSR
> S_IWRITE
>
>     Write permission bit for the owner of the file.  Usually 0200.
>     S_IWRITE is an obsolete synonym provided for BSD compatibility.

Current usage:

    $ git grep -F S_IWRITE -- src | wc -l
    11
    $ git grep -F S_IWUSR -- src | wc -l
    26

    $ git grep -l -z S_IWRITE -- src | xargs -0 -I '{}' sh -c \
      "printf '%s\n' \"\`sed 's/S_IWRITE/S_IWUSR/g' '{}'\`\" >'{}'"

Note: The other related non-POSIX macros are not used anywhere:

    $ git grep -F -e S_IREAD -e S_IEXEC -- src
    $

[1] https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html
[2] https://www.gnu.org/software/libc/manual/html_node/Permission-Bits.html

---

Misc: I noticed this on #4586.

Since this conflicts with #4586, I'm marking this as a draft until that is
merged.

Note to self: This will need to be rebased/reworded.

### DIFF
--- a/src/firejail/chroot.c
+++ b/src/firejail/chroot.c
@@ -86,7 +86,7 @@ static void update_file(int parentfd, const char *relpath) {
 	if (arg_debug)
 		printf("Updating chroot /%s\n", relpath);
 	unlinkat(parentfd, relpath, 0);
-	int out = openat(parentfd, relpath, O_WRONLY|O_CREAT|O_EXCL|O_CLOEXEC, S_IRUSR | S_IWRITE | S_IRGRP | S_IROTH);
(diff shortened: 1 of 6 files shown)


smitsohu:
> Since this conflicts with #4586, I'm marking this as a draft until that is
> merged.

Resolved in that pull request. Looks good!

netblue30:
All fine, the worst that can happen is to break the compilation on some old systems. I'm merging it in.

Issue #4592, opened 2021-10-05
`--build` creates invalid paths with "after,$HOME" $HOME expanded


matu3ba:
### Description

Path handling for `firejail --build nvim` is broken.

### Steps to Reproduce

_Steps to reproduce the behavior_

1. Run in bash `LANG=C firejail --build nvim`
2. See result contains stuff like `allow ${HOME}/.local/share/nvim/site/after,/etc/xdg/nvim/after,/home/misterspoon/.config/kdedefaults/nvim/after,/home/misterspoon/.config/nvim/after/start`
3. Install from git to have profile lookup.
4. firejail nvim
5. Get error shown in actual behavior

### Expected behavior

There should be never invalid paths generated. I hope this code is not used elsewhere.

### Actual behavior

site/after,/ ... nvim/after, and especially `/kdedefaults/nvim/after,` looks broken. Invalid path generated.
```txt
Error: "${HOME}/.local/share/nvim/site/after,/etc/xdg/nvim/after,/home/misterspoon/.config/kdedefaults/nvim/after,/home/misterspoon/.config/nvim/after/start" is an invalid filename: rejected character: ","
```

### Additional context

PATH contains the path to nvim installation from source. 
I have a file $HOME/.local/share/nvim/site/pack/packer/start/coq_nvim/.vars/runtime/lib/python3.9/site-packages/pip/_vendor/tenacity/after.py
and
$HOME/.local/share/nvim/site/pack/packer/start/coq_nvim/.vars/runtime/lib/python3.9/site-packages/pip/_vendor/tenacity/__pycache__/after.cpython-39.pyc.

### Environment

- Linux distribution and version Linux 5.14 Arch (Endeavour)
- Firejail version (`firejail --version`). master with commit 32fb5edd557d64945a39393e8c4fe8ba7fed4d93

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [ ] I can reproduce the issue without custom modifications (e.g. globals.local).
- [ ] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [ ] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [ ] I have performed a short search for similar issues (to avoid opening a duplicate).
[...]

PR #4594, opened 2021-10-06
build: allow building with address sanitizer

reinerh:
This allows building firejail (and other binaries) with address sanitizer (with `--enable-asan`).
I noticed that some were interested in this in #4592 (ping @rusty-snake @smitsohu).

I didn't rebuild `configure`, as I have a different autoconf version installed and it would only blow up the diff. If someone wants to test it, please run `autoreconf -vfi` before. (In my humble opinion generated files like `configure` should anyway not be part of the repository.)

Right now running some of the tools only finds memory leaks. Running firejail itself does not work for me, it fails with this error:
```
Error: proc 17925 cannot sync with peer: unexpected EOF
```
Running with `--noprofile` works, but I didn't figure out yet which setting is the one that breaks it.
Also so far I only tested it with gcc. clang might give different results.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -45,6 +45,15 @@ AS_IF([test "x$enable_analyzer" = "xyes"], [
 	EXTRA_CFLAGS="$EXTRA_CFLAGS -fanalyzer -Wno-analyzer-malloc-leak"
 ])
 
+AC_ARG_ENABLE([sanitizer],
+    AS_HELP_STRING([--enable-sanitizer=@<:@address | memory | undefined@:>@], [enable a compiler-based sanitizer (debug)]), [], [enable_sanitizer=no])
+AS_IF([test "x$enable_sanitizer" != "xno" ],
+	[AX_CHECK_COMPILE_FLAG([-fsanitize=$enable_sanitizer], [
+		EXTRA_CFLAGS="$EXTRA_CFLAGS -fsanitize=$enable_sanitizer -fno-omit-frame-pointer"
+		EXTRA_LDFLAGS="$EXTRA_LDFLAGS -fsanitize=$enable_sanitizer"
+	], [AC_MSG_ERROR([sanitizer not supported: $enable_sanitizer])]
+)])
+
 HAVE_APPARMOR=""
 AC_ARG_ENABLE([apparmor],
     AS_HELP_STRING([--enable-apparmor], [enable apparmor]))
--- a/configure
+++ b/configure
@@ -711,6 +711,7 @@ ac_subst_files=''
 ac_user_opts='
 enable_option_checking
 enable_analyzer
+enable_sanitizer
 enable_apparmor
 enable_selinux
 enable_dbusproxy
@@ -1368,6 +1369,8 @@ Optional Features:
   --disable-FEATURE       do not include FEATURE (same as --enable-FEATURE=no)
   --enable-FEATURE[=ARG]  include FEATURE [ARG=yes]
   --enable-analyzer       enable GCC static analyzer
+  --enable-sanitizer=[address | memory | undefined]
+                          enable a compiler-based sanitizer (debug)
   --enable-apparmor       enable apparmor
   --enable-selinux        SELinux labeling support
   --disable-dbusproxy     disable dbus proxy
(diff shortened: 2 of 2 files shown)


configure.ac
```diff
@@ -64,6 +64,14 @@ AS_IF([test "x$enable_selinux" = "xyes"], [
...
 ])
 
+AC_ARG_ENABLE([asan],
+    AS_HELP_STRING([--enable-asan], [enable address sanitizer]))
+AS_IF([test "x$enable_asan" = "xyes"], [
+	EXTRA_CFLAGS="$EXTRA_CFLAGS -fsanitize=address"
+	EXTRA_LDFLAGS="$EXTRA_LDFLAGS -fsanitize=address"
+])
+
+
```
kmk3, line 74: Extra newline:

```suggestion
```

configure.ac
```diff
@@ -64,6 +64,14 @@ AS_IF([test "x$enable_selinux" = "xyes"], [
 	AC_SUBST(HAVE_SELINUX)
 ])
 
+AC_ARG_ENABLE([asan],
+    AS_HELP_STRING([--enable-asan], [enable address sanitizer]))
+AS_IF([test "x$enable_asan" = "xyes"], [
+	EXTRA_CFLAGS="$EXTRA_CFLAGS -fsanitize=address"
+	EXTRA_LDFLAGS="$EXTRA_LDFLAGS -fsanitize=address"
+])
```
kmk3, line 72: Suggestion: Move it below the `--enable-analyzer` part, as `--enable-asan` is
also mostly just about compiler flags, while apparmor/selinux are things that
are linked in.

Misc: I think that apparmor/selinux might have made more sense using `--with-`
instead of `--enable-`, but I digress.

rusty-snake:
> Right now running some of the tools only finds memory leaks.

IIRC `ASAN_OPTIONS=detect_leaks=0 firejail --…`

rusty-snake:
Hmm if I `./configure --enable-asan && make` it spams stderr full with `AddressSanitizer:DEADLYSIGNAL` (more than 1000 lines per second) after `src/fseccomp/fseccomp default seccomp`. Anyone else getting this?

reinerh:
> Hmm if I `./configure --enable-asan && make` it spams stderr full with `AddressSanitizer:DEADLYSIGNAL` (more than 1000 lines per second) after `src/fseccomp/fseccomp default seccomp`. Anyone else getting this?

No, I don't get that error. It runs without output, procudes a seccomp file and exits with 0.

kmk3:
@reinerh commented [on Oct 7](https://github.com/netblue30/firejail/pull/4594#issuecomment-938173182):

> I'm wondering whether the flag should be changed to be more flexible to
> support different sanitizers. E.g. `--with-sanitizer=` and allow selection of
> either `address`, `memory`

Which flag(s) would `memory` add?

> or `undefined`.

What does it do by itself?  From the docs, to me it looks like each suboption
has to be explicitly enabled:

<https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html>:

> ```plaintext
> -fsanitize=undefined
> 
>     Enable UndefinedBehaviorSanitizer, a fast undefined behavior detector.
>     Various computations are instrumented to detect undefined behavior at
>     runtime.  Current suboptions are:
> 
>     -fsanitize=shift
> 
> 	This option enables checking that the result of a shift operation is
> 	not undefined.  Note that what exactly is considered undefined differs
> 	slightly between C and C++, as well as between ISO C90 and C99, etc.
> 	This option has two suboptions, -fsanitize=shift-base and
> 	-fsanitize=shift-exponent.
> 
>     -fsanitize=shift-exponent
> 
> 	This option enables checking that the second argument of a shift
> 	operation is not negative and is smaller than the precision of the
> 	promoted first argument.
> 
>     -fsanitize=shift-base
> 
> 	If the second argument of a shift operation is within range, check that
> 	the result of a shift operation is not undefined.  Note that what
> 	exactly is considered undefined differs slightly between C and C++, as
> 	well as between ISO C90 and C99, etc.
> 
>     -fsanitize=integer-divide-by-zero
> 
>         Detect integer division by zero.
> 
>     [...]
> ```

Or does it enable all the suboptions at once?

configure.ac
```diff
@@ -45,6 +45,15 @@ AS_IF([test "x$enable_analyzer" = "xyes"], [
...
 
+AC_ARG_ENABLE([sanitizer],
+    AS_HELP_STRING([--enable-sanitizer=@<:@address,memory,undefined@:>@], [enable address/memory/undefined sanitizer (debug)]))
+AS_IF([test "x$enable_sanitizer" = "xaddress" -o "x$enable_sanitizer" = "xmemory" -o "x$enable_sanitizer" = "xundefined"],
+	[AX_CHECK_COMPILE_FLAG([-fsanitize=$enable_sanitizer],[
+		EXTRA_CFLAGS="$EXTRA_CFLAGS -fsanitize=$enable_sanitizer -fno-omit-frame-pointer"
+		EXTRA_LDFLAGS="$EXTRA_LDFLAGS -fsanitize=$enable_sanitizer"
+],[AC_MSG_ERROR([$enable_sanitizer sanitizer not supported])]
+)])
+
```
kmk3, line 56:
```suggestion
AC_ARG_ENABLE([sanitizer],
    AS_HELP_STRING([--enable-sanitizer=@<:@address | memory | undefined@:>@],
        [enable a compiler-based sanitizer (debug)]), [], [enable_sanitizer=no])
AS_IF([test "x$enable_sanitizer" != "xno"], [
	AX_CHECK_COMPILE_FLAG([-fsanitize=$enable_sanitizer], [
		EXTRA_CFLAGS="$EXTRA_CFLAGS -fsanitize=$enable_sanitizer -fno-omit-frame-pointer"
		EXTRA_LDFLAGS="$EXTRA_LDFLAGS -fsanitize=$enable_sanitizer"
	], [AC_MSG_ERROR([sanitizer not supported: $enable_sanitizer])])
])

```

Main suggestion:

* Just test if flag was passed and then let `AX_CHECK_COMPILE_FLAG` deal with
  it (so that there's less need to maintain the sanitizer list)

Nitpicks:

* Make help string more cli-like (with pipes == OR)
* Make error message more greppable/more like perror(3p) messages (static part
  first, etc)

configure.ac
```diff
@@ -45,6 +45,15 @@ AS_IF([test "x$enable_analyzer" = "xyes"], [
 	EXTRA_CFLAGS="$EXTRA_CFLAGS -fanalyzer -Wno-analyzer-malloc-leak"
 ])
 
+AC_ARG_ENABLE([sanitizer],
+    AS_HELP_STRING([--enable-sanitizer=@<:@address | memory | undefined@:>@], [enable a compiler-based sanitizer (debug)]), [], [enable_sanitizer=no])
+AS_IF([test "x$enable_sanitizer" != "xno" ],
+	[AX_CHECK_COMPILE_FLAG([-fsanitize=$enable_sanitizer],[
```
kmk3, line 51:
```suggestion
	[AX_CHECK_COMPILE_FLAG([-fsanitize=$enable_sanitizer], [
```

Space nitpick.

Other than that, LGTM.

kmk3 (approved):

PR #4600, opened 2021-10-09
Add profiles for imv, retroarch, and torbrowser

crocket:
The profiles are based on profile template.

### DIFF
--- a/etc/profile-a-l/imv.profile
+++ b/etc/profile-a-l/imv.profile
@@ -1,0 +1,57 @@
+# Firejail profile for imv
+# Description: imv is an image viewer.
+# This file is overwritten after every install/update
+# Persistent local customizations
+include imv.local
+# Persistent global definitions
+include globals.local
+
+include allow-bin-sh.inc
+
+blacklist /usr/libexec
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
+# Users may want to view images in ${HOME}
+#include disable-xdg.inc
+
+# Users may want to view images in ${HOME}
+#include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+# Users may want to view images in /usr/share
+#include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+net none
+nodvd
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+
+private-bin imv,imv-wayland,imv-x11,sh
+private-cache
+private-dev
+private-tmp
+
+dbus-user none
+dbus-system none
+
+read-only ${HOME}
--- a/etc/profile-m-z/retroarch.profile
+++ b/etc/profile-m-z/retroarch.profile
@@ -1,0 +1,54 @@
+# Firejail profile for retroarch
+# Description: retroarch is a frontend to libretro emulator cores.
+# This file is overwritten after every install/update
+# Persistent local customizations
+include retroarch.local
+# Persistent global definitions
+include globals.local
+
+blacklist /usr/libexec
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/retroarch
+whitelist ${HOME}/.config/retroarch
+whitelist /run/udev
+whitelist /usr/share/retroarch
+whitelist /usr/share/libretro
+include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+netfilter
+nodvd
+nogroups
+nonewprivs
+noroot
+notv
+nou2f
+# If you need access to cameras, add `ignore novideo` to retroarch.local
+novideo
+protocol unix,inet,inet6,netlink
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+
+disable-mnt
+private-bin retroarch
+private-cache
+private-dev
+private-tmp
+
+dbus-user none
+dbus-system none
--- a/etc/profile-m-z/torbrowser.profile
+++ b/etc/profile-m-z/torbrowser.profile
@@ -1,0 +1,26 @@
+# Firejail profile for torbrowser
+# Description: This profile was tested with www-client/torbrowser::torbrowser
+# on Gentoo Linux.
+# This file is overwritten after every install/update
+# Persistent local customizations
+include torbrowser.local
+# Persistent global definitions
+include globals.local
+
+ignore dbus-user none
+
+noblacklist ${HOME}/.cache/mozilla
+noblacklist ${HOME}/.mozilla
+
+blacklist /usr/libexec
+
+mkdir ${HOME}/.cache/mozilla/torbrowser
+mkdir ${HOME}/.mozilla
+whitelist ${HOME}/.cache/mozilla/torbrowser
+whitelist ${HOME}/.mozilla
+include whitelist-usr-share-common.inc
+
+dbus-user filter
+dbus-user.own org.mozilla.torbrowser.*
+
+include firefox-common.profile
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -396,6 +396,7 @@ ideaIC
 imagej
 img2txt
 impressive
+imv
 inkscape
 inkview
 inox
@@ -687,6 +688,7 @@ rednotebook
 redshift
 regextester
 remmina
+retroarch
 rhythmbox
 rhythmbox-client
 ricochet
@@ -804,6 +806,7 @@ tor-browser-tr
 tor-browser-vi
 tor-browser-zh-cn
 tor-browser-zh-tw
+torbrowser
 torbrowser-launcher
 torcs
 totem


etc/profile-a-l/imv.profile
```diff
@@ -0,0 +1,40 @@
+# Persistent local customizations
```
rusty-snake, line 1: Headers

etc/profile-a-l/imv.profile
```diff
@@ -0,0 +1,40 @@
...
+
+include allow-bin-sh.inc
+
+blacklist /usr/libexec
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
```
rusty-snake, line 14:
```suggestion
```

etc/profile-a-l/imv.profile
```diff
@@ -0,0 +1,40 @@
...
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
+
+include whitelist-runuser-common.inc
```
rusty-snake, line 19: Can we use wvc, wrc, wusc?

etc/profile-a-l/imv.profile
```diff
@@ -0,0 +1,40 @@
...
+
+apparmor
+caps.drop all
+net none
+nodvd
+noinput
```
rusty-snake, line 25: `nogroups`?

etc/profile-a-l/imv.profile
```diff
@@ -0,0 +1,40 @@
...
+nonewprivs
+nosound
```
rusty-snake, line 27: `noroot`?

etc/profile-a-l/imv.profile
```diff
@@ -0,0 +1,40 @@
...
+notv
+nou2f
+novideo
+protocol unix
```
rusty-snake, line 31: Does it need `unix`?

etc/profile-a-l/imv.profile
```diff
@@ -0,0 +1,40 @@
...
+seccomp
```
rusty-snake, line 32: `seccomp.block-secondary`?

etc/profile-a-l/imv.profile
```diff
@@ -0,0 +1,40 @@
...
+shell none
```
rusty-snake, line 33: `tracelog`?

etc/profile-a-l/imv.profile
```diff
@@ -0,0 +1,40 @@
...
+
+private-bin imv,imv-wayland,imv-x11,sh
```
rusty-snake, line 35: `private-cache`?

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
+# Persistent local customizations
```
rusty-snake, line 1: Headers

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+include retroarch.local
+# Persistent global definitions
+include globals.local
+
+blacklist /boot
+blacklist /opt
```
rusty-snake, line 7:
```suggestion
```

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+blacklist /usr/libexec
+blacklist /var
```
rusty-snake, line 9:
```suggestion
```

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
```
rusty-snake, line 18:
```suggestion
```

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+blacklist /boot
+blacklist /opt
+blacklist /usr/libexec
+blacklist /var
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
```
rusty-snake, line 15:
```suggestion
```

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
+include disable-xdg.inc
+
+whitelist ${HOME}/.config/retroarch
+whitelist /usr/share/retroarch
+whitelist /usr/share/libretro
+
```
rusty-snake, line 24:
```suggestion
```

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
+include disable-xdg.inc
+
+whitelist ${HOME}/.config/retroarch
```
rusty-snake, line 21: `mkdir`+`blacklist`(in dp.inc)+`noblacklist`

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+whitelist /usr/share/retroarch
+whitelist /usr/share/libretro
+
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
```
rusty-snake, line 27: Can we use wr and wvc?

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+apparmor
+caps.drop all
+netfilter
+nodvd
+nonewprivs
+notv
+nou2f
+protocol unix,inet,inet6
+seccomp
+shell none
```
rusty-snake, line 38: `tracelog`?

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+
+apparmor
+caps.drop all
+netfilter
+nodvd
+nonewprivs
+notv
+nou2f
+protocol unix,inet,inet6
+seccomp
```
rusty-snake, line 37: `seccomp.block-secondary`?

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+whitelist /usr/share/libretro
+
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+
+apparmor
+caps.drop all
+netfilter
+nodvd
```
rusty-snake, line 32: `nogroups`?

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+nonewprivs
+notv
```
rusty-snake, line 34: `noroot`?

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+nou2f
```
rusty-snake, line 35: `novideo`?

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,46 @@
...
+protocol unix,inet,inet6
+seccomp
+shell none
+
+disable-mnt
+private-bin retroarch
```
rusty-snake, line 41: `private-cache`?

etc/profile-m-z/torbrowser.profile
```diff
@@ -0,0 +1,22 @@
...
+mkdir ${HOME}/.cache/mozilla/torbrowser
+mkdir ${HOME}/.mozilla
+whitelist ${HOME}/.cache/mozilla/torbrowser
+whitelist ${HOME}/.mozilla
+include whitelist-usr-share-common.inc
+
+dbus-user filter
+dbus-user.own org.mozilla.torbrowser.*
+
+include firefox-common.profile
```
rusty-snake, line 22:
```suggestion
# Firejail profile for torbrowser
# This file is overwritten after every install/update
# Persistent local customizations
include torbrowser.local
# Persistent global definitions
# added by included profile
#include globals.local

# Redirect
include start-tor-browser.profile
```

etc/profile-a-l/imv.profile
crocket: Why remove disable-passwdmgr?

etc/profile-a-l/imv.profile
crocket: I thought I wanted to view images in var and usr.

etc/profile-a-l/imv.profile
crocket: noroot breaks imv on my gentoo linux system.

etc/profile-m-z/retroarch.profile
crocket: It works fine without /var.

etc/profile-m-z/retroarch.profile
crocket: nogroups breaks my gentoo linux system.

etc/profile-m-z/retroarch.profile
crocket: noroot breaks my gentoo linux system.

etc/profile-m-z/retroarch.profile
crocket: I don't know what exactly `novideo` does. What are video devices?

etc/profile-m-z/torbrowser.profile
crocket: www-client/torbrowser::torbrowser on Gentoo Linux doesn't use torbrowser-launcher. It compiles torbrowser directly from source code.

etc/profile-m-z/retroarch.profile
crocket: Why?

etc/profile-m-z/retroarch.profile
crocket: Added it to my local copy.

etc/profile-a-l/imv.profile
crocket: Added it to my local copy.

etc/profile-m-z/retroarch.profile
crocket: Added headers to my local copy.

etc/profile-m-z/retroarch.profile
crocket: Okay, people may put their ROMs in /mnt. Removed it from my local copy.

etc/profile-m-z/retroarch.profile
crocket: ???

etc/profile-m-z/retroarch.profile
crocket: Done in my local copy.

etc/profile-m-z/retroarch.profile
crocket: I'm going to test it with every libretro core.

etc/profile-a-l/imv.profile
crocket: Why?

etc/profile-a-l/imv.profile
crocket: Done

etc/profile-a-l/imv.profile
kmk3: @crocket [on Oct 9](https://github.com/netblue30/firejail/pull/4600#discussion_r725546759):

> Why remove disable-passwdmgr?

It does not exist anymore.

See #4461 and the current profile.template.

kmk3:
@crocket commented [on Oct 9](https://github.com/netblue30/firejail/pull/4600#issuecomment-939374645):

> Whenever I read `nogroups` and `noroot`, I feel marginalized as someone who
> doesn't use systemd.
> 
> Quite a few linux distributions don't use systemd by default.
> 
> * Void Linux
> * Gentoo Linux
> * Artix Linux
> * GNU Guix System
> * ...
> 
> https://nosystemd.org/
> 
> Don't break linux distributions for systemd.

What do you mean?  I use Artix Linux myself and I don't remember having
problems with either option.

crocket:
I tell you. Either `nogroup` or `noroot` breaks imv and retroarch.

etc/profile-a-l/imv.profile
crocket: After further testing, it turned out that what broke imv was actually file permission on images.

Some image files were not owned by the normal user. `nogroup` and `noroot` don't actually break imv.

etc/profile-a-l/imv.profile
crocket: After further testing, it turned out that what broke imv was actually file permission on images.

Some image files were not owned by the normal user. `nogroup` and `noroot` don't actually break imv.

Added it to my local copy.

etc/profile-a-l/imv.profile
crocket: Removed from my local copy.

etc/profile-m-z/retroarch.profile
crocket: Removed from my local copy.

etc/profile-m-z/retroarch.profile
crocket: It seems `novideo` turns off cameras. I added it to my local copy.

etc/profile-m-z/retroarch.profile
crocket: ????

etc/profile-a-l/imv.profile
crocket: It seems that unix socket is required for wayland.

rusty-snake: > Can net none alone disable unix sockets?

No it is not a syscall related option, it is a namespace related option. To block all protocols, use `seccomp socket`. Anyway if it uses wayland ...

etc/profile-a-l/imv.profile
rusty-snake: Because we add it to every profile where it works.

etc/profile-m-z/retroarch.profile
rusty-snake: `/boot` is a hardcoded blacklist, you don't need to repeat it.

`/opt` might me necessary if someone installs it there ...

etc/profile-m-z/retroarch.profile
rusty-snake: IDC if it works fine for you without `/var` it will not work on systems which use `/var/lib/ca-certificates` for example.

etc/profile-m-z/retroarch.profile
rusty-snake: No.

etc/profile-m-z/retroarch.profile
rusty-snake: Why not?

etc/profile-m-z/retroarch.profile
rusty-snake: Only if you blacklist a file that it tries to access in a loop.

etc/profile-m-z/retroarch.profile
rusty-snake: IDK how common/necessary it is. Maybe add a comment.

etc/profile-m-z/torbrowser.profile
rusty-snake: Ahh, it is only for the gentoo package, then add this information to the description.

crocket: I just did in my local copy. It will be reflected soon in the pull request.

rusty-snake:
> Whenever I read nogroups and noroot, I feel marginalized as someone who doesn't use systemd.
> Don't break linux distributions for systemd.

Users who refuse the technical progress of the last ten years are no reason to reduce the security for all. 

> Quite a few linux distributions don't use systemd by default.

And here's their distrowatch ranking, guess what the majority of users have.

1. MX Linux
3. Manjaro
4. Mint
6. Ubuntu
7. Debian
10. Fedora
12. openSUSE
16. Arch
19. Artix
35. Devuan
41. Void
53. Gentoo

> https://nosystemd.org/

Sorry, but this site is just bullshit.

> List of notable bugs and security issues:

Guess where bugs are found, it the software used by a few  people with a heavily selfconfigured system or in software used by the majority of users and in all enterprise distribution.

> systemd will be able to kill a process if he needs/wants more RAM

Sorry but that's a feature.

> systemd does not respect system wide resource limits

TBH, resource limit configuration is a mess like shell startup configuration. And `/etc/security/limits.conf` isn't "the system-wide resource limits" configuration file. RT(F)M "configuration file for the pam_limits module". If something does not use pam at all of a pam-configuraton without pam_limits ...

etc/profile-a-l/imv.profile
rusty-snake: usr/share makes sense. var, IDK if it contains any images except flatpak icons, run can be added imho.

etc/profile-m-z/retroarch.profile
rusty-snake: https://github.com/netblue30/firejail/blob/07e9029b24e21de4c6ecae3060bed3d2450bb296/etc/templates/profile.template#L62-L64

```suggestion
mkdir ${HOME}/.config/retroarch
whitelist ${HOME}/.config/retroarch
```

crocket:
> In a multi-user system, a remotely logged in user can listen to microphone, camera and input devices when another user is logged in locally. The users can also make set-gid binaries to keep access even if the group would be removed.

That particular attack scenario is rare and difficult to pull off. There is no way that someone is just remotely logging in my computer. Don't assume that remote login is easy. Even after remote login, making a setgid binary as a non-root user isn't trivial. There is no easy shortcut.

Not sure about `nogroups`, but `noroot` is questionable from security standpoint. It uses user namespace which may have more security holes than an SUID binary such as firejail.

> Users who refuse the technical progress of the last ten years are no reason to reduce the security for all.

Some people say systemd represents technical progress, but I don't see much progress being made by systemd.

It doesn't offer enough real advantages over other alternatives to entice me. It is just a big lump of softwares that do a lot of things. Technical progress isn't measured by the scope or size of a software. I used to use systemd on Arch Linux. But, I settled with OpenRC because it strikes a good balance between features and simplicity. OpenRC does what I want without much hassle and without sysvinit.

We haven't had much technical progress in general over the last ten years. CPU speed hasn't increased much. Software quality certainly hasn't increased. Consider the possibility that there hasn't been much technical progress in general over the last 10 years.

I also think tools like firejail shouldn't be too dependent on one specific software such as systemd. Detecting systemd with `NOSYSTEMD?` is like a web page detecting browsers. Nowadays, web pages detect browser capabilities.

To make firejail work smoothly on all linux distributions and perhaps BSDs, firejail should detect capabilities.

topimiettinen:
> > In a multi-user system, a remotely logged in user can listen to microphone, camera and input devices when another user is logged in locally. The users can also make set-gid binaries to keep access even if the group would be removed.
> 
> Actual multi-user system doesn't exist outside shared hosting. Almost everyone has dedicated machines. That particular attack scenario is rare and difficult to pull off. There is no way that someone is just remotely logging in my computer. Don't assume that remote login is easy. Even after remote login, making a setgid binary as a non-root user isn't trivial. There is no easy shortcut.

Multi-user (even multi-seat) systems exist in internet cafés. The totally unprivileged command is `chmod g+s filename`. It's really trivial and easy.

Do you use `elogind`?

crocket:
Internet cafes don't use linux. Almost all of them use windows. Also, internet cafe computers are used by one person at a time.

Firejail targets home market for people who are sensitive to security and privacy.  It's safe to assume that firejail users know how to secure their systems.

topimiettinen:
> Internet cafes don't use linux. Almost all of them use windows. Forget about security in internet cafe computers. It's a lost cause. You are not going to get it. If you don't manage the computer that you are using, don't expect it to be secure and private.

Then there are schools and libraries. The system administrator must of course be trusted, but it's another level when you can't trust other unprivileged users either.

> Restricted access to supplementary groups has yet to be implemented.

Restricted access to devices has been implemented, it's called `uaccess` and ACLs. elogind also implements this.

I think you didn't get my idea of `?NOSYSTEMD`. The configurations would remain exactly same for systemd/elogind and other users. For systemd/elogind users, `?NOSYSTEMD ignore noroot` (or perhaps`?SYSTEMD noroot`) would enable improved security of `noroot`. Other users would also get a working system with the same configuration, but without the extra security. The magic directive `?NOSYSTEMD` would do the work.

> I replaced elogind with seatd

It seems that seatd doesn't implement `uaccess`/ACL stuff. It would probably not be too difficult to add it [here](https://git.sr.ht/~kennylevinsen/seatd/tree/master/item/seatd/seat.c).

crocket:
schools and libraries also use windows or sometimes Mac OS. Linux is only dominant in server market. I'm not counting android as linux.

I think `?NOSYSTEMD` is inaccurate. As far as I know, systemd alone can't give temporary access to audio/input devices.

topimiettinen:
> schools and libraries also use windows or sometimes Mac OS. Linux is only dominant in server market. I'm not counting android as linux.

Wikipedia entry for [Multiseat configuration](https://en.wikipedia.org/wiki/Multiseat_configuration) lists installations in Brazil and Tanzania. There's also a [company](http://www.multi-seat.com/) selling multiseat equipment.

> I think `?NOSYSTEMD` is inaccurate. As far as I know, systemd alone can't give temporary access to audio/input devices. It's the job of pulseaudio or pipewire.

Right, choice of PID1 doesn't have anything to do with this. It's a combination of udevd and systemd-logind/elogind. `?UACCESS` would be more accurate, though a bit cryptic. 

Actually it would be simple to implement this as a `configure` option, for example `--without-systemd` or `--without-uaccess`:
- When Firejail is built with `configure --without-systemd`, `?UACCESS` would drop the rest of the commands. I suppose this could be enabled in Gentoo with USE flags which exclude both systemd and elogind.
- When Firejail is built with  `configure --with-systemd` (the default), `?UACCESS` would execute the commands. In Gentoo, USE flags systemd or elogind would enable this.

I don't know too much about Gentoo (the one time I tried to install it, the kernel which took ages to compile didn't boot), so the examples could be wrong.

rusty-snake:
> schools and libraries also use windows or sometimes Mac OS. Linux is only dominant in server market.

https://www.linuxmuster.net

topimiettinen:
> We already disable `nogroups` for nVidia users at runtime maybe this code could be extended.

I don't know how to detect absence of systemd-logind or elogind at runtime, except maybe by using APIs of `libsystemd` (linking it to Firejail). The nice thing about the build time switch is that anti-systemd distros can get what they want (their users like @crocket could contribute nice profiles that can be useful elsewhere too) but it doesn't reduce security for modern distros. Perhaps it wouldn't work correctly in the case where systemd is installed, but I think in that case Gentoo users could be expected to recompile all things enabling systemd support everywhere anyway and other distros may have separate systemd/non-systemd package variants or whatever.

I think nvidia driver can't use uaccess because of some GPL issues or limitations to APIs/ABIs for non-GPL drivers. Or maybe the driver is just written to work with the lowest common denominator.

crocket:
`noroot` or `nogroups` will break retroarch.profile.
For now, `noroot` and `nogroups` can reside without `?UACCESS` in torbrowser.profile.
`noroot` and `nogroups` didn't break imv.profile.

etc/profile-m-z/retroarch.profile
crocket: Just removed it.

etc/profile-m-z/retroarch.profile
crocket: I don't understand why `whitelist` requires `mkdir`, but I just added it.

etc/profile-m-z/retroarch.profile
crocket: Added with a comment.

etc/profile-m-z/retroarch.profile
crocket: It is fine with

```
include whitelist-common.inc
```

rusty-snake: and wrc?

etc/profile-a-l/imv.profile
```diff
@@ -0,0 +1,52 @@
...
+shell none
+tracelog
+
+private-bin imv,imv-wayland,imv-x11,sh
+private-cache
+private-dev
+private-tmp
+
+dbus-user none
+dbus-system none
```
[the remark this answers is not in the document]
crocket: `private-tmp` makes `read-only /tmp` redundant.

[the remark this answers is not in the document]
rusty-snake: Not necessarily.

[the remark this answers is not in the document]
crocket: By default, `read-only ${HOME}` and `read-only /tmp` don't break imv. But, customized user scripts executed from imv may want to write to ${HOME}. I added `read-only ${HOME}` and `read-only /tmp` to my local copy.

Is `read-only /tmp` useful in the presence of `private-tmp`?

etc/profile-m-z/retroarch.profile
crocket: wrc is `include whitelist-runuser-common.inc`.

rusty-snake: Review comment says

> wrc: whitelist-run-common.inc

and whitelist-runuser-common.inc is wruc

crocket: Yes, I applied it to retroarch.profile.

etc/profile-m-z/retroarch.profile
```diff
@@ -0,0 +1,52 @@
...
+
+apparmor
+caps.drop all
+netfilter
+nodvd
+# nogroups breaks udev input driver which requires access to /dev/input
+#nogroups
+nonewprivs
+# noroot breaks udev input driver which requires access to /dev/input
+#noroot
```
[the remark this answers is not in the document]
davidebeatrici: @crocket How do the permissions in `/dev/input` look like on your system?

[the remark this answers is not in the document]
crocket: I just applied noroot and nogroups to retroarch.profile and pushed the change. Can somebody test retroarch.profile on a system that has (e)logind?

crocket:
I just added a comment to imv.profile. I also added disable-mnt to retroarch.profile because `man hier` says `/mnt` is for temporarily mounted filesystems.

crocket:
I guess the profile already works on logind systems. I suggest just merging this pull request and deal with any problems later.

crocket:
I just noticed that sdl2 controller driver doesn't work inside firejail. It needs a little more work.

/var/log/messages

```
Oct 17 21:35:59 localhost kernel: retroarch[14465]: segfault at 28 ip 00007fc6332f9c2a sp 00007ffcb6bf61f8 error 4 in libSDL2-2.0.so.0.14.0[7fc633214000+e9000]
```

crocket:
I just fixed it by allowing netlink protocol which is somehow needed by sdl2.

PR #4606, opened 2021-10-14
libtrace.c: use realpath instead of readlink to avoid PATH_MAX

kmk3:
PATH_MAX is not guaranteed to be defined and it may be defined to -1.
Avoid depending on it by getting the result directly from realpath.  See
commit 579f856c5 ("firejail.h: add missing linux/limits.h include") /
PR #4583 for details.

Note: This replaces the static char array currently used with a dynamic
one returned from realpath.

Misc: This is a continuation of #4583.

### DIFF
--- a/src/libtrace/libtrace.c
+++ b/src/libtrace/libtrace.c
@@ -18,12 +18,12 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 #define _GNU_SOURCE
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <dlfcn.h>
 #include <sys/types.h>
-#include <limits.h>
 #include <unistd.h>
 #include <sys/socket.h>
 #include <netinet/in.h>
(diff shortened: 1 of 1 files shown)


smitsohu:
> Besides not having to deal with PATH_MAX directly, I also thought that this
> could possibly (untested) allow tracing paths bigger than PATH_MAX, which
> would be kind of interesting.

As far as I know these magic links from /proc are never longer than 4096 bytes, which is the size of a single page. People hardcode this frequently.

Well... let's not complicate it too much. It looks like a simple change, whatever way you choose will be right.

kmk3:
@smitsohu commented [on Oct 16](https://github.com/netblue30/firejail/pull/4606#issuecomment-944997634):

> > Besides not having to deal with PATH_MAX directly, I also thought that this
> > could possibly (untested) allow tracing paths bigger than PATH_MAX, which
> > would be kind of interesting.
> 
> As far as I know these magic links from /proc are never longer than 4096
> bytes, which is the size of a single page. People hardcode this frequently.
> 
> Well... let's not complicate it too much. It looks like a simple change,
> whatever way you choose will be right.

Alright, I did a force-push with the changes from the diff; unless any other
points are raised, that would be the final version.

smitsohu:
Merged.

PR #4612, opened 2021-10-16
blobwars: add path to game assets compatible with Arch

jose1711:


rusty-snake (approved):
I think there are quite a few more games with need `/usr/share/games/foobar` under some distros (e.g. arch/debian).

netblue30:
merged, thanks!

PR #4613, opened 2021-10-17
Drop noinput for games with joystick/gamepad support

jose1711:


etc/profile-m-z/xonotic.profile
```diff
@@ -32,7 +32,6 @@ caps.drop all
 netfilter
 nodvd
 nogroups
-noinput
```
SkewedZeppelin, line 35: Xonotic gamepad support only functions on Windows.

jose1711: I did test all the games I was able to compile/install on my Arch box and Xonotic does have a working gamepad support on Linux as well. Here's a screenshot from a session running on Arch:
obrázok

PR #4624, opened 2021-10-19
Fix warsow profile for Arch users

jose1711:


etc/profile-m-z/warsow.profile
```diff
@@ -40,13 +41,13 @@ noroot
 notv
 nou2f
 novideo
-protocol unix,inet,inet6
+protocol netlink,unix,inet,inet6
```
rusty-snake, line 44:
```suggestion
protocol unix,inet,inet6,netlink
```

etc/profile-m-z/warsow.profile
```diff
@@ -11,6 +11,7 @@ ignore noexec ${HOME}
 noblacklist ${HOME}/.cache/warsow-2.1
 noblacklist ${HOME}/.local/share/warsow-2.1
 
+include allow-bin-sh.inc
```
rusty-snake, line 14:
```suggestion
include allow-bin-sh.inc

```

etc/profile-m-z/warsow.profile
```diff
@@ -11,6 +11,8 @@ ignore noexec ${HOME}
...
```
kmk3, line 14:
```suggestion
# Allow /bin/sh (blacklisted by disable-shell.inc)
include allow-bin-sh.inc
```

See etc/templates/profile.template.

Issue #4603, opened 2021-10-13
Feature Request: Logind conditional


crocket:
### Is your feature request related to a problem? Please describe.

(e)udev attaches uaccess tag to devices.
systemd-logind and elogind grant logged-in users rights to devices with uaccess tag.
With logind, `noroot` and `nogroups` don't break device access.
Without logind, `noroot` or `nogroups` may break device access.

Some linux distributions like Void Linux and Gentoo Linux allow users to replace logind with other seat managers like seatd.

### Describe the solution you'd like

To support all linux distributions and all seat managers without any manual user configuration, a logind conditional like `?SEAT-ACL` is desirable. Then,

```
noroot
nogroups
```

can be replaced with

```
?SEAT-ACL: noroot
?SEAT-ACL: nogroups
```

smitsohu:
@topimiettinen Right. This kind of fine grained control (`groups.drop`) won't be very helpful for upstream profiles. I was thinking from the perspective of an administrator, but a more dynamic solution would be clearly better.

crocket:
It doesn't make much sense to put `ignore nogroups` in allow-audio.inc if allow-audio.inc is included in main profiles.

You should just remove `nogroups`.

rusty-snake:
> If we keep adding stuff without eliminating lines, we may end up with something like `ignore ignore ignore nogroups`

`ignore nogroups` is enough.

----

What about `nogroups <all|auto|none>`?

- `nogroups all` (and alias `nogroups` for backward compatibility) will drop all supplementary groups. (and `noroot` will not change groups #3303)
- `nogroups none` will keep all groups (maybe we don't need this as it is the same as not using `nogroups` and just complicates things)
- `nogroups auto` will autodetect which groups it should keep (whitelisting)
   ```js
   if !arg_nosound:
       drop_group("audio")
   if !arg_novideo:
       drop_group("video")
   if !arg_nodvd:
       drop_group("cdrom")
   if !arg_noinput:
       drop_group("input")
   if !arg_no3d: // /dev/dri/card0 has group video
       drop_group("render")
   if !arg_noprinters: // yes I still need to write a ticket for it
       drop_group("lp")
   ```
   in case other groups names are used, we can provide mappings in firejail.config

crocket:
I think I like `nogroups (all)` and `nogroups auto`.

crocket:
> The users would gain access to this GID, which they didn't have before. Then they could make setgid binaries to keep the rights after Firejail exits.

[...]

Issue #3644, opened 2020-09-24
opengl errors w/nvidia proprietary driver w/firejail 'nogroups', included in chromium-common.profile


hlein:
With recent nvidia+opengl+firejail, opengl apps getting `nogroups` in their profile fail at startup, due to the `video` group being required to access `/dev/nvidia*`. This might be similar to #841

I discovered this with Chromium 85+, on Gentoo, but a very simple reproducer is:

```
$ glxgears     # works
$ firejail --noprofile glxgears     # works
$ firejail --noprofile --nogroups glxgears
Parent pid 5629, child pid 5630
Child process initialized in 10.68 ms
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  149 (GLX)
  Minor opcode of failed request:  3 (X_GLXCreateContext)
  Value in failed request:  0x0
  Serial number of failed request:  50
  Current serial number in output stream:  51
```

With chromium and the default profiles, you will get errors like:

```
$ firejail chromium
Reading profile /etc/firejail/chromium.profile
Reading profile /etc/firejail/chromium-common.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-exec.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-common.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Parent pid 18087, child pid 18088
Child process initialized in 89.08 ms
[36:36:0923/121755.556861:ERROR:angle_platform_impl.cc(43)] initialize(795): ANGLE Display::initialize error 12289: Could not create a backing OpenGL context.
[36:36:0923/121755.556949:ERROR:gl_surface_egl.cc(741)] EGL Driver message (Critical) eglInitialize: Could not create a backing OpenGL context.
[36:36:0923/121755.556973:ERROR:gl_surface_egl.cc(1247)] eglInitialize OpenGL failed with error EGL_NOT_INITIALIZED, trying next display type
[36:36:0923/121755.642851:ERROR:angle_platform_impl.cc(43)] initialize(795): ANGLE Display::initialize error 12289: Could not create a backing OpenGL context.
[36:36:0923/121755.642895:ERROR:gl_surface_egl.cc(741)] EGL Driver message (Critical) eglInitialize: Could not create a backing OpenGL context.
[...]

Issue #841, opened 2016-10-06
nvidia driver and noroot setting


reinerh:
A Debian user reported that Steam was segfaulting when run with firejail after upgrading the nvidia driver to 367.44-2.
(Also glxgears behaves wrong with the steam profile)

The problematic line in the steam profile is the "noroot" setting, though I don't know why exactly this is causing issues.

The complete bug report is [here](https://bugs.debian.org/839868).

netblue30:
I have a fix in https://github.com/netblue30/firejail/commit/40ed53c20b7a1a5569795a858c3bfe100083c666

I don't have a Nvidia card, but I think it will fix the issue. I added video and games to the groups allowed in the user namespace. I close the bug for now and reopen it if necessary. Thanks.

kevinoid:
I'm seeing this same issue with Firejail 0.9.44.8-1 on Debian with an nVidia NVS 5400M.  It looks like 40ed53c either did not fix the issue or it has reappeared.

I compiled from the current master (22414ad) and captured the `strace` output for `glxgears` with and without `--noroot` and posted it as a gist:  https://gist.github.com/kevinoid/cb1c4ed6c8f073d41b1c4e1039e04e99#file-glxgears-strace-diff

Notably, when running with `--noroot` it tries and fails to `chmod` `/dev/nvidia-modeset`, `/dev/nvidia-modeset`, and `/dev/nvidia0` then forks a child which fails to `chmod` the device, then opens it (which succeeds, since the devices are still `root:root` `0666` which `stat` showed before `chmod`).  It also does not try to write anything to `~/.nv/GLCache` (although the train may have left the tracks before this point).  It doesn't make much sense to me.  Any ideas?

Thanks,
Kevin

chiraag-nataraj:
I have an Nvidia card and have used `noroot` with with `nvidia` driver for a while now with no issues. Is anyone still seeing this issue?

kevinoid:
I can confirm that this is still an issue for me on Debian with firejail 0.9.54-1, nvidia 390.77-1, and Linux 4.18.0 (from kernel.org, without Debian patches).  `firejail --noprofile glxgears` works fine, but `firejail --noprofile --noroot glxgears` shows a black window without the spinning gears.  I updated the gist at https://gist.github.com/kevinoid/cb1c4ed6c8f073d41b1c4e1039e04e99#file-glxgears-strace-diff with the `strace` differences between the two.

chiraag-nataraj:
[...]

PR #4632, opened 2021-10-22
Consider nosound and novideo when keeping groups & misc refactors

kmk3:
Even when `nogroups` is not used, avoid keeping the audio and video
groups when `nosound` and `novideo` are used, respectively.

Based on @rusty-snake's suggestion:
https://github.com/netblue30/firejail/issues/4603#issuecomment-944046299

Relates to #4603.

---

Note: I haven't tested this.

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -3144,17 +3144,21 @@ int main(int argc, char **argv, char **envp) {
 			}
 
 			// add audio group
-			g = get_group_id("audio");
-			if (g) {
-				sprintf(ptr, "%d %d 1\n", g, g);
-				ptr += strlen(ptr);
+			if (!arg_nosound) {
+				g = get_group_id("audio");
+				if (g) {
+					sprintf(ptr, "%d %d 1\n", g, g);
+					ptr += strlen(ptr);
+				}
 			}
 
 			// add video group
-			g = get_group_id("video");
-			if (g) {
-				sprintf(ptr, "%d %d 1\n", g, g);
-				ptr += strlen(ptr);
+			if (!arg_novideo) {
+				g = get_group_id("video");
+				if (g) {
+					sprintf(ptr, "%d %d 1\n", g, g);
+					ptr += strlen(ptr);
+				}
 			}
 
 			// add games group
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -3088,94 +3088,94 @@ int main(int argc, char **argv, char **envp) {
 	}
 	EUID_ASSERT();
 
- 	// close each end of the unused pipes
- 	close(parent_to_child_fds[0]);
- 	close(child_to_parent_fds[1]);
+	// close each end of the unused pipes
+	close(parent_to_child_fds[0]);
+	close(child_to_parent_fds[1]);
 
 	// notify child that base setup is complete
- 	notify_other(parent_to_child_fds[1]);
+	notify_other(parent_to_child_fds[1]);
 
- 	// wait for child to create new user namespace with CLONE_NEWUSER
- 	wait_for_other(child_to_parent_fds[0]);
- 	close(child_to_parent_fds[0]);
+	// wait for child to create new user namespace with CLONE_NEWUSER
+	wait_for_other(child_to_parent_fds[0]);
+	close(child_to_parent_fds[0]);
 
- 	if (arg_noroot) {
-	 	// update the UID and GID maps in the new child user namespace
+	if (arg_noroot) {
+		// update the UID and GID maps in the new child user namespace
 		// uid
-	 	char *map_path;
-	 	if (asprintf(&map_path, "/proc/%d/uid_map", child) == -1)
-	 		errExit("asprintf");
+		char *map_path;
+		if (asprintf(&map_path, "/proc/%d/uid_map", child) == -1)
+			errExit("asprintf");
 
-	 	char *map;
-	 	uid_t uid = getuid();
-	 	if (asprintf(&map, "%d %d 1", uid, uid) == -1)
-	 		errExit("asprintf");
- 		EUID_ROOT();
-	 	update_map(map, map_path);
-	 	EUID_USER();
-	 	free(map);
-	 	free(map_path);
+		char *map;
+		uid_t uid = getuid();
+		if (asprintf(&map, "%d %d 1", uid, uid) == -1)
+			errExit("asprintf");
+		EUID_ROOT();
+		update_map(map, map_path);
+		EUID_USER();
+		free(map);
+		free(map_path);
 
-	 	// gid file
+		// gid file
 		if (asprintf(&map_path, "/proc/%d/gid_map", child) == -1)
 			errExit("asprintf");
-	 	char gidmap[1024];
-	 	char *ptr = gidmap;
-	 	*ptr = '\0';
+		char gidmap[1024];
+		char *ptr = gidmap;
+		*ptr = '\0';
 
-	 	// add user group
-	 	gid_t gid = getgid();
-	 	sprintf(ptr, "%d %d 1\n", gid, gid);
-	 	ptr += strlen(ptr);
+		// add user group
+		gid_t gid = getgid();
+		sprintf(ptr, "%d %d 1\n", gid, gid);
+		ptr += strlen(ptr);
 
-	 	if (!arg_nogroups) {
-		 	//  add firejail group
-		 	gid_t g = get_group_id("firejail");
-		 	if (g) {
-		 		sprintf(ptr, "%d %d 1\n", g, g);
-		 		ptr += strlen(ptr);
-		 	}
+		if (!arg_nogroups) {
+			// add firejail group
+			gid_t g = get_group_id("firejail");
+			if (g) {
+				sprintf(ptr, "%d %d 1\n", g, g);
+				ptr += strlen(ptr);
+			}
 
-		 	//  add tty group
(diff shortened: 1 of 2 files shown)


hlein:
> > I'll add some more checks for that.
> 
> I think I got it.
> 
> @rusty-snake
> 
> But are you sure that `novideo` by itself breaks hw-accel on nvidia? Or does that only happen with `nogroups`?
> 
> After re-reading #3644 and the related code, if firejail detects nvidia, then what happens is that `nogroups` is ignored, which prevents not only the "video" group, but also the "render" group from being dropped. It could be that either both are needed or that just the "render" group is needed for hw-accel.
> 
> Cc: @hlein (from #3644)

I don't have the whole thing reloaded into my brain, but I tried the reproducer from #3644 and variants:

```
$ firejail --version
firejail version 0.9.66

$ glxgears   ### Works
Running synchronized to the vertical refresh.  The framerate should be
approximately the same as the monitor refresh rate.
...

$ firejail --noprofile glxgears   ### Works
Parent pid 14201, child pid 14202
Warning: not remounting [redacted]
Child process initialized in 7.98 ms
Running synchronized to the vertical refresh.  The framerate should be
approximately the same as the monitor refresh rate.
...

$ firejail --noprofile --nogroups glxgears   ### Fails
Parent pid 13354, child pid 13355
Warning: not remounting [redacted]
Child process initialized in 8.00 ms
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  149 (GLX)
  Minor opcode of failed request:  3 (X_GLXCreateContext)
  Value in failed request:  0x0
  Serial number of failed request:  51
  Current serial number in output stream:  52

$ firejail --noprofile --novideo glxgears   ### Works
Parent pid 13419, child pid 13420
Warning: not remounting [redacted]
Child process initialized in 7.85 ms
Running synchronized to the vertical refresh.  The framerate should be
approximately the same as the monitor refresh rate.
...
```

So: notwithstanding any auto-detection-induced behavior changes that firejail might be doing, nogroups is enough to cause failure, novideo works.

kmk3:
@hlein commented [on Nov 8](https://github.com/netblue30/firejail/pull/4632#issuecomment-963624437):

> > > I'll add some more checks for that.
> > 
> > 
> > I think I got it.
> > 
> > @rusty-snake
> > 
> > But are you sure that `novideo` by itself breaks hw-accel on nvidia? Or
> > does that only happen with `nogroups`?
> > 
> > After re-reading #3644 and the related code, if firejail detects nvidia,
> > then what happens is that `nogroups` is ignored, which prevents not only
> > the "video" group, but also the "render" group from being dropped. It could
> > be that either both are needed or that just the "render" group is needed
> > for hw-accel.
> > 
> > Cc: @hlein (from #3644)
> 
> I don't have the whole thing reloaded into my brain, but I tried the
> reproducer from #3644 and variants:
> 
> ```
> $ firejail --version
> firejail version 0.9.66
> 
> $ glxgears   ### Works
> Running synchronized to the vertical refresh.  The framerate should be
> approximately the same as the monitor refresh rate.
> ...
> 
> $ firejail --noprofile glxgears   ### Works
> Parent pid 14201, child pid 14202
> Warning: not remounting [redacted]
> Child process initialized in 7.98 ms
> Running synchronized to the vertical refresh.  The framerate should be
> approximately the same as the monitor refresh rate.
> ...
> 
> $ firejail --noprofile --nogroups glxgears   ### Fails
> Parent pid 13354, child pid 13355
> Warning: not remounting [redacted]
> Child process initialized in 8.00 ms
> X Error of failed request:  BadValue (integer parameter out of range for operation)
>   Major opcode of failed request:  149 (GLX)
>   Minor opcode of failed request:  3 (X_GLXCreateContext)
>   Value in failed request:  0x0
>   Serial number of failed request:  51
>   Current serial number in output stream:  52
> 
> 
> $ firejail --noprofile --novideo glxgears   ### Works
> Parent pid 13419, child pid 13420
> Warning: not remounting [redacted]
> Child process initialized in 7.85 ms
> Running synchronized to the vertical refresh.  The framerate should be
> approximately the same as the monitor refresh rate.
> ...
> ```
> 
> So: notwithstanding any auto-detection-induced behavior changes that firejail
> might be doing, nogroups is enough to cause failure, novideo works.

Hello, thanks for re-testing!

Considering the above, it seems that unless `novideo` misses a nvidia device
that is both required for hw-accel and that is also owned by the "video" group
(rather than the usual "render" group) for some reason, then `novideo`
shouldn't affect hw-accel (with or without these changes).

The udev rules don't seem to do anything special for nvidia GPUs either:

```console
$ pacman -Q udev
eudev 249.5-1
$ grep '"video"' /usr/lib/udev/rules.d/*
/usr/lib/udev/rules.d/50-udev-default.rules:SUBSYSTEM=="video4linux", GROUP="video"
/usr/lib/udev/rules.d/50-udev-default.rules:SUBSYSTEM=="graphics", GROUP="video"
/usr/lib/udev/rules.d/50-udev-default.rules:SUBSYSTEM=="drm", KERNEL!="renderD*", GROUP="video"
/usr/lib/udev/rules.d/50-udev-default.rules:SUBSYSTEM=="dvb", GROUP="video"
/usr/lib/udev/rules.d/50-udev-default.rules:SUBSYSTEM=="media", GROUP="video"
/usr/lib/udev/rules.d/50-udev-default.rules:SUBSYSTEM=="cec", GROUP="video"
/usr/lib/udev/rules.d/50-udev-default.rules:SUBSYSTEM=="firewire", TEST=="units", ENV{IEEE1394_UNIT_FUNCTION_VIDEO}=="1", GROUP="video"
$ grep '"render"' /usr/lib/udev/rules.d/*
/usr/lib/udev/rules.d/50-udev-default.rules:SUBSYSTEM=="drm", KERNEL=="renderD*", GROUP="render", MODE="0666"
/usr/lib/udev/rules.d/50-udev-default.rules:SUBSYSTEM=="kfd", GROUP="render", MODE="0666"
$ grep -i nvidia /usr/lib/udev/rules.d/*
/usr/lib/udev/rules.d/51-android.rules:#        Nvidia
/usr/lib/udev/rules.d/51-android.rules:ATTR{idVendor}!="0955", GOTO="not_Nvidia"
/usr/lib/udev/rules.d/51-android.rules:LABEL="not_Nvidia"
/usr/lib/udev/rules.d/70-steam-input.rules:# NVIDIA Shield Portable (2013 - NVIDIA_Controller_v01.01 - In-Home Streaming only)
/usr/lib/udev/rules.d/70-steam-input.rules:# NVIDIA Shield Controller (2015 - NVIDIA_Controller_v01.03 over USB hidraw)
/usr/lib/udev/rules.d/70-steam-input.rules:# NVIDIA Shield Controller (2017 - NVIDIA_Controller_v01.04 over bluetooth hidraw)
```

@rusty-snake What do you think?

rusty-snake:
I still don't understand what all this has to do with nVidia.

kmk3:
@rusty-snake commented [on Nov 9](https://github.com/netblue30/firejail/pull/4632#issuecomment-964571034):

> I still don't understand what all this has to do with nVidia.

It's because of [this code on src/firejail/profile.c](https://github.com/netblue30/firejail/blob/d9403dcdc8cfb23d5350b0488817a5eaa5a0f20d/src/firejail/profile.c#L418-L427):

```C
	else if (strcmp(ptr, "nogroups") == 0) {
		// nvidia cards require video group; disable nogroups
		if (access("/dev/nvidiactl", R_OK) == 0 && arg_no3d == 0) {
			fwarning("Warning: NVIDIA card detected, nogroups command disabled\n");
			arg_nogroups = 0;
		}
		else
			arg_nogroups = 1;
		return 0;
	}
```

It's a workaround for nvidia that mentions "video group" and checks for
`arg_no3d`.

(As the one making the changes, in my mind it seemed obvious, but I should
probably have referenced it sooner)

See the following commits:

* 623e68216 ("`temporary fix for nvidia/nogroups/noroot issue (#3644, #841)`",
  2020-10-02)
* cb460c32c ("`more nvidia (#3644)`", 2020-10-03)

You even suggested the latter one apparently:

* <https://github.com/netblue30/firejail/issues/3644#issuecomment-703067860>

Related issues:

* #841
* #3644

On the latter issue, it appears that either the "video" or "render" group (or
both) were needed to enable hw-accel of OpenGL on Chromium on Gentoo.

After reading some more of #3644, it looks like this PR is probably the
definitive fix for it coincidentally, so we could probably drop the workaround
on profile.c after this PR (provided that it wouldn't cause issues considering
the "drop all groups" instances in other parts of the code), but I'd rather do
that in a new PR.

kmk3:
@kmk3 commented [on Nov 10](https://github.com/netblue30/firejail/pull/4632#issuecomment-964926247):

> After reading some more of #3644, it looks like this PR is probably the
> definitive fix for it coincidentally, so we could probably drop the
> workaround on profile.c after this PR (provided that it wouldn't cause issues
> considering the "drop all groups" instances in other parts of the code), but
> I'd rather do that in a new PR.

Sorry, I was getting ahead of myself.  I mean drop the workaround once we split
the handling of these groups (i.e.: the ones mapped to `no` options) from the
handling of `nogroups` (and maybe also from the handling of `noroot`) on
src/firejail/main.c (WIP).  Then these groups would always be kept, even when
`nogroups` is enabled.

I keep forgetting that this PR only does organizing and some hardening and that
the WIP changes are the ones intended to increase compatibility.

smitsohu:
Merged!

PR #4634, opened 2021-10-24
[minor] update mpv.profile

ghost:
add yt-dlp in private-bin


kmk3 (approved):
> add yt-dlp in private-bin

This is funny, I had tried to use yt-dlp with mpv before and gave up debugging
after a while.  I had searched on many different places but did not check this.
Now it works.

Issue #4440, opened 2021-08-04
[Meta] Handling of background processes / services / agents / deamons


rusty-snake:
#928 Services preventing firejail from stopping
#3949 firejail not terminated after browser is killed
#2433 Use Firejail as shell - shutdown issue
#3042 Add an option to kill all the processes in containers when the initial process finishes
#3491 Firejail breaks daemonizing processes
#3671 GDB a process inside FireJail

PR #4635, opened 2021-10-24
deterministic-shutdown option

### LINKED ISSUE #928: Services preventing firejail from stopping
when agent services like gpg-agent or ssh-agent are started by firejailed applications, firejail will not close because those services will keep running

When gpg-agent is automatically firejailed by firecfg, it will break many things like apt-get update when it is used by install scripts

Is there a way to prevent this? Mabe tell services to close when the application that started them is stopped

### LINKED ISSUE #3042: Add an option to kill all the processes in containers when the initial process finishes
Processes using gpgme start gpg daemons and that are not killed automatically.

smitsohu:
Sometimes sandboxed apps are outlived by their child processes which prevents the sandbox from shutting down.

Add a new option to deal with this situation.

Probably there are better names for the option, I just needed a placeholder. `kill-remains` was suggested by @msva. 

Tracked in #4440
Closes #928 
Closes #3042


netblue30:
all in!

KOLANICH:
Thanks for implementing it.

Issue #408, opened 2016-04-03
firecfg.config location


reinerh:
In the header of firecfg.config you mention a location of /etc/firejail/firecfg.config.
But by default it is installed in $LIBDIR/firejail/firecfg.config.
Which is supposed to be the correct location (i.e., is is intended to be edited by the user?)?

netblue30:
Sorry, it is in $LIBDIR/firejail/firecfg.config

netblue30:
Fixed, thanks!

Vincent43:
@netblue30 why it should be non-editable? I think that enabling/disabling specific apps from firecfg may solve https://github.com/netblue30/firejail/issues/2097 cc @chiraag-nataraj

chiraag-nataraj:
Yes, it's odd that a configuration file is being stored in `$LIBDIR`...why was that chosen in the first place? Why not store it in `${CFG}` like `firejail.config`?

Vincent43:
@netblue30 opinions?

Issue #3665, opened 2020-10-13
With firecfg, how do I configure specific applications to go through firejail?


Joe23232:
I know that with the command `sudo firecfg` will get all the applications to open through firejail, however I don't want this, I want to only specify specific applications such as chromium etc.

How would I achieve this if this is even possible?

rusty-snake:
1. https://github.com/netblue30/firejail/issues/3213 extended with a script (with is started as root). You must verify the .desktop files yourself.
```
#!/bin/bash
apps=(firefox thunderbird libreoffice)
for app in "${apps[@]}"; do
	ln -s /usr/bin/firejail "/usr/local/bin/$app"
done
```

2. Edit `/usr/lib/firejail/firecfg.config` (path differs from distro and configure; use `locate firecfg.config`). You should first create a backup of it.

3. Use [firecfg.py](https://github.com/rusty-snake/firecfg.py) with custom groups (I work on more group control via command-line and config-file).

Joe23232:
Hi thanks for your response,

> 2.  Edit /usr/lib/firejail/firecfg.config (path differs from distro and configure; use locate firecfg.config). You should first create a backup of it.

Sorry but in regards to this step, what should I specifically edit in this file?

rusty-snake:
This file specifies for which programs firecfg creates symlinks (if they installed and have a profile). If a program is not listed there (or commented) no symlinks are created. 

Example (create a backup; only set firefox, thunderbird and libreoffice)
```
$ mv /usr/lib/firejail/firecfg.config /usr/lib/firejail/firecfg.config.bak
$ cat > /usr/lib/firejail/firecfg.config <<EOF
thunderbird
firefox
libreoffice
EOF
```

Joe23232:
Hey man `cat > /usr/lib/firejail/firecfg.config <<EOF` sorry what does `cat >` and  `<<EOF` mean?

rusty-snake:
That's are bash redirections.
https://www.cyberciti.biz/faq/using-heredoc-rediection-in-bash-shell-script-to-write-to-file/
https://wiki.bash-hackers.org/syntax/redirection
https://www.gnu.org/software/bash/manual/html_node/Redirections.html

---

Maybe it is easier to understand for you if it uses `echo`. However, you do not need to understand them. It is enough to know what they do (replace /usr/lib/firejail/firecfg.config with thunderbird, firefox, libreoffice (see below)).

[...]

Issue #2829, opened 2019-07-04
Local option for firecfg.config


Edu4rdSHL:
Actually firejail handle what profiles create using the firecfg utility reading the /usr/lib/firejail/firecfg.config file. For some reasons users will not want to sandbox some program because of functionality or another reason.

Actually you can just comment out the line of the program name that you don't want to generate and ran firefcg again. The problem is that the /usr/lib/firejail/firecfg.config overwrite in every new firejail installation. We can backup that file in the distro package manager but isn't a good idea since that additional profiles are added and more.

Will be good if we can have a /usr/lib/firejail/firecfg.local file that works in the same way that $profile.local and then you can make something like:

```
# Local file configuration for firecfg.config
[exclude] # It will exclude the program when generating the symlinks using firecfg
program1
program2
programN
[/exclude]

[include] # It will include the program when generating the symlinks using firecfg
program1
program2
programN
[/include]
```
Or well:
```
# Local file configuration for firecfg.config

# It will exclude the program when generating the symlinks using firecfg
exclude program1
exclude program2
exclude programN

# It will include the program when generating the symlinks using firecfg
include program1
include program2
include programN
```

Or is already a functionality in firejail for doing that?

FOSSONLY:
Why not:

chattr -i /usr/lib/firejail/firejail.config ? ;)

On the other side, your suggestion would be more comfortable. But don't forget: Yes sandboxing has sometimes a few downsides, but sandbox nearly every program by default is the best possible way of protection. Personally I would write custom rules to some profiles, to circumvent problems.

Nick

Vincent43:
This is duplicate of https://github.com/netblue30/firejail/issues/2097 , please continue discussion there.

I believe the best solution is to treat `firecfg.config` as user editable file similarly to `firejail.config`: https://github.com/netblue30/firejail/issues/408

Edu4rdSHL:
I doesn't see why it's a duplicate, are two different things... I doesn't want it "to be editable by the user" and it isn't the best solution anyways because we are missing the new added profiles. I want a .local option to preserve local changes/configuration maintaining the benefit of the new added profiles.

rusty-snake:
@Edu4rdSHL #408 is about storing it in `/etc/firejail` and not under $HOME. @Vincent43 say it's a duplcate of #2097. #2097 is about ignoring some entrys in firecfg.config and adding some other (firecfg.allow, firecfg.deny, ...)

[...]

Issue #2097, opened 2018-08-27
firecfg: allow for ignoring specific apps


Vincent43:
Currently firecfg symlinks unconditionally creates for all apps it detects. This is problematic if user doesn't want to use specific app with firejail.

I propose allowing users to prevent creating symlinks for chosen apps: `firecfg --ignore <app_name>`. The database of ignored apps could be saved same way as database of allowed users is currently.

This was discussed before: https://github.com/netblue30/firejail/issues/1675

chiraag-nataraj:
How about, instead, have an `-a/--all` flag which would replicate the behaviour seen here and otherwise, only link the apps listed on the command-line?

chiraag-nataraj:
This way, you could also use `--clean` to clean up only specific links or `--clean --all` to clean all links.

Vincent43:
Ability to symlink only chosen apps would be useful too but it's something different. Usecase which I have in mind is ignoring one,two or three apps among dozens which is rather hard to type in cmdline in whitelist mode so I prefer blacklisting.

chiraag-nataraj:
So maybe have files like `firecfg.allow` and `firecfg.deny`? That way, if you only want to allow specific programs, edit `firecfg.allow`. If you only want to deny specific programs, edit `firecfg.deny`. If neither of those exists, we get the existing behavior.

Vincent43:
That sounds good!

hlein:
I propose moving `firecfg.config` under `/etc/` - `/etc/firejail/firecfg.config`, or whatever.

More advanced features like a `.local`, `.deny`, `.allow`, or whatever, are all well and good to keep discussing, but in the meantime I would argue simply: these amount to configuration files, they should go in a path that a) system administrators expect to find configs in and b) distribution packaging tools expect to store configs in, and therefore won't silently overwrite local customizations during version upgrades.

There's multiple issues regarding this - #408  which refers to #2829 which was closed with a comment pointing here.

If I make a PR that relocates `firecfg.config` from `LIBDIR/firejail/` to `SYSCONFDIR/firejail/`, is it likely to be accepted?

Thanks!

rusty-snake:
I am strongly in favour of using /etc instead of /usr/lib64 for this and I don't think anyone is against it.

hlein:
> I am strongly in favour of using /etc instead of /usr/lib64 for this and I don't think anyone is against it.

Sweet, thanks! I will work on this as soon as I get a chance.

[...]

PR #4669, opened 2021-11-06
Relocate firecfg.config to /etc/firejail/

hlein:
This should make it easier for users, and distributions, to customize
which programs they want firejail to wrap.

Closes: https://github.com/netblue30/firejail/issues/408
Bug: https://github.com/netblue30/firejail/issues/2097
Bug: https://github.com/netblue30/firejail/issues/2829
Bug: https://github.com/netblue30/firejail/issues/3665

### DIFF
--- a/src/firecfg/main.c
+++ b/src/firecfg/main.c
@@ -171,17 +171,17 @@ static void set_file(const char *name, const char *firejail_exec) {
 	free(fname);
 }
 
-// parse /usr/lib/firejail/firecfg.cfg file
+// parse /etc/firejail/firecfg.config file
 static void set_links_firecfg(void) {
 	char *cfgfile;
-	if (asprintf(&cfgfile, "%s/firejail/firecfg.config", LIBDIR) == -1)
+	if (asprintf(&cfgfile, "%s/firecfg.config", SYSCONFDIR) == -1)
 		errExit("asprintf");
 
 	char *firejail_exec;
 	if (asprintf(&firejail_exec, "%s/bin/firejail", PREFIX) == -1)
 		errExit("asprintf");
 
-	// parse /usr/lib/firejail/firecfg.cfg file
+	// parse /etc/firejail/firecfg.config file
 	FILE *fp = fopen(cfgfile, "r");
 	if (!fp) {
 		perror("fopen");
@@ -440,7 +440,7 @@ int main(int argc, char **argv) {
 	// clear all symlinks
 	clean();
 
-	// set new symlinks based on /usr/lib/firejail/firecfg.cfg
+	// set new symlinks based on /etc/firejail/firecfg.config
 	set_links_firecfg();
 
 	if (getuid() == 0) {
--- a/Makefile.in
+++ b/Makefile.in
@@ -116,7 +116,7 @@ endif
 	install -m 0755 src/jailcheck/jailcheck $(DESTDIR)$(bindir)
 	# libraries and plugins
 	install -m 0755 -d $(DESTDIR)$(libdir)/firejail
-	install -m 0644 -t $(DESTDIR)$(libdir)/firejail $(MYLIBS) $(SECCOMP_FILTERS) src/firecfg/firecfg.config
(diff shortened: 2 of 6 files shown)


hlein:
N.B. my testing and initial patch was actually against release 0.9.66, but I think this does the right things.

src/firecfg/main.c
```diff
@@ -171,10 +171,10 @@ static void set_file(const char *name, const char *firejail_exec) {
 	free(fname);
 }
 
-// parse /usr/lib/firejail/firecfg.cfg file
+// parse /etc/firejail/firecfg.cfg file
```
kmk3, line 174:
```suggestion
// parse /etc/firejail/firecfg.config file
```

The filename is currently misspelled.

kmk3 (commented):
There are a few more places to change:

```diff
diff --git a/src/firecfg/main.c b/src/firecfg/main.c
index b8fdbcda0..1a05efd29 100644
--- a/src/firecfg/main.c
+++ b/src/firecfg/main.c
@@ -181,7 +181,7 @@ static void set_links_firecfg(void) {
        if (asprintf(&firejail_exec, "%s/bin/firejail", PREFIX) == -1)
                errExit("asprintf");

-       // parse /usr/lib/firejail/firecfg.cfg file
+       // parse /etc/firejail/firecfg.config file
        FILE *fp = fopen(cfgfile, "r");
        if (!fp) {
                perror("fopen");
@@ -440,7 +440,7 @@ int main(int argc, char **argv) {
        // clear all symlinks
        clean();

-       // set new symlinks based on /usr/lib/firejail/firecfg.cfg
+       // set new symlinks based on /etc/firejail/firecfg.config
        set_links_firecfg();

        if (getuid() == 0) {
```

Misc: The filename is misspelled in these comments as well.

src/firecfg/main.c
hlein: Aha, good catch, I will clean up these misspellings of `firecfg.config` and update their paths too.

hlein: > Aha, good catch, I will clean up these misspellings of `firecfg.config` and update their paths too.

Done, I think.

topimiettinen:
This reminds me that Firejails should support this kind of administrative hierarchy:
- `/usr/lib/firejail/profiles` or `/usr/share/firejail/profiles` is for distros, based on our set of profiles
- `/etc/firejail/profiles` can be used by local system admin to override distro defaults, initially completely empty
- `$XDG_CONFIG_HOME/firejail/profiles` (or `$HOME/.config/firejail/profiles`) can be used by users to override system admin when allowed, initially doesn't exist

Finally, users can override their own config files and other configs with command line switches and environment variables, again when allowed.

In theory, there could be another element for dynamic config generation in `/run/firejail/profiles`, for example in case the version of the kernel or loading of kernel modules could influence the configs somehow (optimize allowed system calls in seccomp lists? only allow some apps when a USB key is plugged in, kill the app when unplugged?).

src/firecfg/main.c
kmk3: @hlein [on Nov 6](https://github.com/netblue30/firejail/pull/4669#discussion_r744079948):

> Aha, good catch, I will clean up these misspellings of `firecfg.config` and
> update their paths too.

Thanks.

From the new commit message:

> Also fixed some firecfg.cfg -> firecfg.config references.

One minor non-blocking nitpick: I'd `s/fixed/fix/` to make it more similar to
this:

kmk3 (approved):
LGTM.

netblue30:
Thanks!

kmk3:
Note: This could be considered a breaking change and is probably relevant to
packagers either way, so I added it to the RELNOTES on commit c374c0c5e
("RELNOTES: mention move of firecfg.config to /etc/firejail/").

By the way, on the note I included the PR number to make it easier to find.
Let me know if this is undesirable for some reason.  If not, maybe this could
be done by default (where applicable) on the notes of future releases, but
that's probably worth its own discussion.

PR #4675, opened 2021-11-10
more ssh fixes

glitsj16:
See https://github.com/netblue30/firejail/commit/9a81078ddbbb4215d06f7d1861481ece05ebda99#commitcomment-59921139 for details. Basically this PR (no)blacklists /usr/lib/{open,}ssh instead of referencing all possible individual files that might get installed there depending on OS and versioning.

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -497,6 +497,7 @@ blacklist ${PATH}/xinput
 blacklist ${PATH}/ssh
 blacklist /usr/lib/openssh
 blacklist /usr/lib/ssh
+blacklist /usr/libexec/openssh/ssh-keysign
 blacklist ${PATH}/passwd
 blacklist /usr/lib/xorg/Xorg.wrap
 blacklist /usr/lib/policykit-1/polkit-agent-helper-1
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -495,7 +495,8 @@ blacklist ${PATH}/xev
 blacklist ${PATH}/xinput
 # from 0.9.67
 blacklist ${PATH}/ssh
-blacklist /usr/lib/openssh/ssh-keysign
+blacklist /usr/lib/openssh
+blacklist /usr/lib/ssh
 blacklist ${PATH}/passwd
 blacklist /usr/lib/xorg/Xorg.wrap
 blacklist /usr/lib/policykit-1/polkit-agent-helper-1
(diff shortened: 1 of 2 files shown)


rusty-snake (commented):
Fedora: `/usr/libexec/openssh`

glitsj16:
@rusty-snake I've added a fix for Fedora in allow-ssh.inc. The disable-common.inc part might be a bit more complex, due to several other profiles touching /usr/libexec (and being unfamiliar with Fedora). Feel free to make suggestions!

etc/inc/disable-common.inc
```diff
@@ -495,7 +495,9 @@ blacklist ${PATH}/xev
 blacklist ${PATH}/xinput
 # from 0.9.67
 blacklist ${PATH}/ssh
-blacklist /usr/lib/openssh/ssh-keysign
```
reinerh, line 498: `/usr/lib/openssh/ssh-keysign` is the path in Debian.
Please keep it. :-)

reinerh: ah, the directory is now blacklisted.
but then why not blacklist the whole `/usr/libexec/openssh` as well?

glitsj16:
> but then why not blacklist the whole /usr/libexec/openssh as well?

@reinerh Thanks for your review. I've brought in your suggestion. Not too clearheaded at the moment, great to see your in much better shape :-)

reinerh (approved):

rusty-snake (approved):

kmk3 (commented):
The code changes LGTM.

Could you fixup/squash the commits?

Example: With `git rebase -i "$(git merge-base master HEAD)"`.

Also, the first commit has a (non-auto-generated) line that's over 500 chars
long, which makes it hard to read (and makes it stand out) on git log.
Suggestion: Replace the message with a formatted version:

```
disable-common.inc: blacklist more ssh lib paths

After seeing 9a81078 it dawned on me that Arch Linux doesn't have
/usr/lib/openssh, but uses /usr/lib/ssh instead.  That's a different
path than what's referenced in our current
{allow-ssh,disable-common}.inc files.  Some very superficial checks
revealed that OpenSSH seems to be packaged quite differently, at least
on Debian/Ubuntu and Arch Linux.  And then there's version differences
on non-rolling distros to consider.  All in all IMO it makes more sense
to (no)blacklist /usr/lib/openssh and /usr/lib/ssh instead of
referencing all the possible individual files that live under those
paths.

Also, add Fedora paths as per:
https://github.com/netblue30/firejail/pull/4675#pullrequestreview-802438767
```

(The above would be the message of the final commit, after all the other commits
are fixedup/squashed)

netblue30:
Fixed, thanks!

Issue #4677, opened 2021-11-11
Add openstego profile


Rult:
UPD: Java-specific error was resolved.
All that's need now for OpenStego's profile is to deny Internet access.
_____
[OpenStego in AUR.](https://aur.archlinux.org/packages/openstego)
Currently gives this when run with `--net=none`:
```
java.lang.NullPointerException: Cannot invoke "java.io.InputStream.read(byte[])" because "tmplIS" is null
        at com.openstego.desktop.util.UserPreferences.init(UserPreferences.java:57)
        at com.openstego.desktop.OpenStego.main(OpenStego.java:429)
```
I suppose there are should be some Java paths whitelisted, but I don't know which ones are needed to be specified.

jmetrius:
Hey,

I recently updated the OpenStego PKGBUILD since it has been terribly outdated. I tried to reproduce your issue by running `firejail --net=none openstego` and all seems fine.

Can you try again with the latest version and report if the issue persists?

Rult:
@jmetrius hey, that's pretty neat! The program works as intended.

Though it would be nice if firejail would disable Internet access to it by default as it's not needed and the package comes precompiled on AUR.

PR #4681, opened 2021-11-12
Add OpenStego profile

jmetrius:
This PR adds a profile for OpenStego as requested in #4677 

Advice regarding the proper whitelisting of XDG user directories would be welcome since I am not sure how restrictive this is supposed to be. A user probably would like to files access in all user dirs for convenience. Is there a standard regarding that?

I couldn't write a sensible `private-etc` since this seems very dependent on the specific Java environment used and is probably hard to maintain.

Have a nice day!

### DIFF
--- a/etc/profile-m-z/openstego.profile
+++ b/etc/profile-m-z/openstego.profile
@@ -15,8 +15,8 @@ include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
+include disable-proc.inc
 include disable-programs.inc
-include disable-passwdmgr.inc
 
 mkfile ${HOME}/openstego.ini
 whitelist ${HOME}/openstego.ini
@@ -24,20 +24,14 @@ whitelist ${HOME}/.java
 whitelist ${PICTURES}
 whitelist ${DOCUMENTS}
 whitelist ${DESKTOP}
-include whitelist-common.inc
-
 whitelist /usr/share/java
+include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
-# AppArmor breaks Java interpreter
-ignore apparmor
-
 caps.drop all
-
-# Makes fonts look grainy
-#ipc-namespace
-
 machine-id
 net none
 no3d
@@ -50,6 +44,7 @@ notv
 nou2f
 novideo
 seccomp
+seccomp.block-secondary
 shell none
 tracelog
 
--- a/etc/profile-m-z/openstego.profile
+++ b/etc/profile-m-z/openstego.profile
@@ -1,0 +1,63 @@
+# Firejail profile for OpenStego
+# Description: Steganography application that provides data hiding and watermarking functionality
+# This file is overwritten after every install/update
+# Persistent local customizations
+include openstego.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/openstego.ini
+
+# Allow java (blacklisted by disable-devel.inc)
+include allow-java.inc
+
+include disable-common.inc
+include disable-devel.inc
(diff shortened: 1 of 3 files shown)


etc/profile-m-z/openstego.profile
```diff
@@ -0,0 +1,63 @@
...
+include whitelist-var-common.inc
+
+# AppArmor breaks Java interpreter
+ignore apparmor
+
+caps.drop all
+
+# Makes fonts look grainy
+#ipc-namespace
+
```
rusty-snake, line 40:
```suggestion
```

etc/profile-m-z/openstego.profile
```diff
@@ -0,0 +1,63 @@
...
+no3d
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+seccomp
```
rusty-snake, line 52: `seccomp.block-secondary`?

etc/profile-m-z/openstego.profile
```diff
@@ -0,0 +1,63 @@
...
+whitelist ${DESKTOP}
+include whitelist-common.inc
+
+whitelist /usr/share/java
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+# AppArmor breaks Java interpreter
+ignore apparmor
+
```
rusty-snake, line 35:
```suggestion
```

Never set.

etc/profile-m-z/openstego.profile
```diff
@@ -0,0 +1,63 @@
...
+# Allow java (blacklisted by disable-devel.inc)
+include allow-java.inc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-programs.inc
+include disable-passwdmgr.inc
```
rusty-snake, line 19:
```suggestion
```

etc/profile-m-z/openstego.profile
```diff
@@ -0,0 +1,63 @@
...
+whitelist ${HOME}/openstego.ini
+whitelist ${HOME}/.java
+whitelist ${PICTURES}
+whitelist ${DOCUMENTS}
+whitelist ${DESKTOP}
+include whitelist-common.inc
+
+whitelist /usr/share/java
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
```
rusty-snake, line 31:
```suggestion
whitelist /usr/share/java
include whitelist-common.inc
include whitelist-usr-share-common.inc
include whitelist-var-common.inc
```

rusty-snake, line 31: Can we add wruc, wru?

rusty-snake (commented):
Looks good.

etc/profile-m-z/openstego.profile
jmetrius: 👍

etc/profile-m-z/openstego.profile
jmetrius: 👍

etc/profile-m-z/openstego.profile
jmetrius: 👍

etc/profile-m-z/openstego.profile
jmetrius: Yes, works fine with that addition 👍

rusty-snake (approved):

etc/profile-m-z/openstego.profile
jmetrius: Ah, didn't catch that. I included disable-proc.inc instead

etc/profile-m-z/openstego.profile
jmetrius: Yes, works fine with that

netblue30:
merged, thanks

PR #4688, opened 2021-11-14
Update firejail-local for Brave + ipfs

Bundy01:
Hi,

The Ipfs network of Brave is broken by firejail if in the profile of chromium-common 'apparmor' is activated.

Regards.

### DIFF
--- a/etc/apparmor/firejail-local
+++ b/etc/apparmor/firejail-local
@@ -8,6 +8,9 @@
 #owner @HOME/bin/** ix
 #owner @HOME/.local/bin/** ix
 
+# Uncomment to opt-in to apparmor for brave + ipfs
+#owner @{HOME}/.config/BraveSoftware/Brave-Browser/oecghfpdmkjlhnfpmmjegjacfimiafjp/*/** ix,
+
 # Uncomment to opt-in to apparmor for brave + tor
 #owner @{HOME}/.config/BraveSoftware/Brave-Browser/biahpgbdmdkfgndcmfiipgcebobojjkp/*/** ix,
 


etc/apparmor/firejail-local
```diff
@@ -8,6 +8,9 @@
 #owner @HOME/bin/** ix
 #owner @HOME/.local/bin/** ix
 
+# Uncomment to opt-in to apparmor for brave + ipfs
+#owner @{HOME}/.config/BraveSoftware/Brave-Browser/oecghfpdmkjlhnfpmmjegjacfimiafjp/*/** ix,
```
reinerh, line 12: these paths are probably randomly generated and unique to your current profile, or?
can you please replace them with a different more generic placeholder?

Bundy01: For this PR, I was inspired by the line below (brave + tor). I'm not sure if Brave uses a random path because the Tor extension directory works for me too. It would be nice if some users of this browser could confirm this.
If what I think is wrong, I can delete both lines (ipfs and tor) and use the Brave directory.
example: 
`#owner @{HOME}/.config/BraveSoftware/Brave-Browser/*/** ix,`

If you prefer.

rusty-snake: At least the Tor extension has the same ID for me (I couldn't make brave installing the ipfs extension). Which makes sense given that extensions IDs are unique to the extensions in chrome\* (unlike firefox where they are unique to the profile).

reinerh: ok, TIL. it looked random to me. :-)
if they are unique to the extensions then feel free to keep it.

netblue30:
merged, thanks!

PR #4690, opened 2021-11-15
Fix inconsistent descriptions of machine-id option

kmk3:
Some places say that it "preserves" the file and other places say that
it "spoofs" the file.  Based on the fs_machineid function on
src/firejail/fs_etc.c, the latter one is correct.

This amends commit d0cc960c9 ("spoof machine-id", 2016-12-05).

Fixes #4689.

Reported-by: @svc88


rusty-snake (approved):

netblue30:
merged!

Issue #4642, opened 2021-10-26
Switch from autotools to meson


rusty-snake:
This came up in https://github.com/netblue30/firejail/issues/4633#issuecomment-950317130. I like the idea. Meson is faster and has an simpler and less error-prone syntax. Is this something we want to do?

CMake would be a second alternative.

kmk3:
> Switch from autotools to meson

@rusty-snake commented [on Oct 26](https://github.com/netblue30/firejail/issues/4642#issue-1036210905):

> This came up in [#4633
> (comment)](https://github.com/netblue30/firejail/issues/4633#issuecomment-950317130).
> I like the idea. Meson is faster

> and has an simpler and less error-prone syntax.

I find the core of the syntax of autotools and make to be rather simple.  In
the case of make there are some terse operator-looking internal macros (e.g.:
`$<`) that are easy to forget, but usually only a few of them are used and IMO
they aren't overused in this repo anyway.

I'll post a `make` guide later in a new thread to show what I mean.

> Is this something we want to do?
> 
> CMake would be a second alternative.

I think it would be a great loss to break building with the traditional/more
portable build tools (I'll expand on this later).

The benefits of supporting alternative build systems isn't very clear either.
As for the speed, what are the workflows in question and the benchmark results?
Was `make` tested with something like `-j$(nproc)`?

rusty-snake:
> I find the core of the syntax of autotools and make to be rather simple.

The syntax of `Makefile`s is simple but
1. it is not always intuitive to read, example: `$(subst test-,,$@)` vs. `'foobar'.replace("test-", "")
2. can be very detailed
3. maintaining (dist)clean and uninstall targets is often forgotten
4. configure.ac is ugly

> I'll post a make guide later in a new thread to show what I mean.

:+1: 

Suggestion: https://github.com/netblue30/firejail/discussions/categories/show-and-tell

> I think it would be a great loss to break building with the traditional/more
portable build tools (I'll expand on this later).

Two things to add (before you expand this)
1. We can keep autotools (i.e. you can choose which one you use)
2. IMHO portability isn't a problem. I wouldn't say meson has high requirements and we do not need to support legacy systems on which firejail doesn't work.

> As for the speed, what are the workflows in question and the benchmark results?
Was make tested with something like -j$(nproc)?

https://mesonbuild.com/Simple-comparison.html

rusty-snake:
> The benefits of supporting alternative build systems isn't very clear either.

[...]

PR #4695, opened 2021-11-16
build: Stop linking pthread

kmk3:
Added on commit 137985136 ("Baseline firejail 0.9.28", 2015-08-08).  See
also commit ad6bb83fa ("consolidate makefiles", 2018-03-31).

It is not used anywhere.  And it looks like it has never been used
anywhere:

    $ git log --oneline -Gpthread.h 137985136..master
    $

Issue mentioned by @rusty-snake:
https://github.com/netblue30/firejail/issues/4642#issuecomment-955795463


kmk3:
By the way, I have some more autoconf-related improvements/refactorings, but
one commit changes much more stuff, so I'll post them after this PR.

reinerh (approved):

topimiettinen (approved):
Looks good, tests pass and it can always be reverted if there are obscure problems.

Issue #4607, opened 2021-10-15
`noprinters`


rusty-snake:
### Describe the solution you'd like

`noprinters` command to disable printing. It should `blacklist /dev/lp*` stuff and `/run/cups/cups.sock`.

This idea is based on the observation that wrc has
- `/run/**/resolv.conf` which is either needed or useless because of `net none`
- `/run/dbus/system_bus_socket` which is further handled by `dbus-system`
- `/run/media` which is further handled by `disable-mnt`/`disable-write-mnt.inc`
- `/run/cups/cups.sock` which is not handled elsewhere.

rusty-snake:
FTR: bd15e763eab8c07072f8ef495a17d387c26150c7

WhyNotHugo:
How about disabling printers by default and having a `printers` command to enable them when desired?

It's a bit easier to think about permissions in terms of "profile doesn't allow anything, unless otherwise stated", rather than the other way around.

rusty-snake:
While I generally agree with you to list everything that is required rather than listing everything which is not required (enumerating badness vs. goodness), I don't feel good about changing the logic how firejail works.

PR #4724, opened 2021-11-29
Add a profile for Flatseal

WhyNotHugo:
Flatseal is a GUI tool to configure permissions for Flatpak applications.

This restricts permissions as much as possible without affecting functionality.

### DIFF
--- a/etc/profile-a-l/com.github.tchx84.Flatseal.profile
+++ b/etc/profile-a-l/com.github.tchx84.Flatseal.profile
@@ -1,0 +1,65 @@
+# Firejail profile for flatseal
+# This file is overwritten after every install/update
+# Persistent local customizations
+include com.github.tchx84.Flatseal.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.local/share/flatpak/overrides
+noblacklist /var/lib/flatpak/app
+
+# Allow gjs (blacklisted by disable-interpreters.inc)
+include allow-gjs.inc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.local/share/flatpak/overrides
+whitelist ${HOME}/.local/share/flatpak/overrides
+include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+net none
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noprinters
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+
+disable-mnt
+private-bin com.github.tchx84.Flatseal,gjs
+private-cache
+private-dev
+private-etc alternatives,dconf,fonts,gtk-3.0,ld.so.cache,ld.so.preload
+private-tmp
+
+dbus-user filter
+dbus-user.own com.github.tchx84.Flatseal
+dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.freedesktop.impl.portal.PermissionStore
+dbus-user.talk org.gnome.Software
+dbus-system none
+
+read-write ${HOME}/.local/share/flatpak/overrides
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -157,6 +157,7 @@ com.github.bleakgrey.tootle
 com.github.dahenson.agenda
 com.github.johnfactotum.Foliate
 com.github.phase1geo.minder
+com.github.tchx84.Flatseal
 com.gitlab.newsflash
 conkeror
 conky


etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+# This file is overwritten after every install/update
+# Persistent local customizations
+include com.github.tchx84XXgg12uC.Flatseal.local
+# Persistent global definitions
+include globals.local
+
+# Required for GTK theem to work:
+read-only ${HOME}/.config/dconf/
+noblacklist ${HOME}/.config/dconf/
+whitelist ${HOME}/.config/dconf/
```
WhyNotHugo, line 11: The current GTK theme didn't pick up without this, but I expected it to after granting access to dconf's D-Bus interface.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+
+# TODO: Should be relative to xdg-data-home
+mkdir ${HOME}/.local/share/flatpak/overrides
+noblacklist ${HOME}/.local/share/flatpak/overrides
+whitelist ${HOME}/.local/share/flatpak/overrides
```
WhyNotHugo, line 16: These should be relative to `XDG_DATA_HOME`, but I don't think it's usable in this context.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+private-etc dconf,fonts,gtk-3.0
+private-tmp
+
+dbus-user filter
+# FIXME: Shouln't this be enough for theme to work?
+dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.freedesktop.impl.portal.PermissionStore
+dbus-user.talk org.gnome.Software
+dbus-user.talk org.a11y.Bus
+dbus-user.own com.github.tchx84.Flatseal
```
rusty-snake, line 54:
```suggestion
dbus-user.own com.github.tchx84.Flatseal
dbus-user.talk ca.desrt.dconf
dbus-user.talk org.freedesktop.impl.portal.PermissionStore
dbus-user.talk org.gnome.Software
```

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+
```
rusty-snake, line 55:
```suggestion
```

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+dbus-system none
+ignore dbus-user none
```
rusty-snake, line 57:
```suggestion
```

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+nodvd
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+shell none
```
rusty-snake, line 39: What about `tracelog`?

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
```
rusty-snake, line 38: What about `protocol`? `seccomp`? `seccomp.block-secondary`?

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+whitelist ${HOME}/.local/share/flatpak/overrides
+
+include disable-devel.inc
+include disable-exec.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
+include disable-xdg.inc
+
```
rusty-snake, line 25: What about wruc? wrc? wusc? wvc?

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+mkdir ${HOME}/.local/share/flatpak/overrides
+noblacklist ${HOME}/.local/share/flatpak/overrides
+whitelist ${HOME}/.local/share/flatpak/overrides
+
+include disable-devel.inc
+include disable-exec.inc
+include disable-passwdmgr.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
```
rusty-snake, line 23:
```suggestion
```

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+whitelist ${HOME}/.config/dconf/
+
+# TODO: Should be relative to xdg-data-home
+mkdir ${HOME}/.local/share/flatpak/overrides
+noblacklist ${HOME}/.local/share/flatpak/overrides
+whitelist ${HOME}/.local/share/flatpak/overrides
+
+include disable-devel.inc
+include disable-exec.inc
+include disable-passwdmgr.inc
```
rusty-snake, line 20:
```suggestion
```

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+read-only ${HOME}/.config/dconf/
+noblacklist ${HOME}/.config/dconf/
+whitelist ${HOME}/.config/dconf/
+
+# TODO: Should be relative to xdg-data-home
+mkdir ${HOME}/.local/share/flatpak/overrides
+noblacklist ${HOME}/.local/share/flatpak/overrides
+whitelist ${HOME}/.local/share/flatpak/overrides
+
+include disable-devel.inc
```
rusty-snake, line 18: What about `disable-common`? `disable-proc`? `disable-interpreters`?

etc/profile-a-l/com.github.tchx84.Flatseal.profile
rusty-snake, line 16: Move the whitelist+mkdir down as per template.

rusty-snake, line 16: Use `whitelist-common.inc`.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
...
+include com.github.tchx84XXgg12uC.Flatseal.local
+# Persistent global definitions
+include globals.local
+
+# Required for GTK theem to work:
+read-only ${HOME}/.config/dconf/
+noblacklist ${HOME}/.config/dconf/
+whitelist ${HOME}/.config/dconf/
+
+# TODO: Should be relative to xdg-data-home
```
rusty-snake, line 13:
```suggestion
```

xdg base dirs are not supported by firejail so far.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
rusty-snake, line 11:
```suggestion
```

Use `whitelist-commonc.inc`.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
+# Firejail profile for flatseal
+# This file is overwritten after every install/update
+# Persistent local customizations
+include com.github.tchx84XXgg12uC.Flatseal.local
```
rusty-snake, line 4:
```suggestion
include com.github.tchx84.Flatseal.local
```

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,57 @@
+# Firejail profile for flatseal
```
rusty-snake, line 1: Add it to firecfg.config

rusty-snake (dismissed):

etc/profile-a-l/com.github.tchx84.Flatseal.profile
WhyNotHugo: Thanks for the clarification, I'd no idea.

WhyNotHugo (commented):
I've updated based on most feedback.

Is there any way to, given a specific profile, make a list of all resulting whitelisted paths and permissions?

Rather than an overview of what's restricted, I'd kind want to see just what's being allowed.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
rusty-snake: If you don't include them, you grant full access (instead of restricted access). Include them if they work (wusc maybe need `whitelist /usr/share/flatseal`).

etc/profile-a-l/com.github.tchx84.Flatseal.profile
rusty-snake: In master shouldn't be any files referencing it. If you find one, open a PR. It was merged into disable-programs/disable-common.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
rusty-snake: Then try to figure out which `noblacklist`s are necessary.

For di you will need to `include allow-gjs.inc`.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
rusty-snake: Include it. Or you will not get an approving review from me. wc has been well tested across different distro/versions/configurations/… and non of these paths are really sensitive.

netblue30:
> Is there any way to, given a specific profile, make a list of all resulting whitelisted paths and permissions?

I put in this morning a tool (profstats) to print whitelsts/blacklists and several stats for profiles. It gets installed by default in /etc/firejail directory. As a regular user, go in /etc/firejail and start it:
`````
$ cd /etc/firejail
$ ./profstats --print-whitelist firefox.profile
firefox.profile: whitelist ${HOME}/.cache/mozilla/firefox
firefox.profile: whitelist ${HOME}/.mozilla
firefox.profile: whitelist /usr/share/doc
firefox.profile: whitelist /usr/share/firefox
firefox.profile: whitelist /usr/share/gnome-shell/search-providers/firefox-search-provider.ini
firefox.profile: whitelist /usr/share/gtk-doc/html
firefox.profile: whitelist /usr/share/mozilla
firefox.profile: whitelist /usr/share/webext
firefox.profile: whitelist ${RUNUSER}/*firefox*
firefox-common.profile: whitelist ${DOWNLOADS}
firefox-common.profile: whitelist ${HOME}/.pki
firefox-common.profile: whitelist ${HOME}/.local/share/pki
whitelist-run-common.inc: whitelist /run/NetworkManager/resolv.conf
whitelist-run-common.inc: whitelist /run/cups/cups.sock
whitelist-run-common.inc: whitelist /run/dbus/system_bus_socket
whitelist-run-common.inc: whitelist /run/media
whitelist-run-common.inc: whitelist /run/resolvconf/resolv.conf
whitelist-run-common.inc: whitelist /run/shm
whitelist-run-common.inc: whitelist /run/systemd/journal/dev-log
whitelist-run-common.inc: whitelist /run/systemd/journal/socket
whitelist-run-common.inc: whitelist /run/systemd/resolve/resolv.conf
whitelist-run-common.inc: whitelist /run/systemd/resolve/stub-resolv.conf
whitelist-run-common.inc: whitelist /run/udev/data

For blacklists use --print-blacklists

**Review by rusty-snake — COMMENTED** [review]
Forgotten to submit.

**Comment by rusty-snake** [comment]
> Is there any way to, given a specific profile, make a list of all resulting whitelisted paths and permissions?

Permissions is not a concrete defined concept in firejail, it's only used in an abstract way.
Regarding whitelisted paths, I think `fjp generate-standalone gnome-2048.profile | grep "^whitelist" | cut -d" " -f2 | sort -u` is the easiest, but with official tools IDK.

**Comment by glitsj16** [comment]
> For blacklists use --print-blacklists

Correction for https://github.com/netblue30/firejail/pull/4724#issuecomment-982638979: the flag is `--print-blacklist` (singular)

File: etc/profile-a-l/com.github.tchx84.Flatseal.profile [hunk 15]
```diff
@@ -0,0 +1,55 @@
...
+# Persistent local customizations
+include com.github.tchx84.Flatseal.local
+# Persistent global definitions
+include globals.local
+
+# Required for GTK theem to work:
+noblacklist ${HOME}/.config/dconf/
+noblacklist ${HOME}/.local/share/flatpak/overrides
+
+# Required for GTK theem to work:
```
glitsj16, line 12: Syntax: ... GTK theme ...

etc/profile-a-l/com.github.tchx84.Flatseal.profile
WhyNotHugo: Why grant access to read all these directories (and even write access to some of them) when it's not needed? Even if you haven't found any potential exploits, I really don't see a reason to give so much access.

I'd rather follow the principle of least privileged: to give access to strictly what's needed to do the job at hand, and nothing more.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,55 @@
...
+seccomp
+seccomp.block-secondary
+shell none
+tracelog
+
+disable-mnt
+private-bin com.github.tchx84.Flatseal,gjs
+private-cache
+private-dev
+private-etc dconf,fonts,gtk-3.0
```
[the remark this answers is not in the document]
WhyNotHugo: Oh, I see CI requires them. What's `alternatives`? It's not present on my system.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
rusty-snake: _`s|/usr|/usr/share|`_

Yes, that's how whitelisting works.
You can try it with wusc included and not included: `firejail --ignore=private-bin --profile=/path/to/flatseal.profile ls /usr/share`

glitsj16 (approved):
LGTM. Can be merged when requested change by @rusty-snake is resolved.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,55 @@
+# Firejail profile for flatseal
+# This file is overwritten after every install/update
+# Persistent local customizations
+include com.github.tchx84.Flatseal.local
+# Persistent global definitions
+include globals.local
+
+# Required for GTK theem to work:
+noblacklist ${HOME}/.config/dconf/
```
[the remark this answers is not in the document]
rusty-snake: If you go and add `blacklist ${HOME}/.config/dconf` to your `disable-common.local`, so many programs would break …

etc/profile-a-l/com.github.tchx84.Flatseal.profile
WhyNotHugo: Manage to figure this one out, easier that expected.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,62 @@
...
+
+include allow-gjs.inc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-proc.inc
```
rusty-snake, line 21:
```suggestion
include disable-proc.inc
include disable-programs.inc
include disable-shell.inc
```

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,62 @@
...
+
+dbus-user filter
+dbus-user.own com.github.tchx84.Flatseal
+dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.freedesktop.impl.portal.PermissionStore
+dbus-user.talk org.gnome.Software
+dbus-system none
+
+read-only ${HOME}/.config/dconf/
+read-write ${HOME}/.local/share/flatpak/overrides
```
rusty-snake, line 62: Never `read-only`ed

WhyNotHugo: Removing this line, I get an error:

```
(com.github.tchx84.Flatseal:12): Gjs-CRITICAL **: 16:20:46.526: JS ERROR: GLib.FileError: Failed to create file “/home/hugo/.local/share/flatpak/overrides/com.github.Eloston.UngoogledChromium.XYTID1”: Read-only file system
_saveOverrides@resource:///com/github/tchx84/Flatseal/js/models/permissions.js:182:21
_updateModels@resource:///com/github/tchx84/Flatseal/js/models/permissions.js:222:14
main@resource:///com/github/tchx84/Flatseal/js/main.js:34:24
run@resource:///org/gnome/gjs/modules/script/package.js:206:19
@/usr/local/bin/com.github.tchx84.Flatseal:9:17
```

rusty-snake: Interesting, `${HOME}/.local/share/flatpak/exports` is the only `read-only`ed dir in `~/.local/share/flatpak`. What could cause this?

etc/profile-a-l/com.github.tchx84.Flatseal.profile
WhyNotHugo: Oh, so access to `/usr/share` is unrestricted by default, but if anything inside that directory is whitelisted, then _only_ the whitelisted directory is visible, and everything else is hidden.

Right, re-reading the man page now, that's actually what is says, but I somehow didn't quite understand that the first time, thanks for the clarification.

WhyNotHugo: Updated this bit.

Honestly, I think the whitelisting mechanic is a bit hard to fully understand. On the other hand, I can't think of any obvious way of making it simpler, so don't take this as a hard criticism.

rusty-snake: `include whitelist-run-common.inc`?

etc/profile-a-l/com.github.tchx84.Flatseal.profile
WhyNotHugo: `whitelist-common.inc` seems to give read-write access to a bunch of locations, right. Shouldn't these be read-only? It seems that any application with those has permissions to mess with quite wide-scoped configurations files (e.g.: proxies used by KDE applications).

etc/profile-a-l/com.github.tchx84.Flatseal.profile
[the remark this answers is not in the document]
WhyNotHugo: I don't think flatseal can in any way spawn other applications, so that might be unnecessary here.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
rusty-snake: > Honestly, I think the whitelisting mechanic is a bit hard to fully understand. On the other hand, I can't think of any obvious way of making it simpler, so don't take this as a hard criticism.

Yes, it's probably the hardest one (if we ignore this like seccomp which go deeper in how linux works). Based on the discussion in #3447, the idea is to have `mask /usr/share` to enable/disable this behaviour (if anyone ever implements this).

etc/profile-a-l/com.github.tchx84.Flatseal.profile
WhyNotHugo: It's set read-only in `disable-common.inc`.

rusty-snake: `flatpak/exports` yes but `flatpak/overrides`?

```
$ grep "read-only .*flatpak" /etc/firejail/*.inc
/etc/firejail/disable-common.inc:read-only ${HOME}/.local/share/flatpak/exports
```

WhyNotHugo: Somewhat out-of-scope: what do you think about whitelisted locations being read-only by default, and requiring that read-write be set explicitly?

Currently, a lot of locations are whitelisted, and all those are made read-write implicitly, which really increases the scope of how much access each sandbox has (and how much impact they can have messing around).

rusty-snake: > Somewhat out-of-scope: what do you think about whitelisted locations being read-only by default, and requiring that read-write be set explicitly?

Generally I thinks it's a bad idea.
1. It's a not small breaking change used in hundreds of profiles and locals.
2. It's not what users expect if they `whitelist ${HOME}/my-documents` in a local.

In specific of wc, we could add `read-only` all `whitelist`ed paths (which don't need to be read-write).

WhyNotHugo: True, it'd be a mess in terms of compat and breaking people's overrides too.

I think `whitelist-ro` makes sense too. My train of though is mostly that, if I want to grant just read-only access to a directory right now, I need to `whitelist`, which grants write access _and then_ remove the write access, which just sounds... not right. I'll experiment a bit with that and see if I can produce a viable PR.

rusty-snake: Maybe it's just a `profile_add("whitelist {path}"); profile_add("read-only {path}");`, IDK.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
WhyNotHugo: The thing that doesn't fully convince me about the whitelist mechanic, is that by default, all access to `/run` is allowed, so we use `include whitelist-run-common.inc`, but that still grants access to various unnecessary items (including printing!?) which this application doesn't **need**.

Ideally, `/run` should be entirely hidden here.

rusty-snake: > including printing?

Then add `noprinters`

> Ideally, /run should be entirely hidden here.

Not possible `/run/user` is required and `/run/firejail` too.

>  but that still grants access to various unnecessary items which this application doesn't need.

In specific of wrc, see #4607.

rusty-snake (approved):

WhyNotHugo:
I think all comments have been addressed, but lemme know if I've missed anything.

etc/profile-a-l/com.github.tchx84.Flatseal.profile
```diff
@@ -0,0 +1,63 @@
+# Firejail profile for flatseal
+# This file is overwritten after every install/update
+# Persistent local customizations
+include com.github.tchx84.Flatseal.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.local/share/flatpak/overrides
+noblacklist /var/lib/flatpak/app
+
```
kmk3, line 10: Missing comment from profile.template:

```
# Allow gjs (blacklisted by disable-interpreters.inc)
```

WhyNotHugo: Added

etc/profile-a-l/com.github.tchx84.Flatseal.profile
WhyNotHugo: Added `noprinters`.

kmk3 (approved):
LGTM.

WhyNotHugo:
Anything blocking this merge?

rusty-snake:
No.

Merged, thanks.

PR #4727, opened 2021-11-30
additional electron blacklists

ghost:
Two additional blacklist items for electron:
- ${HOME}/.config/electron-flag*.conf (see [this](https://wiki.archlinux.org/title/Wayland#Electron) for rationale)
- ${HOME}/.config/Electron

### DIFF
--- a/etc/profile-a-l/electron.profile
+++ b/etc/profile-a-l/electron.profile
@@ -6,8 +6,7 @@ include electron.local
 
 noblacklist ${HOME}/.config/Electron
 noblacklist ${HOME}/.config/electron-flag*.conf
-whitelist ${HOME}/.config/Electron
-whitelist ${HOME}/.config/electron-flag*.conf
+
 include disable-common.inc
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/electron.profile
```diff
@@ -4,6 +4,8 @@
 # Persistent local customizations
 include electron.local
 
+whitelist ${HOME}/.config/electron-flag*.conf
+whitelist ${HOME}/.config/Electron
```
rusty-snake, line 8: Move down to the other whitelists.

Also shouldn't we `noblacklist` them here?

glitsj16: Added the `noblacklist`s, thanks for reminding me about those!

> Move down to the other whitelists.
Has this changed recently? As the blacklisting is done in `disable-programs.inc` I added them above the block that includes that file...

rusty-snake (approved):

PR #4735, opened 2021-12-03
etc/profile-a-l/display.profile: additions needed on Gentoo

hlein:
Various .so's are needed to allow execution, /etc/ImageMagick-7/ is
needed for various policy XML files, and /usr/$(libdir)/ImageMagick-x.y.z/
is needed in order to have access to decoders.

Tested on Gentoo; I don't know if other distros put the relevant bits
in different paths.

### DIFF
--- a/etc/profile-a-l/display.profile
+++ b/etc/profile-a-l/display.profile
@@ -40,7 +40,8 @@ shell none
 private-bin display,python*
 private-dev
 # On Debian-based systems, display is a symlink in /etc/alternatives
-private-etc alternatives,ld.so.cache,ld.so.preload
+private-etc alternatives,ImageMagick-6,ImageMagick-7,ld.so.cache,ld.so.preload
+private-lib gcc/*/*/libgcc_s.so.*,gcc/*/*/libgomp.so.*,ImageMagick*,libfreetype.so.*,libltdl.so.*,libMagickWand-*.so.*,libXext.so.*
 private-tmp
 
 dbus-user none


etc/profile-a-l/display.profile
```diff
@@ -40,7 +40,8 @@ shell none
 private-bin display,python*
 private-dev
 # On Debian-based systems, display is a symlink in /etc/alternatives
-private-etc alternatives,ld.so.cache,ld.so.preload
+private-etc alternatives,ImageMagick*,ld.so.cache,ld.so.preload
```
glitsj16, line 43: private-etc doesn't have globbing support (yet):
```
$ firejail /usr/bin/display
Error: "ImageMagick*" is an invalid filename: rejected character: "*"
Error: proc 612189 cannot sync with peer: unexpected EOF
Peer 612190 unexpectedly exited with status 1
```
Using `ImageMagick-7` in the private-etc line works fine here on Arch Linux.

glitsj16 (changes requested):
Suggestion:
`private-etc alternatives,ImageMagick-7,ld.so.cache,ld.so.preload`

etc/profile-a-l/display.profile
hlein: > private-etc doesn't have globbing support (yet):

Ah, right you are. I had it correct in my `display.local` for testing and then did it wrong when preparing the PR :/

Updated.

glitsj16 (approved):
LGTM. Thanks for the PR!

PR #4736, opened 2021-12-03
Stop creating unused ~/.TelegramDesktop

WhyNotHugo:
Telegram does not use this directory, and these directives result in
filejail creating an empty directory in `$HOME` each time telegram is run.

### DIFF
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -19,7 +19,6 @@ blacklist ${HOME}/.Sayonara
 blacklist ${HOME}/.Steam
 blacklist ${HOME}/.Steampath
 blacklist ${HOME}/.Steampid
-blacklist ${HOME}/.TelegramDesktop
 blacklist ${HOME}/.VSCodium
 blacklist ${HOME}/.ViberPC
 blacklist ${HOME}/.VirtualBox
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -5,7 +5,6 @@ include telegram.local
 # Persistent global definitions
 include globals.local
 
-noblacklist ${HOME}/.TelegramDesktop
 noblacklist ${HOME}/.local/share/TelegramDesktop
 
 include disable-common.inc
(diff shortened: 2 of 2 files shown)


etc/inc/disable-programs.inc
```diff
@@ -19,7 +19,6 @@ blacklist ${HOME}/.Sayonara
 blacklist ${HOME}/.Steam
 blacklist ${HOME}/.Steampath
 blacklist ${HOME}/.Steampid
-blacklist ${HOME}/.TelegramDesktop
```
rusty-snake, line 22: This directory might still exist on some systems and it maybe contains things like chat protocols? We should keep the `blacklist` of it.

etc/profile-m-z/telegram.profile
```diff
@@ -5,7 +5,6 @@ include telegram.local
 # Persistent global definitions
 include globals.local
 
-noblacklist ${HOME}/.TelegramDesktop
```
rusty-snake, line 8: Hence we keep the `noblacklist`.

etc/profile-m-z/telegram.profile
```diff
@@ -16,9 +15,7 @@ include disable-programs.inc
 include disable-shell.inc
 include disable-xdg.inc
 
-mkdir ${HOME}/.TelegramDesktop
 mkdir ${HOME}/.local/share/TelegramDesktop
-whitelist ${HOME}/.TelegramDesktop
```
rusty-snake, line 21: If you did not started telegram for a long time, this directory might still be read if you start telegram to migrate the settings.

etc/profile-m-z/telegram.profile
```diff
@@ -16,9 +15,7 @@ include disable-programs.inc
 include disable-shell.inc
 include disable-xdg.inc
 
-mkdir ${HOME}/.TelegramDesktop
```
rusty-snake, line 19: There might still telegram versions in use which do use this directory (https://packages.debian.org/buster/telegram-desktop).

rusty-snake (changes requested):

glitsj16:
@WhyNotHugo You can add `ignore mkdir ${HOME}/.TelegramDesktop` to your telegram.local instead.

etc/profile-m-z/telegram.profile
WhyNotHugo: Oh, I tried to figure out in what conditions Telegram would use that directory, but found nothing relevant. Makes sense.

WhyNotHugo:
Would simply removing `mkdir ${HOME}/.TelegramDesktop` be acceptable?

If it's only read to migrate legacy user data, then it never needs to be created; only mounted if it exists, right?

PR #4740, opened 2021-12-03
Implement a `whitelist-ro` command

WhyNotHugo:
This is a shortcut to:

    whitelist $PATH
    read-only $PATH

Ideally, a great deal of usages of `whitelist` should be replaced with this instead.


WhyNotHugo:
I'm a bit hesitant on whether a `whitelist-rw` makes sense, and then discourage using `whitelist`.

Specifying whether it should be read-only or read-write explicitly could reduce surprises and avoid unintentionally granting write access when trying to just grant read access.

src/firejail/profile.c
```diff
@@ -1789,6 +1789,17 @@ void profile_read(const char *fname) {
 			free(ptr);
 			ptr = tmp;
 		}
+		else if (strncmp(ptr, "whitelist-ro ", 13) == 0) {
```
rusty-snake, line 1792:
```C
while {
    …
}
else if
```

???

Anyway I can not review code changes in firejail.

rusty-snake (commented):
IMHO `--whitelist-ro` should be implemented too.

> If you add a new command, here's the checklist:
>
> - [ ] Update manpages: firejail(1) and firejail-profile(5)
> - [ ] Update shell completions
> - [ ] Update vim syntax files
> - [ ] Update --help

kmk3 (dismissed):
Leaving a preliminary NAK, as I have some concerns related to whitelisting and
new commands.  I'll expand on this tomorrow.

kmk3 (commented):
After some consideration, I'm not 100% sold on it, but the concerns below are
not necessarily blockers.  I'm still posting them as I think they could be
useful.

---

Firstly, I think that the name itself looks fine and if adding a shortcut
command, I do agree that it would be better to make it simply "decay" to
separate commands rather than being an actual new command, as it makes using
`ignore` easier.  Example:

With `whitelist-ro PATH` == `whitelist PATH` + `read-only PATH`:

* Ignore only the whitelist: `ignore whitelist PATH`
* Ignore only read-only: `ignore read-only PATH`

With `whitelist-ro PATH` being its own command:

* Ignore only the whitelist: `ignore whitelist-ro PATH` + `read-only PATH`
* Ignore only read-only: `ignore whitelist-ro PATH` + `whitelist PATH` 

This is minor, but for completeness: Note that (unless `ignore` calls are also
translated), some people could be confused that `ignore whitelist-ro PATH`
would not work.

---

@WhyNotHugo [on Dec 3](https://github.com/netblue30/firejail/pull/4724#discussion_r762103002):

> I think `whitelist-ro` makes sense too. My train of though is mostly that, if
> I want to grant just read-only access to a directory right now, I need to
> `whitelist`, which grants write access _and then_ remove the write access,
> which just sounds... not right. I'll experiment a bit with that and see if I
> can produce a viable PR.

If it helps, `whitelist` is about visibility, not read/write access, so it does
not technically grant write access per se.  And firejail profiles are written
in a "blacklist mode" rather than "whitelist mode" (i.e.: blocking things is
explicit), so it kind of makes sense that there are multiple commands to do
different things.  But I get what you mean usability-wise when trying to grant
limited access to many paths at once.

Though note that there could be shortcuts for many combinations of path-related
commands, such as ones involving the following commands:

* noblacklist
* mkdir
* mkfile
* whitelist
* read-only

So technically we could add things like `whitelist-ro`, `mkdir-whitelist`,
`mkdir-whitelist-ro`, `noblacklist-whitelist`, `noblacklist-mkfile-whitelist`,
`noblacklist-mkdir-whitelist-ro`, etc (though hopefully with better names).  I
don't think it would be a good idea to have (/maintain and support) too many of
these (it would be confusing IMO), so it could be useful to take a look at
which combinations would help the most.

`noblacklist` is a prerequisite to whitelisting on all/most committed profiles
anyway and `mkdir` / `mkfile` is usually used on the same paths as well, so a
shortcut for these would be my first choice (and I do have a WIP proposal for
something like that).

I also have a WIP proposal for adding two new commands that split the
functionality of `whitelist` (keep + enforce), to be able to whitelist things
anywhere.  I was going to say that maybe `whitelist-ro` could mean
`whitelist.keep` + `read-only` rather than `whitelist` + `read-only`, but I
assume that it is intended mostly as syntatic sugar for end-users (rather than
for upstream profiles), so having it mean the latter probably makes more sense.
Also, a `whitelist-ro.keep` command could be added later anyway if that were to
be desired.

rusty-snake:
> ignore whitelist PATH

Does `ignore` work with `profile_add`?

> ignore whitelist-ro PATH

I think this work, it does not work?

kmk3:
@rusty-snake commented [on Dec 6](https://github.com/netblue30/firejail/pull/4740#issuecomment-987102519):

> > ignore whitelist PATH
> 
> Does `ignore` work with `profile_add`?
> 
> > ignore whitelist-ro PATH
> 
> I think this work, it does not work?

I assumed that `profile_add("whitelist PATH"); profile_add("read-only PATH");`
would be completely equivalent to writing those commands directly on a profile.
But now that you mention it, I'm not so sure.

@WhyNotHugo Can you test which `ignore` commands would work for `whitelist-ro`:

* `ignore whitelist PATH`
* `ignore read-only PATH`
* `ignore whitelist-ro PATH`

src/firejail/profile.c
WhyNotHugo: I think that's just github being weird when rendering:

https://github.com/netblue30/firejail/blob/894bf81a595304b0c1a843c85a607003fd419d1b/src/firejail/profile.c#L1785-L1802

WhyNotHugo:
I'd rather not think of it in these terms:

    whitelist-ro PATH == whitelist PATH + read-only PATH

But these:

    whitelist == whitelist-ro PATH + read-write PATH

I understand that a Linux process has, by default, access to the whole filesystem, and can assume all capabilities, etc, but when it comes to security, it's sometimes easier to think about it the opposite way: start with no access to anything, and grant access to individual items one-by-one. `whitelist-ro` lets a process _read_ a path, and `read-write` lets it read+write.

Of course, this somewhat means that the high level concepts are exactly the opposite of what's happening under the hood. For example, if a process has "only" read-access to one file, in reality, the actions that firejail does should reflect the exact opposite: deny access to all others, and then remove write access to that particular one.

This makes the internal translation a bit more complex, but makes the API exposed more security-friendly.

---

I didn't try ignore. I'll give those a shot and report back on how those work, plus previous comments.

netblue30:
Let's go for it, I'll put something in the man page. Thanks!

rusty-snake:
https://github.com/netblue30/firejail/pull/4740#pullrequestreview-822995437

> IMHO `--whitelist-ro` should be implemented too.
> 
> > If you add a new command, here's the checklist:
> > 
> > * [ ]  Update manpages: firejail(1) and firejail-profile(5)
> > * [ ]  Update shell completions
> > * [ ]  Update vim syntax files
> > * [ ]  Update --help

PR #4743, opened 2021-12-06
Add CachyBrowser profile

vnepogodin:


etc/profile-a-l/cachy-browser.profile
```diff
@@ -0,0 +1,57 @@
...
+#dbus-user.talk org.kde.JobViewServer
+#dbus-user.talk org.kde.kuiserver
+# Add the next three lines to your cachy-browser.local to allow screensharing under Wayland.
+#whitelist ${RUNUSER}/pipewire-0
+#whitelist /usr/share/pipewire/client.conf
+#dbus-user.talk org.freedesktop.portal.*
+# Also add the next line to your cachy-browser.local if screensharing does not work with
+# the above lines (depends on the portal implementation).
+#ignore noroot
+ignore dbus-user none
```
rusty-snake, line 54: Should be based on master librewolf.profile.

rusty-snake (commented):
LGTM

etc/profile-a-l/cachy-browser.profile
vnepogodin: it is actually
or you mean to include librewolf.profile instead of firefox
`include librewolf.profile`

btw will it lead to any problem for example with cachy-browser{-beta,-nightly,etc}?

rusty-snake: > it is actually

That's the current block on master:
https://github.com/netblue30/firejail/blob/9b1daa6b3abecfa0b270e9202106575cd49aeacc/etc/profile-a-l/librewolf.profile#L38-L53

It does not contain `whitelist /usr/share/pipewire/client.conf` (for example).

> btw will it lead to any problem for example with cachy-browser{-beta,-nightly,etc}?

You mean
- having both installed? No (firecfg does not create symlinks anyway)
- running -beta... with this profile? No (except if paths a different or you use private-bin)

vnepogodin: no. I meant that you do `include cachy-browser.profile` in cachy-browser-beta(for example).

rusty-snake: You can do it just like firefox or librewolf handle nightly/beta/dev/esr versions.

vnepogodin: ah ok. just wanted to be sure

vnepogodin: how to determine string for `dbus-user.own`? Where is it located?

vnepogodin: oh nevermind. I did find it

rusty-snake: Without source-code digging: Run cachy without dbus filtering and look at
- d-feet (GTK)
- qdbusviewer (QT)
- `busctl --user list` (CLI)
- or call ListNames
  - `dbus-send --print-reply --dest=org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus.ListNames`
  - `gdbus call --session --dest org.freedesktop.DBus --object-path /org/freedesktop/DBus --method org.freedesktop.DBus.ListNames`
  - …

vnepogodin: yeah thx I already did it

vnepogodin: btw It should be the issue as firefox-developer-edition or librewolf-nightly differ from the parent

for example for librewolf-nightly:
should be `dbus-user.own org.mozilla.librewolfnightly.*` instead of `dbus-user.own org.mozilla.librewolf.*`

netblue30:
all merged, thanks~

PR #4744, opened 2021-12-07
dino.profile: netlink protocol is required for audio/video calls.

crocket:
According to my tests, without netlink protocol, audio/video calls keep trying to connect for ever without really connecting.


glitsj16 (approved):
LGTM

PR #4747, opened 2021-12-07
Skype profile tweaks

WhyNotHugo:
I hope these are in-order, since this profile is a bit of an unusual one.

I added Skype's config directory to `whitelist` and `mkdir`, since otherwise the session is lost when closing and re-running it, requiring logging in again each time it's run.

I also hardened the D-Bus aspect of the profile: I started out blocking everything, and then unblocked the necessary bit. Surprisingly, Skype will log out immediately after starting _without_ `org.kde.StatusNotifierWatcher`, I've no idea about the correlation. It's probably good to keep it around for tray users too anyway.

Ran a test call and it worked fine. User session is retained when quitting/re-running too.

### DIFF
--- a/etc/profile-m-z/skypeforlinux.profile
+++ b/etc/profile-m-z/skypeforlinux.profile
@@ -14,8 +14,8 @@ ignore include whitelist-var-common.inc
 ignore nou2f
 ignore novideo
 ignore private-dev
+
 ignore dbus-user none
-ignore dbus-system none
 
 # breaks Skype
 ignore apparmor
@@ -27,6 +27,12 @@ mkdir ${HOME}/.config/skypeforlinux
 whitelist ${HOME}/.config/skypeforlinux
 
 # private-dev - needs /dev/disk
+
+dbus-user filter
+dbus-user.talk org.freedesktop.Notifications
+dbus-user.talk org.freedesktop.secrets
+# Note: Skype will log out the current session on start-up without this:
+dbus-user.talk org.kde.StatusNotifierWatcher
 
 # Redirect
 include electron.profile
--- a/etc/profile-m-z/skypeforlinux.profile
+++ b/etc/profile-m-z/skypeforlinux.profile
@@ -23,6 +23,9 @@ ignore noexec /tmp
 
 noblacklist ${HOME}/.config/skypeforlinux
 
+mkdir ${HOME}/.config/skypeforlinux
+whitelist ${HOME}/.config/skypeforlinux
+
 # private-dev - needs /dev/disk
 
 # Redirect


etc/profile-m-z/skypeforlinux.profile
```diff
@@ -15,15 +15,22 @@ ignore nou2f
 ignore novideo
 ignore private-dev
 ignore dbus-user none
```
rusty-snake, line 17: Move this (`ignore dbus-user none`) out of the "Disabled until someone reported positive feedback" block.

etc/profile-m-z/skypeforlinux.profile
```diff
@@ -15,15 +15,22 @@ ignore nou2f
...
 
+mkdir ${HOME}/.config/skypeforlinux
+whitelist ${HOME}/.config/skypeforlinux
+
 # private-dev - needs /dev/disk
 
+dbus-user filter
+dbus-user.talk org.freedesktop.Notifications
+dbus-user.talk org.freedesktop.secrets
+dbus-user.talk org.kde.StatusNotifierWatcher
```
rusty-snake, line 33:
```suggestion
?ALLOW_TRAY: dbus-user.talk org.kde.StatusNotifierWatcher
```

rusty-snake (changes requested):
LGTM. Nitpicking `?ALLOW_TRY:`.

If you use skype, you can try to reduce the "Disabled until someone reported positive feedback" if you want.

etc/profile-m-z/skypeforlinux.profile
rusty-snake: OT: We should add a profile check for this.

WhyNotHugo: FYI: Skype will immediately log you out if this is not present, so I might be worth leaving a comment....?

rusty-snake: In that case, I would say leave it enabled but add a comment that we do not add `?ALLOW_TRAY:` later on profile clean ups.

WhyNotHugo: Oh, I've already added it. Given that it's enabled by default it sounds kinda safe...? 
But let me know your preference, either way sounds okay to me.

rusty-snake: > Given that it's enabled by default it sounds kinda safe...?

What? allow-tray or StatusNotifierWatcher?

rusty-snake: >  But let me know your preference, either way sounds okay to me (I don't use a tray anyway).

If skype does not work without `dbus-user.talk org.kde.StatusNotifierWatcher`, we should allow it.

rusty-snake (approved):

WhyNotHugo:
Squashed intermediate commits.

netblue30:
all set, thanks!

PR #4755, opened 2021-12-09
yt-dlp: add missing paths & mpv.profile: whitelist paths for yt-dlp

### LINKED ISSUE #4754: Support for youtube-dl forks in e.g mpv
### Description

youtube-dl was abandoned and yt-dlp is the most popular alternative, but there are loads of other forks but I do not know how relevant they are to us. Since youtube-dl has not got any updates since 4 months it might stop working very soon and another path is needed in `private-bin` so it can keep working with e.g yt-dlp. It might be counterproductive to just randomly add the 5 most popular (which is also hard to define) yt-dl forks into there, so what do we do? For now I got yt-dlp added into an override but it would be great if there would be some default solution since it will potentially affect many.

Potentially relevant: #4539

kmk3:
Fixes #4754.

Relates to #3472, #4634 and <https://github.com/yt-dlp/yt-dlp/issues/19>.

Cc: @Fred-Barclay @NetSysFire @pirate486743186

### DIFF
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -637,6 +637,7 @@ blacklist ${HOME}/.config/youtube-music-desktop-app
 blacklist ${HOME}/.config/youtube-viewer
 blacklist ${HOME}/.config/youtubemusic-nativefier-040164
 blacklist ${HOME}/.config/yt-dlp
+blacklist ${HOME}/.config/yt-dlp.conf
 blacklist ${HOME}/.config/zathura
 blacklist ${HOME}/.config/zim
 blacklist ${HOME}/.config/zoomus.conf
@@ -1128,6 +1129,7 @@ blacklist ${HOME}/mps
 blacklist ${HOME}/openstego.ini
 blacklist ${HOME}/wallet.dat
 blacklist ${HOME}/yt-dlp.conf
+blacklist ${HOME}/yt-dlp.conf.txt
 blacklist ${RUNUSER}/*firefox*
 blacklist /tmp/.wine-*
 blacklist /tmp/akonadi-*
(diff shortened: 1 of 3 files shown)


rusty-snake (approved):
Probably there are some more:

```
$ grep -l 'noblacklist ${HOME}/.config/youtube-dl' /etc/firejail/*.profile
/etc/firejail/celluloid.profile
/etc/firejail/firefox-common-addons.profile
/etc/firejail/mpsyt.profile
/etc/firejail/mpv.profile
/etc/firejail/smplayer.profile
/etc/firejail/youtube-dl-gui.profile
/etc/firejail/youtube-dl.profile
```

glitsj16 (approved):
LGTM

kmk3:
@rusty-snake [on Dec 9](https://github.com/netblue30/firejail/pull/4755#pullrequestreview-827354646):

> Probably there are some more:
> 
> ```
> $ grep -l 'noblacklist ${HOME}/.config/youtube-dl' /etc/firejail/*.profile
> /etc/firejail/celluloid.profile
> /etc/firejail/firefox-common-addons.profile
> /etc/firejail/mpsyt.profile
> /etc/firejail/mpv.profile
> /etc/firejail/smplayer.profile
> /etc/firejail/youtube-dl-gui.profile
> /etc/firejail/youtube-dl.profile
> ```

Indeed; I'll change those on a new PR.

By the way, does any of those programs create/modify youtube-dl/yt-dlp config?

I think maybe only youtube-dl-gui would do so.

If we can be reasonably sure that they don't, more `mkdir`s could be removed
and the config paths could be marked `read-only`.

netblue30:
merged, thanks!

PR #4771, opened 2021-12-10
Revert allow/deny leftovers

kmk3:
Remove the remaining code that was not reverted on #4410.

Note: If you want to keep profcleaner, I can drop the last commit.


netblue30:
merged, thanks!

PR #4774, opened 2021-12-13
CI: pin GitHub actions to SHAs and use Dependabot to update them

topimiettinen:
See:
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
and:
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/workflows/build-extra.yml
+++ b/.github/workflows/build-extra.yml
@@ -30,7 +30,7 @@ jobs:
   build-clang:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
     - name: configure
       run: CC=clang-11 ./configure --enable-fatal-warnings
     - name: make
@@ -38,7 +38,7 @@ jobs:
   scan-build:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
     - name: install clang-tools-11
       run: sudo apt-get install clang-tools-11
     - name: configure
@@ -48,7 +48,7 @@ jobs:
   cppcheck:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
     - name: install cppcheck
       run: sudo apt-get install cppcheck
     - name: cppcheck
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
   build_and_test:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
     - name: install dependencies
       run: sudo apt-get install gcc-11 libapparmor-dev libselinux1-dev expect xzdec
     - name: configure
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 2
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -43,11 +43,11 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v1
+      uses: github/codeql-action/init@e095058bfa09de8070f94e98f5dc059531bc6235
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
@@ -58,7 +58,7 @@ jobs:
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
     - name: Autobuild
-      uses: github/codeql-action/autobuild@v1
+      uses: github/codeql-action/autobuild@e095058bfa09de8070f94e98f5dc059531bc6235
 
     # ℹ️ Command-line programs to run using the OS shell.
     # 📚 https://git.io/JvXDl
@@ -72,4 +72,4 @@ jobs:
     #   make release
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v1
+      uses: github/codeql-action/analyze@e095058bfa09de8070f94e98f5dc059531bc6235
--- a/.github/workflows/profile-checks.yml
+++ b/.github/workflows/profile-checks.yml
@@ -20,7 +20,7 @@ jobs:
   profile-checks:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
     - name: sort.py
       run: ./ci/check/profiles/sort.py etc/inc/*.inc etc/{profile-a-l,profile-m-z}/*.profile
     - name: private-etc-always-required.sh


reinerh:
Is this really necessary for actions that come from Github?
I wouldn't expect them to be malicious (and even if they were, we don't use them to ship artefacts).

I personally would be more annoyed by regular merge requests from a bot that just bumps commit IDs.

topimiettinen:
> Is this really necessary for actions that come from Github? I wouldn't expect them to be malicious (and even if they were, we don't use them to ship artefacts).

Maybe, but there might still be problems with forks and PRs. We already have also LGTM, which is 3rd party. I'd expect the CI setup to expand to further 3rd party systems (for example Coveralls could be interesting) and maybe one day there are Python dependencies (like Meson and Ninja). The small downside is that simple version IDs like `v1` became much longer hashes. The benefit (although a bit theoretical ATM) is that it's hard to forge SHAs while simple strings can be copied easily.

> I personally would be more annoyed by regular merge requests from a bot that just bumps commit IDs.

The frequency can be tuned to monthly but I haven't found weekly check to be too noisy. The PRs are rather trivial and it's also easy to be vigilant and check the SHAs.

rusty-snake:
Did I miss understand something or where's the issue with semver dependencies if we are not using `${{ secrets.GITHUB_TOKEN }}`?

topimiettinen:
> Did I miss understand something or where's the issue with semver dependencies if we are not using `${{ secrets.GITHUB_TOKEN }}`?

There are other problems mentioned in the page, like "pull_request_target still has the read/write repository token in memory that is potentially available to any running program". Maybe other problems will rise in the future.

The actions could be also buggy and keeping them relatively recent could help with that, like with any software. Using unmaintained 'v1' forever is asking for problems.

kmk3:
@topimiettinen commented [on Dec 14](https://github.com/netblue30/firejail/pull/4774#issuecomment-993249757):

> > Is this really necessary for actions that come from Github? I wouldn't
> > expect them to be malicious (and even if they were, we don't use them to
> > ship artefacts).
> 
> Maybe, but there might still be problems with forks and PRs. We already have
> also LGTM, which is 3rd party. I'd expect the CI setup to expand to further
> 3rd party systems (for example Coveralls could be interesting) and maybe one
> day there are Python dependencies (like Meson and Ninja). The small downside
> is that simple version IDs like `v1` became much longer hashes. The benefit
> (although a bit theoretical ATM) is that it's hard to forge SHAs while simple
> strings can be copied easily.

Since AFAIK GitHub does not verify the GPG signature of a tag after cloning it,
pinning to a SHA does seem more secure.

> > I personally would be more annoyed by regular merge requests from a bot
> > that just bumps commit IDs.
> 
> The frequency can be tuned to monthly but I haven't found weekly check to be
> too noisy. The PRs are rather trivial and it's also easy to be vigilant and
> check the SHAs.

The dependabot spam is also a concern of mine, though I suppose that it's
partly because I've mostly seen it on e.g.: nodejs-based projects that have
hundreds of dependencies, which are also updated rather frequently.  If it can
be made to alert only weekly/monthly, then that doesn't seem too bad,
especially when using only a handful of GitHub Actions.

.github/workflows/build-extra.yml
```diff
@@ -30,15 +30,15 @@ jobs:
   build-clang:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@@ec3a7ce113134d7a93b817d10a8272cb61118579
```
kmk3, line 33: It's unclear which version a SHA represents.

How about putting each intended version in a comment?

```suggestion
    # v2
    - uses: actions/checkout@@ec3a7ce113134d7a93b817d10a8272cb61118579
```

Or maybe:

```suggestion
    # actions/checkout@v2
    - uses: actions/checkout@@ec3a7ce113134d7a93b817d10a8272cb61118579
```

topimiettinen: Would the comment be updated by Dependabot? Otherwise it will bitrot.

topimiettinen, line 33: Two @-signs...

.github/workflows/build-extra.yml
```diff
@@ -30,15 +30,15 @@ jobs:
...
     - name: configure
       run: CC=clang-11 ./configure --enable-fatal-warnings
     - name: make
       run: make
   scan-build:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@@ec3a7ce113134d7a93b817d10a8272cb61118579
```
topimiettinen, line 41: Also here @@

.github/workflows/build-extra.yml
```diff
@@ -48,7 +48,7 @@ jobs:
   cppcheck:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@@ec3a7ce113134d7a93b817d10a8272cb61118579
```
topimiettinen, line 51: Here too...

.github/workflows/build.yml
```diff
@@ -22,7 +22,7 @@ jobs:
   build_and_test:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@@ec3a7ce113134d7a93b817d10a8272cb61118579
```
topimiettinen, line 25: ditto

topimiettinen:
> @topimiettinen commented [on Dec 14](https://github.com/netblue30/firejail/pull/4774#issuecomment-993249757):
> 
> > > Is this really necessary for actions that come from Github? I wouldn't
> > > expect them to be malicious (and even if they were, we don't use them to
> > > ship artefacts).
> > 
> > 
> > Maybe, but there might still be problems with forks and PRs. We already have
> > also LGTM, which is 3rd party. I'd expect the CI setup to expand to further
> > 3rd party systems (for example Coveralls could be interesting) and maybe one
> > day there are Python dependencies (like Meson and Ninja). The small downside
> > is that simple version IDs like `v1` became much longer hashes. The benefit
> > (although a bit theoretical ATM) is that it's hard to forge SHAs while simple
> > strings can be copied easily.
> 
> Since AFAIK GitHub does not verify the GPG signature of a tag after cloning it, pinning to a SHA does seem more secure.

A GPG signature is also a hash, signed with a key. Instead of pinning the SHA, only the public key used for signing would need to be pinned. But which key? It would also need updating when the keys expire, there's maybe also CRL, OCSP, master keys, CAs etc. Cryptographic signatures could be better but they're also more complex.

> > > I personally would be more annoyed by regular merge requests from a bot
> > > that just bumps commit IDs.
> > 
> > 
> > The frequency can be tuned to monthly but I haven't found weekly check to be
> > too noisy. The PRs are rather trivial and it's also easy to be vigilant and
> > check the SHAs.
> 
> The dependabot spam is also a concern of mine, though I suppose that it's partly because I've mostly seen it on e.g.: nodejs-based projects that have hundreds of dependencies, which are also updated rather frequently. If it can be made to alert only weekly/monthly, then that doesn't seem too bad, especially when using only a handful of GitHub Actions.

Yeah, I can only imagine the noise with JS or Python. This shouldn't be as bad. Continuing the comparison of the Actions to other software: an LTS version of the Action with only security updates could be nice but I guess such things don't exist.

.github/workflows/build-extra.yml
kmk3: @topimiettinen [on Dec 14](https://github.com/netblue30/firejail/pull/4774#discussion_r769078738):

> Would the comment be updated by Dependabot? Otherwise it will bitrot.

With the current configuration, does Dependabot put something like "update
foo@v1 to foo@v2" in the commit message (I know that it can do that, but I'm
not sure if it's the default)?  If the equivalent tags can be discovered
locally/without having to search for them on GitHub, I think that would be good
enough.

If not, instead of a comment on each item, to avoid bitrot and duplication I
think we could have a mapping on a separate file.  It would only need updating
when adding a completely new "action@version" combination (i.e.: one that was
never part of any committed workflow file).  Kind of like a "package.lock"
file.  Example:

```
actions/checkout@v2: ec3a7ce113134d7a93b817d10a8272cb61118579
actions/checkout@v3: 1231231231231231231231231231231231231231
github/codeql-action/init@v1: e095058bfa09de8070f94e98f5dc059531bc6235
```

By the way, is Dependabot smart enough to update a SHA to a new SHA or would it
open a PR to bump `actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579`
to `actions/checkout@v3` in this case?

topimiettinen: > @topimiettinen [on Dec 14](https://github.com/netblue30/firejail/pull/4774#discussion_r769078738):
> 
> > Would the comment be updated by Dependabot? Otherwise it will bitrot.
> 
> With the current configuration, does Dependabot put something like "update foo@v1 to foo@v2" in the commit message (I know that it can do that, but I'm not sure if it's the default)? If the equivalent tags can be discovered locally/without having to search for them on GitHub, I think that would be good enough.

Yes, here's an example commit/PR with the same settings: https://github.com/topimiettinen/libaslrmalloc/commit/0aeb6639ebfda800ace6cd36306d0869e7ce207e

> 
> If not, instead of a comment on each item, to avoid bitrot and duplication I think we could have a mapping on a separate file. It would only need updating when adding a completely new "action@version" combination (i.e.: one that was never part of any committed workflow file). Kind of like a "package.lock" file. Example:
> 
> ```
> actions/checkout@v2: ec3a7ce113134d7a93b817d10a8272cb61118579
> actions/checkout@v3: 1231231231231231231231231231231231231231
> github/codeql-action/init@v1: e095058bfa09de8070f94e98f5dc059531bc6235
> ```

For Python, there's `requirements.txt` with similar syntax (https://github.com/topimiettinen/libaslrmalloc/blob/master/.github/workflows/requirements.txt) but I don't know if that would work for GH actions. Example of Dependabot updating it: https://github.com/topimiettinen/libaslrmalloc/commit/248495bd96599b59771bd9edcd4e6e510c4e97db

> By the way, is Dependabot smart enough to update a SHA to a new SHA or would it open a PR to bump `actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579` to `actions/checkout@v3` in this case?

Only SHA is updated. There's also a GH action to verify that PRs don't change SHAs to semvers: https://github.com/zgosalvez/github-actions-ensure-sha-pinned-actions but I don't know if there's an action to verify that it actually does what it says it does and so forth.

kmk3: Sounds good then; thanks for the examples.

kmk3 (approved):
+1 for having Dependabot check the dependencies.

As for the pinning, maybe using exact versions (such as `@2.1.0`) would be a
secure enough alternative (not sure), but I don't feel strongly about it either
way.  Might expand on this on a discussion later.

topimiettinen:
> +1 for having Dependabot check the dependencies.

Great! What about @reinerh and @rusty-snake, are you now OK with this or do you have further questions?

> As for the pinning, maybe using exact versions (such as `@2.1.0`) would be a secure enough alternative (not sure), but I don't feel strongly about it either way. Might expand on this on a discussion later.

The problem is that PRs and forks could still forge the same exact version numbers but with SHAs its very hard.

rusty-snake:
I still don't understand why it is necessary (we don't use pull_request_target either), but don't take this as blocking a merge.

reinerh (approved):

PR #4779, opened 2021-12-14
Fix teams ignoring input sources e.g. microphones

seonwoolee:
`teams-for-linux.profile` and `teams.profile` both include `electron.profile`, which has the `noinput` option, which blocks microphones and other input sources, so this option needs to be ignored.

### DIFF
--- a/etc/profile-m-z/teams.profile
+++ b/etc/profile-m-z/teams.profile
@@ -10,9 +10,10 @@ include globals.local
 ignore include disable-xdg.inc
 ignore include whitelist-runuser-common.inc
 ignore include whitelist-usr-share-common.inc
-ignore noinput
 ignore novideo
 ignore private-tmp
+
+ignore novideo
 
 # see #3404
 ignore apparmor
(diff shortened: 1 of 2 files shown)


rusty-snake (changes requested):
Don't put them in the "Disabled until someone reported positive feedback" block.

---

Is this special for teams or does it also happen with e.g. firefox?

seonwoolee:
Sorry, I was trying to respect the ordering rules for pull requests.

It's not a general problem, just any electron app that needs microphone input. Not applicable to Firefox

rusty-snake (approved):

etc/profile-m-z/teams.profile
```diff
@@ -13,6 +13,8 @@ ignore include whitelist-usr-share-common.inc
 ignore novideo
 ignore private-tmp
 
+ignore novideo
```
kmk3, line 16: I think you meant `ignore noinput`.

kmk3:
Also, could you squash the commits and push?  Example:

```sh
git stash; git tag backup # just in case
git rebase -i "$(git merge-base master HEAD)..HEAD"
# Change all occurrences of "pick" to "fixup", except for the first one
# (This makes the first commit "absorb" all the other commits)
# Check if it looks good (if not, `git reset --hard backup` and try again)
# Then just push
git push --force-with-lease
```

netblue30:
merged, thanks!

etc/profile-m-z/teams.profile
rusty-snake: Still unresolved.

Issue #3434, opened 2020-05-26
Default shell is guessed from $SHELL, despite manpage specifying /bin/bash


iltep64:
**Bug and expected behavior**
Firejail's man page describes the default shell for execution as `/bin/bash`.
However, if the given command line contains bash-isms(?) such as adding `--` in `firejail -- ls`, and an incompatible shell is currently used and present in `$SHELL` (e.g. fish), then the command fails.

This issue was encountered in the `zoom-firejail` AUR package.

**Steps and Output**
The logs are collapsed for brevity.

*Unexpected*
The external shell environment has `$SHELL=/bin/fish` set, unless stated otherwise.
<details>
<summary>Running without any flags</summary>

```
$ export SHELL=/bin/fish
$ firejail -- ls
Reading profile /etc/firejail/default.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc

** Note: you can use --noprofile to disable default.profile **

Parent pid 143487, child pid 143488
Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Child process initialized in 32.74 ms
fish: Unknown command: --
fish: 
--
^

Parent is shutting down, bye...
```

</details>

<details>
<summary>Running with <code>--noprofile</code></summary>

```
 firejail --noprofile -- ls
Parent pid 145677, child pid 145678
Child process initialized in 7.24 ms
fish: Unknown command: --
fish: 
--
^

Parent is shutting down, bye...
```

</details>

*Expected*

<details>
<summary>Direct execution</summary>

```
$ which ls
/usr/bin/ls
$ /usr/bin/ls
a  b  c
```
</details>

With `firejail` + a workaround, other than the command line itself, the results of execution are the same.
<details>
<summary>Running with <code>SHELL=/bin/fish</code> and <code>--shell=none</code> flag</summary>

```
$ firejail --shell=none -- ls
Reading profile /etc/firejail/default.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc

** Note: you can use --noprofile to disable default.profile **

Parent pid 143496, child pid 143497
Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Child process initialized in 36.39 ms
a  b  c

[...]

PR #4783, opened 2021-12-19
Allow telegram to open hyperlinks

YorkZ:
Without this fix, clicking hyperlinks in Telegram had no effect.

### DIFF
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -8,6 +8,9 @@ include globals.local
 noblacklist ${HOME}/.TelegramDesktop
 noblacklist ${HOME}/.local/share/TelegramDesktop
 
+# Allow opening hyperlinks
+include allow-bin-sh.inc
+
 include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
@@ -41,7 +44,7 @@ seccomp.block-secondary
 shell none
 
 disable-mnt
-private-bin telegram,Telegram,telegram-desktop
+private-bin bash,sh,telegram,Telegram,telegram-desktop,xdg-open
 private-cache
 private-dev
 private-etc alsa,alternatives,ca-certificates,crypto-policies,fonts,group,ld.so.cache,ld.so.preload,localtime,machine-id,os-release,passwd,pki,pulse,resolv.conf,ssl,xdg


SkewedZeppelin:
This doesn't seem right?

YorkZ:
@SkewedZeppelin What did you mean?

glitsj16:
> Without this fix, clicking hyperlinks in Telegram had no effect.

The [telegram profile](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/telegram.profile#L44) uses `private-bin telegram,Telegram,telegram-desktop`. Allowing ${PATH}/bash and ${PATH}/sh doesn't change that. I'm wondering how this change can fix anything if these aren't added to `private-bin` too. Maybe that's what @SkewedZeppelin's hinting at.

BTW, we have [allow-bin-sh.inc](https://github.com/netblue30/firejail/blob/master/etc/inc/allow-bin-sh.inc) to override 'include disable-shell.inc' in a more maintainable way compared to adding seperate noblacklist paths.

To get to the bottom of this it might be helpful to open an issue ticket where we can try to reproduce and discuss things before jumping in with a PR that - for now - offers little context. Just my 2 cents, nothing personal.

YorkZ:
@glitsj16 Open an issue would be a good idea, and I can do that. Meanwhile, I did have the issue that I wasn't able to open hyperlinks in Telegram because when Telegram tried to open a link, it invokes `xdg-open` which wouldn't work if `bash` and `sh` have been blacklisted. After `noblacklist` `bash` and `sh` the problem went away.
If you wondered about the issue and the proposed fix, can you try it out on your system (if you use Telegram)?

glitsj16:
I don't use telegram. But I can definately see how it would rely on `xdg-open`, like so many apps do, to open URI's in your user's preferred application. And the latter indeed needs bash/sh, there's no confusion about that part. What I cannot (yet) understand is how unblocking `bash` and `sh` can help if they are not also added to `private-bin`. You should need `xdg-open` there too.

Usually I don't mind installing an app I will never use myself for firejail testing purposes, done so many times. As far as I know Telegram requires a phone number for account creation. So to test its hyperlink functionality under firejail I'd have to supply something I personally regard as very private indeed and that's where I draw the testing line. Maybe other collaborators use it, I don't know. Would be helpful if they chimed in :-)

YorkZ:
> As far as I know Telegram requires a phone number for account creation
Fair enough. I've opened the issue #4784.

YorkZ:
@glitsj16 

> The telegram profile uses private-bin telegram,Telegram,telegram-desktop. Allowing ${PATH}/bash and ${PATH}/sh doesn't change that. I'm wondering how this change can fix anything if these aren't added to private-bin too.

You are actually right. The reason this PR worked for me was because in my version (installed with Arch Linux pacman), the `private-bin` command has been commented out:

    #private-bin telegram,Telegram,telegram-desktop
    
But in the git version, the above line was uncommented:

    private-bin telegram,Telegram,telegram-desktop
    
When I uncommented this command, the PR didn't fix the issue.

YorkZ:
Adding `bash` and `sh` to `private-bin` worked. I've updated the commit.

YorkZ:
Can anyone explain to me why "Profile Checks" failed? Thank you.

glitsj16:
Crossing communications... I noticed the differences between the git profile and the one from 0.9.66 too. Bottomline, change the private-bin line to:
`private-bin bash,sh,telegram,Telegram,telegram-desktop,xdg-open`

> I don't know why it tried to execute xdg-open in /usr/local/bin which didn't exist. Any idea?

One thing I noticed that might explain this is this warning on your issue debug log:

```
[...]
Warning: an existing sandbox was detected. /usr/bin/telegram-desktop will run without any additional sandboxing features
[...]
```

That warning usually is an indication that there's been a mix-up in the paths when starting a sandbox. If you use firecfg, that places a telegram symlink under /usr/local/bin for 'ease-of-use' or what's refered to as `desktop integration` in man firejail. During your testing you might have used `$ firejail --debug telegram`, which boils down to trying to sandbox telegram twice (hence the warning). Try adding the full path to the telegram executable: `$ firejail --debug /usr/bin/telegram` and notice the difference.

etc/profile-m-z/telegram.profile
```diff
@@ -8,6 +8,10 @@ include globals.local
 noblacklist ${HOME}/.TelegramDesktop
 noblacklist ${HOME}/.local/share/TelegramDesktop
 
+# Allow opening hyperlinks
```
glitsj16, line 11: Suggestion: use our inc file specifically designed for situations like this (instead of the two noblacklists)

include allow-bin-sh.inc

etc/profile-m-z/telegram.profile
```diff
@@ -41,7 +45,7 @@ seccomp.block-secondary
 shell none
 
 disable-mnt
-private-bin telegram,Telegram,telegram-desktop
+private-bin bash,telegram,Telegram,telegram-desktop,sh
```
glitsj16, line 48: add xdg-open too and respect ordering:
private-bin bash,sh,telegram,Telegram,telegram-desktop,xdg-open

glitsj16 (changes requested):
Just some minor changes needed befor we can commit this.

YorkZ:
Thank you very much for the help @glitsj16, I'm very new to `firejail`.

You are right about the reason of the warning:

> Warning: an existing sandbox was detected. /usr/bin/telegram-desktop will run without any additional sandboxing features

which, as you pointed out, was because I had a symbolic link in `/usr/local/bin` and I tested according to the suggestion of the issue template.

>add xdg-open too

Do we still need to add `xdg-open` to `private-bin` list? I asked because it worked here without having to add it.

> and respect ordering:
> private-bin bash,sh,telegram,Telegram,telegram-desktop,xdg-open

This order worked, but I don't understand the reason behind this order, could you explain?

etc/profile-m-z/telegram.profile
YorkZ: > Suggestion: use our inc file specifically designed for situations like this (instead of the two noblacklists)
> 
> include allow-bin-sh.inc

Nice suggestion, thank you @glitsj16.

glitsj16:
> Thank you very much for the help @glitsj16, I'm very new to firejail.

You're welcome. We've all been new to firejail and it's a complex application (code-wise). So keep asking for help whenever the need arises, that's very wise in regards to a security/privacy related app IMO. Also consult the wiki pages and issue tracker etc. and you'll get the hang of it soon enough.

> This order worked, but I don't understand the reason behind this order, could you explain?

There's no strictlu technical reason for ordering option values `alphabetically`. Just one of these social conventions to help avoiding profile breakage (typo's, doubled entrees, ...) and make things more manageable I guess.

YorkZ:
> There's no strictly technical reason for ordering option values `alphabetically`

Have no idea why I thought `s` was after `t` in `alphabetical` ordering. Thank you.

glitsj16:
> Have no idea why I thought s was after t in alphabetical ordering :-). Thank you.

Heh, happens to all of us. That's exactly the reason for having a CI check on stuff like this :-)

glitsj16:
> Do we still need to add xdg-open to private-bin list? I asked because it worked here without having to add it.

Yep, we need it there too. Run a search in your editor of choice on xdg-open in /etc/firejail/. You'll notice we add it in lots of profiles to guarantee hyperlink functionality.

etc/profile-m-z/telegram.profile
```diff
@@ -41,7 +44,7 @@ seccomp.block-secondary
 shell none
 
 disable-mnt
-private-bin telegram,Telegram,telegram-desktop
+private-bin bash,sh,telegram,Telegram,telegram-desktop
```
glitsj16, line 47: private-bin bash,sh,telegram,Telegram,telegram-desktop,xdg-open

glitsj16 (changes requested):
Just the one xdg-open in private-bin left before we can merge. Thanks!

etc/profile-m-z/telegram.profile
YorkZ: > private-bin bash,sh,telegram,Telegram,telegram-desktop,xdg-open

Done. Thank you.

YorkZ:
> Just the one xdg-open in private-bin left before we can merge. Thanks!

Done. Thank you.

glitsj16 (approved):
LGTM

YorkZ:
@glitsj16 BTW, where do I ask question?

glitsj16:
@YorkZ You can ask questions via the [discussions](https://github.com/netblue30/firejail/discussions) section here on GH. Not a perfectly labelled section but well, that's out of our control :-)

YorkZ:
> @YorkZ You can ask questions via the [discussions](https://github.com/netblue30/firejail/discussions) section here on GH. Not a perfectly labelled section but well, that's out of our control :-)

Very nice, thank you.

etc/profile-m-z/telegram.profile
```diff
@@ -41,7 +44,7 @@ seccomp.block-secondary
 shell none
 
 disable-mnt
-private-bin telegram,Telegram,telegram-desktop
+private-bin bash,sh,telegram,Telegram,telegram-desktop,xdg-open
```
rusty-snake, line 47: Why bash?

xdg-open is a shell-script, consider `sed,xdg-mime,which,mimeopen,grep,egrep,printf,cut,uname,dbus-send,xprop,dirname,cat,…`, `exo-open,gio,gvfs-open,mate-open,enlightenment_open,gnome-open,dde-open,kde-open,kde-open*,kfmclient,cygstart,kde-config,gnome-default-applications-properties,…` or just drop private-bin.

glitsj16:
> Why bash?

Good question and fair points. The private-bin enablement after 0.9.66 caused some confusion as you can see (both here and in the issue). I can only find [this](https://github.com/netblue30/firejail/commit/3492d15b77de60bce87c1e2ae0c9be2db3166823#diff-c3a5169dd043c70b61fc3b488ea507fbf0148e83154227981ec025349afb3d4a) in the history of telegram.profile. No info on why private-bin got enabled exactly, but I assume it was just good hardening practice. @netblue30 Any thoughts?

etc/profile-m-z/telegram.profile
YorkZ: > Why bash?

I believe enabling `bash` is necessary because in Arch Linux, `/usr/bin/sh` has been symlink'ed to `/usr/bin/bash`. Some customized systems also use `bash` as `sh`. I personally frequently use `bash` as `sh` in production systems too because `bash` is so much more featureful than `sh`. In fact, `bash` has a `sh` compatible mode which is probably why I've never had any issue when replacing `sh` with `bash`. Finally, in this case, I tested only enabling `sh`, and confirmed that it didn't work.

rusty-snake: > I believe enabling bash is necessary because in Arch Linux, `/usr/bin/sh` has been symlink'ed to `/usr/bin/bash`

Do not believe, test :wink:.

private-bin follows symlinks IIRC and even if it don't there are other shells commonly used as `/bin/sh` (namely `dash` by Debian and Ubuntu(?)).

> Finally, in this case, I tested only enabling sh, and confirmed that it didn't work.

(Fedora Linux 35; `sh->bash`):

```console
$ firejail --noprofile --private-bin=sh,ls ls -l /usr/bin
bash
ls
sh -> /usr/bin/bash
```

I guess you had tested with only `sh` in `private-bin` and only `noblacklist ${PATH}/sh`. This does not work `blacklist ${PATH}/bash` will blacklist the binary used by `/usr/bin/sh`. FWIW, `blacklist` follows symlinks too.

rusty-snake: > xdg-open is a shell-script, consider …

*On my system* for xdg-open to work at all: `sh,xdg-open,grep,egrep,<A installed well known browser, e.g. firefox>`.
*On my system* for xdg-open to work correct: `sh,xdg-open,grep,egrep,xdg-mime,sed,tr,awk,cut,head,basename,which,readlink,<My browser, e.g. firefox>`
For the firefox start script: `...`

> or just drop private-bin.

yeah, I think this is the way to go.

YorkZ: > I guess you had tested with only `sh` in `private-bin` and only `noblacklist ${PATH}/sh`

That's correct, I think tested with only `sh` in `private-bin` and I only `noblacklist` `sh`.

> there are other shells commonly used as /bin/sh (namely dash by Debian and Ubuntu(?))

I think `dash` is indeed an implementation of POSIX `sh` which is actually `sh`.

kmk3: @rusty-snake [on Dec 19](https://github.com/netblue30/firejail/pull/4783#discussion_r771977694):

> > I believe enabling bash is necessary because in Arch Linux, `/usr/bin/sh`
> > has been symlink'ed to `/usr/bin/bash`
> 
> Do not believe, test wink.
> 
> private-bin follows symlinks IIRC and even if it don't there are other shells
> commonly used as `/bin/sh` (namely `dash` by Debian and Ubuntu(?)).
> 
> > Finally, in this case, I tested only enabling sh, and confirmed that it didn't work.
> 
> (Fedora Linux 35; `sh->bash`):
> 
> ```
> $ firejail --noprofile --private-bin=sh,ls ls -l /usr/bin
> bash
> ls
> sh -> /usr/bin/bash
> ```

This just lists the files; /bin/sh could point to `foobar` and it would not
fail because of that.  On Artix it fails, but I'm not sure why:

```console
$ firejail --quiet --noprofile --private-bin=sh,ls ls -l /usr/bin
```

A better test would be to actually run /bin/sh:

```console
$ firejail --quiet --noprofile --private-bin=sh /bin/sh -c 'echo yes'
$ firejail --quiet --noprofile --private-bin=bash,sh /bin/sh -c 'echo yes'
yes
$ readlink /bin/sh
dash
```

So in my case it does require `private-bin bash`, even though /bin/sh is
/bin/dash.

IIRC this is because firejail hardcodes bash as the default shell.

Related:

* <https://github.com/netblue30/firejail/issues/3434#issuecomment-923552299>

Maybe firejail could just use `readlink /bin/sh` once on startup and consider
that the default shell.

rusty-snake: > This just lists the files;

@kmk3 it clearly shows that the file `/bin/sh` points to is copied/bind-mounted (what does it do?) if you only list `sh` but not `bash` (assuming `/bin/sh` points to somewhere in `{/usr,}/bin`).

rusty-snake: > IIRC this is because firejail hardcodes bash as the default shell.

It don't, read the issues title in your comment:

> - Default shell is guessed from $SHELL, despite manpage specifying /bin/bash

---

> On Artix it fails, but I'm not sure why:

If you get your right (sh->dash; `getent passwd $USER | cut -d: -f7`: bash), you try to start a program (bash) which isn't inside the sandbox.

IMHO `--shell=none` should be the default (if you specify a program).

kmk3: @rusty-snake [on Dec 19](https://github.com/netblue30/firejail/pull/4783#discussion_r771991938):

> > This just lists the files;
> 
> @kmk3 it clearly shows that the file `/bin/sh` points to is
> copied/bind-mounted (what does it do?) if you only list `sh` but not `bash`
> (assuming `/bin/sh` points to somewhere in `{/usr,}/bin`).

I don't understand what you mean.  A symlink points to a (text) path; it
doesn't actually point to an inode.  For example, if /bin/sh points to just
"dash" and you copy the symlink with to /tmp, the copy will not work:

```console
$ readlink /bin/sh
dash
$ /bin/sh -c 'echo yes'
yes
$ cd /tmp
$ cp -P /bin/sh .
$ readlink ./sh
dash
$ ./sh -c 'echo yes'
bash: ./sh: No such file or directory
$ echo 'echo hello world' >dash
$ chmod +x dash
$ ./dash
hello world
$ ./sh
hello world
```

Do you mean that sh points to just "bash" outside of firejail but inside of it
it points to "/usr/bin/bash"?  That might mean that firejail hardcodes its own
/bin/sh symlink inside the sandbox, but whether /bin/bash is bind-mounted or
not is unclear from your example.

rusty-snake: What I want to tell is that `private-bin=sh` will copy/bind-mount `sh` *and* the program it points to (e.g. bash or dash). No matter if you list `bash/dash` or not.

I don't care about `/bin/sh` pointing to somewhere else (e.g. `/mnt/extra-program/mysh`).

YorkZ: I just tried removing `bash` from `private-bin`, and it worked (the link could be opened from Telegram). @kmk3 do you still have problem if removing `bash` from `private-bin`?

kmk3: (Continued on #4790)

PR #4802, opened 2021-12-26
Fix clipgrab profile (yt-dlp requires python)

jose1711:


etc/profile-a-l/clipgrab.profile
```diff
@@ -6,10 +6,14 @@ include clipgrab.local
 # Persistent global definitions
 include globals.local
 
+noblacklist ${HOME}/.config/ClipGrab
```
glitsj16, line 9: Add a counterpart `blacklist ${HOME}/.config/ClipGrab` to disable-programs.inc so the noblacklist makes sense.

glitsj16 (changes requested):
LGTM, just one additional change and we can merge this. Thanks!

glitsj16 (approved):
All good. Thanks!

PR #4807, opened 2021-12-30
skypeforlinux: Whitelist downloads directory

WhyNotHugo:
The downloads directory is used by skype to download files send by other parties.

### DIFF
--- a/etc/profile-m-z/skypeforlinux.profile
+++ b/etc/profile-m-z/skypeforlinux.profile
@@ -6,7 +6,6 @@ include skypeforlinux.local
 include globals.local
 
 # Disabled until someone reported positive feedback
-ignore whitelist ${DOWNLOADS}
 ignore include whitelist-common.inc
 ignore include whitelist-runuser-common.inc
 ignore include whitelist-usr-share-common.inc


etc/profile-m-z/skypeforlinux.profile
```diff
@@ -6,7 +6,6 @@ include skypeforlinux.local
 include globals.local
 
 # Disabled until someone reported positive feedback
-ignore whitelist ${DOWNLOADS}
 ignore include whitelist-common.inc
```
rusty-snake, line 9: We should remove this too.

rusty-snake (approved):

etc/profile-m-z/skypeforlinux.profile
WhyNotHugo: `whitelist-common.inc` grants write access to `.config/dconf`, and this sounds like it could be abused to run arbitrary commands (for example, by changing the default terminal).

rusty-snake: 1. https://github.com/netblue30/firejail/blob/facf03c54108ff7180339f09425c2b3fdc651efd/etc/inc/disable-common.inc#L155
2. Who cares about such hacks for a profile which allows `systemd-run` & co.

netblue30:
Thanks for the fix!

PR #4824, opened 2022-01-05
add notable profile

glitsj16:
See discussion [here](https://github.com/netblue30/firejail/discussions/4812).

### DIFF
--- a/etc/profile-m-z/notable.profile
+++ b/etc/profile-m-z/notable.profile
@@ -1,0 +1,37 @@
+# Firejail profile for notable
+# Description: The Markdown-based note-taking app that doesn't suck
+# This file is overwritten after every install/update
+# Persistent local customizations
+include notable.local
+# Persistent global definitions
+include globals.local
+
+# Note: On debian-based distributions the binary might be located in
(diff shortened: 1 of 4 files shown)


etc/profile-m-z/notable.profile
```diff
@@ -0,0 +1,37 @@
...
+ignore dbus-user none
+
+# Notable keeps claiming it is started for the first time when whitelisting - see #4812.
+ignore whitelist ${DOWNLOADS}
+ignore whitelist ${HOME}/.config/Electron
+ignore whitelist ${HOME}/.config/electron-flag*.conf
+ignore include whitelist-common.inc
+ignore include whitelist-runuser-common.inc
+ignore include whitelist-usr-share-common.inc
+ignore include whitelist-var-common.inc
```
rusty-snake, line 34: Even for wruc, wusc, wvc?

glitsj16: Yep, even those set notable into a fit and it always resets some internal pref so users are faced with a nagging screen on each invocation to select the main data dir (which is ${HOME}/.notable). I've tried to make it a whitelisting profile but didn't succeed. Closed-source app...

PR #4828, opened 2022-01-06
profiles: allow /usr/share/webext in chromium profile

reinerh:
Debian packages can distribute browser extensions in `/usr/share/webext`.
The directory is already allowed in Firefox (and related) profiles, but not yet for chromium.

For now I have added it to the chromium.profile, as this profile also contains whitelists another extension directory (`/usr/share/mozilla/extensions`), but maybe it would make sense to move both to chromium-common.profile?

Bug-Debian: https://bugs.debian.org/1003234

### DIFF
--- a/etc/profile-a-l/chromium-common.profile
+++ b/etc/profile-a-l/chromium-common.profile
@@ -29,6 +29,8 @@ mkdir ${HOME}/.local/share/pki
 whitelist ${DOWNLOADS}
 whitelist ${HOME}/.pki
 whitelist ${HOME}/.local/share/pki
+whitelist /usr/share/mozilla/extensions
+whitelist /usr/share/webext
 include whitelist-common.inc
 include whitelist-run-common.inc
 include whitelist-runuser-common.inc
--- a/etc/profile-a-l/chromium.profile
+++ b/etc/profile-a-l/chromium.profile
@@ -16,8 +16,6 @@ whitelist ${HOME}/.cache/chromium
 whitelist ${HOME}/.config/chromium
 whitelist ${HOME}/.config/chromium-flags.conf
 whitelist /usr/share/chromium
-whitelist /usr/share/mozilla/extensions
-whitelist /usr/share/webext
 
 # private-bin chromium,chromium-browser,chromedriver
 
--- a/etc/profile-a-l/chromium.profile
+++ b/etc/profile-a-l/chromium.profile
@@ -17,6 +17,7 @@ whitelist ${HOME}/.config/chromium
 whitelist ${HOME}/.config/chromium-flags.conf
 whitelist /usr/share/chromium
 whitelist /usr/share/mozilla/extensions
+whitelist /usr/share/webext
 
 # private-bin chromium,chromium-browser,chromedriver
 


glitsj16:
> For now I have added it to the chromium.profile, as this profile also contains whitelists another extension directory (/usr/share/mozilla/extensions), but maybe it would make sense to move both to chromium-common.profile?

Not that I have particularly strong views on this, but we have several profiles that include chromium-common.profile. And I'm not sure if (all) those apps support browser extensions. More of a reminder remark though. I don't think moving both to chromium-common.profile will break anything. Worst-case scenario is another dir exposed in the sandbox that isn't particularly relevant for the app in question. The PR itself is straight-up.

glitsj16 (approved):
LGTM

kmk3 (approved):
LGTM

kmk3:
@reinerh commented [on Jan 6](https://github.com/netblue30/firejail/pull/4828#issue-1095579431):

> Debian packages can distribute browser extensions in `/usr/share/webext`. The
> directory is already allowed in Firefox (and related) profiles, but not yet
> for chromium.
> 
> For now I have added it to the chromium.profile, as this profile also
> contains whitelists another extension directory
> (`/usr/share/mozilla/extensions`), but maybe it would make sense to move both
> to chromium-common.profile?

Agreed.

@glitsj16 commented [on Jan 6](https://github.com/netblue30/firejail/pull/4828#issuecomment-1006841766):

> Not that I have particularly strong views on this, but we have several
> profiles that include chromium-common.profile. And I'm not sure if (all)
> those apps support browser extensions. More of a reminder remark though. I
> don't think moving both to chromium-common.profile will break anything.
> Worst-case scenario is another dir exposed in the sandbox that isn't
> particularly relevant for the app in question. The PR itself is straight-up.

AFAICT most/all appear to be web browsers:

```console
$ git grep -l '^include chromium-common.profile' -- etc
etc/profile-a-l/bnox.profile
etc/profile-a-l/brave.profile
etc/profile-a-l/chromium.profile
etc/profile-a-l/dnox.profile
etc/profile-a-l/enox.profile
etc/profile-a-l/flashpeak-slimjet.profile
etc/profile-a-l/google-chrome-beta.profile
etc/profile-a-l/google-chrome-unstable.profile
etc/profile-a-l/google-chrome.profile
etc/profile-a-l/inox.profile
etc/profile-a-l/iridium.profile
etc/profile-m-z/microsoft-edge-beta.profile
etc/profile-m-z/microsoft-edge-dev.profile
etc/profile-m-z/min.profile
etc/profile-m-z/opera-beta.profile
etc/profile-m-z/opera.profile
etc/profile-m-z/snox.profile
etc/profile-m-z/vivaldi.profile
etc/profile-m-z/yandex-browser.profile
```

In which case I think they probably do (or should) support webextensions.

Allowing the path probably wouldn't make sense for electron apps (which are
usually proprietary and restrictive), but luckily electron.profile does not
include chromium-common.profile.

reinerh:
Okay, I've moved both whitelists to chromium-common.profile.

glitsj16:
I just realised that [min](https://github.com/minbrowser/min) is Electron-based. When I originally added the min.profile I glossed over that fact and based its profile entirely on chromium's... FTR, I build min from git and it's the only non-gecko web browser tolerated on my machine. The dev has been [reluctant](https://github.com/minbrowser/min/issues/1) on adding extension support and it still hasn't. The [min](https://archlinux.org/packages/community/x86_64/min/) package from Arch Linux Community repo depends on electron15 and probably never worked with firejail. I'm in the proces of confirming that and will do a PR based on the outcome.

So I guess there's nothing holding off moving the extensions dir whitelisting to chromium-common.profile.

glitsj16 (commented):
Let's merge, LGTM.

PR #4831, opened 2022-01-07
Blacklist rxvt after the blacklist of Perl.

vinc17fr:
rxvt needs Perl modules, thus does not work. And its blacklist is needed so that Firefox can run applications with `Terminal=true` in their .desktop file (depending on what is installed). The reason is that Firefox does this via glib, which currently uses a [hardcoded list of terminal emulators](https://gitlab.gnome.org/GNOME/glib/-/issues/338). And in this list, rxvt comes before xterm.

### DIFF
--- a/etc/inc/disable-interpreters.inc
+++ b/etc/inc/disable-interpreters.inc
@@ -40,6 +40,15 @@ blacklist /usr/lib/perl*
 blacklist /usr/lib64/perl*
 blacklist /usr/share/perl*
 
+# rxvt needs Perl modules, thus does not work. In particular, blacklisting
+# it is needed so that Firefox can run applications with Terminal=true in
+# their .desktop file (depending on what is installed). The reason is that
+# this is done via glib, which currently uses a hardcoded list of terminal
+# emulators:
+#   https://gitlab.gnome.org/GNOME/glib/-/issues/338
+# And in this list, rxvt comes before xterm.
+blacklist ${PATH}/rxvt
+
 # PHP
 blacklist ${PATH}/php*
 blacklist /usr/lib/php*


netblue30:
Thanks for the fix!

kmk3 (commented):
The reverse code (`noblacklist ${PATH}/rxvt`) should probably be added to
allow-perl.inc or else rxvt will always be blocked.

kmk3:
(Offtopic rant)

@vinc17fr commented [on Jan 6](https://github.com/netblue30/firejail/pull/4831#issue-1095923989):

> rxvt needs Perl modules, thus does not work. And its blacklist is needed so
> that Firefox can run applications with `Terminal=true` in their .desktop file
> (depending on what is installed). The reason is that Firefox does this via
> glib, which currently uses a [hardcoded list of terminal
> emulators](https://gitlab.gnome.org/GNOME/glib/-/issues/338). And in this
> list, rxvt comes before xterm.

The state of that bug is very unfortunate.  A hardcoded list is used, but MRs
that add common missing terminal emulators are rejected because the MRs are
considered to be a "workarounds" (but arguably so is the list itself).  Why
even have a list at all if it's going to be deliberately unmaintained?  Also,
the bug is over 11 years old and the accepted approaches seem unlikely to be
implemented any time soon.

It might be better for Firefox to ignore/work around glib and just source the
terminal emulator from `$TERM`.

vinc17fr:
> The reverse code (`noblacklist ${PATH}/rxvt`) should probably be added to allow-perl.inc or else rxvt will always be blocked.

Yes, this would probably make sense (but on the other hand, if I understand correctly, there is no way to disable a `noblacklist` if a user would really want to allow perl, but blacklist rxvt, though I think that this is rather unlikely).

BTW, I think that the existence of these `allow-*.inc` files should be documented for each interpreter in `disable-interpreters.inc`.

About the Firefox/glib bug, yet, this is very unfortunate. But anyway, the user may also want to test in his own scripts (e.g. reused on different machines) whether some program is available, and leaving a program that doesn't work in the path would be rather bad.

rusty-snake:
>  if I understand correctly, there is no way to disable a noblacklist if a user would really want to allow perl, but blacklist rxvt, 

You can `blacklist ${PATH}/rxvt` or `ignore noblacklist ${PATH}/rxvt`.

vinc17fr:
> You can `blacklist ${PATH}/rxvt` or `ignore noblacklist ${PATH}/rxvt`.

OK, I got confused by the documentation: since the firejail(1) says for `--noblacklist` "Disable blacklist for this directory or file.", I thought that `noblacklist` always had the precedence over `blacklist`. But with

    firejail --blacklist='${PATH}/rxvt' --noblacklist='${PATH}/rxvt' --noprofile rxvt

one can see that rxvt is really blacklisted.

PR #4832, opened 2022-01-07
Add DBus filter rules specific to firefox-developer-edition

ammongit:
Firefox Developer Edition has a slightly different DBus path than stock Firefox. Without these filters, a firejailed browser instance cannot permit other programs to open tabs within it.


rusty-snake (approved):

glitsj16 (approved):
LGTM

Issue #4839, opened 2022-01-07
Fix Mediathekview


Spacewalker2:
Hello,

I added the stuff below to my `mediathekview.profile` and it started working again.

```
ignore noexec /tmp

noblacklist ${HOME}/.openjfx
read-only ${HOME}/.openjfx
whitelist ${HOME}/.mediathek3
whitelist ${DOWNLOADS}
```

kmk3:
@Spacewalker2 commented [on Jan 7](https://github.com/netblue30/firejail/issues/4839#issue-1096675377):

> Hello,
> 
> I added the stuff below to my `mediathekview.profile` and it started working
> again.
> 
> ```
> ignore noexec /tmp
> 
> noblacklist ${HOME}/.openjfx
> read-only ${HOME}/.openjfx
> whitelist ${HOME}/.mediathek3
> whitelist ${DOWNLOADS}
> ```

Nice.  Would you like to open a pull request for that?

I'm not sure about `whitelist ${DOWNLOADS}`, but the rest seems alright.

glitsj16:
Odd, we don't blacklist ${HOME}/.openjfx anywhere AFAICT. Is that something you have added to a globals.local or any of the disable-foo.local files?

Spacewalker2:
You are right. It works without the OpenJFX lines. My fault this was probably a leftover from testing the profile sorry. About `whitelist ${DOWNLOADS}` your are right this isn't necessary either but I thought since MediathekView is also for downloading stuff having your download folder around couldn't hurt. I would have probably created a pull request in the first place but my branching skill could be more sophisticated. I would give it a try if you want but hopefully I don't create a mess.

I guess this reduces everything to two line then.

```
ignore noexec /tmp
whitelist ${HOME}/.mediathek3
```

glitsj16:
That happens, no poblem. And I personally understand whitelisting ${DOWNLOADS} with an app like this, but that's something you can easily do via mediathekview.local.

> I would have probably created a pull request in the first place but my branching skill could be more sophisticated. I would give it a try if you want but hopefully I don't create a mess.

[...]

PR #4840, opened 2022-01-08
Fix MediathekView Profile

Spacewalker2:
Related to issue #4839.

### DIFF
--- a/etc/profile-m-z/mediathekview.profile
+++ b/etc/profile-m-z/mediathekview.profile
@@ -17,6 +17,8 @@ noblacklist ${HOME}/.mediathek3
 noblacklist ${HOME}/.mplayer
 noblacklist ${VIDEOS}
 
+ignore noexec /tmp
+
 # Allow java (blacklisted by disable-devel.inc)
 include allow-java.inc
 
@@ -27,6 +29,7 @@ include disable-interpreters.inc
 include disable-programs.inc
 include disable-xdg.inc
 
+whitelist ${HOME}/.mediathek3
 include whitelist-var-common.inc
 
 caps.drop all
--- a/etc/profile-m-z/mediathekview.profile
+++ b/etc/profile-m-z/mediathekview.profile
@@ -29,6 +29,7 @@ include disable-interpreters.inc
 include disable-programs.inc
 include disable-xdg.inc
 
+mkdir ${HOME}/.mediathek3
 whitelist ${HOME}/.mediathek3
 include whitelist-var-common.inc
 


etc/profile-m-z/mediathekview.profile
```diff
@@ -27,6 +29,7 @@ include disable-interpreters.inc
 include disable-programs.inc
 include disable-xdg.inc
 
+whitelist ${HOME}/.mediathek3
```
glitsj16, line 32: Suggestion: add this just above the whitelist
`mkdir ${HOME}/.mediathek3`

etc/profile-m-z/mediathekview.profile
```diff
@@ -27,6 +29,7 @@ include disable-interpreters.inc
...
 include whitelist-var-common.inc
```
glitsj16, line 33: Does it still work if we add the below to the profile? If not, leave them out. We can always try to tighten the profile in a follow-up PR.

include whitelist-common.inc

glitsj16 (changes requested):
Looks OK. Just one nitpick suggestion about the mkdir...

etc/profile-m-z/mediathekview.profile
rusty-snake: We should also (no)blacklist this path.

etc/profile-m-z/mediathekview.profile
rusty-snake: If we start whitelisting, we should include wc.

etc/profile-m-z/mediathekview.profile
```diff
@@ -27,6 +29,7 @@ include disable-interpreters.inc
 include disable-programs.inc
```
glitsj16, line 29: Can we add `include disable-shell.inc` without breaking things? Like the comment below on the other potential includes, just test, and if it breaks mediathekview leave it out.

etc/profile-m-z/mediathekview.profile
glitsj16: @rusty-snake We already do so...

rusty-snake: Ohh yes, but then we should whitelist all noblacklisted paths, otherwise it is a bit useless to noblacklist them.

netblue30: I'll do:

add blacklist ${HOME}/.mediathek3 in disable-programs.inc

noblacklist ${HOME}/.mediathek3

etc/profile-m-z/mediathekview.profile
netblue30: Goes in a new pr. For now I'll pull this in to get it out of the way.

etc/profile-m-z/mediathekview.profile
netblue30: New pr.

Issue #4845, opened 2022-01-11
Options to deal with open file descriptors


smitsohu:
Currently Firejail sandboxes inherit all open file descriptors from their parent process.

As far as standard streams (stdin, stdout, stderr) are concerned this is probably desired behaviour. On the other hand, passing non-standard file descriptors (> 2) into the sandbox is only rarely necessary in my experience.

IMHO Firejail should provide a mechanism to prevent accidental leakage of open file descriptors, either by default with a way to opt-out, or alternatively as an opt-in. I would like to know if there is general interest in options like

`inheritfd` (don't close one or more file descriptors > 2)
`discard-stdin` (-> /dev/null)
`discard-stdout` (-> /dev/null)
`discard-stderr` (-> /dev/null)

rusty-snake:
Is there anything bad expect that you can read/write on that fd?

> Just out of curiosity, could you provide an example of such a case?

`firejail --noprofile --shell=none --private-bin=gnome-2048 /proc/self/fd/5 -l /usr/bin 5</usr/bin/ls`

reinerh:
> Is there anything bad expect that you can read/write on that fd?

Open FDs are bad in simple chroots, as you can use them to escape with `fchdir(fd)` (if `fd` is outside). There is probably other bad stuff you can do with open FDs to get information from outside the jail.

Not sure how firejail behaves with `--chroot`. It's probably best to close everything.

rusty-snake:
Yeah, if the open fd refers to a directory you can `fchdir`, `openat`, `execveat`, ... which is close to full filesystem access. But if it refers to a file? Is there something like `get_basedirfd_of_fd`? If anyone knows more, I'm interested.

smitsohu:
> Is there anything bad expect that you can read/write on that fd?

It depends on the file descriptor. As a rule of thumb everything that can be done with a file descriptor outside the sandbox can be done also inside the sandbox. This includes reading/writing to files or interprocess communication.

Many file descriptors (those that are obtained by actually opening a file) are bound to a mount point, and inherited file descriptors necessarily refer to mount points outside the sandbox. So a worst case scenario would be a file descriptor that allows to bypass a `blacklist` or `read-only` directive, or `fchdir` out of the sandbox entirely.

Pythons `subprocess` module by the way closes all non-standard file descriptors by default, or at least that is my understanding. If this is ok for a general purpose Python module, I think it should be okay'ish for Firejail as well.

[...]

PR #4856, opened 2022-01-14
keep-fd option (#4845)

smitsohu:
(I had to move setrlimit code because I am using an open /proc/self/fd file descriptor to close everything else, and it was getting in the way of `rlimit-nofile`)

Stuff related to stdin, stdout, sterr is supposed to follow.


netblue30:
merged, thanks!

PR #4862, opened 2022-01-17
steam.profile: allow /etc/vulkan

antonv6:
AMD Open Source Driver For Vulkan (amdvlk) installs ICD files to /etc/vulkan.

### DIFF
--- a/etc/profile-m-z/steam.profile
+++ b/etc/profile-m-z/steam.profile
@@ -157,7 +157,7 @@ shell none
 private-dev
 # private-etc breaks a small selection of games on some systems. Add 'ignore private-etc'
 # to your steam.local to support those.
-private-etc alsa,alternatives,asound.conf,bumblebee,ca-certificates,crypto-policies,dbus-1,drirc,fonts,group,gtk-2.0,gtk-3.0,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,localtime,lsb-release,machine-id,mime.types,nvidia,os-release,passwd,pki,pulse,resolv.conf,services,ssl
+private-etc alsa,alternatives,asound.conf,bumblebee,ca-certificates,crypto-policies,dbus-1,drirc,fonts,group,gtk-2.0,gtk-3.0,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,localtime,lsb-release,machine-id,mime.types,nvidia,os-release,passwd,pki,pulse,resolv.conf,services,ssl,vulkan
 private-tmp
 
 # dbus-user none


kmk3 (approved):
LGTM.

I'll add it to other relevant profiles later.

By the way, what happens if it cannot access the path?

antonv6:
> By the way, what happens if it cannot access the path?

Depends on how you switch the drivers. Under the hood you're still using ICD files to pick the right .so file with, so if you cannot access that file, well, I'm not sure what you can do. But there are 2 ways to switch drivers with an environment variable.

If you use AMD_VULKAN_ICD=AMDVLK, it will simply skip to the next available driver (e.g. RADV), because it won't find AMDVLK. So forcing a driver like this is just confusing because it won't work without /etc/vulkan.

If you use VK_ICD_FILENAMES, which lets you directly specify the .icd file (e.g. `VK_ICD_FILENAMES=/etc/vulkan/icd.d/amd_icd64.json vkcube`), then you get
```
vkEnumerateInstanceExtensionProperties failed to find the VK_KHR_surface extension.

Do you have a compatible Vulkan installable client driver (ICD) installed?
Please look at the Getting Started guide for additional information.
```

Mesa's RADV doesn't need /etc/vulkan because it puts the ICD file in a different place: `VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/radeon_icd.x86_64.json vkcube`.

kmk3:
@antonv6 commented [on Jan 18](https://github.com/netblue30/firejail/pull/4862#issuecomment-1015059095):

> > By the way, what happens if it cannot access the path?
> 
> Depends on how you switch the drivers. Under the hood you're still using ICD
> files to pick the right .so file, so if you cannot access that file, well,
> I'm not sure what you can do. But there are 2 ways to switch drivers with an
> environment variable.
> 
> If you use AMD_VULKAN_ICD=AMDVLK, it will simply skip to the next available
> driver (e.g. RADV), because it won't find AMDVLK. So forcing a driver like
> this is just confusing because it won't work without /etc/vulkan.
> 
> If you use VK_ICD_FILENAMES, which lets you directly specify the ICD file
> (e.g. `VK_ICD_FILENAMES=/etc/vulkan/icd.d/amd_icd64.json vkcube`), then you
> get
> 
> ```
> vkEnumerateInstanceExtensionProperties failed to find the VK_KHR_surface extension.
> 
> Do you have a compatible Vulkan installable client driver (ICD) installed?
> Please look at the Getting Started guide for additional information.
> ```
> 
> Mesa's RADV doesn't need /etc/vulkan because it puts the ICD file in a
> different place:
> `VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/radeon_icd.x86_64.json vkcube`.

Good to know about these options; thanks for the detailed explanation.

PR #4863, opened 2022-01-17
{lutris,wine}.profile: allow ~/.cache/wine

antonv6:
~/.cache/wine is a directory where wine stores .msi files for wine-gecko and wine-mono that it may download (with user's permission) and reuse every time a new prefix is created.

### DIFF
--- a/etc/profile-m-z/wine.profile
+++ b/etc/profile-m-z/wine.profile
@@ -6,6 +6,7 @@ include wine.local
 # Persistent global definitions
 include globals.local
 
+noblacklist ${HOME}/.cache/wine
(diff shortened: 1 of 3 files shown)


etc/profile-m-z/wine.profile
```diff
@@ -6,6 +6,7 @@ include wine.local
 # Persistent global definitions
 include globals.local
 
+noblacklist ${HOME}/.cache/wine
```
rusty-snake, line 9: Never blacklisted.

antonv6: It is now.

netblue30:
merged, thanks!

PR #4864, opened 2022-01-17
steam.profile: allow ~/.config/MangoHud

antonv6:
MangoHud is a Vulkan and OpenGL overlay for monitoring FPS, temperatures, CPU/GPU load and more, and it can be configured by user in ~/.config/MangoHud/MangoHud.conf.

This probably could also go into profiles for lutris and wine... But also for every game as well. So maybe ~/.config/MangoHud can be allowed in general in one of the whitelist-*.inc files instead?

### DIFF
--- a/etc/profile-m-z/steam.profile
+++ b/etc/profile-m-z/steam.profile
@@ -8,6 +8,7 @@ include globals.local
 
 noblacklist ${HOME}/.config/Epic
 noblacklist ${HOME}/.config/Loop_Hero
+noblacklist ${HOME}/.config/MangoHud
 noblacklist ${HOME}/.config/ModTheSpire
 noblacklist ${HOME}/.config/RogueLegacy
 noblacklist ${HOME}/.config/RogueLegacyStorageContainer
@@ -55,6 +56,7 @@ include disable-programs.inc
 
 mkdir ${HOME}/.config/Epic
 mkdir ${HOME}/.config/Loop_Hero
+mkdir ${HOME}/.config/MangoHud
 mkdir ${HOME}/.config/ModTheSpire
 mkdir ${HOME}/.config/RogueLegacy
 mkdir ${HOME}/.config/unity3d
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/steam.profile
```diff
@@ -8,6 +8,7 @@ include globals.local
 
 noblacklist ${HOME}/.config/Epic
 noblacklist ${HOME}/.config/Loop_Hero
+noblacklist ${HOME}/.config/MangoHud
```
rusty-snake, line 11: Never blacklisted.

SkewedZeppelin:
perhaps also use `whitelist-ro`?

rusty-snake:
> perhaps also use `whitelist-ro`?

:+1: It's turing complete.


----

Just to quote me again:

https://github.com/netblue30/firejail/pull/4740#pullrequestreview-822995437

> IMHO `--whitelist-ro` should be implemented too.
> 
> > If you add a new command, here's the checklist:
> > 
> > * [ ]  Update manpages: firejail(1) and firejail-profile(5)
> > * [ ]  Update shell completions
> > * [ ]  Update vim syntax files
> > * [ ]  Update --help

kmk3:
@SkewedZeppelin commented [on Jan 22](https://github.com/netblue30/firejail/pull/4864#issuecomment-1019324306):

> perhaps also use `whitelist-ro`?

Can we just treat it as an undocumented feature and avoid using it in profiles,
at least in this release?

I think that aliases should be added sparingly (e.g.: being able to just grep
for `read-only foo` vs also having to grep for `whitelist-ro foo`) and that
there are command combinations that make more sense to become aliases
(considering how common their occurrences are).  I'll expand on this on #4740.

SkewedZeppelin:
I don't really care how it is done.
mangohud is ld_preloaded into whatever uses it, its config should be read-only.

antonv6:
I went with `read-only` because I saw examples in profiles shipped with 0.9.66.

netblue30:
Merged, thanks!

Issue #3491, opened 2020-07-04
Firejail breaks daemonizing processes


NetSysFire:
The following snippet works without problems when executed without firejail. However, when run with firejail (`firejail bash snippet.sh`), it keeps this process in the foreground.
This causes problems with the SSH connection multiplexing feature and causes ansible and even git over SSH to hang, because SSH can not daemonize and lingers indefinitely.

```bash
#!/bin/bash

some_loop(){
  while true; do
    echo "looped"
    sleep 3
  done
}

echo "some interactive stuff"
echo -n "enter something: "
read -r spam
echo "you entered $spam"
echo "forking loop and exiting"
(some_loop&)
```

Related: https://github.com/netblue30/firejail/issues/1518

rusty-snake:
Duplicate of #2433?

glitsj16:
@rusty-snake I was talking with the OP on IRC about this for quite a while. I think there are a few other potentially related issues. The fact there's seemingly something keeping firejail from backgrounding processes might be the root cause or the common-denominator for issues that go beyond SSH and git. I'll add proper links to those i could find as soon as i can track them down again.

DanSman:
Might be not the right issues here? I have problems with ansible. Might be a similar problem to the other issues. 

Unlink `/usr/local/bin/ssh` works but if I'm using firejail then I can't use Ansible. The `Gathering Facts` already can't reach the hosts:  
`{"changed": false, "msg": "Failed to connect to the host via ssh: ", "unreachable": true}`.

But I can connect just fine with ssh.

ndorf:
To daemonize, a process must sever its relationship with its parent (e.g, the shell). This is done by forking and exiting; the new, orphaned process is usually assigned PID 1 (init) as a parent.

Under firejail, it seems that orphaned processes are instead inherited by the same firejail process that spawned their parents, so it's not possible to "fork away" from your parent -- you'll just inherit the same one again.

Could this possibly be solved by having a second firejail process, that itself is daemonized (detached from both its original parent process and the tty), and inherits orphaned processes instead of the main one?

itoffshore:
I came across this issue with `ansible` - I did notice that connections to `localhost` worked & it was only remote `ssh` connections that failed.

Mentioning here in case helpful as the [other issue](https://github.com/netblue30/firejail/issues/1518#issuecomment-955721870) is closed.

kmk3:
See the following thread for a potential workaround:

[...]

Issue #1518, opened 2017-09-02
ControlSocket failed behaviour with ssh


xundeenergie:
Hi there!

I want to git push my repo to github.com, and i have ssh in firejail with standard-profile.
git hangs in case of hanging ssh.

ssh has a ControlMaster for multiplexed ssh-connections. And i have more than one github-repo which i update in a script one after another. So a Control-Socket is a good way, to reuse the connection.

With firejail, the first connection starts the ControlSocket, and the ssh-session for the ControlSocket (the Master) stays in foreground, so the first ssh-process stays in foreground and github cannot come to an end, because it waits for a finnished ssh-process.

If i start in another terminal another git push, it will finnish as expected.
I also can start a "ssh git@github.com" in a second terminal first, then my script with severals git push-comands will finnish.

My ~/.ssh/config contains:
```
Host *
ControlMaster auto
ControlPath /tmp/ssh-%r@%h:%p
```

But i don't know, if there is an option to send the master-connection to background.
Without firejail, the multiplexed connections will run without problems, the master connection goes to background.

chiraag-nataraj:
@xundeenergie Is this still an issue?

chiraag-nataraj:
I did a different test for various reasons:

1. `sudo ln -s /usr/local/bin/firejail /usr/local/bin/ssh`
2. `sshfs user@hostname /path/to/mountpoint`
3. Verify that `ssh` is being run through firejail with `firejail --list` (it is)
4. `ssh user@hostname` to test if the connection is reused
5. Note that no authentication is requested and I'm automatically logged in

It seems like the connection is re-used, and `ssh` is backgrounded by the `sshfs` process. I don't know if this is completely equivalent to the scenario listed here though?

chiraag-nataraj:
So I followed the same process with `git` and had no issues. I linked `/usr/local/bin/ssh -> /usr/local/bin/firejail` and ran `git push` normally and didn't run into any issues. I'm going to close this for now, but @xundeenergie, please feel free to re-open if you still have this issue.

mtth-bfft:
Issue is 100% reproducible, and could be renamed "Firejail makes ssh with control sockets hard to use (and breaks Ansible)". For example, see the time it takes to get a prompt back, corresponding to the control process timeout:

```
user@localhost$ time ssh -C -o ControlMaster=auto -o ControlPersist=60s -o ControlPath=/tmp/test admin@something '/bin/sh -c "echo hello"'
[...]

PR #4870, opened 2022-01-19
profiles: enable deterministic shutdown for ssh

reinerh:
ssh can start in master mode, which will spawn an additional long
running process, which keeps connections to a server open, so that
it can be reused by later connection attempts.

But the lingering master process will prevent the jail from shutting
down, when `firejail ssh` tries to exit.
This breaks for example ansible when using a firejailed ssh, as it
calls ssh with ControlMaster flags.

deterministic-shutdown will kill the other process when the parent
exits.

---

does anyone see something negative with enabling this flag?

### DIFF
--- a/etc/profile-m-z/ssh.profile
+++ b/etc/profile-m-z/ssh.profile
@@ -50,4 +50,5 @@ writable-run-user
 dbus-user none
 dbus-system none
 
+deterministic-shutdown
 memory-deny-write-execute


rusty-snake (approved):

rusty-snake:
Maybe related #3491

reinerh:
Also related: #1518, #4440

kmk3 (commented):
@reinerh commented [on Jan 19](https://github.com/netblue30/firejail/pull/4870#issue-1108597125):

> ssh can start in master mode, which will spawn an additional long running
> process, which keeps connections to a server open, so that it can be reused
> by later connection attempts.
> 
> But the lingering master process will prevent the jail from shutting down,
> when `firejail ssh` tries to exit.
> This breaks for example ansible when using a firejailed ssh, as it calls ssh
> with ControlMaster flags.
> 
> deterministic-shutdown will kill the other process when the parent exits.
> 
> does anyone see something negative with enabling this flag?

No idea, but I think it makes sense to add it.

glitsj16:
I've been running with deterministic-exit-code in my globals.local from the moment it came in. So far I haven't seen any negative side-effects. So FWIW I see nothing against adding d-e-c to the ssh profile.

glitsj16 (approved):
All good.

rusty-snake:
> I've been running with deterministic-exit-code in my globals.local from the moment it came in. So far I haven't seen any negative side-effects. So FWIW I see nothing against adding d-e-c to the ssh profile.

`deterministic-shutdown`

glitsj16:
> `deterministic-shutdown`

Now I'm going blind besides deaf & dumb arghhhhh... Ah well, it's on record as of now :-)

glitsj16:
> does anyone see something negative with enabling this flag?

@reinerh After my initial mixup with deterministic-exit-code I took a step back to test all my apps/shell scripts that involve SSH. For me deterministic-shutdown breaks SSH tunneling (which I happen to use a lot). Easily fixed by adding `ignore deterministic-shutdown` to my ssh.local. But I wonder how many setups this will break (unexpectedly) after the next future upgrade comes down through their package manager. Not sure if it needs to be reverted though, the explanation for adding it seems very reasonable. Perhaps we need a comment to inform users and/or a specific mention of this potential issue in the release notes for 0.9.68. Thoughts?

reinerh:
@glitsj16 do you know what exactly breaks in your tunnel? are you using `-f` to background it maybe?

glitsj16:
@reinerh What breaks is the connection :-) Indeed, I'm using `-f` to background it. Been doing that forever and it never let me down before. According to my notes I got inspired by [this gist](https://gist.github.com/scy/6781836). Do you happen to know another way?

Thanks for your reply. I've got to go out IRL for the rest of the evening. But I'll dive into this tomorrow and report back here.

reinerh:
@glitsj16 I just reproduced it, with `deterministic-shutdown` backgrounding a tunnel is also not working for me. But when I disable `deterministic-shutdown`, it still does not behave as one would expect. `ssh -fND 1234 $server` does NOT go to background here.
So somehow the `-f` has no effect when firejail keeps running processes open.

rusty-snake:
#3491 ?

glitsj16:
@reinerh @rusty-snake @kmk3 Should we open a seperate issue/discussion on this? Might be too late for 0.9.68, but IMO we need to fix these daemonizing issues properly if at all possible. Just don't want to see this bitrot because the conversation is on a MR...

glitsj16:
> I just reproduced it, with deterministic-shutdown backgrounding a tunnel is also not working for me. But when I disable deterministic-shutdown, it still does not behave as one would expect. ssh -fND 1234 $server does NOT go to background here.
So somehow the -f has no effect when firejail keeps running processes open.

@reinerh Correct. It only works for me because I background the ssh call myself in a wrapper script. Hacky but at least that way I can firejail ssh tunneling.

Issue #4780, opened 2021-12-16
private-cwd leaks access to the entire filesystem


WhyNotHugo:
### Description

Using `firejail --private --private-cwd=. /usr/bin/sh` leaks access to the entire filesystem.

### Steps to Reproduce

1. `cd` into some subdirectory of `$HOME`.
2. `firejail --private --private-cwd=. /usr/bin/sh
3. `ls ../../some-file` (for a relative path that exists). `cat` also works.

### Expected behavior

Access to these files should not be possible.

### Actual behavior

Access to files using relative paths is permitted, allowing access to the entire filesystem.

Note: changing directory into those locations does not work, but reading files without changing directory does.

### Behavior without a profile

n/a

### Additional context

-

### Environment

```
➜ uname -sro
Linux 5.15.8-arch1-1 GNU/Linux

➜ firejail --version
firejail version 0.9.67

Compile time support:
	- always force nonewprivs support is disabled
	- AppArmor support is enabled
	- AppImage support is enabled
	- chroot support is enabled
	- D-BUS proxy support is enabled
	- file transfer support is enabled
	- firetunnel support is enabled
	- networking support is enabled
	- output logging is enabled
	- overlayfs support is disabled
	- private-home support is enabled
	- private-cache and tmpfs as user enabled
	- SELinux support is disabled
	- user namespace support is enabled
	- X11 sandboxing support is enabled
```

### Checklist

(removed some non-applicable items)

- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.

glitsj16:
@WhyNotHugo Nice find! Thanks for reporting.

@rusty-snake I'm new to marking issues to a `milestone`. Any thoughts on adding this to [0.9.68](https://github.com/netblue30/firejail/milestone/1)? Even when this isn't fixed by then (whenever it comes), it's a nice way to keep track of things.

netblue30:
Fixed: https://github.com/netblue30/firejail/commit/d2e10f8b728eb83f05c1c57cf06a28a6cd48f58f

Very cool bug, thanks!

WhyNotHugo:
The command in the example above no longer works:

```console
$ firejail --private --private-cwd=. /usr/bin/sh
Error: invalid private working directory
```

kmk3:
[...]

PR #4748, opened 2021-12-07
README.md: Mention security situation on Ubuntu and recommend PPA

kmk3:
Add the information posted by @reinerh on #4666 (related to
CVE-2021-26910 and Ubuntu's security policy) and also the instructions
from #4663 for installing from the PPA.

See also https://bugs.launchpad.net/ubuntu/+source/firejail/+bug/1916767

PR #4889, opened 2022-02-01
RELNOTES: add security-related items

kmk3:
Note: #4780 was closed by commit d2e10f8b7 ("fix --private-cwd problem",
2021-12-19).

Relates to #4748 #4780.

Cc: @WhyNotHugo

### DIFF
--- a/RELNOTES
+++ b/RELNOTES
@@ -1,5 +1,9 @@
 firejail (0.9.68rc2) baseline; urgency=low
   * work in progress
+  * security: on Ubuntu, the PPA is now recommended over the distro package
+    (see README.md) (#4748)
+  * security: bugfix: private-cwd leaks access to the entire filesystem
+    (#4780); reported by Hugo Osvaldo Barrera
   * exit code: distinguish fatal signals by adding 128 (#4533)
   * close file descriptors greater than 2 (--keep-fd) (#4845)
   * intrusion detection system (--ids-init, --ids-check)


RELNOTES
```diff
@@ -1,5 +1,10 @@
 firejail (0.9.68rc2) baseline; urgency=low
   * work in progress
+  * security: the firejail package on Ubuntu was left vulnerable to
```
reinerh, line 3: i'm wondering if this should really go to the release notes. it's nothing specific to this release. and users obtaining the software via source or their distro probably don't care about that.

RELNOTES
```diff
@@ -1,5 +1,10 @@
...
+    CVE-2021-26910 for months after a patch for it was made available; the PPA
+    is now recommended on README.md instead (#4748)
```
reinerh, line 5: I don't understand the `on README.md` in this sentence. Did you add this by accident?

RELNOTES
rusty-snake: I had the same concern we reading it.

RELNOTES
kmk3: @reinerh [on Feb 1](https://github.com/netblue30/firejail/pull/4889#discussion_r797061501):

> I don't understand the `on README.md` in this sentence. Did you add this by
> accident?

I tried to make the entry look consistent by itself and also look similar to
what the PR says.  But indeed it looks better without that part; removed.

RELNOTES
kmk3: @reinerh [on Feb 1](https://github.com/netblue30/firejail/pull/4889#discussion_r797060998):

> i'm wondering if this should really go to the release notes. it's nothing
> specific to this release. and users obtaining the software via source or
> their distro probably don't care about that.

The CVE seems like an important issue and I couldn't think of a better way to
try to make users aware of how security is handled in the package vs in the
PPA.  Users that already installed the distro package might just keep using it
without even knowing that the PPA exists.  If some users know that and don't
care as you say that's fine; but for someone to decide whether they care about
something or not, they need to be aware of it in the first place.

This information is currently only on README.md, which is something that I
wouldn't expect anyone to check periodically or to even read more than once.

Personally, for the projects I'm interested in, I usually don't go around
checking the README/issues/discussions on every release, but I might read the
release notes/announcement, which is where I would hope to find at least the
breaking changes.  And I think it would be nice to have important suggestions
in there as well.

> it's nothing specific to this release.

Well, it's not a code change, but the PPA recommendation was added to README.md
in this release.  I think of this entry as a documentation change + a
security-related (and packaging-related) announcement.

kmk3: Is the issue mostly that the entry is verbose?

If so, what if it was reduced to:

```
  * security: on Ubuntu, the PPA is now recommended over the distro package
    (see README.md) (#4748)
```

reinerh: I like your suggested wording much better. The old one sounds (in my opinion) a bit too accusing.

reinerh (approved):

RELNOTES
kmk3: @reinerh [on Feb 2](https://github.com/netblue30/firejail/pull/4889#discussion_r797932049):

> I like your suggested wording much better. The old one sounds (in my opinion)
> a bit too accusing.

Yes; I just wanted to make the entry self-contained (that is, not vague) and I
had attempted to simplify and write it in different ways, but all of them also
ended up sounding bad, so I just went with the original matter-of-fact way.

Anyway, changed to the new text.

Issue #4891, opened 2022-02-02
LibreWolf - Can't launch additional WebPages into New Tabs via Firejail Command


Lonniebiz:
firejail version 0.9.64.4
Debian 11
LibreWolf [v96.0.3-2](https://gitlab.com/librewolf-community/browser/appimage/-/releases/v96.0.3-2)

This issue is similar to [one reported for Firefox](https://github.com/netblue30/firejail/issues/4670).

A list of LibreWolf AppImages can be found [here](https://gitlab.com/librewolf-community/browser/appimage/-/releases). Here's the command I'm using to run LibreWolf:
`firejail --appimage --profile=/etc/firejail/librewolf.profile ~/AppImages/LibreWolf.x86_64.AppImage`

The problem occurs after first launch, when I'm trying to open additional web pages. For example, let's say LibreWolf is already running, and I want to open [Ebay](https://www.ebay.com) from the command line:
`firejail --appimage --profile=/etc/firejail/librewolf.profile ~/AppImages/LibreWolf.x86_64.AppImage https://www.ebay.com`

Instead of opening Ebay into a new tab, I get this error:
![LW](https://user-images.githubusercontent.com/151919/152204433-8623598b-5dc9-42ed-9967-75af612cc128.png)

In a prior AppImage of LibreWolf, [version 96.0.2-1](https://gitlab.com/librewolf-community/browser/appimage/-/releases/v96.0.2-1), I was able to get past this issue using a custom profile:

**~/.config/firejail/librewolf.profile**
```
# Firejail profile for Librewolf
# Description: Firefox fork based on privacy
# This file is overwritten after every install/update
# Persistent local customizations
include librewolf.local
# Persistent global definitions
include globals.local

noblacklist ${HOME}/.cache/librewolf
noblacklist ${HOME}/.librewolf

mkdir ${HOME}/.cache/librewolf
mkdir ${HOME}/.librewolf
whitelist ${HOME}/.cache/librewolf
whitelist ${HOME}/.librewolf

whitelist /usr/share/doc
whitelist /usr/share/gtk-doc/html
whitelist /usr/share/mozilla
whitelist /usr/share/webext
include whitelist-usr-share-common.inc

dbus-user filter
dbus-user.own org.mozilla.librewolf.*
ignore dbus-user none

# Redirect
include firefox-common.profile
```
[...]

PR #4897, opened 2022-02-04
Update librewolf.profile: use new d-bus message bus

Eklektisk:
Starting Librewolf 96.0, Librewolf switched from using d-bus `org.mozilla.librewolf.*` to `io.gitlab.librewolf.*` for communicating with open browser sessions.

### DIFF
--- a/etc/profile-a-l/librewolf.profile
+++ b/etc/profile-a-l/librewolf.profile
@@ -37,7 +37,7 @@ include whitelist-usr-share-common.inc
 #private-etc librewolf
 
 dbus-user filter
-dbus-user.own org.mozilla.librewolf.*
+dbus-user.own io.gitlab.librewolf.*
 # Add the next line to your librewolf.local to enable native notifications.
 #dbus-user.talk org.freedesktop.Notifications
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/librewolf.profile
```diff
@@ -37,7 +37,7 @@ include whitelist-usr-share-common.inc
 #private-etc librewolf
 
 dbus-user filter
-dbus-user.own org.mozilla.librewolf.*
+dbus-user.own io.gitlab.librewolf.*
```
kmk3, line 40: I'd keep both entries so that the profile works with both old and new versions
of LibreWolf.

kmk3:
Related comment:

<https://github.com/netblue30/firejail/issues/4891#issuecomment-1028231449>

rusty-snake (changes requested):
https://github.com/netblue30/firejail/pull/4897#discussion_r799159868

netblue30:
I'll merge it and add back the old dbus-user command as suggested.

netblue30:
Fix in https://github.com/netblue30/firejail/commit/4211ee323fc7a5b57f89e50eeb813de398b902bd

PR #4903, opened 2022-02-06
keepass*: remove nou2f & add note about private-dev

### LINKED ISSUE #4883: Hardware key not detected on keepassxc
@miragy69 at https://github.com/keepassxreboot/keepassxc/issues/7317#issue-1109477546:

> ## Summary
> 
> Keeppassxc can't detect any hardware key after get in the jail. Hardware serial number is detectable. (Edited) Please reassign the label if needed.
> ## Temporary solution
> 
> The issue can be solved by commented out these applications under `/usr/lib/firejail/firecfg.config`. Then `# firecfg --clean && firecfg` for recreate links again image for commenting out the keepassxc-proxy can solve the issue [#6230](https://github.com/keepassxreboot/keepassxc/issues/6230), i.e. you don't have to manually click reload on browser plugin for password autoloading.
> ## Looking for future support on firejail
> 
> It will make safer for running keepassxc in the firejail, if the above issue can be solved. Many thanks.

----

Possible solutions:
1. Add a comment (`Put 'ignore nou2f' and 'ignore private-dev' in your keepassxc.local`).
2. Remove `nou2f` and a a comment for `private-dev` (`If you need to plugin devices while kpxc is running add ...`)
3. Remove both, `nou2f` and `private-dev`.

I'm not sure which one is the best, but I think we should fix this for 0.9.68.

_Originally posted by @rusty-snake in https://github.com/netblue30/firejail/discussions/4770#discussioncomment-2057394_

kmk3:
Fixes #4883.

### DIFF
--- a/etc/profile-a-l/keepassxc.profile
+++ b/etc/profile-a-l/keepassxc.profile
@@ -86,6 +86,9 @@ shell none
 tracelog
 
 private-bin keepassxc,keepassxc-cli,keepassxc-proxy
+# Note: private-dev prevents the program from seeing new devices (such as
+# hardware keys) on /dev after it has already started; add "ignore nou2f" to
+# keepassxc.local if this is an issue (see #4883).
 private-dev
 private-etc alternatives,fonts,ld.so.cache,ld.so.preload,machine-id
 private-tmp
--- a/etc/profile-a-l/keepassxc.profile
+++ b/etc/profile-a-l/keepassxc.profile
@@ -78,7 +78,6 @@ nonewprivs
 noroot
 nosound
 notv
-nou2f
 novideo
 protocol unix
 seccomp !name_to_handle_at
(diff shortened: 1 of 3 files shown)


rusty-snake:
Does keepass and keepassx support hardware keys?

etc/profile-a-l/keepassxc.profile
```diff
@@ -87,6 +86,9 @@ shell none
 tracelog
 
 private-bin keepassxc,keepassxc-cli,keepassxc-proxy
+# Note: private-dev prevents the program from seeing new devices (such as
+# hardware keys) on /dev after it has already started; add "ignore nou2f" to
```
rusty-snake, line 90:
```suggestion
# hardware keys) on /dev after it has already started; add "ignore prevate-dev" to
```

netblue30:
I'll grab it as is and make the release. You guys decide if we change the notes from "ignore nou2f" into "ignore private-dev" after that.

PR #4912, opened 2022-02-07
CI: replace centos (EOL) with almalinux

reinerh:


kmk3 (approved):
+1 for fixing GitLab CI.

PR #4732, opened 2021-12-01
Fix keeping certain groups with nogroups

kmk3:
This amends commit b828a9047 ("Keep audio and video groups regardless of
nogroups", 2021-11-28) from PR #4725.

The commit above did not change the behavior (the groups are still not
kept).  With this commit, it appears to work properly:

    $ groups | grep audio >/dev/null && echo kept
    kept
    # with check_can_drop_all_groups == 0
    $ firejail --quiet --noprofile --nogroups groups |
      grep audio >/dev/null && echo kept
    kept
    # with check_can_drop_all_groups == 1
    $ firejail --quiet --noprofile --nogroups groups |
      grep audio >/dev/null && echo kept
    $

Add a new check_can_drop_all_groups function to check whether the
supplementary groups can be safely dropped without potentially causing
issues with audio, 3D hardware acceleration or input (and maybe more).
It returns false if nvidia (and no `no3d`) is used or if (e)logind is
not running, as in either case the supplementary groups might be needed.

Note: With this, the behavior from before #4725 is restored on (e)logind
systems (when not using nvidia), as it makes the supplementary groups
always be dropped.

Note2: Even with the static variable, these checks still happen at least
once per translation unit (so twice in total per my testing).

This also amends (/kind of reverts) commit 6ddedeba0 ("Make nogroups
work on nvidia again", 2021-11-29) from PR #4725, as it restores the
nvidia check from it into the new check_can_drop_all_groups function.

PR #4933, opened 2022-02-11
Disable/comment message about nogroups being ignored

### LINKED ISSUE #4930: `nogroups` + wrc prints confusing messages
### Description

`nogroups` + wrc prints confusing messages.

### Steps to Reproduce

1. Create test.profile (ordering doesn't matter)
```
nogroups
```

2. test, run and see
```console
$ firejail --profile=./test.profile groups
Reading profile ./test.profile
Parent pid 1234, child pid 1235
Warning: logind not detected, nogroups command ignored     <--- is a lie
Child process initialized in 30.00 ms
rusty-snake    <---- running `groups` outside of the sandbox shows more so groups are actually cleaned

```

### Expected behavior

1. `Warning: logind not detected` is only shown when there is no logind
2. `Warning: nogroups command ignored` is only shown when it is rally ignored

### Actual behavior

1. `Warning: logind not detected` is shown on logind systems
2. `Warning: nogroups command ignored; cleaning all supplementary groups` makes no sense and confuses users.

### Additional context

#4725
#4732

### Environment

- Fedora Linux 35 (of course with systemd/logind)

```
firejail version 0.9.69 (03395e1)

Compile time support:
	- AppImage support is enabled
	- chroot support is disabled
	- file transfer support is disabled
	- firetunnel support is disabled
	- networking support is disabled
	- output logging is disabled
	- private-home support is disabled
	- SELinux support is enabled
	- X11 sandboxing support is disabled
```

kmk3:
Added on commit 7abce0b4c ("Fix keeping certain groups with nogroups",
2021-11-30) / PR #4732.

As reported by @rusty-snake on #4930, conflicting messages are printed
when using whitelist-run-common.inc with nogroups:

    $ cat test.profile
    nogroups
    $ firejail --profile=./test.profile groups
    Reading profile ./test.profile
    Parent pid 1234, child pid 1235
    Warning: logind not detected, nogroups command ignored     <--- is a lie
    Child process initialized in 30.00 ms
    rusty-snake    <---- running `groups` outside of the sandbox shows more so groups are actually cleaned

This probably happens because wrc causes /run/systemd to be hidden in
the sandbox and because check_can_drop_all_groups is called multiple
times, seemingly both before and after the whitelisting goes into
effect.  So disable the message about nogroups being ignored, but keep
the message about cleaning all supplementary groups (which is unlikely
to be printed unless it really happens).

Fixes #4930.

### DIFF
--- a/src/firejail/util.c
+++ b/src/firejail/util.c
@@ -125,7 +125,8 @@ int check_can_drop_all_groups() {
 	 * and #4603.
 	 */
 	if (access("/run/systemd/seats/", F_OK) != 0) {
-		fwarning("logind not detected, nogroups command ignored\n");
+		// TODO: wrc causes this to be printed even with (e)logind (see #4930)
+		//fwarning("logind not detected, nogroups command ignored\n");
 		can_drop_all_groups = 0;
 		goto out;
 	}


netblue30:
Merging for now.

Issue #4508, opened 2021-09-04
telegram.profile breaks download, open links und tray icon


Schweber:
Using the default telegram.profile, i cannot download files, cannot open links and another tray icon is added every time i re-open the telegram window.

I expect all of this to work out of the box.

I created a telegram.local file whitelisting my download directory but this doesn't change anything. Running telegram-desktop without firejail works fine.

I am running Manjaro Cinnamon 21.1.1 with Kernel 5.10.60-1 and firejail 0.9.66 (no git-version).

### Checklist

- [X] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [X] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [X] I have performed a short search for similar issues (to avoid opening a duplicate).
- [X] If it is a AppImage, `--profile=PROFILENAME` is used to set the right profile.
- [X] Used `LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 PROGRAM` to get english error-messages.
- [X] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [X] This is not a question. Questions should be asked in https://github.com/netblue30/firejail/discussions.

### Log

[debug output.txt](https://github.com/netblue30/firejail/files/7109396/debug.output.txt)

jose1711:
Unable to reproduce the issue with tray icon on Arch Linux and `firejail` from git (a6672757d1d884d02538ac3f92b13f997ee3efbb).

davidebeatrici:
Reproduced on KDE, Telegram version: 3.4.8

```
error: : cannot open
error: : cannot open
error: : cannot open
kf.notifications: env says KDE is running but SNI unavailable -- check KDE_FULL_SESSION and XDG_CURRENT_DESKTOP
Corrupt JPEG data: premature end of data segment
qt.svg: Error while inflating gzip file: SVG format check failed
qt.svg: Error while inflating gzip file: SVG format check failed
qt.svg: Error while inflating gzip file: SVG format check failed
qt.svg: Error while inflating gzip file: SVG format check failed
qt.svg: Error while inflating gzip file: SVG format check failed
qt.svg: Error while inflating gzip file: SVG format check failed
qt.svg: Error while inflating gzip file: SVG format check failed
qt.svg: Error while inflating gzip file: SVG format check failed
[...]

PR #4946, opened 2022-02-15
firejail.config: add warning about allow-tray

kmk3:
According to #4053, there is currently no safe (in the sense of not
allowing to escape the sandbox) implementation of
`org.kde.StatusNotifierWatcher`, but it is required by multiple programs
for tray functionality.  Users may not be aware of this (for example,
see https://github.com/netblue30/firejail/issues/4508), so add a warning about it.

Note: allow-tray was added on commit c86cae2d0 ("Add new condition
ALLOW_TRAY", 2021-09-04) / PR #4510.

Cc: @davidebeatrici

### DIFF
--- a/etc/firejail.config
+++ b/etc/firejail.config
@@ -2,7 +2,8 @@
 # keyword-argument pairs, one per line. Most features are enabled by default.
 # Use 'yes' or 'no' as configuration values.
 
-# Allow programs to display a tray icon
+# Allow programs to display a tray icon (warning: allows escaping the sandbox;
+# see https://github.com/netblue30/firejail/discussions/4053)
 # allow-tray no
 
 # Enable AppArmor functionality, default enabled.


rusty-snake (commented):
> dbus-user.talk org.kde.StatusNotifierWatcher is
unsafe and allows escaping the sandbox

In theory it could be safe, but I'm not aware of a safe implementation.

The unsafer thing with tray icons is `dbus-user.own org.kde.*`.

> WARNING: allows escaping the sandbox;

How cares? There are enough other ways to escape. Anyway let's add it.

> `see https://github.com/netblue30/firejail/discussions/4053`

Do we need the link? IMHO there isn't much information other than "it's unsafe!!!".

kmk3:
@rusty-snake left a comment:

> > dbus-user.talk org.kde.StatusNotifierWatcher is
> unsafe and allows escaping the sandbox
> 
> In theory it could be safe, but I'm not aware of a safe implementation.
> 
> The unsafer thing with tray icons is `dbus-user.own org.kde.*`.

Reworded it (and made the warning lowercase).

Does it make sense now?  Also, what are the implementations of
`org.kde.StatusNotifierWatcher` that you have in mind?

> > WARNING: allows escaping the sandbox;
> 
> How cares? There are enough other ways to escape. Anyway let's add it.

Users can only choose to care if they know about it.  I think it would be
useful to document every common case of sandbox escaping (such as with #4760).

> > `see https://github.com/netblue30/firejail/discussions/4053`
> 
> Do we need the link? IMHO there isn't much information other than "it's
> unsafe!!!".

I don't know much about the details, so to me it's much more information than
nothing.  It also centralizes the discussion and the link might help attract
someone who would care enough to fix the implementation.  But if you know of a
better resource we could use that instead.

rusty-snake:
> Also, what are the implementations of
org.kde.StatusNotifierWatcher that you have in mind?

IIRC I tested the native kde implementation, gnome-shell-extension-appindicator and the xfce implementation.

rusty-snake (approved):

PR #4950, opened 2022-02-16
man.profile: remove read-only tmp to fix mandoc

### LINKED ISSUE #4927: broken man.profile in 0.9.68
### Description

Using firejail 0.9.68 on Arch Linux with "firecfg" run after the upgrade, I'm no longer able to view man pages with the man command. The issue is likely that "man" can no longer spawn the PAGER process. "man ls" prints the man page all at once without piping it to more/less. Running "man ls | less" works fine.

### Steps to Reproduce

Install firejail, use the mandoc package (rather than GNU man, may be related) for the "man" command, and run it.

### Expected behavior

Paging through a man page.

### Actual behavior

Man page is printed all at once.

Works normally.

### Checklist

- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).

kmk3:
Having `read-only /tmp` yields the following:

    $ man ls
    [...]
    man: /usr/share/man/man1/ls.1.gz: SYSERR: mkstemp: /tmp/man.XXXXxxxxxx: Read-only file system
    [...]

It also causes the pager (e.g.: less(1)) to not be called, which means
that the entire man page is just printed all at once on the terminal.

Environment: mandoc 1.14.6-1 on Artix Linux.

Fixes #4927.

Reported-by: @hyder365

### DIFF
--- a/etc/profile-m-z/man.profile
+++ b/etc/profile-m-z/man.profile
@@ -66,4 +66,4 @@ dbus-system none
 
 memory-deny-write-execute
 read-only ${HOME}
-read-only /tmp
+#read-only /tmp # breaks mandoc (see #4927)


rusty-snake (approved):

reinerh:
Btw was someone able to reproduce the issue with mandoc? For me it was working with mandoc and read-only /tmp. Could also have been an issue in the reporter's setup (maybe additional scripts in the pipeline).

reinerh (approved):

rusty-snake:
`mkstemp("/tmp/man.XXXXXXX")` seems to be used by term_tag:

https://cvsweb.bsd.lv/mandoc/term_tag.c?rev=1.6&content-type=text/x-cvsweb-markup

reinerh:
Thanks for checking ti!

PR #4958, opened 2022-02-18
drop private-dev from wireshark.profile

glitsj16:
This came up in https://github.com/netblue30/firejail/discussions/4949. I can reproduce that `private-dev` stops tshark/wireshark from showing interfaces.

### DIFF
--- a/etc/profile-m-z/wireshark.profile
+++ b/etc/profile-m-z/wireshark.profile
@@ -45,6 +45,9 @@ tracelog
 
 # private-bin wireshark
 private-cache
+# private-dev prevents (some) interfaces from being shown.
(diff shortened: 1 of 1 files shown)


glitsj16:
> As I understand @crocket in #4949 some interfaces are shown with private-dev, so maybe keep it as optional hardening?

That's the problem, it's not optional IMO. For me `private-dev` stops the app from showing **any** interfaces, reagardless of the caps.keep that @crocket is trying to deal with in #4949. I've thorougly tested this the last couple of days. We can put this on hold, no problem. But it is a profile bug in my opinion/testing.

rusty-snake (approved):

PR #4964, opened 2022-02-20
qbittorrent.profile: fix data directory location

avallach2000:
Since version 4.3.3 qBittorrent switched to new data directory location (`~/.local/share/qBittorrent` instead of `~/.local/share/data/qBittorrent`).
ChangeLog says that legacy 'data' directory is now used only as a fallback. Also, when using current profile the following warning appears on the terminal:
```
The legacy data directory '/home/avallac_h/.local/share/data/qBittorrent' is used. It is recommended to move its content to '/home/avallac_h/.local/share/qBittorrent'
```
The proposed pull request should fix this issue.

### DIFF
--- a/etc/profile-m-z/qbittorrent.profile
+++ b/etc/profile-m-z/qbittorrent.profile
@@ -10,6 +10,7 @@ noblacklist ${HOME}/.cache/qBittorrent
 noblacklist ${HOME}/.config/qBittorrent
 noblacklist ${HOME}/.config/qBittorrentrc
 noblacklist ${HOME}/.local/share/data/qBittorrent
+noblacklist ${HOME}/.local/share/qBittorrent
 
 # Allow python (blacklisted by disable-interpreters.inc)
 include allow-python2.inc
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/qbittorrent.profile
```diff
@@ -10,6 +10,7 @@ noblacklist ${HOME}/.cache/qBittorrent
 noblacklist ${HOME}/.config/qBittorrent
 noblacklist ${HOME}/.config/qBittorrentrc
 noblacklist ${HOME}/.local/share/data/qBittorrent
+noblacklist ${HOME}/.local/share/qBittorrent
```
rusty-snake, line 13: Never blacklisted ;)

rusty-snake (changes requested):

etc/profile-m-z/qbittorrent.profile
avallach2000: Do you mean that I should also change `disable-programs.inc`?

rusty-snake: Yes

rusty-snake (approved):

netblue30:
Merged!

PR #4966, opened 2022-02-20
Add support for changing appearance of the Qt6 apps with qt6ct

avallach2000:
In Arch this applies at least to telegram-desktop, qmmp, qbittorrent and strawberry which are now built using Qt6 libraries. But I'm unsure about `whitelist-1793-workaround.inc` which is currently used by `neochat.profile` only.

Also I found a strange artifact in zeal.profile: https://github.com/netblue30/firejail/blob/070e78a8892d86687a1a3e74262628ee9c562c46/etc/profile-m-z/zeal.profile#L22

### DIFF
--- a/etc/inc/whitelist-1793-workaround.inc
+++ b/etc/inc/whitelist-1793-workaround.inc
@@ -23,6 +23,7 @@ noblacklist ${HOME}/.config/kio_httprc
 noblacklist ${HOME}/.config/kioslaverc
 noblacklist ${HOME}/.config/ksslcablacklist
 noblacklist ${HOME}/.config/qt5ct
+noblacklist ${HOME}/.config/qt6ct
 noblacklist ${HOME}/.config/qtcurve
 
 blacklist ${HOME}/.config/*
--- a/etc/inc/whitelist-common.inc
+++ b/etc/inc/whitelist-common.inc
@@ -69,6 +69,7 @@ whitelist ${HOME}/.config/kio_httprc
 whitelist ${HOME}/.config/kioslaverc
 whitelist ${HOME}/.config/ksslcablacklist
 whitelist ${HOME}/.config/qt5ct
+whitelist ${HOME}/.config/qt6ct
 whitelist ${HOME}/.config/qtcurve
 whitelist ${HOME}/.kde/share/config/kdeglobals
 whitelist ${HOME}/.kde/share/config/kio_httprc
@@ -83,3 +84,4 @@ whitelist ${HOME}/.kde4/share/config/ksslcablacklist
 whitelist ${HOME}/.kde4/share/config/oxygenrc
 whitelist ${HOME}/.kde4/share/icons
 whitelist ${HOME}/.local/share/qt5ct
+whitelist ${HOME}/.local/share/qt6ct
--- a/etc/inc/whitelist-usr-share-common.inc
+++ b/etc/inc/whitelist-usr-share-common.inc
@@ -55,6 +55,7 @@ whitelist /usr/share/qt
 whitelist /usr/share/qt4
 whitelist /usr/share/qt5
 whitelist /usr/share/qt5ct
+whitelist /usr/share/qt6ct
 whitelist /usr/share/sounds
 whitelist /usr/share/tcl8.6
 whitelist /usr/share/tcltk
--- a/etc/profile-a-l/bibletime.profile
+++ b/etc/profile-a-l/bibletime.profile
@@ -49,7 +49,7 @@ seccomp !chroot
 shell none
 
 disable-mnt
-# private-bin bibletime,qt5ct
+# private-bin bibletime
 private-cache
 private-dev
 private-etc alternatives,ca-certificates,crypto-policies,fonts,ld.so.cache,ld.so.preload,login.defs,machine-id,passwd,pki,resolv.conf,ssl,sword,sword.conf
--- a/etc/profile-m-z/zeal.profile
+++ b/etc/profile-m-z/zeal.profile
@@ -19,7 +19,6 @@ include disable-shell.inc
 include disable-xdg.inc
 
 mkdir ${HOME}/.cache/Zeal
-mkdir ${HOME}/.config/qt5ct
 mkdir ${HOME}/.config/Zeal
 mkdir ${HOME}/.local/share/Zeal
 whitelist ${HOME}/.cache/Zeal


glitsj16 (approved):
LGTM

glitsj16:
> Also I found a strange artifact in zeal.profile:
 mkdir ${HOME}/.config/qt5ct

Nice catch. We should remove that line. Can you take care of that please?

avallach2000:
Let it be in the same PR: everything related to qt5ct/ct6ct

kmk3:
```console
070e78a88 2022-02-20 08:20:59 -0500 Merge branch 'master' of ssh://github.com/netblue30/firejail
$ git grep qt5ct -- etc
etc/inc/whitelist-1793-workaround.inc:noblacklist ${HOME}/.config/qt5ct
etc/inc/whitelist-common.inc:whitelist ${HOME}/.config/qt5ct
etc/inc/whitelist-common.inc:whitelist ${HOME}/.local/share/qt5ct
etc/inc/whitelist-usr-share-common.inc:whitelist /usr/share/qt5ct
etc/profile-a-l/bibletime.profile:# private-bin bibletime,qt5ct
etc/profile-m-z/zeal.profile:mkdir ${HOME}/.config/qt5ct
```

Can you add it to bibletime.profile as well?

@glitsj16 commented [on Feb 20](https://github.com/netblue30/firejail/pull/4966#issuecomment-1046339246):

> > Also I found a strange artifact in zeal.profile:
> > 
> > mkdir ${HOME}/.config/qt5ct
> 
> Nice catch. We should remove that line. Can you take care of that please?

Are you sure it's safe to remove it?  Apparently it was required in order to
fix #3389:

@OndrejMalek commented [on Apr 29, 2020](https://github.com/netblue30/firejail/issues/3389#issue-609379203):

> ```
> $ firejail --noprofile zeal
> Parent pid 17556, child pid 17557
> Child process initialized in 9.53 ms
> qt5ct: using qt5ct plugin
> Qt: Session management error: None of the authentication protocols specified are supported
> zeal.core.applicationsingleton: Singleton ID: GNxAletmknejwIs_KXZiYaAzw9OzTsTq2ggd_l64LPQ
> zeal.core.applicationsingleton: Starting as a primary instance. (PID: 6)
> qt5ct: D-Bus global menu: no
> libpng warning: iCCP: known incorrect sRGB profile
> libpng warning: iCCP: known incorrect sRGB profile
> qt5ct: D-Bus system tray: no
> inotify_add_watch("/home/malekon/.config/qt5ct") failed: "No such file or directory"
> 
> ```

* 8c69eab21 ("fixes for zeal.profile", 2020-04-30)

glitsj16:
> Are you sure it's safe to remove it? Apparently it was required in order to
fix #3389

@kmk3 Thanks for spotting that. Yet, after another round of installing/testing zeal I can confirm the `mkdir ${HOME}/.config/qt5ct` is not needed at all, so yes, IMO we can safely drop it. But let's ping the OP of that issue to double-check...
@OndrejMalek If you're still using zeal it would be nice if you could check this too. The original discussion in #3389 might have been a bit uninformed due to my lack of experience with QT back at the time.

On a side note, during re-testing zeal it turned out the profile can be updated/hardened somewhat, which I'll take care of in a separate PR.

rusty-snake:
> Apparently it was required in order to
fix https://github.com/netblue30/firejail/issues/3389

Which is likely for all Qt programs that use qt5ct needed and not only zeal?

kmk3:
@rusty-snake commented [on Feb 21](https://github.com/netblue30/firejail/pull/4966#issuecomment-1046624257):

> > Apparently it was required in order to
> > fix #3389
> 
> Which is likely for all Qt programs that use qt5ct needed and not only zeal?

It depends, was inotify being called by zeal or by Qt?

```
inotify_add_watch("/home/malekon/.config/qt5ct") failed: "No such file or directory"
```

glitsj16:
Can't find any reference to qt5ct in zeal's source code: https://github.com/zealdocs/zeal/search?q=qt5ct. Distro's might package zeal differently, but IMO the qt5ct use is a user thing. There are several [QT environment vars](https://wiki.archlinux.org/title/Qt) that a user can set to change theming, wayland support etcetera. For me that has been working for a long time just fine when firejailing the apps in question, without ever needing to `mkdir` ${HOME}/config/qt5ct. We whitelist the relevant qt5ct paths and that's all that firejail needs AFAICT (besides qt6ct support which this clean PR now adds).

avallach2000:
I compiled Zeal on Arch and can confirm that it does not require to create qt5ct directory by itself (probably like any other Qt5 application). It works fine with or without that directory.

Also, I've seen the described behaviour - the main window disappeance - just recently. That is how qbittorrent behaves without the changes proposed by this PR. It is about Qt6 but still.

About bibletime. I saw that, but that line was commented, so I decided not to touch it. I've tested it on Arch, and it turns out that even in private-bin mode it doesn't require to run qt5ct (which is actually a configuration tool, because all work is done by env. vars, libqt5ct.so and libqt5ct-style.so).

Important note: I'm using Arch/LXQt, maybe I should perform some checks with some KDE distro (like Kubuntu) before merging this.

glitsj16:
@avallach2000 Thanks for confirming, that hopefully helps to clear up any confusion about the mkdir.

avallach2000:
Sorry to keep you waiting. I've tested both Zeal and BibleTime under Kubuntu 21.10 Live DVD with recent firejail from deki's PPA. As for me everything seems to be ok:
- Zeal doesn't need `~/.config/qt5ct` directory to work properly.
- BibleTime in private-bin mode doesn't need to call qt5ct directly.

Btw, while doing this I've found a possible bug in zeal.profile. I had to add `machine-id` to `private-etc` statement. Without doing this Zeal fails to start with the following error:
```
Child process initialized in 149.95 ms
dbus[28]: D-Bus library appears to be incorrectly set up: see the manual page for dbus-uuidgen to correct this issue. (Failed to open "/var/lib/dbus/machine-id": No such file or directory; Failed to open "/etc/machine-id": No such file or directory)
 
kubuntu@kubuntu:~$ echo $?
134
```
But I'm unsure if this is related only to the live session or not.

glitsj16:
> Sorry to keep you waiting.

That's quite allright, you're doing a great job by opening the PR and testing things thoroughly. Much appreciated!

Like I said above, while testing zeal it turned out we could add a few extra hardening options/nits. Perhaps you can test those while you're at it. Here's what I had in mind:

```
[...]
# fix ordering
noblacklist ${HOME}/.cache/Zeal
noblacklist ${HOME}/.config/Zeal
[...]
include disable-proc.inc
[...]
[...]
# zeal can use the system default web browser, let's assume that's Firefox
dbus-user filter
dbus-user.talk org.mozilla.Firefox.*
dbus-user.talk org.mozilla.firefox.*
```

@rusty-snake @kmk3 Anything holding back merging this in your opinion?

glitsj16 (approved):
Looks great. Let's wait a bit to give other contributors a chance to review, but IMO this is a very nice contribution, thanks for that!

rusty-snake:
> @rusty-snake @kmk3 Anything holding back merging this in your opinion?

No

> zeal can use the system default web browser, let's assume that's Firefox
> dbus-user filter
dbus-user.talk org.mozilla.Firefox.*
dbus-user.talk org.mozilla.firefox.*

1. This means to remove `dbus-user none`
2. This requires `whitelist ${HOME}/.mozilla/...`, it don't?
3. I think only one of the talk rules is requeired.


We can also `?ALLOW_TRAY: .. org.kde.StatusNotifierWatcher ...`

> extra hardening 

```
seccomp.block-secondary
```

glitsj16:
@rusty-snake

> 1. This means to remove dbus-user none

Indeed, forgot to mention that!

> 2. This requires whitelist ${HOME}/.mozilla/..., it don't?

Indeed it does need that. We can use the exact same block as we have in {signal-desktop,thunderbird}.profile:

```
# These lines are needed to allow Firefox to open links
noblacklist ${HOME}/.mozilla
```

> 3. I think only one of the talk rules is requeired.

Noticed that too on several visits to the firefox related profiles, but I'm not sure on this one. Might be the case that older FF versions and/or ESR editions still use different syntax here, I'd have to dig around in documentation to check. BTW, we (still) use both formats in current {firefox{,-developer-edition},signal-desktop}.profile so if we decide we need only one, those files need to be edited too (more of a note-to-self remark).

Your other extra's look fine too. Perhaps we should keep the extra hardening for a seperate PR. @avallach2000 has been very kind already to accomodate our remarks, and they are not related to qt{5,6}ct anyway. Thanks for the input though, always a pleasure collaborating!

PR #4981, opened 2022-02-25
ffplay: fix private-etc

glitsj16:
ffplay redirects to ffmpeg, which already has several of the same items in its private-bin. I suspect this is fallout from https://github.com/netblue30/firejail/commit/23c4234d709fccc4a78d64aec0adbd6a22db5909. There might be similar potential cleanups, I'll look into that.

### DIFF
--- a/etc/profile-a-l/ffplay.profile
+++ b/etc/profile-a-l/ffplay.profile
@@ -14,7 +14,7 @@ ignore nogroups
 ignore nosound
 
 private-bin ffplay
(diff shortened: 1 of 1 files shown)


rusty-snake:
> - private-etc-always-required.sh […] It can not check redirect profiles, therefore some files are double allowed if both profile (redirect profile and redirected profile) contain a `private-etc` line.

#4643

glitsj16:
@rusty-snake Whow, I never realised that. Thanks for pointing that out here. I'll drop the PR asap.

PR #4985, opened 2022-02-25
whitelist restructuring

smitsohu:
Some reorganization/cleanup of the whitelist code.  Also a new mechanism to detect earlier whitelist mounts.

There should be no functional changes.

### DIFF
--- a/src/firejail/fs_whitelist.c
+++ b/src/firejail/fs_whitelist.c
@@ -49,20 +49,24 @@ static void whitelist_error(const char *path) {
 	exit(1);
 }
 
-static int whitelist_mkpath(const char* path, mode_t mode) {
+static int whitelist_mkpath(const char *parentdir, const char *relpath, mode_t mode) {
+	// starting from top level directory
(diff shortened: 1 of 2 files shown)


glitsj16:
I didn't respond sooner as I noticed you were actively working on this and regularly pushing changes. Having some free time over the weekend I was planning on checking out your [whitelist branch](https://github.com/smitsohu/firejail/tree/whitelist) and give it a test run. Might be redundant to ask for a branch described as 'no functional changes', but if there's anything specific to look out for, feel free to throw in pointers.

smitsohu:
@glitsj16 thanks for trying it out :)

If you find any functional change it means that I got something wrong. If you do things like `--whitelist=~/.local --whitelist=~/.local/share` there will be a new debug message `Debug 108: whitelisted already`.

That's about it. Other than that it is just writing stuff in a different way but do the same thing. I was just hoping to organize the code a bit better.

netblue30:
all in!

PR #4990, opened 2022-02-28
Add ability to disable user profiles at compile time.

chestnykh:


configure
```diff
@@ -3434,8 +3438,8 @@ if test "x$enable_apparmor" = "xyes"; then :
 	HAVE_APPARMOR="-DHAVE_APPARMOR"
 
 pkg_failed=no
-{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for libapparmor" >&5
-$as_echo_n "checking for libapparmor... " >&6; }
+{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for AA" >&5
+$as_echo_n "checking for AA... " >&6; }
```
dseight, line 3442: It is better to remove any unrelated changes from commit. Both this renaming and some whitespace changes.

dseight (changes requested):

configure
chestnykh: I removed unrelated changes

netblue30:
Thanks for the patch!

Issue #4982, opened 2022-02-25
Geary fails to fully start and burns CPU


spantaleev:
### Description

geary (`1:40.0-6` on Archlinux) starts, but the UI is frozen.

### Steps to Reproduce

I'm using the default geary profile. The one deployed by the Archlinux `firejail` package seems to be up to date with current `master`.

_Steps to reproduce the behavior_

1. `LC_ALL=C firejail geary`
2. Observe geary's frozen / empty UI and 100% CPU usage

Geary's dialog window still remains running after that, but nothing is clickable.

See the log below for some errors.

### Behavior without a profile

_What changed calling `LC_ALL=C firejail --noprofile /usr/bin/geary` in a terminal?_

Geary starts up normally.

### Additional context

_Any other detail that may help to understand/debug the problem_

### Environment

- Archlinux
- Firejail version (`firejail --version`): 0.9.68

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [ ] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ ] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

<details>
<summary>Output of <code>LC_ALL=C firejail /path/to/program</code></summary>
<p>

```
Reading profile /etc/firejail/geary.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-exec.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-shell.inc
Reading profile /etc/firejail/disable-xdg.inc
Reading profile /etc/firejail/whitelist-common.inc
Reading profile /etc/firejail/whitelist-runuser-common.inc
Reading profile /etc/firejail/whitelist-usr-share-common.inc
[...]

PR #4992, opened 2022-03-01
geary fixes

glitsj16:
Fixes for https://github.com/netblue30/firejail/issues/4982.

### DIFF
--- a/etc/profile-a-l/geary.profile
+++ b/etc/profile-a-l/geary.profile
@@ -53,7 +53,7 @@ include whitelist-var-common.inc
 
 apparmor
 caps.drop all
-ipc-namespace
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/geary.profile
```diff
@@ -47,15 +53,16 @@ include whitelist-var-common.inc
 
 apparmor
 caps.drop all
-machine-id
+ipc-namespace
```
rusty-snake, line 56: `ipc-namespace` may cause issues with X11.

glitsj16: I commented the ipc-namespace option. Thanks!

rusty-snake (approved):
LGTM

PR #4999, opened 2022-03-02
opera fixes/hardening

glitsj16:
The URL for the 'Disable for now' comment seems out of place. After testing it looks like we can actually harden the profiles by dropping the disabled includes alltogether.

### DIFF
--- a/etc/profile-m-z/opera-beta.profile
+++ b/etc/profile-m-z/opera-beta.profile
@@ -5,11 +5,6 @@ include opera-beta.local
 # Persistent global definitions
 include globals.local
 
-# Disable for now, see https://www.tutorialspoint.com/difference-between-void-main-and-int-main-in-c-cplusplus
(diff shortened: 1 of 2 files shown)


rusty-snake:
Oops, wrong URL in click-board. Anyway #3688 is it:

> chromium-common.profile: add wruc and wusc, but disable it for the following profiles until tested. tests welcome.

So if you tested this (e.g. there is no `/usr/share/opera`) we can remove the `ignore`s.

glitsj16:
> So if you tested this (e.g. there is no /usr/share/opera) we can remove the ignores.

Yes I've been testing opera as a fall-back browser for a couple of weeks now. The ignores can indeed be dropped, and on Arch Linux at least there's no sign of /usr/share/opera.

glitsj16:
Okay, I've tracked down more info and there's no trace of /usr/share/opera AFAICT. If it turns up we can always add it.

Issue #4995, opened 2022-03-01
problems with sylpheed


Rosika2:
Hi all,

I´ve got a dedicated work-directory which I want the e-mail-client `sylpheed` to use with the option `--private=[PATH to the DIR]` .
This directory is located on my system on: **/media/rosika/f14a27c2-0b49-4607-94ea-2e56bbf76fe1/DATEN-PARTITION/Dokumente/kgw_work3_sylpheed/** .

Strangely it seems to work just with the "--noprofile" option:
As an example I cd-ed into the respective directory and then issued the command

`firejail --noprofile --private=(pwd) sylpheed`

which works well.

`firejail --private=(pwd) sylpheed`

however won´t even open sylpheed´s GUI and shuts down the sandbox immediately:

```
firejail --private=(pwd) sylpheed
Reading profile /etc/firejail/sylpheed.profile
Reading profile /etc/firejail/email-common.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-exec.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-xdg.inc
Reading profile /etc/firejail/whitelist-common.inc
Reading profile /etc/firejail/whitelist-runuser-common.inc
Reading profile /etc/firejail/whitelist-usr-share-common.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Warning: networking feature is disabled in Firejail configuration file
Parent pid 11866, child pid 11869
Warning: cleaning all supplementary groups
Warning: cleaning all supplementary groups
Warning: An abstract unix socket for session D-BUS might still be available. Use --net or remove unix from --protocol set.
Warning: skipping crypto-policies for private /etc
Warning: skipping gcrypt for private /etc
Warning: skipping gnupg for private /etc
Warning: skipping groups for private /etc
Warning: skipping hosts.conf for private /etc
Warning: skipping mailname for private /etc
Private /etc installed in 47.83 ms
Private /usr/etc installed in 0.01 ms
Blacklist violations are logged to syslog
Warning: cleaning all supplementary groups
Child process initialized in 199.78 ms
[...]

PR #5003, opened 2022-03-03
fix sylpheed

glitsj16:
Sylpheed needs access to /var/lib/dbus/machine-id, which usually is a symlink to /etc/machine-id. Fix for https://github.com/netblue30/firejail/issues/4995. Technically speaking we could add a private-etc to sylpheed.profile. I decided to add it in email-common.profile instead for ease-of-use/maintainability.

### DIFF
--- a/etc/profile-a-l/email-common.profile
+++ b/etc/profile-a-l/email-common.profile
@@ -66,7 +66,7 @@ tracelog
 # disable-mnt
 private-cache
 private-dev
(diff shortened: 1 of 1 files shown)


kmk3 (commented):
@glitsj16 commented [on Mar 3](https://github.com/netblue30/firejail/pull/5003#issue-1158381799):

> Sylpheed needs access to /var/lib/dbus/machine-id, which usually is a symlink
> to /etc/machine-id. Fix for #4995. Technically speaking we could add a
> private-etc to sylpheed.profile. I decided to add it in email-common.profile
> instead for ease-of-use/maintainability.

On future commits, whenever feasible please include the reasoning in the commit
message itself (the above message is only present on GitHub), as that enables
reading it locally on the git log.  Also, having the issue/PR references in the
commit message makes it easier to grep for related commits.

PR #5009, opened 2022-03-04
drop redundant ignore in chromium-based browsers

glitsj16:
The `ignore whitelist /usr/share/chromium` can be dropped from profiles redirecting to chromium-common.profile, as that path is only whitelisted in chromium.profile.


netblue30:
all in!

PR #5013, opened 2022-03-05
Blacklist scala devel stuff

rusty-snake:


glitsj16 (approved):
LGTM

etc/inc/disable-programs.inc
```diff
@@ -466,6 +467,7 @@ blacklist ${HOME}/.config/iridium
 blacklist ${HOME}/.config/itch
 blacklist ${HOME}/.config/jami
 blacklist ${HOME}/.config/jd-gui.cfg
+blacklist ${HOME}/.config/jgit
```
kmk3, line 470: AFAIK jgit is the generic git implementation used by Java programs, so this
could possibly affect programs like Eclipse and SonarQube.

etc/inc/disable-programs.inc
```diff
@@ -714,6 +717,7 @@ blacklist ${HOME}/.icedove
 blacklist ${HOME}/.imagej
 blacklist ${HOME}/.inkscape
 blacklist ${HOME}/.itch
+blacklist ${HOME}/.ivy2
```
kmk3, line 720: This is used by Apache Ivy (and maybe also maven and gradle) and so could
potentially impact any Java development tool (including IDEs).

kmk3 (commented):
This seems to affect more than just Scala.

netblue30:
merged, thanks!

Issue #5012, opened 2022-03-05
Visual Studio Code profile cannot access some directories


Gerenuk:
### Description

Running visual studio code seems to have important directories inaccessible unless the home directory is explicitely whitelisted.

### Steps to Reproduce

I'm running `firejail /usr/bin/code --no-sandbox --unity-launch` as these options for `code` were already preconfigured by the usual desktop link. Without `--no-sandbox` it would not start.

### Expected behavior

It should start and load my theme from the extensions.

### Actual behavior

Visual studio code starts, but apparently does not see a lot of required files. In particular the directory `~/.vscode/extensions` appears empty when trying to open a file from vscode whereas in reality there are extension folders inside. It does not load the theme therein either. None of my usual home folders are visible. (Is that intended?)

However, everything seems to work if I white list `whitelist ~`

### Behavior without a profile

`--noprofile` seems fine.

### Environment

Linux: Manjaro (Archlinux based)
firejail 0.9.68

### Log

```
> firejail  /usr/bin/code --no-sandbox --unity-launch
Reading profile /etc/firejail/code.profile
Reading profile /etc/firejail/allow-common-devel.inc
Reading profile /etc/firejail/electron.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-programs.inc
Parent pid 56475, child pid 56476
Warning: NVIDIA card detected, nogroups command ignored
Warning: NVIDIA card detected, nogroups command ignored
Warning: NVIDIA card detected, nogroups command ignored
Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Warning: NVIDIA card detected, nogroups command ignored
Child process initialized in 25.91 ms
```

`firejail --debug`:
```
Autoselecting /bin/bash as shell
Building quoted command line: '/usr/bin/code' '--no-sandbox' '--unity-launch' 
Command name #code#
Found code.profile profile in /etc/firejail directory
Found allow-common-devel.inc profile in /etc/firejail directory
Found electron.profile profile in /etc/firejail directory
Found disable-common.inc profile in /etc/firejail directory
Found disable-programs.inc profile in /etc/firejail directory
Using the local network stack
Using the local network stack
Initializing child process
PID namespace installed
Mounting tmpfs on /run/firejail/mnt directory
Creating empty /run/firejail/mnt/seccomp directory
Creating empty /run/firejail/mnt/seccomp/seccomp.protocol file
[...]

PR #5026, opened 2022-03-09
Electron app fixes

smitsohu:
From discussion in #5012

skypeforlinux is whitelist profile, so adding more whitelisted paths will not break anything. Just cleaning up that profile.

notable.profile has 
```
ignore whitelist ${HOME}/.config/Electron
ignore whitelist ${HOME}/.config/electron-flag*.conf
```
already.


glitsj16 (approved):
LGTM

rusty-snake (approved):

PR #5028, opened 2022-03-09
gcov: fix gcov functions always declared as dummy

kmk3:
Currently, the check to enable gcov relies on a non-existent macro due
to a typo, which looks like it would cause the dummy/empty versions of
the gcov functions to always be declared (even with --enable-gcov),
instead of the real ones from gcov.h.  This commit fixes the typo
(HAS_GCOV -> HAVE_GCOV).  See configure.ac for the macro declaration.

This amends commit 5106b2ec4 ("gcov: use no-op functions if not
enabled", 2021-06-20) / PR #4376.

Occurrences of each macro with this commit applied:

    $ git grep -F HAVE_GCOV | wc -l
    16
    $ git grep -F HAS_GCOV | wc -l
    0

### DIFF
--- a/src/include/gcov_wrapper.h
+++ b/src/include/gcov_wrapper.h
@@ -21,7 +21,7 @@
 #ifndef GCOV_WRAPPER_H
 #define GCOV_WRAPPER_H
 
-#ifdef HAS_GCOV
+#ifdef HAVE_GCOV
 #include <gcov.h>
 
 /*
@@ -41,6 +41,6 @@ static void __gcov_flush(void) {
 #define __gcov_dump() ((void)0)
 #define __gcov_reset() ((void)0)
 #define __gcov_flush() ((void)0)
-#endif /* HAS_GCOV */
+#endif /* HAVE_GCOV */
 
 #endif /* GCOV_WRAPPER_H */


kmk3:
> This amends commit 5106b2e ("gcov: use no-op functions if not enabled",
> 2021-06-20) / PR #4376.

The commit that introduces the typo is from over 8 months ago and the only
other gcov-related commit since then that I found is commit b8d282c7d ("gcov",
2022-01-17) by @smitsohu.

So it looks like either the typo wasn't an issue or that no one has tried using
gcov since #4376.

IIRC I tried testing gcov when working on #4373/#4376 but couldn't get it to
work.

And with that I wonder, does anyone actually test firejail with gcov?  When was
the last time anyone tried doing that?  It does not seem to be hooked up in
tests or ci, so I assume that it has to be done manually.

smitsohu:
@kmk3 TBH I sent b8d282c7d53bbb0876ee02b10bf7b487932cb80e just because I realized that I had broken a pattern.

> IIRC I tried testing gcov when working on https://github.com/netblue30/firejail/pull/4373/https://github.com/netblue30/firejail/pull/4376 but couldn't get it to
> work.

Same here. If someone knows how to hold it, any hints would be appreciated.

netblue30:
merged, thanks!

PR #5042, opened 2022-03-13
mupdf refactoring cfr. https://github.com/netblue30/firejail/discussions/4993

glitsj16:


etc/profile-m-z/mupdf-x11-curl.profile
```diff
@@ -8,9 +8,9 @@ include mupdf-x11-curl.local
 #include globals.local
 
 ignore net none
-
 netfilter
 protocol unix,inet,inet6
+ignore protocol
```
rusty-snake, line 13: Not needed, `protocol` accumulates.

etc/profile-m-z/mupdf-x11-curl.profile
```diff
@@ -8,9 +8,9 @@ include mupdf-x11-curl.local
 #include globals.local
 
 ignore net none
-
```
rusty-snake, line 11: I think this is intentionally to keep `ignore`s in it's own block.

etc/profile-m-z/mupdf.profile
```diff
@@ -41,3 +42,6 @@ private-tmp
 
 dbus-user none
 dbus-system none
+
+memory-deny-write-execute
```
rusty-snake, line 46: Sure that this does not break `mupdf-gl`?

etc/profile-m-z/mupdf-x11-curl.profile
glitsj16: Fixed, thanks.

etc/profile-m-z/mupdf.profile
glitsj16: I've checked several times and both mupdf and mupdf-gl both work with mdwe. At least they do on Arch Linux...

rusty-snake (approved):

PR #5051, opened 2022-03-15
[WIP] Node.js refactoring - again

glitsj16:


kmk3 (commented):
> ```
> D       etc/profile-m-z/nodejs-common.profile
> D       etc/profile-m-z/npm.profile
> D       etc/profile-m-z/nvm.profile
> D       etc/profile-m-z/yarn.profile
> ```

Note that this would be a (potentially hidden) breaking change.  If anyone is
relying on something like npm.local, that wouldn't get picked up by
node.profile.

Also, if you have a shell script wrapper for any of these (e.g.: on ~/bin/npm),
it wouldn't be fully sandboxed anymore, but only the part that executes `node`.

To avoid that, rather than deleting them, they could be turned into redirects
to node.profile instead (so they would keep working the same way as they
currently do).

kmk3:
Relates to:

* #5050

etc/profile-m-z/node.profile
```diff
@@ -1,11 +1,104 @@
...
-include nodejs-common.profile
+# NOTE: gulp, node-gyp, npm, npx, semver and yarn are all node scripts
+# using the `#!/usr/bin/env node` shebang. By sandboxing node the full
+# node.js stack will be firejailed. The only exception is nvm, which is implemented
+# as a sourced shell function, not an executable binary. Hence it is not
+# directly firejailable. You can work around this by sandboxing the programs
+# used by nvm: curl, sha256sum, tar and wget. We have comments in these
+# profiles on how to enable nvm support via local overrides.
+
+blacklist /tmp/.X11-unix
```
rusty-snake, line 18: `s/blacklist /tmp/.X11-unix/include disable-X11.inc/`

glitsj16: Done, thanks!

PR #5054, opened 2022-03-16
cmake: fix local override & wusc

glitsj16:


etc/profile-a-l/cmake.profile
```diff
@@ -1,12 +1,15 @@
-# Firejail profile for cargo
-# Description: The Rust package manager
+# Firejail profile for cmake
+# Description: A cross-platform open-source make system
 # This file is overwritten after every install/update
 quiet
 # Persistent local customizations
-include cargo.local
+include cmake.local
```
rusty-snake, line 6: OT: It's always interesting how often you forget to change stuff if you do the same task multiple times.

glitsj16: Very true :-) Not checking up on your work or anything. Firejailing the build systems had been on my to-do list for a while and I started to implement them one by one only recently. Thanks for the reviews, always a pleasure!

PR #5056, opened 2022-03-17
ocenaudio hardening

glitsj16:


etc/profile-m-z/ocenaudio.profile
```diff
@@ -18,38 +19,44 @@ include disable-programs.inc
...
 nogroups
 noinput
 nonewprivs
+noprinters
 noroot
 notv
 nou2f
 novideo
-protocol unix
+# Add `protocol unix` to your ocenaudio.local to disable networking.
```
rusty-snake, line 48:
```suggestion
# Add `protocol unix\nignore protocol` to your ocenaudio.local to disable networking.
```

rusty-snake (commented):
:heavy_check_mark: when below is fixed.

etc/profile-m-z/ocenaudio.profile
glitsj16: Nice catch! Thanks, fixed it.

PR #5058, opened 2022-03-20
Node.js stack refactoring

glitsj16:
My n-th attempt at implementing a fully sandboxed Node.js stack. I tried to accomodate earlier review comments/suggestions and only dropped nvm.profile as that cannot work as expected anyway (see specific commit comment for context). Although this could be regarded as a breaking change IMO it's better to remove it and motivate users to reflect on its peculiarity instead of suggesting a sense of security simply by having a profile.

Relates to:
- #5050 
- #5051

Actionables:
- close #4172 after merge (I think everything in there is incorporated here)
- drop #5051 (older, not-that-well-informed attempt)


etc/profile-m-z/nodejs-common.profile
```diff
@@ -7,7 +7,14 @@ include nodejs-common.local
 # added by caller profile
 #include globals.local
 
-blacklist /tmp/.X11-unix
+NOTE: gulp, node-gyp, npm, npx, semver and yarn are all node scripts
```
rusty-snake, line 10:
```suggestion
# NOTE: gulp, node-gyp, npm, npx, semver and yarn are all node scripts
```

glitsj16: Ouch. Fixed now, thanks!

netblue30:
merged!

PR #5061, opened 2022-03-21
ping: (extra) hardening  

glitsj16:
This started out as a simple typo fix. By sheer coincidence I noticed there were some comments in the profile that seemed odd to me and I started experimenting. Turns out quite a lot of hardening could be done. I've been running for a while with all of these added options without hiccups.

Makes me wonder if there are any distro differences out there that could explain the former comments. I've left them untouched, because I don't know where they stem from.

### DIFF
--- a/etc/profile-m-z/ping.profile
+++ b/etc/profile-m-z/ping.profile
@@ -7,23 +7,30 @@ include ping.local
 # Persistent global definitions
 include globals.local
 
-blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}
 
 include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
+include disable-proc.inc
 include disable-programs.inc
+include disable-X11.inc
 include disable-xdg.inc
 
 include whitelist-common.inc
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/ping.profile
```diff
@@ -7,52 +7,65 @@ include ping.local
...
 include disable-xdg.inc
 
 include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
+# Add the next line to your ping.local if your kernel allows unprivileged userns clone.
+include ping-hardened.inc.profile
```
rusty-snake, line 28: Did you mean?
```suggestion
#include ping-hardened.inc.profile
```

glitsj16: Yeah, that's what I meant. Fixed, thanks.

kmk3 (commented):
Works for me, so LGTM.

Also, it seems like it can be hardened further; I'll test some more and post
the results.

netblue30:
all in

PR #5066, opened 2022-03-23
megaglest.profile: Add allow-lua.inc

NetSysFire:
It would not start otherwise. It reports:

```
megaglest: error while loading shared libraries: liblua5.1.so.5.1: cannot open shared object file: Permission denied
```

This also fixes a tiny bit of whitespace in the same file. `megaglest_editor` does not appear to be affected so I did not have to modify it. There is also a new binary, `megaglest_g3dviewer`, which does not have a profile yet.

### DIFF
--- a/etc/profile-m-z/megaglest.profile
+++ b/etc/profile-m-z/megaglest.profile
@@ -21,7 +21,8 @@ include disable-xdg.inc
 mkdir ${HOME}/.megaglest
 whitelist ${HOME}/.megaglest
 whitelist /usr/share/megaglest
-whitelist /usr/share/games/megaglest # Debian version
+# Debian version
+whitelist /usr/share/games/megaglest
 include whitelist-common.inc
 include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
--- a/etc/profile-m-z/megaglest.profile
+++ b/etc/profile-m-z/megaglest.profile
@@ -8,6 +8,8 @@ include globals.local
 
 noblacklist ${HOME}/.megaglest
 
+include allow-lua.inc
+
 include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/megaglest.profile
```diff
@@ -19,7 +21,7 @@ include disable-xdg.inc
 mkdir ${HOME}/.megaglest
 whitelist ${HOME}/.megaglest
 whitelist /usr/share/megaglest
-whitelist /usr/share/games/megaglest	# Debian version
+whitelist /usr/share/games/megaglest # Debian version
```
rusty-snake, line 24: It's common to separate same-line comments with more than one space.

rusty-snake: [PEP 8](https://peps.python.org/pep-0008/) for example

> An inline comment is a comment on the same line as a statement. Inline comments should be separated by at least two spaces from the statement. They should start with a # and a single space.

NetSysFire: It actually looks like it was even a tab. In any case, I did a quick regex grep and it would be the only comment after a directive, so I moved it to a dedicated line.

rusty-snake (approved):
LGTM

---

IMHO we don't need comments like this anyway.

etc/profile-m-z/megaglest.profile
```diff
@@ -19,7 +21,8 @@ include disable-xdg.inc
 mkdir ${HOME}/.megaglest
 whitelist ${HOME}/.megaglest
 whitelist /usr/share/megaglest
-whitelist /usr/share/games/megaglest	# Debian version
+# Debian version
```
glitsj16, line 24: As already suggested, let's just drop this comment.

glitsj16 (changes requested):
Nitpick: drop the comment alltogether.

etc/profile-m-z/megaglest.profile
NetSysFire: I am against this since I am on a distribution that is not debian and the comment helps understand why this path is whitelisted

glitsj16: Alrighty, point taken.

glitsj16 (approved):
Let's merge it. Thanks!

Issue #1646, opened 2017-11-14
Add p11-kit support


lhernanz:
Hi,

I would love to have p11-kit support added in the whitelist-common file. I would propose the following:
```
whitelist ~/.config/pkcs11
read-only ~/.config/pkcs11
```
From their web [https://p11-glue.freedesktop.org/](https://p11-glue.freedesktop.org/):
> This is an effort to use and promote PKCS#11 as glue between crypto libraries and security applications on the open source desktop.

As a security related project, I think it would be of interest for firejail users for it to be supported. The above directory is where the user configuration is stored. All the applications using the p11 library (browsers, ssh, etc.) need to access to those files. This is the reason I am proposing those lines to go to the common file. The configurations file there are not very sensitive, so having global access should not be a huge problem.

Thanks
Regards

netblue30:
Will do!

PR #5069, opened 2022-03-24
whitelist-common.inc: remove redundant read-only entries

kmk3:
They are already present on disable-common.inc.

Misc: read-only entries were first added to disable-common.inc on commit
2b37849db ("Protect shell startup files", 2015-10-30) and to
whitelist-common.inc on commit 695b67f43 ("handle
~/.config/user-dirs.dirs", 2015-11-17).

### DIFF
--- a/etc/inc/whitelist-common.inc
+++ b/etc/inc/whitelist-common.inc
@@ -12,14 +12,11 @@ whitelist ${HOME}/.config/mimeapps.list
 whitelist ${HOME}/.config/pkcs11
 read-only ${HOME}/.config/pkcs11
 whitelist ${HOME}/.config/user-dirs.dirs
-read-only ${HOME}/.config/user-dirs.dirs
 whitelist ${HOME}/.config/user-dirs.locale
-read-only ${HOME}/.config/user-dirs.locale
 whitelist ${HOME}/.drirc
 whitelist ${HOME}/.icons
 ?HAS_APPIMAGE: whitelist ${HOME}/.local/share/appimagekit
 whitelist ${HOME}/.local/share/applications
-read-only ${HOME}/.local/share/applications
 whitelist ${HOME}/.local/share/icons
 whitelist ${HOME}/.local/share/mime
 whitelist ${HOME}/.mime.types


kmk3:
Note that there is still one read-only entry on whitelist-common.inc:

```firejail
whitelist ${HOME}/.config/pkcs11
read-only ${HOME}/.config/pkcs11
```

Added on commit f74cfd07c ("add p11-kit support - #1646", 2017-11-23) by
@smitsohu.

Why is ~/.config/pkcs11 read-only?  What kinds of files are supposed to be in
it?  Do they allow arbitrary command execution?  Can the entry be moved to
disable-common.inc?

glitsj16:
I don't have all the answers, but here's my take on them. It all starts out with the importance of [PKCS11](https://wikiless.org/wiki/PKCS_11) for cryptography, including hardware security modules (HSM) and smart cards.

> Why is ~/.config/pkcs11 read-only?

Safe programming? Any process that might try to write to that location (to change loaded crypto modules for example) should be stopped in its tracks. So IMO it's a very good rule to have in a sandbox.

> What kinds of files are supposed to be in it?

Per-user overrides/additions to /etc/pkcs11.

> Do they allow arbitrary command execution?

Don't know. But knowing what sort of changes might be possible cfr. the above I think it's safe to assume a YES here.

> Can the entry be moved to disable-common.inc?

IMO it can be yes. Did a quick check and haven't found any profile that uses whitelist-common.inc without including disable-common.inc (which would present a case for keeping it in whitelist-common.inc I suppose).

PR by itself looks good :+1: .

glitsj16 (approved):
LGTM

rusty-snake:
```console
$ grep -L "^include disable-common.in" $(grep -l "^include whitelist-common.inc" /etc/firejail/*.profile)
~/etc/firejail/fossamail.profile~
~/etc/firejail/icedove.profile~
/etc/firejail/Xephyr.profile
~/etc/firejail/xlinks.profile~
/etc/firejail/Xvfb.profile
```

kmk3:
@glitsj16 commented [on Mar 24](https://github.com/netblue30/firejail/pull/5069#issuecomment-1077040002):

> I don't have all the answers, but here's my take on them. It all starts out
> with the importance of [PKCS11](https://wikiless.org/wiki/PKCS_11) for
> cryptography, including hardware security modules (HSM) and smart cards.
> 
> > Why is ~/.config/pkcs11 read-only?
> 
> Safe programming? Any process that might try to write to that location (to
> change loaded crypto modules for example) should be stopped in its tracks. So
> IMO it's a very good rule to have in a sandbox.
> 
> > What kinds of files are supposed to be in it?
> 
> Per-user overrides/additions to /etc/pkcs11. Some countries use e-ID
> (electronic identity card) viewers and Firefox extensions for example. I
> happen to have one of those and it's reassuring to know loading of its module
> can't be changed in a firejail sandbox. Just one example I'm aware of.

You mention overrides/additions to /etc/pkcs11 and I see that [pkcs11(5)][1]
mentions ~/.config/pkcs11/pkcs11.conf.  So does it store only config files?
Would it also store something like user certificates or maybe symlinks to
hardware keys in /dev?

It would be nice to know the extent of what exactly is supposed to be in there,
for example, to see where it would make sense to add the entry on
disable-common.inc.

> > Do they allow arbitrary command execution?
> 
> Don't know. But knowing what sort of changes might be possible cfr. the above
> I think it's safe to assume a YES here.
> 
> > Can the entry be moved to disable-common.inc?
> 
> IMO it can be yes. Did a quick check and haven't found any profile that uses
> whitelist-common.inc without including disable-common.inc (which would
> present a case for keeping it in whitelist-common.inc I suppose).
> 
> PR by itself looks good +1 .

Thanks for all the details.

Considering that the following packages depend on it:

```console
$ pacman -Qo /etc/pkcs11/
/etc/pkcs11/ is owned by p11-kit 0.24.1-1
$ pactree -ars -d 1 p11-kit
p11-kit
|-ca-certificates-utils
|-gcr
|-gnutls
|-nss
`-softhsm
```

It could potentially be used by a ton of programs.

So how common is it (/what would be the use case) for programs that depend on
these (such as firefox, which depends on nss) to create/modify files in
~/pkcs11?

Additionally, disable-common.inc is way more commonly included than
whitelist-common.inc, so making it read-only on the former could end up
affecting/breaking many programs:

```console
$ git grep '^include disable-common\.inc' -- etc | wc -l
608
$ git grep '^include whitelist-common\.inc' -- etc | wc -l
293
```

[1]: https://man.archlinux.org/man/core/p11-kit/pkcs11.conf.5.en

kmk3:
@rusty-snake commented [on Mar 24](https://github.com/netblue30/firejail/pull/5069#issuecomment-1077929349):

> ```
> $ grep -L "^include disable-common.in" $(grep -l "^include whitelist-common.inc" /etc/firejail/*.profile)
> ~/etc/firejail/fossamail.profile~
> ~/etc/firejail/icedove.profile~
> ```

These redirect to firefox.profile -> firefox-common.profile, which does include
disable-common.inc.

> ~/etc/firejail/xlinks.profile~

Redirects to links.profile -> links-common.profile, which does include
disable-common.inc.

> /etc/firejail/Xephyr.profile
> /etc/firejail/Xvfb.profile

Since these whitelist things in the homedir anyway, why don't they also include
disable-common.inc?  Or is it supposed to (for some reason) protect only the X
server and nothing else in the homedir?

Misc: What's going on with the `~`s?  Is it supposed to mean
strikethrough/ignore?  If so, I'd suggest leaving these lines out of the
codeblock (i.e.: leave only the grep line in it) to make it clearer.

rusty-snake:
> Misc: What's going on with the ~s? Is it supposed to mean
strikethrough/ignore? If so, I'd suggest leaving these lines out of the
codeblock (i.e.: leave only the grep line in it) to make it clearer.

FYI
Actually you strikethrough in codeblocks but is more work.

```markdown

<pre><code>A
<s>B</s>
C</code></pre>
```

<pre><code>A
<s>B</s>
C</code></pre>

kmk3:
> > /etc/firejail/Xephyr.profile
> > /etc/firejail/Xvfb.profile
> 
> Since these whitelist things in the homedir anyway, why don't they also
> include disable-common.inc? Or is it supposed to (for some reason) protect
> only the X server and nothing else in the homedir?

Nevermind, I see now that they use `private`.  Still, any specific reason not
to use disable-common.inc?

Maybe avoiding a large include makes the startup a bit faster, though I think
that it would make more sense for something named whitelist-common.inc to only
allow things rather both allow and restrict (and it would also result in less
duplication of entries).

kmk3:
@rusty-snake commented [on Mar 24](https://github.com/netblue30/firejail/pull/5069#issuecomment-1077965429):

> > Misc: What's going on with the ~s? Is it supposed to mean
> > strikethrough/ignore? If so, I'd suggest leaving these lines out of the
> > codeblock (i.e.: leave only the grep line in it) to make it clearer.
> 
> FYI Actually you strikethrough in codeblocks but is more work.
> 
> ```gfm
> <pre><code>A
> <s>B</s>
> C</code></pre>
> ```
> 
> ```
> A
> B
> C
> ```

Nice, never thought of that.  For reference, I meant something like this:

```
$ grep -L "^include disable-common.in" $(grep -l "^include whitelist-common.inc" /etc/firejail/*.profile)
```

* ~/etc/firejail/fossamail.profile~
* ~/etc/firejail/icedove.profile~
* /etc/firejail/Xephyr.profile
* ~/etc/firejail/xlinks.profile~
* /etc/firejail/Xvfb.profile

But I would probably just do this for simplicity:

> ```
> $ grep -L "^include disable-common.in" $(grep -l "^include whitelist-common.inc" /etc/firejail/*.profile)
> /etc/firejail/fossamail.profile # ignore
> /etc/firejail/icedove.profile # ignore
> /etc/firejail/Xephyr.profile
> /etc/firejail/xlinks.profile # ignore
> /etc/firejail/Xvfb.profile
> ```

glitsj16:
> You mention overrides/additions to /etc/pkcs11 and I see that pkcs11(5)
mentions ~/.config/pkcs11/pkcs11.conf. So does it store only config files?
Would it also store something like user certificates or maybe symlinks to
hardware keys in /dev?

> It could potentially be used by a ton of programs.
So how common is it (/what would be the use case) for programs that depend on
these (such as firefox, which depends on nss) to create/modify files in
~/pkcs11?

~/.config/pkcs11 contains only config files. Maybe you were thinking about ~/.pki from [NSS](https://wiki.archlinux.org/title/Network_Security_Services) here?

kmk3:
@glitsj16 commented [on Mar 24](https://github.com/netblue30/firejail/pull/5069#issuecomment-1078418130):

> > You mention overrides/additions to /etc/pkcs11 and I see that pkcs11(5)
> > mentions ~/.config/pkcs11/pkcs11.conf. So does it store only config files?
> > Would it also store something like user certificates or maybe symlinks to
> > hardware keys in /dev?
> 
> > It could potentially be used by a ton of programs.
> > So how common is it (/what would be the use case) for programs that depend on
> > these (such as firefox, which depends on nss) to create/modify files in
> > ~/pkcs11?
> 
> ~/.config/pkcs11 contains only config files.

Nice, so if no one is aware of any program that modifies this path, then I
suppose that there's indeed nothing blocking it from being added to
disable-common.inc.

Nevertheless, I still want to see what can be done about Xephyr.profile and
Xvfb.profile beforehand. 

> Maybe you were thinking about ~/.pki from
> [NSS](https://wiki.archlinux.org/title/Network_Security_Services) here?

Yes, this is the closest thing that I know of, so I thought that
~/.config/pkcs11 could contain similar files.  Interestingly, it looks like
certain nss-based programs support (or at least did support) pkcs11-related
config files in ~/.pki and /etc/pki:

* ~/.pki/nssdb/pkcs11.txt
* /etc/pki/nssdb/pkcs11.txt

kmk3:
Closing for now, as both Xephyr.profile and Xvfb.profile include
whitelist-common.inc but not disable-common.inc.

PR #5092, opened 2022-04-04
harden vlc

smitsohu:
`apparmor` doesn't disable D-Bus anymore, so add it back to the profile. Remove `memory-deny-write-execute` comment, as this also breaks JIT compiled QtQuick nowadays.

### DIFF
--- a/etc/profile-m-z/vlc.profile
+++ b/etc/profile-m-z/vlc.profile
@@ -27,9 +27,11 @@ whitelist ${HOME}/.config/aacs
 whitelist ${HOME}/.local/share/vlc
 include whitelist-common.inc
 include whitelist-player-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
 include whitelist-var-common.inc
 
-#apparmor - on Ubuntu 18.04 it refuses to start without dbus access
+apparmor
 caps.drop all
 netfilter
 nogroups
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/vlc.profile
```diff
@@ -45,9 +47,9 @@ private-bin cvlc,nvlc,qvlc,rvlc,svlc,vlc
...
-# dbus-user none
-# dbus-system none
-
-# mdwe is disabled due to breaking hardware accelerated decoding
-#memory-deny-write-execute
+dbus-user filter
+dbus-user.talk org.freedesktop.Notifications
+dbus-user.talk org.freedesktop.ScreenSaver
+dbus-user.talk org.mpris.MediaPlayer2.Player
+dbus-user.own org.mpris.MediaPlayer2.vlc
```
glitsj16, line 54: nitpick: add this line above the 'dbus-user.talk <foo>' lines

glitsj16 (approved):
LGTM. Just one nitpick remark...

etc/profile-m-z/vlc.profile
smitsohu: Will do, thanks!

smitsohu:
Added `?ALLOW_TRAY: dbus-user.talk org.kde.StatusNotifierWatcher`

PR #4725, opened 2021-11-30
Keep some groups regardless of nogroups and restore nogroups on nvidia

kmk3:
```console
$ git log --reverse --pretty='* %s' master..
* Keep audio and video groups regardless of nogroups
* Keep render, lp, input and other groups regardless of nogroups
* Make nogroups work on nvidia again
* etc: Remove comments about nogroups and noroot on nvidia
```

This should fix audio on seatd (#4531/#4603) and also make `nogroups` work
regardless of `no3d` on nvidia (#841/#3644).

Note: This is a continuation of #4632.

Note2: I have not tested this yet.

Cc: @crocket (from #4603) @hlein (from #4632)

PR #5114, opened 2022-04-22
Stop warning on safe supplementary group clean

kmk3:
When nogroups is used, the following warning may be issued (potentially
multiple times, as drop_privs may be called more than once):

But the warning is being shown even when it seems that all supplementary
groups can be safely dropped (and are thus dropped), which is likely a
common scenario.  This commit prevents the warning from being printed in
that case, making it so that it is only shown in the non-happy paths (as
was the case on firejail 0.9.66).

Misc: The added code was copied from drop_privs.

This amends commit 7abce0b4c ("Fix keeping certain groups with
nogroups", 2021-11-30) / PR #4732.

Kind of relates to #4930.

### DIFF
--- a/src/firejail/util.c
+++ b/src/firejail/util.c
@@ -173,11 +173,17 @@ static void clean_supplementary_groups(gid_t gid) {
 	assert(cfg.username);
 	gid_t groups[MAX_GROUPS];
 	int ngroups = MAX_GROUPS;
+
+	if (arg_nogroups && check_can_drop_all_groups()) {
+		if (setgroups(0, NULL) < 0)
+			errExit("setgroups");
+		if (arg_debug)
+			printf("No supplementary groups\n");
+		return;
+	}
+
 	int rv = getgrouplist(cfg.username, gid, groups, &ngroups);
 	if (rv == -1)
-		goto clean_all;
-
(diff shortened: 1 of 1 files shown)


glitsj16 (approved):
Can't really be sure on the coding side of this PR, but it definately deserves full credit for the effort to do something about these superfluous (hence potentially confusing) warnings: +1!

kmk3:
@glitsj16 left a comment:

> Can't really be sure on the coding side of this PR, but it definately
> deserves full credit for the effort to do something about these superfluous
> (hence potentially confusing) warnings: +1!

Thanks; sorry for taking so long to fix this.

This is the one part of the code that I changed without having a good enough
understanding of it and now it's rather messy, but I'm working on fixing that.

There's one more notable bugfix to post and a WIP refactor/unification of group
handling, which should simplify some things a lot.

Though I can't see an easy fix for one key part, which is the fact that
`drop_privs` may be called during argument parsing but (since #4725/#4732) it
depends on `arg_` variables.

If all arguments/profile entries were parsed before doing anything "important",
I think it would make using `arg_` variables a lot safer/more robust.  I
remember someone suggesting something like this in an issue (or maybe in a
comment), but I can't find it at the moment.

netblue30:
all in, thanks!

PR #5115, opened 2022-04-23
Whitelist electron-flags.conf for all versions of electron

m00nwtchr:
Different versions of electron load flags from different files.

### DIFF
--- a/etc/profile-a-l/electron.profile
+++ b/etc/profile-a-l/electron.profile
@@ -5,7 +5,7 @@
 include electron.local
 
 noblacklist ${HOME}/.config/Electron
-noblacklist ${HOME}/.config/electron-flag*.conf
+noblacklist ${HOME}/.config/electron*-flag*.conf
 
 include disable-common.inc
 include disable-devel.inc
@@ -16,7 +16,7 @@ include disable-xdg.inc
 
 whitelist ${DOWNLOADS}
 whitelist ${HOME}/.config/Electron
-whitelist ${HOME}/.config/electron-flag*.conf
+whitelist ${HOME}/.config/electron*-flag*.conf
 include whitelist-common.inc
 include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
--- a/etc/profile-a-l/electron.profile
+++ b/etc/profile-a-l/electron.profile
@@ -5,7 +5,7 @@
 include electron.local
 
 noblacklist ${HOME}/.config/Electron
-noblacklist ${HOME}/.config/electron*-flag*.conf
+noblacklist ${HOME}/.config/electron[0-9]*-flag*.conf
 
 include disable-common.inc
 include disable-devel.inc
@@ -16,7 +16,7 @@ include disable-xdg.inc
 
 whitelist ${DOWNLOADS}
 whitelist ${HOME}/.config/Electron
-whitelist ${HOME}/.config/electron*-flag*.conf
+whitelist ${HOME}/.config/electron[0-9]*-flag*.conf
 include whitelist-common.inc
 include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
--- a/etc/profile-a-l/electron.profile
+++ b/etc/profile-a-l/electron.profile
@@ -5,7 +5,7 @@
 include electron.local
 
 noblacklist ${HOME}/.config/Electron
-noblacklist ${HOME}/.config/electron*([0-9])-flag*.conf
+noblacklist ${HOME}/.config/electron*-flag*.conf
 
 include disable-common.inc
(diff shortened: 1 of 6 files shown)


etc/profile-a-l/electron.profile
```diff
@@ -5,7 +5,7 @@
 include electron.local
 
 noblacklist ${HOME}/.config/Electron
-noblacklist ${HOME}/.config/electron-flag*.conf
+noblacklist ${HOME}/.config/electron[0-9]*-flag*.conf
```
rusty-snake, line 8: Never blacklisted (update disable-programs.inc).

etc/profile-a-l/electron.profile
```diff
@@ -16,7 +16,7 @@ include disable-xdg.inc
 
 whitelist ${DOWNLOADS}
 whitelist ${HOME}/.config/Electron
-whitelist ${HOME}/.config/electron-flag*.conf
+whitelist ${HOME}/.config/electron[0-9]*-flag*.conf
```
rusty-snake, line 19: This will break all program which `ignore` the "old" line.

Update

```console
$ grep 'ignore whitelist ${HOME}/.config/electron-flag\*.conf' /etc/firejail/*
/etc/firejail/atom.profile:ignore whitelist ${HOME}/.config/electron-flag*.conf
/etc/firejail/code.profile:ignore whitelist ${HOME}/.config/electron-flag*.conf
/etc/firejail/github-desktop.profile:ignore whitelist ${HOME}/.config/electron-flag*.conf
/etc/firejail/notable.profile:ignore whitelist ${HOME}/.config/electron-flag*.conf
```

rusty-snake, line 19: What is the format for electron-flag files? You edit looks like you maybe tried to use a regexp. Firejail uses `glob(3)` for globbing, the format is described in `glob(7)`.

This will match `electron` + one digit + anything (zero or more of anything) + `-flag` + anything + `.conf`.

rusty-snake (changes requested):

glitsj16:
Thanks for the PR. I like your format over ${HOME}/.config/electron-flag*.conf. The latter though is also present in:

- [atom.profile](https://github.com/netblue30/firejail/blob/master/etc/profile-a-l/atom.profile)
- [code.profile](https://github.com/netblue30/firejail/blob/master/etc/profile-a-l/code.profile)
- [disable-programs.inc](https://github.com/netblue30/firejail/blob/master/etc/inc/disable-programs.inc)
- [github-desktop.profile](https://github.com/netblue30/firejail/blob/master/etc/profile-a-l/github-desktop.profile)
- [notable.profile](https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/notable.profile)

Can you change those files as well please?

etc/profile-a-l/electron.profile
m00nwtchr: It is `electron<version>-flags.conf`, i wasn't sure if maybe some older version also used `-flag.conf` so i kept the `*` there.
Yes, that was the intended pattern. (Since it needs to match both `electron-flags and `electron16-flags` for example, because 1. some versions of electron do use the `electron-flags` file IIRC and 2. the `electron<version>-flags` files are often symlinked to this one common file for convenience).

etc/profile-m-z/notable.profile
```diff
@@ -27,7 +27,7 @@ ignore dbus-user none
 # Notable keeps claiming it is started for the first time when whitelisting - see #4812.
 ignore whitelist ${DOWNLOADS}
 ignore whitelist ${HOME}/.config/Electron
-ignore whitelist ${HOME}/.config/electron-flag*.conf
+ignore ${HOME}/.config/electron*([0-9])-flag*.conf
```
rusty-snake, line 30:
```suggestion
ignore whitelist ${HOME}/.config/electron*([0-9])-flag*.conf
```

rusty-snake:
Did you actually tested this?

Note again firejail uses `glob(3)` with the rules described in `glob(7)`. Neither regex nor bash's extglob is supported.

---

I think we have to go with

```
whitelist ${HOME}/.config/electron-flag*.conf
whitelist ${HOME}/.config/electron[[:digit:]]-flag.conf
whitelist ${HOME}/.config/electron[[:digit:]][[:digit:]]-flag.conf
```

or just

```
whitelist ${HOME}/.config/electron*-flag*.conf
```

which would also cover things like `electron14.5a1-flags.conf` if such a format is ever introduced.

rusty-snake (approved):

glitsj16 (approved):
LGTM

netblue30:
all in, thanks!

PR #5131, opened 2022-05-06
Whitelist for NixOS to resolve binary paths in user environment

onny:
This PR fixes an issue on NixOS where you cannot use firejail like this:
```
firejail chromium
```
it fails with `Error: no suitable chromium executable found`.

This is because `chromium` is usually a symlink in `~/.nix-profile` to the binary in the NixOS store. After whitelisting `~/.nix-profile` running 
```
firejail chromium
```
works :)

Solves downstream issue https://github.com/NixOS/nixpkgs/issues/170784


rusty-snake (approved):

glitsj16 (approved):
LGTM. Nice work!

PR #5133, opened 2022-05-07
configure*: remove ultimately unused INSTALL and RANLIB check macros

kmk3:
From the manual of GNU Autoconf (version 2.69):

>  -- Macro: AC_PROG_INSTALL
>      Set output variable 'INSTALL' to the name of a BSD-compatible
>      'install' program, if one is found in the current 'PATH'.
>      Otherwise, set 'INSTALL' to 'DIR/install-sh -c', checking the
>      directories specified to 'AC_CONFIG_AUX_DIR' (or its default
>      directories) to determine DIR (*note Output::).  Also set the
>      variables 'INSTALL_PROGRAM' and 'INSTALL_SCRIPT' to '${INSTALL}'
>      and 'INSTALL_DATA' to '${INSTALL} -m 644'.

>  -- Macro: AC_PROG_RANLIB
> 	 Set output variable 'RANLIB' to 'ranlib' if 'ranlib' is found,
> 	 and otherwise to ':' (do nothing).

None of the aforementioned variables are used:

    $ git grep -F -e '${INSTALL}' -e INSTALL_PROGRAM -e INSTALL_SCRIPT \
      -e INSTALL_DATA -e RANLIB
    $

So remove the macros that define them.

Misc: The macros in question have been present on configure.ac since it
was created, on commit 137985136 ("Baseline firejail 0.9.28",
2015-08-08).  And while the install command is called multiple times,
ranlib is not used anywhere (and it seems that it was never used):

    $ git grep -E '^[[:blank:]]+install ' -- '*Makefile*' '*.mk*' |
      wc -l
    32
    $ git grep -F ranlib | wc -l
    0
    $ git log --pretty= --name-only -G'RANLIB|ranlib' \
      137985136..master | sort -u
    README.md

Kind of relates to #4695.


kmk3:
By the way, I'm working on other build-related improvements (involving the
makefiles).

PR #5140, opened 2022-05-12
build: deduplicate configure-time vars into new config files

kmk3:
makefiles: deduplicate configure-time vars into new config.mk.in

Currently, the configure-time variables (that is, the ones that assign
to placeholders, such as "@HAVE_MAN@", which are set/replaced at
configure-time) are defined on multiple files (such as on Makefile.in
and on common.mk.in).

To avoid duplication, centralize these variables on a single file
(config.mk.in) and replace all of the other definitions of them with an
include of config.mk.

---

mkdeb.sh.in: move configure-time vars into new config.sh.in

For better organization and so that they can be used by other shell
scripts by just sourcing config.sh.

---

Relates to #646.


kmk3:
@reinerh Could you test whether `mkdeb.sh.in` still works as intended in this
branch?

reinerh:
@kmk3 I don't use the `mkdeb.sh` script, but I just tried to run it and got an error:
```
$ ./configure
...
$ make deb
...
./mkdeb.sh
./mkdeb.sh: 11: .: cannot open ./config.sh: No such file
Makefile:211: recipe for target 'deb' failed
```

kmk3:
@reinerh commented [on May 13](https://github.com/netblue30/firejail/pull/5140#issuecomment-1126338707):

> @kmk3 I don't use the `mkdeb.sh` script, but I just tried to run it and got an error:
> 
> ```
> $ ./configure
> ...
> $ make deb
> ...
> ./mkdeb.sh
> ./mkdeb.sh: 11: .: cannot open ./config.sh: No such file
> Makefile:211: recipe for target 'deb' failed
> ```

Good catch; with #5142 this should not be an issue anymore.

Considering #5142, #5154 and contrib/fj-mkdeb.py, changing mkdeb.sh.in is more
complicated than I anticipated, so I'll leave that for later.  This branch only
focuses on makefiles now (which was the original idea anyway).

kmk3:
> Considering #5142, #5154 and contrib/fj-mkdeb.py, changing mkdeb.sh.in is
> more complicated than I anticipated, so I'll leave that for later.

Actually, I now realized that contrib/fj-mkdeb.py does more than a work around;
it also automates building the .deb package, so I'll try #5154 one more time
before this PR.

kmk3:
Depends on #5142 (again) and on #5182.

Rebased to master to include the changes from those PRs.

kmk3:
So, the next PR is the one that does the key changes and I want to reference it
in a few (more) comments.

Since this was opened over a month ago (though admittedly it stayed most of the
time as a draft) and since there have been no reviews for about the same amount
of time, I'm going to go ahead and merge this.

kmk3:
(Rebased to master)

PR #5142, opened 2022-05-13
Makefile.in: stop running distclean on dist

kmk3:
distclean erases user-made modifications to the files generated by the
configure script; it is not obvious that a non-"clean" target would ever
run it.

That is, instead of:

    ./configure --enable-foo && make distclean && make dist

It would probably be safer (and more maintainable) to do:

    make distclean && ./configure --enable-foo && make dist

So drop the distclean call and the file-moving hacks around it.

Added on commit 137985136 ("Baseline firejail 0.9.28", 2015-08-08).

Relates to commit 684919100 ("bug: preserve config.status during make
dist", 2016-09-19) and commit 31dc1218a ("fixups", 2020-07-19).


reinerh (approved):

kmk3:
Misc: Used a normal merge for this because I refer to this PR on other commits.

PR #5158, opened 2022-05-22
seamonkey.profile: support enigmail/gpg

hknaack:
Changes inspired by Thunderbird profile.


hknaack:
The Thunderbird profile also contains `whitelist /usr/share/gnupg`, but that causes Seamonkey to fail with a segmentation fault on startup:

```
/etc/magic, 3: Warning: using regular magic file `/usr/share/misc/magic'
/usr/bin/seamonkey: line 140:    12 Segmentation fault      (core dumped) $MOZ_PROGRAM "$@"
```

rusty-snake (changes requested):
Remove the unused `ignore`s and move `writeable-run-user` down as per template.

hknaack:
> Remove the unused `ignore`s and move `writeable-run-user` down as per template.

I removed all ignore statements, and it seems to work fine, too. Though I'm not experienced enough to assess if that will cause some side effects or if they are also unneeded in the Thunderbird profile.

etc/profile-m-z/seamonkey.profile
```diff
@@ -50,6 +53,7 @@ notv
 protocol unix,inet,inet6,netlink
 seccomp
 tracelog
+writable-run-user
 
 disable-mnt
 # private-etc adobe,alternatives,asound.conf,ca-certificates,crypto-policies,firefox,fonts,group,gtk-2.0,hostname,hosts,iceweasel,localtime,machine-id,mailcap,mime.types,nsswitch.conf,pango,passwd,pki,pulse,resolv.conf,ssl
```
rusty-snake, line 59:
```suggestion

disable-mnt
# private-etc adobe,alternatives,asound.conf,ca-certificates,crypto-policies,firefox,fonts,group,gtk-2.0,hostname,hosts,iceweasel,localtime,machine-id,mailcap,mime.types,nsswitch.conf,pango,passwd,pki,pulse,resolv.conf,ssl
writable-run-user
```

hknaack: Done.

rusty-snake (approved):

netblue30:
thanks for the patch!

PR #5161, opened 2022-05-23
Update transmission profile for DBus notifications

sinner-:
While investigating why transmission-gtk seems slow when launched through firejail, I noticed the current profile blocks access to org.freedesktop.Notifications.

This profile PR fixes that, although transmission-gtk still seems to take ages to load when firejail'ed

### DIFF
--- a/etc/profile-m-z/transmission-common.profile
+++ b/etc/profile-m-z/transmission-common.profile
@@ -48,7 +48,9 @@ private-dev
 private-lib
 private-tmp
 
-dbus-user none
+dbus-user filter
+dbus-user.own com.transmissionbt.Transmission.*
+dbus-user.talk org.freedesktop.Notifications
 dbus-system none
 
 memory-deny-write-execute


etc/profile-m-z/transmission-common.profile
```diff
@@ -48,7 +48,9 @@ private-dev
 private-lib
 private-tmp
 
-dbus-user none
+dbus-user filter
+dbus-user.own com.transmissionbt.Transmission.*
+dbus-user.talk org.freedesktop.Notifications
```
rusty-snake, line 53: Is unsafe and can be used to escape the sandbox (on the most systems). Do we want it by default or as an `# If you need native notifications, add the next line to your transmission.local`? Also do we want it for all transmission programs or only the qt/gtk version?

sinner-: Oh if you can use it for sandbox escape then I don't want it!

Do you have any links about how allowing it to make notifications allows for sandbox escape (just for my own curiosity)?

rusty-snake: No links (except https://github.com/netblue30/firejail/wiki/Restrict-DBus), but the issues is rather simple. The notifications API (and spec) for it's own is save AFAICTY. However the most common implementations expose all their objects on all their names. Save is only GNOME >= 3.36.1 and (unchecked) WM focused implementations like dunst.

If you want to check your system: `busctl --user tree org.freedesktop.Notifications` shoud only show 
```
└─/org
  └─/org/freedesktop
    └─/org/freedesktop/Notifications
```

sinner-: Thanks @rusty-snake ...mine (on Fedora 36 running XFCE) looks like 

```
$ busctl --user tree org.freedesktop.Notifications
└─/org
  └─/org/freedesktop
    └─/org/freedesktop/Notifications
```

does that mean having 
```
dbus-user filter
dbus-user.own com.transmissionbt.Transmission.*
```

in my Transmission profile is safe?

rusty-snake: Yes. I've even added `dbus-user.talk org.freedesktop.Notifications` to my globals.local.

Would be interesting if KDE, Mate, Cinnamon are safe nowadays too.

glitsj16: > Also do we want it for all transmission programs or only the qt/gtk version?

IMO it doesn't make much sense in the CLI parts of transmission, so I'd go for touching gtk/qt only. Just my two cents.

sinner-:
I'm going to close this PR because I don't want to cause sandbox escapes for distros that might have more exposed DBus settings.

PR #5159, opened 2022-05-22
Kate fixes

hknaack:
Reduce restrictions, so that Kate can be used as intended by the developers:

- allow usage of Git
- permit Bash completion in built-in Konsole
- permit the requested access to kwinrc configuration

PR #5167, opened 2022-05-30
profiles: move blacklist of /etc/profile.d & blacklist /etc/profile

kmk3:
disable-common.inc: move blacklist of /etc/profile.d

To disable-shell.inc.

Interactive shells can be executed from certain development-related
programs (such as IDEs) and the shells themselves are not blocked by
default, but this shell startup directory currently is.  To avoid
running a shell without access to potentially needed startup files, only
blacklist /etc/profile.d when interactive shells are also blocked.

Note that /etc/profile.d should only be of concern to interactive
shells, so a profile that includes both disable-shell.inc and
allow-bin-sh.inc (which likely means that it needs access to only
non-interactive shells) should not be affected by the blacklisting.

Relates to #3411 #5159.

Cc: @hknaack (from #5159).

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -261,7 +261,6 @@ blacklist /etc/grub*
 blacklist /etc/kernel*
 blacklist /etc/logrotate*
 blacklist /etc/modules*
-blacklist /etc/profile.d
 blacklist /etc/rc.local
 # rc1.d, rc2.d, ...
 blacklist /etc/rc?.d
--- a/etc/inc/disable-shell.inc
+++ b/etc/inc/disable-shell.inc
@@ -13,4 +13,5 @@ blacklist ${PATH}/sh
 blacklist ${PATH}/tclsh
 blacklist ${PATH}/tcsh
 blacklist ${PATH}/zsh
+blacklist /etc/profile
 blacklist /etc/profile.d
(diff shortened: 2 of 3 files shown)


glitsj16 (approved):
LGTM

rusty-snake (approved):
LGTM

- Does anyone know why we blacklist them?
- Should we add more shell startup files like `/etc/bashrc`, `/etc/zpfoile`, ...?

kmk3:
@rusty-snake left a comment:

> LGTM
> 
> *  Does anyone know why we blacklist them?

Not exactly sure; it was added on commit 5db7520b2 ("profile work", 2015-09-22)
by @netblue30.

Maybe it's done as an extra layer of defense, as a program overwriting these
files is likely as bad as it overwriting ~/.bashrc.

> *  Should we add more shell startup files like `/etc/bashrc`, `/etc/zpfoile`,
> ...?

I think that would make sense.

Note that /etc/profile is listed on etc/ids.config, along with other such
paths:

* <https://github.com/netblue30/firejail/blob/7cf7108e25ce73fe6c26169c6e4886466f287760/etc/ids.config#L61>

```
### shells global ###
# all
/etc/dircolors
/etc/environment
/etc/profile
/etc/profile.d
/etc/shells
/etc/skel
# bash
/etc/bash_completion*
/etc/bash.bashrc
/etc/bashrc
# fish
/etc/fish
# ksh
/etc/ksh.kshrc
# tcsh
/etc/complete.tcsh
/etc/csh.cshrc
/etc/csh.login
/etc/csh.logout
# zsh
/etc/zlogin
/etc/zlogout
/etc/zprofile
/etc/zshenv
/etc/zshrc
```

netblue30:
Short attention span here, doesn't go as far back as 2015! So let's move them!

PR #5176, opened 2022-06-05
mkdeb.sh.in: stop enabling apparmor

kmk3:
Since `make deb-apparmor` already exists, use that for now instead of
changing what `make deb` does.

This fixes CI.

Added on commit 494b26d50 ("adding --enable-apparmor by default for make
deb - most Debian-based distros have apparmor enabled by default",
2022-06-03).

Kind of relates to #5154.

### DIFF
--- a/README
+++ b/README
@@ -34,7 +34,7 @@ $ sudo apt-get install git build-essential libapparmor-dev pkg-config gawk
 For --selinux option, add libselinux1-dev (libselinux-devel for Fedora).
 
 We build our release firejail.tar.xz and firejail.deb packages using the following command:
-$ make distclean && ./configure && make deb
+$ make distclean && ./configure && make deb-apparmor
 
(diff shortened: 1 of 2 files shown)


reinerh (approved):

kmk3:
Note: I think it would be nice to add `--enable-apparmor` to mkdeb.sh.in and to
then remove `make deb-apparmor`, as it would make things simpler.  I thought
about doing that on #5154, but I wasn't too sure about it and I wanted to avoid
making any breaking changes in that PR.

Since the upcoming release is near and since it is meant to be about important
bugfixes, I think it would be better to leave this change for later, to
minimize breaking changes and also because it would probably require some
reorganization in .gitlab-ci.yml to make it work without it looking confusing.

netblue30:
Fixed!

PR #5182, opened 2022-06-08
Revert "Makefile.in: stop running distclean on dist"

kmk3:
This reverts commit 1fb814e51149d105233f1edc1abb0de202f71b4d.

If distclean is not executed before copying the files on dist, then the
generated files inside src/ are included in the dist archive:

    $ ./configure >/dev/null && make distclean >/dev/null &&
      ./configure >/dev/null && make dist | grep 'Makefile$' | wc -l
    26

This happens because src/ is copied wholesale on dist (see DISTFILES).
Revert the commit to ensure that only the input files (such as the
"Makefile.in" files) are archived.

Related discussion:
https://github.com/netblue30/firejail/pull/5154#pullrequestreview-980810845

Relates to #5142.


netblue30:
merged!

PR #5187, opened 2022-06-09
steam.profile : add additional folder for Paradox Interactive launcher 

swimik:
On Debian bullseye with Steam installed via Debian repo and a game installed that uses Paradox Interactive Launcher a folder is required to be whitelisted for the launcher to open.
Without the folder being whitelisted the Paradox Launcher will fail to load with a 'permission denied' error and it will ask to reinstall the launcher, which also fails.

Adding this folder to the whitelist corrects the problem

whitelist ${HOME}/.paradoxlauncher


Serphentas:
Can I submit a new PR with the proper basis or do we fix this one ?

PR #5202, opened 2022-06-14
profiles: nodejs-common: fix node-gyp failing to build canvas

skrat:
So there is a commonly use node (npm) package called "canvas", which uses node-gyp to bind to cairo (2D canvas drawing library). And I assume this is useful outside of this particular case. node-gyp (run by `yarn install` or `npm install`) needs access to `/usr/share/pkgconfig` otherwise it fails to buold the native extensions.

So it's either this proposed change, or it can be move to `yarn.profile` or `npm.profile`. I'm not sure it will work when moved to `node-gyp.profile` as that one is invoked as a subprocess.

It can also, instead of including `build-systems-common.profile` just whitelist `/usr/share/pkgconfig`.

### DIFF
--- a/etc/profile-m-z/nodejs-common.profile
+++ b/etc/profile-m-z/nodejs-common.profile
@@ -67,6 +67,7 @@ include disable-xdg.inc
 whitelist /usr/share/doc/node
 whitelist /usr/share/nvm
 whitelist /usr/share/systemtap/tapset/node.stp
+include build-systems-common.profile
 include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc


etc/profile-m-z/nodejs-common.profile
```diff
@@ -67,6 +67,7 @@ include disable-xdg.inc
 whitelist /usr/share/doc/node
 whitelist /usr/share/nvm
 whitelist /usr/share/systemtap/tapset/node.stp
+include build-systems-common.profile
```
rusty-snake, line 70: - This should go to the end of the file, see https://github.com/netblue30/firejail/blob/master/etc/templates/redirect_alias-profile.template.
- All command set by `build-systems-common.profile` should be removed here.
- All command unintentional introduced by this must be `ignore`d.

> > It can also, instead of including `build-systems-common.profile` just whitelist `/usr/share/pkgconfig`.

Likely easier than the above.

rusty-snake (changes requested):

ghost (changes requested):
There's considerable overlap between nodejs-common and build-systems-common. IMO it's asking for trouble by including the latter in the former. I'll do some testing shortly but adding `whitelist /usr/share/pkgconfig` to nodejs-common.profile is the cleaner option.

> Fix node-gyp failing to build canvas

Do you have a reproducer? Or a more detailed log of what is actually failing?

rusty-snake:
> I'm fine with just whitelist /usr/share/pkgconfig. Should I redo?

Yes

glitsj16:
> Any yarn npm project that has canvas as a dependency. I'm fine with just whitelist /usr/share/pkgconfig. Should I redo?

If you mean redo the PR to add whitelist /usr/share/pkgconfig to nodejs-common.profile instead of what you have now: yes please do :-) You might need to rebase for the CI check.

ghost (changes requested):
Can you redo? Add `whitelist /usr/share/pkgconfig` and we can bring this in.

PR #5206, opened 2022-06-16
build: fix file mode of shell scripts (644 -> 755)

kmk3:
Some shell scripts are not executable, so fix their file modes:

    $ git grep -Elz '^#!/bin/(ba)?sh' | xargs -0 -I '{}' \
      chmod +x '{}'

Files changed:

* src/fgit/fgit-uninstall.sh
* src/tools/extract_errnos.sh

Note: I have manually checked that the files above are indeed intended
to be executable directly and not just sourced, as a script of the
latter kind could also contain a shebang (for example, to help ensure
proper syntax highlighting).

Misc: The affected files were added on commit e46dd3e95 ("git-install",
2017-02-04) and on commit 081d1fbf2 ("Add seccomp errno filter support",
2015-09-23) / PR #66, respectively.

Cc: @netblue30 @wader


wader:
LGTM, note that extract_errnos.sh is mostly a dev-time thing and not that useful for a end user i guess

netblue30:
I have to do some cleanup there, we disconnected src/fgit from the interface a long time ago, The files in src/tools are for development only.

kmk3:
@wader commented [on Jun 17](https://github.com/netblue30/firejail/pull/5206#issuecomment-1158620234):

> LGTM, note that extract_errnos.sh is mostly a dev-time thing and not that
> useful for a end user i guess

@netblue30 commented [on Jun 17](https://github.com/netblue30/firejail/pull/5206#issuecomment-1158823425):

> The files in src/tools are for development only.

I don't understand why you mention this.

The file mode changes were made because regardless of the target audience, the
point of having `#!/bin/foo` in a file is so that it can be executed as `./bar`
(rather than using `$SHELL bar`), which requires execute permissions.  This is
also the case for the other shell scripts in src/tools.

Misc: I mostly created a PR (instead of committing directly) as a reminder to
maybe create a CI check for this later.

kmk3:
@netblue30 commented [on Jun 17](https://github.com/netblue30/firejail/pull/5206#issuecomment-1158823425):

> I have to do some cleanup there, we disconnected src/fgit from the interface
> a long time ago

For reference: commit 756fed6bd ("remving src/fgit", 2022-06-17).

I was going to commit a removal of the leftover `--enable-git-install` /
`--git-uninstall` flags from the fgit files after this PR, but that's even
better.  The commit message showed that all indeed appeared to be removed (and
that's still the case with 756fed6bd), so I'll paste it here for future
reference:

---

build: remove remnants of --git-install / --git-uninstall

This amends commit 23d10c8fa ("deprecated --git-install and
--git-uninstall", 2018-04-04).

Relevant options and variables:

* --disable-git-install
* --enable-git-install
* --git-install
* --git-uninstall
* HAVE_GIT_INSTALL
* enable_git_install

With this commit, it looks like all traces of the above are fully removed:

    $ git grep -F -e GIT_INSTALL -e git-install \
      -e git-uninstall -e git_install
    README: - compile/install scripts for --git-install/--git-uninstall commands
    RELNOTES:  * modif: --git-install and --git-uninstall removed
    src/fgit/fgit-install.sh:echo "           $(dirname "$0")/fgit-uninstall.sh"

PR #5219, opened 2022-06-25
build: reduce autoconf input files from 32 to 2

kmk3:
Configure summary: autoconf essentially only parses configure.ac and
generates the configure script (that is, the "./configure" shell
script).  The latter is what actually checks what is available on the
system and internally sets the value of the output variables.  It then,
for every filename foo in AC_CONFIG_FILES (and for every output variable
name BAR in AC_SUBST), reads foo.in, replaces every occurrence of
`@BAR@` with the value of the shell variable `$BAR` and generates the
file foo from the result.  After this, configure is finished and `make`
could be executed to start the build.

Now that (as of #5140) all output variables are only defined on
config.mk.in and on config.sh.in, there is no need to generate any
makefile nor any other mkfile or shell script at configure time.  So
rename every "Makefile.in" to "Makefile", mkdeb.sh.in to mkdeb.sh,
src/common.mk.in to src/common.mk and leave just config.mk and config.sh
as the files to be generated at configure time.

This allows editing and committing all makefiles directly, without
potentially having to run ./configure in between.

Commands used to rename the makefiles:

    $ git ls-files -z -- '*Makefile.in' | xargs -0 -I '{}' sh -c \
      "git mv '{}' \"\$(dirname '{}')/Makefile\""

Additionally, from my (rudimentary) testing, this commit reduces the
time it takes to run ./configure by about 20~25% compared to commit
72ece92ea ("Transmission fixes: drop private-lib (#5213)", 2022-06-22).
Environment: dash 0.5.11.5-1, gcc 12.1.0-2, Artix Linux, ext4 on an HDD.

Commands used for benchmarking each commit:

    $ : >time_configure && ./configure && make distclean &&
      for i in $(seq 1 10); do
      { time -p ./configure; } 2>>time_configure; done
    $ grep real time_configure |
      awk '{ total += $2 } END { print total/NR }'


kmk3:
To clarify, here is a summary of how the output files are generated from the
input files.  Using the following format:

* input-file(s) -> `generator-command` -> output-file(s)

Firstly, the configure script is still generated in the same as always:

* configure.ac -> `autoconf` -> configure

As of firejail 0.9.70 (that is, prior to #5140), this is what happens:

* Makefile.in (x28)
* mkdeb.sh.in
* src/common.mk.in

-> `./configure` ->

* Makefile (x28)
* mkdeb.sh
* src/common.mk

That is, every makefile is generated, regardless of whether it depends on any
output variable (such as `@CFLAGS@`) or not.

---

With #5140 and this PR:

* config.mk.in
* config.sh.in

-> `./configure` ->

* config.mk
* config.sh

Which means that the output variables (such as `@NAME@` and `@VERSION@`) are
all kept on language-specific include files, which allows all makefiles to
include config.mk (and all shell scripts to include config.sh) without having
to care about where config.mk (or config.sh) comes from (that is, whether it is
a normal file or a generated file).

reinerh (approved):

kmk3:
(Force-pushed to add more information to the first commit message)

netblue30:
all merged, thanks!

PR #5278, opened 2022-07-29
neomutt: Avoid creating empty files/directories

WhyNotHugo:
neomutt won't write to these locations. Processes it spawns might read
to some of them, but creating an empty file doesn't help. This just
pollutes user's $HOME with empty files and directories.

I've kept a few paths that MAY be written to by neomutt; it's not ideal,
but I want to minimise the risk of potential data loss, even if it is
corener cases.

See: https://github.com/netblue30/firejail/discussions/5276


kmk3 (approved):
LGTM.

Please do it for mutt.profile as well (either in this PR or in a new one).

kmk3:
Related discussion from ~1.5 year ago about email client profiles polluting the
home directory and also about maintaining whitelists for the combinations of
text editors paths x email client profiles:

* <https://github.com/netblue30/firejail/pull/3849#issuecomment-752222345>

Misc: I think had made some changes similar to this PR locally, though I wasn't
sure about which paths to remove and the changes were done together with a big
refactoring (yak shaving, etc), which wasn't finished.

etc/profile-m-z/neomutt.profile
```diff
@@ -50,31 +50,11 @@ include disable-programs.inc
...
 mkdir ${HOME}/Mail
 mkdir ${HOME}/mail
 mkdir ${HOME}/postponed
 mkdir ${HOME}/sent
-mkfile ${HOME}/.emacs
-mkfile ${HOME}/.mailcap
-mkfile ${HOME}/.msmtprc
-mkfile ${HOME}/.muttrc
-mkfile ${HOME}/.nanorc
-mkfile ${HOME}/.neomuttrc
```
rusty-snake, line 74: Who is `neomuttrc` supposed to be edited? Is there something like an "Open config file" options?

WhyNotHugo: > Who is neomuttrc supposed to be edited?

Usually it's created manually with a text editor.

> Is there something like an "Open config file" options?

Not that I know of. I think if you're very proficient you might figure out a way to possibly open neovim and then edit the config file manually, but that's _very_ unorthodox. I can't imagine someone doing this on a first run when no config file exists.

WhyNotHugo:
Anything pending from me here?

kmk3:
Merged, thanks!

PR #5283, opened 2022-08-01
build: Add files `make uninstall` forgot to remove

remyabel2:
There were a couple of files leftover after `make uninstall`. This
patch fixes that.


kmk3 (approved):
Good catch.

kmk3:
Merged, thanks!

PR #5285, opened 2022-08-02
vmware.profile: snapshot requires /etc/mtab

ra1nb0w:
This patch avoid the following error:
Error: One of the parameters supplied is invalid

Tested with VMware Workstation 16.2.4


rusty-snake (approved):

netblue30:
Thanks for the patch!

ra1nb0w:
thank you

PR #5289, opened 2022-08-02
new profile: gdu

glitsj16:
Add new profile for `gdu` (cfr. ncdu but optimized for SSD's):
https://github.com/dundee/gdu
https://archlinux.org/packages/community/x86_64/gdu/

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,4 +221,4 @@ Stats:
 
 ### New profiles:
 
-onionshare, onionshare-cli, opera-developer, songrec
+onionshare, onionshare-cli, opera-developer, songrec, gdu
--- a/etc/profile-a-l/gdu.profile
+++ b/etc/profile-a-l/gdu.profile
@@ -1,0 +1,37 @@
+# Firejail profile for gdu
+# Description: Fast disk usage analyzer with console interface
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include gdu.local
+# Persistent global definitions
+include globals.local
+
+blacklist ${RUNUSER}/wayland-*
+
+include disable-exec.inc
+
+caps.drop all
+ipc-namespace
+net none
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+x11 none
+
+private-dev
+# private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
--- a/etc/profile-a-l/gdu.profile
+++ b/etc/profile-a-l/gdu.profile
@@ -39,8 +39,8 @@ dbus-system none
 memory-deny-write-execute
 
 # gdu has built-in delete (d), empty (e) dir/file support and shell spawning (b) features.
-# To harden the sandbox we cripple those here.
-# Add ignore statements in your gdu.local if you need/want these functionalities.
-include disable-shell.inc
-private-bin gdu
-read-only ${HOME}
+# Depending on workflow and use case the sandbox can be hardened by adding the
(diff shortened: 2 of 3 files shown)


rusty-snake (commented):
LGTM, but I think there is space to harden.
- seccomp.block-secondary
- `seccomp socket` instead of `protocol unix`?
- `read-only`?

glitsj16:
@rusty-snake

> seccomp.block-secondary

Added!

> `seccomp socket` instead of `protocol unix`?

The app works fine without socket access, I've updated the PR accordingly.

> `read-only`?

We need a decision on this one (also affects `ncdu`). gdu has built-in delete (d), empty (e) dir/file support and shell spawning (b) features. We can cripple that functionality by adding:
```
include disable-shell.inc
private-bin gdu
read-only ${HOME}
```

That's how I use these disk usage analyzers BTW. But that's only my personal taste obviously. So what do you think? Should I add a comment on this and tighten up the sandbox or add the comment while commenting these hardenings? I edited the profile to give you an idea. Thanks for the valuable suggestions, as always!

rusty-snake:
> > `seccomp socket` instead of `protocol unix`?
> 
> The app works fine without socket access, I've updated the PR accordingly.

If you do not use `protocol ...` every call to `socket` is allowed. If it works w/o `socket`, add `socket` to `seccomp`.

https://github.com/netblue30/firejail/blob/316749ae4640f41324e7e5c6ec5e2c90c46581a5/etc/profile-m-z/pngquant.profile#L42

rusty-snake:
Does it not work with `seccomp socket`?

glitsj16:
> Does it not work with seccomp socket?

It does. Wasn't aware of doing it via seccomp filtering, thanks for the suggestion!

kmk3:
@glitsj16 commented [on Aug 3](https://github.com/netblue30/firejail/pull/5289#issuecomment-1204201546):

> > `read-only`?
> 
> We need a decision on this one (also affects `ncdu`). gdu has built-in delete
> (d), empty (e) dir/file support and shell spawning (b) features. We can
> cripple that functionality by adding:
> 
> ```
> include disable-shell.inc
> private-bin gdu
> read-only ${HOME}
> ```
> 
> That's how I use these disk usage analyzers BTW. But that's only my personal
> taste obviously. So what do you think? Should I add a comment on this and
> tighten up the sandbox or add the comment while commenting these hardenings?
> I edited the profile to give you an idea. Thanks for the valuable
> suggestions, as always!

I don't know about (e) and (b), but I usually use `ncdu` as a special-purpose
file manager, to find and delete large files and directories (d).  I think that
would be one of the most common use cases for it, so I would leave out at least
`read-only` (as with general-purpose file managers).

Also, not sure about `gdu`, but `ncdu` itself already has safety-related flags.
From ncdu(1):

```
       --enable-shell, --disable-shell
	   Enable or disable shell spawning from the browser. This feature is
	   enabled by default when scanning a live directory and disabled when
	   importing from file.

       --enable-delete, --disable-delete
	   Enable or disable the built-in file deletion feature. This feature
	   is enabled by default when scanning a live directory and disabled
	   when importing from file. Explicitly disabling the deletion feature
	   can work as a safeguard to prevent accidental data loss.

       --enable-refresh, --disable-refresh
	   Enable or disable directory refreshing from the browser. This
	   feature is enabled by default when scanning a live directory and
	   disabled when importing from file.

       -r  Read-only mode. When given once, this is an alias for
	   "--disable-delete", when given twice it will also add
	   "--disable-shell", thus ensuring that there is no way to modify the
	   file system from within ncdu.
```

glitsj16:
@kmk3 Thanks for your input. I will flip things around shortly to reflect your described use case, which makes sense. Those ncdu flags I'm familiar with, but gdu doesn't have -r (and -rr).

PR #5299, opened 2022-08-07
makepkg: add description

pirate486743186:
missing description

### DIFF
--- a/etc/profile-m-z/makepkg.profile
+++ b/etc/profile-m-z/makepkg.profile
@@ -1,4 +1,5 @@
 # Firejail profile for makepkg
+# Description: A utility to automate the building of Arch Linux packages
 # This file is overwritten after every install/update
 quiet
 # Persistent local customizations


kmk3 (changes requested):
### Review

Regarding the commit message:

> [makepkg] add descripton

Change it to:

> makepkg: add description

### Background

Git strips all occurrences of subject prefixes when applying a commit as a
patch, so it would be better to avoid using `[]` as a prefix in commit message
titles.  From `git-am(1)`:

```
DISCUSSION
       The commit author name is taken from the "From: " line of the message,
       and commit author date is taken from the "Date: " line of the message.
       The "Subject: " line is used as the title of the commit, after
       stripping common prefix "[PATCH <anything>]". The "Subject: " line is
       supposed to concisely describe what the commit is about in one line of
       text.
```

For example, given a commit with any of the following titles:

* `[PATCH] add description`
* `[makepkg] add description`
* `[PATCH makepkg] add description`
* `[PATCH] [makepkg] add description`

Applying it as a patch would change the title to just `add description`:

<details>
<summary>prefixtest</summary>
<p>

Code:

```sh
prefixtest() {
set -e

msgs='[PATCH] add description
[makepkg] add description
[PATCH makepkg] add description
[PATCH] [makepkg] add description'

git checkout -B prefixtest master 2>/dev/null
printf '%s\n' "$msgs" | while read -r msg
do
	echo 1 >>README.md
	git add -u
	git commit -m "$msg" >/dev/null
done

printf '(original)\n'
git log --pretty=%s master..

head="$(git rev-list HEAD~1..HEAD)"
git reset --hard master >/dev/null

printf '\n(git am output)\n'
git format-patch --stdout "master..$head" | git am

printf '\n(applied commits)\n'
git log --pretty=%s master..
}
prefixtest
```

Output:

```
(original)
[PATCH] [makepkg] add description
[PATCH makepkg] add description
[makepkg] add description
[PATCH] add description

(git am output)
Applying: add description
Applying: add description
Applying: add description
Applying: add description

(applied commits)
add description
add description
add description
add description
```

</p>
</details>

From what I've seen, usually the point of using a subject prefix is to specify
the intended project/repository when a single mailing list is used to receive
patches for multiple projects.  Additionally, the prefix is usually added to
the patch file with `git-format-patch(1)` (which can then be sent with
`git-send-email(1)`), rather than being written on the commit message itself.

From `git-format-patch(1)`:

```
       --subject-prefix=<subject prefix>
           Instead of the standard [PATCH] prefix in the subject line, instead
           use [<subject prefix>]. This allows for useful naming of a patch
           series, and can be combined with the --numbered option.
```

etc/profile-m-z/makepkg.profile
```diff
@@ -1,4 +1,5 @@
 # Firejail profile for makepkg
+# Description: An utility to automate the building of Arch Linux packages
```
kmk3, line 2:
```suggestion
# Description: A utility to automate the building of Arch Linux packages
```

Grammar: "utility" starts with a consonant sound ("you"), so it's "a utility".

Related:

* <https://github.com/netblue30/firejail/pull/5290#discussion_r936162181>

kmk3 (changes requested):
> makepkg.profile, add description

s/,/:/

kmk3 (approved):
LGTM.

Issue #5281, opened 2022-08-01
audacity  its broken


a61kt:
### Description

```
firejail audacity
Reading profile /etc/firejail/audacity.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-exec.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-shell.inc
Reading profile /etc/firejail/disable-xdg.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Parent pid 67427, child pid 67428
1 program installed in 112.37 ms
Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Warning: not remounting /run/user/1000/gvfs
Warning: not remounting /run/user/1000/doc
Blacklist violations are logged to syslog
Warning: Cannot confine the application using AppArmor.
Maybe firejail-default AppArmor profile is not loaded into the kernel.
As root, run "aa-enforce firejail-default" to load it.
Child process initialized in 418.40 ms
audacity: error while loading shared libraries: lib-screen-geometry.so: cannot open shared object file: No such file or directory

Parent is shutting down, bye...
```

### Steps to Reproduce

_open firejail audacity  / see it_

1. Run in bash `LC_ALL=C firejail PROGRAM` (`LC_ALL=C` to get a consistent output in English that can be understood by everybody)
4. See error `ERROR`

### Expected behavior

_get my audacity_

### Actual behavior

_its broken_

### Behavior without a profile

<details>
<summary> $ firejail --noprofile --net=none audacity </summary>

```
Parent pid 67492, child pid 67493
Child process initialized in 36.74 ms

(process:3): Gdk-CRITICAL **: 23:36:36.123: gdk_screen_get_root_window: assertion 'GDK_IS_SCREEN (screen)' failed

(process:3): Gdk-CRITICAL **: 23:36:36.125: gdk_window_get_display: assertion 'GDK_IS_WINDOW (window)' failed

(process:3): Gdk-CRITICAL **: 23:36:36.125: gdk_cursor_new_from_pixbuf: assertion 'GDK_IS_DISPLAY (display)' failed

(audacity:3): Gtk-WARNING **: 23:36:36.182: Theme parsing error: gtk.css:63:28: The :prelight pseudo-class is deprecated. Use :hover instead.

[...]

PR #5300, opened 2022-08-08
fix(audacity): !5281 sharedlib bug on Arch/Fedora

TopherIsSwell:
removed `private-bin` line from audacity profile as it appears to block
access to shared libraries needed to start audacity on some
distributions.

Relates to github issue #5281

### DIFF
--- a/etc/profile-a-l/audacity.profile
+++ b/etc/profile-a-l/audacity.profile
@@ -20,7 +20,8 @@ include disable-xdg.inc
 
 include whitelist-var-common.inc
 
-apparmor
+## Enabling App Armor appears to break some Fedora / Arch installs
+#apparmor
 caps.drop all
 net none
 no3d
@@ -36,7 +37,7 @@ protocol unix,inet
 seccomp
 tracelog
 
-#private-bin audacity
+private-bin audacity
 private-dev
 private-tmp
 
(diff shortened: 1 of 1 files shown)


TopherIsSwell:
Update. It appears that this bug isn't always reproduceable (possibly related to whether apparmor is installed?). And on systems where it occurs (Confirmed on some Arch and Fedora installs), you can remove the `private-bin` line _OR_ the `apparmor` line to fix this issue. 

So this bug _might_ be bigger than this single profile, but it's wildly unclear to me.

glitsj16:
See #5281, especially https://github.com/netblue30/firejail/issues/5281#issuecomment-1211329314. IMO it would be better to keep private-bin and drop apparmor to fix this.

etc/profile-a-l/audacity.profile
```diff
@@ -36,7 +36,7 @@ protocol unix,inet
 seccomp
 tracelog
 
-private-bin audacity
```
glitsj16, line 39: As per conversation, let's fix audacity by dropping apparmor and keep private-bin intact.

glitsj16 (changes requested):
See https://github.com/netblue30/firejail/issues/5281#issuecomment-1211329314.

etc/profile-a-l/audacity.profile
TopherIsSwell: How now, brown cow?

glitsj16 (approved):
LGTM. Thanks for your speedy responses and the PR!

reinerh (approved):

PR #5302, opened 2022-08-08
makedeb profile creation

pirate486743186:
makedeb is a makepkg fork for debian.

I didn't add it in firecfg because makepkg isn't there either.

### DIFF
--- a/etc/profile-m-z/makedeb.profile
+++ b/etc/profile-m-z/makedeb.profile
@@ -1,0 +1,13 @@
+# Firejail profile for makedeb
+# Description: A utility to automate the building of Debian packages
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include makedeb.local
+# Persistent global definitions
+#include globals.local
+
+ignore noblacklist /var/lib/pacman
+
+# Redirect
+include makepkg.profile


etc/profile-m-z/makedeb.profile
```diff
@@ -0,0 +1,24 @@
...
+# about makepkg for potential issues and their solutions when Firejailing makedeb, 
+
+# This profile could be significantly strengthened by adding the following to makedeb.local
+# whitelist ${HOME}/<Your Build Folder>
+# whitelist ${HOME}/.gnupg
+
+# noroot is only disabled to allow the creation of kernel headers from an official PKGBUILD.
+#noroot
+
+ignore include makepkg.local
```
rusty-snake, line 20: We don't do so anywhere else, do we?

kmk3, line 20:
```suggestion
```

Since this profile redirects to makepkg.profile, leave these lines only in the
latter to avoid duplication and make maintenance easier.

kmk3:
(Offtopic)

@pirate486743186 commented [on Aug 8](https://github.com/netblue30/firejail/pull/5302#issue-1331839697):

> makedeb is a makepkg fork for debian.

Link to the source code for reference:

* <https://github.com/makedeb/makedeb>

Nice, I always wanted to be able to use makepkg and PKGBUILDs on apt-based
systems, as they make creating and building packages a breeze.

See also also pacapt:

* <https://github.com/icy/pacapt>

> An ArchLinux's pacman-like shell wrapper for many package managers. 56KB and
> run anywhere.

> ### Basic operations
> 
> For system package manager
> 
> * Update package database: `pacapt -Sy`, or `pacapt update`
> * Install a package: `pacapt -S foo`, or `pacapt install foo`
> * Search a package: `pacapt -Ss foo`, or `pacapt search foo`
> * Remove a package: `pacapt -R foo`, or `pacapt remove foo`
> * Upgrade system: `pacapt -Su`, or `pacapt upgrade`
> * Remove orphans: `pacapt -Sc`, or `pacapt autoremove foo`
> * Clean up: `pacapt -Scc` or `pacapt -Sccc`, or `pacapt clean`

With `alias pacact=pacman`, you can run the exact same commands as on Arch to
install/remove packages and update/upgrade the system on apt-based and
rpm-based distributions (among many others).

A non-rolling distro with the DFSG (and separate free/non-free repos) + Arch
tooling sounds pretty interesting.

Maybe even the real pacman could be ported eventually (or a non-rolling
pacman-based distro could be made), as even though Arch officially supports
only x86_64, MSYS2 on Windows shows that pacman can work with multi-arch
(including `mingw-w64-i686`, `mingw-w64-x86_64`, `clang-w64-i686` and
`clang-w64-x86_64`).

Lastly, if someone were to make an AUR for Debian or Ubuntu, that would be
amazing.  Like PPAs, but in source form, using git and PKGBUILDs.

glitsj16:
(Offtopic)

> Nice, I always wanted to be able to use makepkg and PKGBUILDs on apt-based
systems, as they make creating and building packages a breeze.

> Lastly, if someone were to make an AUR for Debian or Ubuntu, that would be
amazing. Like PPAs, but in source form, using git and PKGBUILDs.

Wasn't this formerly called the 'Debian User Repository' (DUR)? Personally I wouldn't hold too many breaths on any of the above promises. It sure sounds nice on paper. Let's hope it doesn't turn into broken frankendebian installations.

etc/profile-m-z/makedeb.profile
kmk3: @pirate486743186 [on Aug 9](https://github.com/netblue30/firejail/pull/5302#discussion_r941111298):

> The comments are meant to be read from users. So i think they should stay
> there.

If users read the comments on redirect profiles, they should eventually come
across the `# Redirect` comment, in which case they can follow the include to
see what else the profile does.

Also, avoiding such duplication is how it's done in other redirect profiles.

See for example the commented whitelisting of the `${HOME}/Projects` directory,
which only exists in build-systems-common.profile:

```console
b18120cc2 2022-08-09 21:11:18 +0000 new profile: gdu (#5289)
$ git grep -F -i projects -- etc
etc/profile-a-l/build-systems-common.profile:#whitelist ${HOME}/Projects
etc/profile-a-l/gitg.profile:#whitelist ${HOME}/YOUR_GIT_PROJECTS_DIRECTORY
etc/profile-m-z/nodejs-common.profile:# If you want whitelisting, change ${HOME}/Projects below to your node projects directory
etc/profile-m-z/nodejs-common.profile:#whitelist ${HOME}/Projects
$ git grep -F build-systems-common.profile -- etc
etc/profile-a-l/bundle.profile:include build-systems-common.profile
etc/profile-a-l/cargo.profile:include build-systems-common.profile
etc/profile-a-l/cmake.profile:include build-systems-common.profile
etc/profile-m-z/make.profile:include build-systems-common.profile
etc/profile-m-z/meson.profile:include build-systems-common.profile
etc/profile-m-z/pip.profile:include build-systems-common.profile
$ git grep -F nodejs-common.profile -- etc
etc/profile-m-z/node-gyp.profile:include nodejs-common.profile
etc/profile-m-z/node.profile:include nodejs-common.profile
etc/profile-m-z/npm.profile:include nodejs-common.profile
etc/profile-m-z/npx.profile:include nodejs-common.profile
etc/profile-m-z/semver.profile:include nodejs-common.profile
etc/profile-m-z/yarn.profile:include nodejs-common.profile
```

Or the comment about whitelisting paths other than the `${HOME}/Mail`
directory, which only exists in email-common.profile:

```console
$ git grep -F -C 1 'whitelist ${HOME}/Mail' -- etc
etc/profile-a-l/email-common.profile-# when storing mail outside the default ${HOME}/Mail path, 'whitelist' the custom path in your email-common.local
etc/profile-a-l/email-common.profile:whitelist ${HOME}/Mail
etc/profile-a-l/email-common.profile-whitelist ${RUNUSER}/gnupg
--
etc/profile-m-z/mutt.profile-whitelist ${HOME}/.w3m
etc/profile-m-z/mutt.profile:whitelist ${HOME}/Mail
etc/profile-m-z/mutt.profile-whitelist ${HOME}/mail
--
etc/profile-m-z/neomutt.profile-whitelist ${HOME}/.w3m
etc/profile-m-z/neomutt.profile:whitelist ${HOME}/Mail
etc/profile-m-z/neomutt.profile-whitelist ${HOME}/mail
$ git grep -F email-common.profile -- etc
etc/profile-a-l/claws-mail.profile:include email-common.profile
etc/profile-m-z/sylpheed.profile:include email-common.profile
```

> Also "makepkg" and "makedeb" are easy to confuse.

Do you mean because this comment in makepkg.profile mentions makepkg.local?

```firejail
# This profile could be significantly strengthened by adding the following to makepkg.local
# whitelist ${HOME}/<Your Build Folder>
# whitelist ${HOME}/.gnupg
```

I would assume that whatever applies to a given profile also applies to
whatever includes it.

And indeed, adding that line to makepkg.local would affect both profiles, since
makedeb.profile ultimately also includes makepkg.local.

> Also, unless you do weird things, makedeb and makepkg are mutually exclusive.
> They are meant to be installed on very different systems and things will
> diverge.

I don't see why either case would matter with regards to profile
maintainability.  As it stands on this PR, one profile already redirects to the
other anyway.

If you're takling about keeping `ignore include makepkg.local`, then I would be
curious as to when it could cause issues (considering that profiles are
generally written to be compatible with a wide variety of distros).

Note: I'm not 100% against keeping it, but I would like to at least understand
in which scenarios would it matter.

And if it is kept, it would be good to add a comment explaining why it's there
(since it's unusual to ignore a .local profile, as @rusty-snake mentioned).

> It's better to fork the comments.

A profile that looks like half of a fork of another profile while redirecting
to it at the same time seems confusing to me.

If the profiles diverge too much, why not fork the entire profile?

If they don't, why not write on the includer profile only what is actually
different about it (which usually just means `ignore` commands), like most
other redirect profiles?

etc/profile-m-z/makedeb.profile
```diff
@@ -0,0 +1,24 @@
+# Firejail profile for makedeb
+# Description: An utility to automate the building of Debian packages
```
kmk3, line 2:
```suggestion
# Description: A utility to automate the building of Debian packages
```

Grammar; same as the following:

* <https://github.com/netblue30/firejail/pull/5299#discussion_r941730616>

kmk3 (approved):
LGTM.

kmk3:
Relates to #5299.

netblue30:
all in, thanks!

Issue #5297, opened 2022-08-06
Add electronapps-common


pirate486743186:
### Is your feature request related to a problem? Please describe.

electronmail can't open the file picker. by disabling noroot, it works again. It's an electron app. There's a profile named electron, i assume it's the SDK, noroot is not there.

I think i remember, there was a sticky about electron apps. In general, these are tricky to handle.

### Describe the solution you'd like

There should be a electronapps-common to minimize this kind of stuff.

### Describe alternatives you've considered

Maybe also use electronapps-common in chromium-common. That's used by a lot of web browsers. Technically, they are a category of electron apps.

So electron-common could be build as a cut down version of chromium-common. That should keep complexity manageable.

rusty-snake:
Here it is: https://github.com/netblue30/firejail/blob/master/etc/profile-a-l/electron.profile
it's just not named `-common`.

That electronmail needs to be refactored as electron redirect profile I said somewhere else already (but could not find it). 

---

Technical electron is chromium and not the other ways around.

---

chromium + file-chooser + noroot (i.e. portals with noroot) is tracked in #4716.

kmk3:
(Re-closing as "not planned" since nothing was changed in firejail)

PR #5304, opened 2022-08-09
profiles: clarify userns comments & fix comment on electron.profile

ghost:
clarified description and fixed a comment.

I think it should be renamed to electron-common.profile ...


etc/profile-a-l/electron.profile
```diff
@@ -22,7 +22,7 @@ include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
-# Add the next line to your chromium-common.local if your kernel allows unprivileged userns clone.
+# Add the next line to your electron.local if your kernel allows unprivileged users clone.
```
rusty-snake, line 25: You can not clone users.

rusty-snake (changes requested):

etc/profile-a-l/electron.profile
kmk3: To clarify, "userns" here means user namespaces.

See also namespaces(7) and the `userns` option in firejail.config:

```
# Enable or disable user namespace support, default enabled.
# userns yes
```

kmk3:
@pirate486743186 commented [on Aug 7](https://github.com/netblue30/firejail/issues/5297#issuecomment-1207350896):

> And renaming it to electron-common is probably better. You should follow
> conventions, or people get confused. Just call all profiles that are meant to
> be redirects -common.

@pirate486743186 commented [on Aug 9](https://github.com/netblue30/firejail/pull/5304#issue-1333086455):

> clarified description and fixed a comment.
> 
> I think it should be renamed to electron-common.profile ...

Makes sense to me.

Feel free to submit it in a new PR after this PR is merged (to avoid
conflicts).

etc/profile-a-l/chromium-common.profile
```diff
@@ -37,7 +37,7 @@ include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
-# Add the next line to your chromium-common.local if your kernel allows unprivileged userns clone.
+# Add the next line to your chromium-common.local if your kernel allows "unprivileged userns clone".
```
kmk3, line 40:
```suggestion
# Add the next line to your chromium-common.local if your kernel allows
# unprivileged userns clone (see user_namespaces(7)).
```

etc/profile-a-l/electron.profile
```diff
@@ -22,7 +22,7 @@ include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
-# Add the next line to your chromium-common.local if your kernel allows unprivileged userns clone.
+# Add the next line to your electron.local if your kernel allows "unprivileged userns clone".
```
kmk3, line 25:
```suggestion
# Add the next line to your electron.local if your kernel allows
# unprivileged userns clone (see user_namespaces(7)).
```

etc/profile-m-z/ping.profile
```diff
@@ -24,7 +24,7 @@ include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
-# Add the next line to your ping.local if your kernel allows unprivileged userns clone.
+# Add the next line to your ping.local if your kernel allows "unprivileged userns clone".
```
kmk3, line 27:
```suggestion
# Add the next line to your ping.local if your kernel allows
# unprivileged userns clone (see user_namespaces(7)).
```

kmk3 (commented):
Unprivileged userns clone is not a thing unto itself, so quoting it makes it
more confusing IMO.  I think having a reference to the relevant man page should
provide enough context.

Also, an additional suggestion for etc/firejail.config:

```diff
 # This logging feature is disabled by default in our implementation.
 # seccomp-log no
 
-# Enable or disable user namespace support, default enabled.
+# Enable or disable user namespace support, default enabled. See
+# user_namespaces(7).
 # userns yes
 
 # Disable whitelist top level directories, in addition to those
```

Lastly, the commit message could be clearer.  Example:

> profiles minor edits

->

> docs: clarify meaning of userns & fix comment on electron.profile

rusty-snake:
> Also, an additional suggestion for etc/firejail.config:

We could be more clear that this maps to `noroot`.
- `dbus` maps to `dbus*`
- `bind` maps to `bind*`
- `apparmor` maps to `apparmor`
- `network` maps to `net*`, `ip*`, `defaultgw` and a few other
- `x11` maps to `x11`
- `seccomp` maps to `seccomp*`
- ...
- *`userns` maps to `noroot`*

kmk3:
@rusty-snake commented [on Aug 11](https://github.com/netblue30/firejail/pull/5304#issuecomment-1212216757):

> > Also, an additional suggestion for etc/firejail.config:
> 
> We could be more clear that this maps to `noroot`.
> 
> * `dbus` maps to `dbus*`
> * `bind` maps to `bind*`
> * `apparmor` maps to `apparmor`
> * `network` maps to `net*`, `ip*`, `defaultgw` and a few other
> * `x11` maps to `x11`
> * `seccomp` maps to `seccomp*`
> * ...
> * _`userns` maps to `noroot`_

If by "`foo` maps to `bar`" you mean "`bar` depends on `foo` being enabled",
then I agree.

Though I would add a comment to each command in the man pages instead of in
firejail.config, to avoid potentially having to update a comment in
firejail.config whenever a new command is added (and to avoid adding another
item to the new command checklist).

rusty-snake:
1. You speak to an advanced user.
2. You can not simplify complex topics to a level where everyone understand it without making the statement useless.

kmk3:
@pirate486743186 commented [on Aug 12](https://github.com/netblue30/firejail/pull/5304#issuecomment-1213086852):

> You are overcomplicating it. You need to speak in stupid. You don't need all
> these details, you speak to a user, not a CPU.
> 
> All that is needed is simply to make it clear it's not a typo.
> 
> What about "unprivileged user namespaces (userns) clone", without the quotes?

This is as verbose as I can make it in 3 lines:

```diff
-# Add the next line to your chromium-common.local if your kernel allows "unprivileged userns clone".
+# If your kernel allows the creation of user namespaces by unprivileged users
+# (that is, if `sysctl kernel.unprivileged_userns_clone` is >= 1), you can add
+# the next line to your chromium-common.local.
```

By the way, I could not find the above sysctl variable being documented:

```console
$ man -w 7 user_namespaces | xargs pacman -Qo
/usr/share/man/man7/user_namespaces.7.gz is owned by man-pages 5.13-1
$ man 7 user_namespaces | grep -F unprivileged_userns
$
$ pacman -Q linux-docs
linux-docs 5.18.16.artix1-1
$ pacman -Qlq linux-docs | grep -v '/$' | xargs \
  grep -F unprivileged_userns 2>/dev/null
$
```

So I think that it would be good to have it written down somewhere in firejail,
even if it's just in the profiles.

rusty-snake:
`kernel.unprivileged_userns_clone` is a Debian patch (which is also used by Arch and others) and does not exists in mainline kernels or RedHat distributions (RHEL, Fedora, Cent OS Stream). If it does not exist userns is either disabled for all users or enabled.

kmk3:
@rusty-snake commented [on Aug 13](https://github.com/netblue30/firejail/pull/5304#issuecomment-1213839313):

> `kernel.unprivileged_userns_clone` is a Debian patch (which is also used by
> Arch and others) and does not exists in mainline kernels or RedHat
> distributions (RHEL, Fedora, Cent OS Stream). If it does not exist userns is
> either disabled for all users or enabled.

I see, thanks.

Do you know of a more portable way to check if it's allowed?

glitsj16:
> Do you know of a more portable way to check if it's allowed?

Perhaps testing `sysctl user.max_user_namespaces`? If that's equal to 0 (zero) one can assume unprivileged namespaces are disallowed.

rusty-snake:
@glitsj16 then are all userns disallowed, not only unprivileged. And if it it >0, they can still be disabled for unprivileged processes.

Maybe running `unshare -U` can work to find it out.

kmk3:
@rusty-snake commented [on Aug 13](https://github.com/netblue30/firejail/pull/5304#issuecomment-1213999645):

> @glitsj16 then are all userns disallowed, not only unprivileged. And if it it
> >0, they can still be disabled for unprivileged processes.
> 
> Maybe running `unshare -U` can work to find it out.

Nice, that seems to work:

```console
$ sudo sysctl kernel.unprivileged_userns_clone=1
kernel.unprivileged_userns_clone = 1
$ unshare -U true; echo $?
0
$ unshare -U whoami
nobody
$ sudo sysctl kernel.unprivileged_userns_clone=0
kernel.unprivileged_userns_clone = 0
$ unshare -U true; echo $?
unshare: unshare failed: Operation not permitted
1
$ unshare -U whoami
unshare: unshare failed: Operation not permitted
```

How about this then?

```diff
-# Add the next line to your chromium-common.local if your kernel allows "unprivileged userns clone".
+# If your kernel allows the creation of user namespaces by unprivileged users
+# (for example, if running `unshare -U whoami` prints "nobody"), you can add
+# the next line to your chromium-common.local.
```

rusty-snake:
Is there a way `whoami` can return something else? Maybe we could use `unshare -U echo enabled`.

kmk3:
@rusty-snake commented [on Aug 28](https://github.com/netblue30/firejail/pull/5304#issuecomment-1229458708):

> Is there a way `whoami` can return something else?

I wondered about that too, but not sure.

> Maybe we could use `unshare -U echo enabled`.

Agreed; updated diff:

```diff
-# Add the next line to your chromium-common.local if your kernel allows "unprivileged userns clone".
+# If your kernel allows the creation of user namespaces by unprivileged users
+# (for example, if running `unshare -U echo enabled` prints "enabled"), you
+# can add the next line to your chromium-common.local.
```

PR #5305, opened 2022-08-09
profiles: rename email-common to claws-sylpheed-common

ghost:
renaming because too generic and misleading.

### DIFF
--- a/etc/profile-a-l/claws-mail.profile
+++ b/etc/profile-a-l/claws-mail.profile
@@ -1,5 +1,5 @@
 # Firejail profile for claws-mail
-# Description: Fast, lightweight and user-friendly GTK+2 based email client
+# Description: Fast, lightweight and user-friendly GTK+ based email client
 # This file is overwritten after every install/update
 # Persistent local customizations
 include claws-mail.local
@@ -27,4 +27,4 @@ dbus-user.talk org.gnome.keyring.SystemPrompter
 # dbus-user.talk org.freedesktop.Notifications
 
 # Redirect
-include email-common.profile
+include claws-sylpheed-common.profile
--- a/etc/profile-a-l/claws-sylpheed-common.profile
+++ b/etc/profile-a-l/claws-sylpheed-common.profile
@@ -1,8 +1,8 @@
-# Firejail profile for email-common
+# Firejail profile for claws-sylpheed-common
 # Description: Common profile for claws-mail and sylpheed email clients
 # This file is overwritten after every install/update
 # Persistent local customizations
-include email-common.local
+include claws-sylpheed-common.local
 # Persistent global definitions
 # added by caller profile
 #include globals.local
@@ -11,7 +11,7 @@ noblacklist ${HOME}/.bogofilter
 noblacklist ${HOME}/.gnupg
 noblacklist ${HOME}/.mozilla
 noblacklist ${HOME}/.signature
-# when storing mail outside the default ${HOME}/Mail path, 'noblacklist' the custom path in your email-common.local
+# when storing mail outside the default ${HOME}/Mail path, 'noblacklist' the custom path in your claws-sylpheed-common.local
 # and 'blacklist' it in your disable-common.local too so it is kept hidden from other applications
 noblacklist ${HOME}/Mail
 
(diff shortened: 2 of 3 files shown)


kmk3 (changes requested):
>     # Description: Common profile for claws-mail and sylpheed email clients

> rename email-common to claws-sylpheed-common

@pirate486743186 commented [on Aug 9](https://github.com/netblue30/firejail/pull/5305#issue-1333126898):

> renaming because too generic and misleading.

I think only the profile's description is misleading; I don't think that
email-common.profile is intended to support only those two programs.

IMO the proper fix would be to make email-common.profile more generic if needed
and then to make more profiles include it.  For example:

* alpine.profile
* mutt.profile
* neomutt.profile

Also, I don't think that it necessarily has to support every email client
profile; I'd say that it just has to be common enough to be useful.  For
example, compared to the aforementioned programs, at first glance
thunderbird.profile seems to diverge a bit much from email-common.profile.

rusty-snake:
> usually *-common are for applications that are related

They *are related*. They are all for emails.

> For example there's firefox-common and chromium-common not browser-common.

For example there are archiver-common, latex-common, file-manager-common, build-systems-common, ….

kmk3:
@pirate486743186 commented [on Aug 12](https://github.com/netblue30/firejail/pull/5305#issuecomment-1213145647):

> that's too ambitious

Why is that?

While it includes many commands, the most important aspect of
email-common.profile IMO is the path commands in it, to avoid maintaining the
same list of paths in multiple places.  Also, not all email client profiles are
whitelisting profiles (e.g.: alpine.profile), so maybe a new
"whitelist-email-common.profile" would be needed.  But other commands that are
not 100% common could be moved directly into claws.profile and
sylpheed.profile.

All of this seems doable, though it is not a trivial change and would require
attention to detail to keep the final profiles essentially the same.  If you
don't want to do it, then don't do it.

For now, this would be my suggestion:

```diff
-# Description: Common profile for claws-mail and sylpheed email clients
+# Description: Common profile for email clients (TODO: include it in more
+# profiles)
```

If the point of this PR is to make things clearer, then that should be good
enough.

> For example there's firefox-common and chromium-common not browser-common.
> 
> Just rename to claws-sylpheed-common. And also create a few other -common
> profiles for groups of email clients, then you see if it makes sense to
> create email-common.

It seems wasteful to have a -common profile intended for just two profiles.
Also, it seems oddly specific in this case (as in "why does this exist?").  It
would make more sense to either move its contents directly into claws.profile
and sylpheed.profile or to make it a more common include.  And unless there is
a blocker for the latter, it seems to be the better approach for
maintainability.

glitsj16:
> It seems wasteful to have a -common profile intended for just two profiles.
Also, it seems oddly specific in this case (as in "why does this exist?").

Let me provide some context. claws-mail is a direct evolution of the original sylpheed email client, family let's say. When I added email-common.profile I picked that more generic name with the intent to enable refactoring other email-related apps in the future. With hindsight I could have given it a more suitable, less confusing name. I don't use apps like alpine, mutt and neomutt and the refactoring never happened. Personally I don't hold strong opinions about any of this.

kmk3:
@glitsj16 commented [on Aug 12](https://github.com/netblue30/firejail/pull/5305#issuecomment-1213590768):

> > It seems wasteful to have a -common profile intended for just two profiles.
> > Also, it seems oddly specific in this case (as in "why does this exist?").
> 
> Let me provide some context. claws-mail is a direct evolution of the original
> sylpheed email client, family let's say. When I added email-common.profile I
> picked that more generic name with the intent to enable refactoring other
> email-related apps in the future. With hindsight I could have given it a more
> suitable, less confusing name.

Just to be clear, I think that the current name is fine; my comment was about
the proposal of renaming it to claws-sylpheed-common.profile.

> I don't use apps like alpine, mutt and neomutt and the refactoring never
> happened. Personally I don't hold strong opinions about any of this.

Yes; these programs seem to be rather flexible and people might use them in
very different ways and so I don't expect anyone to be proficient enough in all
of them to test them all with certainty.  Which is why I suggested doing
basically only mechanical/organizational changes (that is, to mostly just move
the existing commands around), so that even with an extra include, each profile
ends up executing effectively the same commands as before.

glitsj16:
@kmk3 Thanks for the clarifications. And I fully agree with your stance on this :-)

PR #5309, opened 2022-08-11
microsoft-edge.profile rewritten for stable channel and moved microsoft-edge{,-beta,-dev} from private-opt to whitelist

qretornaz-adapei42:
I did rewrite microsoft-edge.profile from dev/beta for stable channel that is using only this name as executable binary unlike the other two.

About what we were discussing in #5307, I did replace private-opt by whitelist for stable, beta and dev profiles.
Tell me if this is not the way you would like it to be and I will follow your recommendations..


rusty-snake (approved):

glitsj16 (approved):
LGTM

netblue30:
merged!

PR #5317, opened 2022-08-15
Fix an AppArmor profile denial issue with ptrace reading and signals

### LINKED ISSUE #5316: peer="firejail-default//&unconfined" - multiple apparmor DENIED entrees in audit log after merging #5274
Firejail from git master currently shows multiple entrees in /var/log/audit/audit.log on my arch linux box. I've bisected and this regression stems from [commit](https://github.com/netblue30/firejail/commit/b987cf001af3e3a88d73d49117ddf7ccad0cbc5e), which added #5274:

```
type=AVC msg=audit(1660549572.602:83): apparmor="DENIED" operation="ptrace" profile="firejail-default" pid=6723 comm="ps" requested_mask="read" denied_mask="read" peer="firejail-default//&unconfined"
type=AVC msg=audit(1660549572.602:83): apparmor="DENIED" operation="ptrace" profile="firejail-default" pid=6723 comm="ps" requested_mask="readby" denied_mask="readby" peer="firejail-default//&unconfined"
```

@ChrysoliteAzalea If there's anything I can post to help debugging this issue, feel free to ping me.

OS: Arch Linux
apparmor: 3.0.7-1
firejail-git: 0.9.71.r8626.5ab4aeb35-1

ChrysoliteAzalea:
Hello everyone!

Recently, a [pull request](https://github.com/netblue30/firejail/pull/5274) was merged that added support for custom AppArmor profiles. However, there was a use-case I haven't tested -- a case where an application confined by a default profile uses ptrace or sends signals. In the original profile, it was allowed only for apps with the same security label. However, I've missed that, while **firejail-default//&unconfined** and **firejail-default** have equal AppArmor permissions, ptrace reading and sending signals were previously allowed for peers only with the latter security label. I've fixed it in this PR by allowing both.

Fixes #5316

### DIFF
--- a/etc/apparmor/firejail-default
+++ b/etc/apparmor/firejail-default
@@ -33,6 +33,7 @@ owner /{,var/}run/firejail/dbus/[0-9]*/[0-9]*-user w,
 #ptrace,
 # Allow obtaining some process information, but not ptrace(2)
 ptrace (read,readby) peer=@{profile_name},
+ptrace (read,readby) peer=@{profile_name}//&unconfined,
 
 ##########
 # Allow read access to whole filesystem and control it from firejail.
@@ -123,6 +124,7 @@ network packet,
 ##########
 # There is no equivalent in Firejail for filtering signals.
 ##########
+signal (send) peer=@{profile_name}//&unconfined,
 signal (send) peer=@{profile_name},
 signal (receive),
 


glitsj16 (approved):
LGTM. Thanks for the fix!

Issue #5321, opened 2022-08-18
Double invocation breaks certain applications


NetSysFire:
### Description

When `firecfg` has been run, so `/usr/local/bin/man` is symlinked and therefore automagically firejailed, `firejail man firejail` does not work while just `man firejail` works.
However, this specifically only breaks man so far. `firejail drill example.com` for example works.

### Steps to Reproduce

1. Make sure `/usr/bin/man` points to `/usr/bin/firejail`.
2. Run `firejail man firejail` (you may choose any other manpage of course)
3. `Error getpwuid: main.c:245 init_cfg: No such file or directory`

### Expected behavior

I know this is a bit of an edge case, but other applications do work with that. E.g `firejail drill example.com` works.

### Actual behavior

It does not work.

```
Active seccomp files:
cat /run/firejail/mnt/seccomp/seccomp.list
/run/firejail/mnt/seccomp/seccomp.protocol
/run/firejail/mnt/seccomp/seccomp.32
/run/firejail/mnt/seccomp/seccomp
/run/firejail/mnt/seccomp/seccomp.mdwx
/run/firejail/mnt/seccomp/seccomp.mdwx.32
Create the new ld.so.preload file
Blacklist violations are logged to syslog
Mount the new ld.so.preload file
Dropping all capabilities
nogroups command not ignored
noroot user namespace installed
Dropping all capabilities
NO_NEW_PRIVS set
Drop privileges: pid 1, uid 1000, gid 1000, force_nogroups 0
nogroups command not ignored
No supplementary groups
AppArmor enabled
Closing non-standard file descriptors
Starting application
LD_PRELOAD=(null)
execvp argument 0: man
execvp argument 1: firejail
Child process initialized in 157.39 ms
Searching $PATH for man
trying #/usr/local/sbin/man#
trying #/usr/local/bin/man#
Installing /run/firejail/mnt/seccomp/seccomp.mdwx.32 seccomp filter
Installing /run/firejail/mnt/seccomp/seccomp.mdwx seccomp filter
Installing /run/firejail/mnt/seccomp/seccomp seccomp filter
Installing /run/firejail/mnt/seccomp/seccomp.32 seccomp filter
Installing /run/firejail/mnt/seccomp/seccomp.protocol seccomp filter
Error getpwuid: main.c:245 init_cfg: No such file or directory
monitoring pid 15

Sandbox monitor: waitpid 15 retval 15 status 256

Parent is shutting down, bye...
```

### Behavior without a profile

[...]

PR #5322, opened 2022-08-18
man.profile needs additional private-etc items on arch

glitsj16:
Came up on IRC when talking about #5321.


rusty-snake (approved):
Not arch specific but LGTM

PR #5298, opened 2022-08-06
electron-mail.profile refactoring

pirate486743186:
title.

~~noroot is missing from all the chromium browsers that i checked and electron the electron sdk....~~

Now refactoring with electron.profile

PR #5325, opened 2022-08-20
tutanota-desktop.profile improvements

pirate486743186:
saving to disk is broken. Fixes other minor errors

### DIFF
--- a/etc/profile-m-z/tutanota-desktop.profile
+++ b/etc/profile-m-z/tutanota-desktop.profile
@@ -1,5 +1,5 @@
 # Firejail profile for tutanota-desktop
-# Description: Encrypted email client
+# Description: Official desktop client for the Tutanota E2E encrypted email provider
 # This file is overwritten after every install/update
 # Persistent local customizations
 include tutanota-desktop.local
@@ -9,7 +9,12 @@ include globals.local
 noblacklist ${HOME}/.config/tuta_integration
 noblacklist ${HOME}/.config/tutanota-desktop
 
+ignore dbus-user none
+ignore disable-mnt
 ignore noexec /tmp
+
+# sh is needed to allow Firefox to open links
+include allow-bin-sh.inc
 
 include disable-shell.inc
 
@@ -18,14 +23,26 @@ mkdir ${HOME}/.config/tutanota-desktop
 whitelist ${HOME}/.config/tuta_integration
 whitelist ${HOME}/.config/tutanota-desktop
 
-# These lines are needed to allow Firefox to open links
+# The lines below are needed to find the default Firefox profile name, to allow
+# opening links in an existing instance of Firefox (note that it still fails if
+# there isn't a Firefox instance running with the default profile; see #5352)
 noblacklist ${HOME}/.mozilla
 whitelist ${HOME}/.mozilla/firefox/profiles.ini
 read-only ${HOME}/.mozilla/firefox/profiles.ini
 
+machine-id
+nosound
+
 ?HAS_APPIMAGE: ignore private-dev
 private-etc @tls-ca
 private-opt tutanota-desktop
+
+dbus-user filter
+dbus-user.talk org.freedesktop.Notifications
+dbus-user.talk org.freedesktop.secrets
+dbus-user.talk org.gnome.keyring.SystemPrompter
+# allow D-Bus communication with firefox for opening links
+dbus-user.talk org.mozilla.*
 
 # Redirect
 include electron.profile


etc/profile-m-z/tutanota-desktop.profile
```diff
@@ -23,8 +26,12 @@ noblacklist ${HOME}/.mozilla
 whitelist ${HOME}/.mozilla/firefox/profiles.ini
 read-only ${HOME}/.mozilla/firefox/profiles.ini
 
+no3d
+
 ?HAS_APPIMAGE: ignore private-dev
-private-etc alternatives,ca-certificates,crypto-policies,fonts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,machine-id,nsswitch.conf,pki,resolv.conf,ssl
+private-bin tutanota-desktop
+private-etc alternatives,ca-certificates,crypto-policies,fonts,gtk-2.0,gtk-3.0,ld.so.cache,ld.so
```
glitsj16, line 33: You can't split these private-etc items out on separate lines, parsing will break.

glitsj16 (dismissed):
private-etc needs changes

etc/profile-m-z/tutanota-desktop.profile
```diff
@@ -9,6 +9,9 @@ include globals.local
 noblacklist ${HOME}/.config/tuta_integration
 noblacklist ${HOME}/.config/tutanota-desktop
 
+# breaks tray functionality and secret storage
```
glitsj16, line 12: Is it possible to use dbus filtering instead of opening it all up? You can use d-feet for example to check what tutanota exposes.

glitsj16 (commented):
Can we use dbus-filtering?

etc/profile-m-z/tutanota-desktop.profile
kmk3: > ```diff
> +# breaks tray functionality and secret storage
> +ignore dbus-user none
> ```

See `allow-tray` in firejail.config and `?ALLOW_TRAY` in profiles.

By "breaks secret storage" do you mean that it cannot access things like KDE
Wallet?

glitsj16: So even with something like the below, that 'secret storage' is broken?
```
dbus-user filter
dbus-user.talk org.gnome.keyring.SystemPrompter
#dbus-user.talk org.mozilla.Firefox.*
#dbus-user.talk org.mozilla.firefox.*
ignore dbus-user none
```

glitsj16: I think that xdg-open fails because it's a shell script and the `include disable-shell.inc` blocks that. You can try adding `include allow-bin-sh.inc` to get that working. This would obviously imply adding `sh` to private-bin.

One more remark. I'm trying to register a free account to further test the tutanota-desktop app on my side. I've installed it on my arch linux box and it starts okay, but due to https://github.com/tutao/tutanota/issues/4342 I haven't gotten a working login yet. Contacted their support about this and will update my comment as soon as I get a reply from them.

etc/profile-m-z/tutanota-desktop.profile
```diff
@@ -23,9 +27,17 @@ noblacklist ${HOME}/.mozilla
 whitelist ${HOME}/.mozilla/firefox/profiles.ini
 read-only ${HOME}/.mozilla/firefox/profiles.ini
 
+no3d
+
 ?HAS_APPIMAGE: ignore private-dev
-private-etc alternatives,ca-certificates,crypto-policies,fonts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,machine-id,nsswitch.conf,pki,resolv.conf,ssl
+private-bin basename,cut,gio,grep,mawk,readlink,sh,tr,tutanota-desktop,xdg-open,xdg-settings
+private-etc alternatives,ca-certificates,crypto-policies,fonts,gtk-2.0,gtk-3.0,ld.so.cache,ld.so.preload,nsswitch.conf,pki,resolv.conf,selinux,ssl,xdg
```
rusty-snake, line 34: Can we remove selinux and xdg?

etc/profile-m-z/tutanota-desktop.profile
```diff
@@ -23,9 +27,17 @@ noblacklist ${HOME}/.mozilla
 whitelist ${HOME}/.mozilla/firefox/profiles.ini
 read-only ${HOME}/.mozilla/firefox/profiles.ini
 
+no3d
```
rusty-snake, line 30: Add `machine-id` and `nosound`.

etc/profile-m-z/tutanota-desktop.profile
```diff
@@ -23,9 +27,17 @@ noblacklist ${HOME}/.mozilla
...
+
 ?HAS_APPIMAGE: ignore private-dev
-private-etc alternatives,ca-certificates,crypto-policies,fonts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,machine-id,nsswitch.conf,pki,resolv.conf,ssl
+private-bin basename,cut,gio,grep,mawk,readlink,sh,tr,tutanota-desktop,xdg-open,xdg-settings
```
rusty-snake, line 33: How do you create this list?

etc/profile-m-z/tutanota-desktop.profile
rusty-snake, line 30: So tutanota does not use hardware accelerated rendering?

etc/profile-m-z/tutanota-desktop.profile
```diff
@@ -23,9 +27,17 @@ noblacklist ${HOME}/.mozilla
...
+private-etc alternatives,ca-certificates,crypto-policies,fonts,gtk-2.0,gtk-3.0,ld.so.cache,ld.so.preload,nsswitch.conf,pki,resolv.conf,selinux,ssl,xdg
 private-opt tutanota-desktop
 
+dbus-user filter
+?ALLOW_TRAY: dbus-user.talk org.freedesktop.Notifications
```
rusty-snake, line 38:
```suggestion
dbus-user.talk org.freedesktop.Notifications
```

rusty-snake (changes requested):

etc/profile-m-z/tutanota-desktop.profile
rusty-snake: No, your PR adds it.

Well, Firefox does, GTK app do, chromium does, ... why should tutanota-desktop override it to use software rendering?

rusty-snake: We use a version control system called "git". Instead of saying that some old version ahd it, you can show that it had it and link to the commit which removed it.

etc/profile-m-z/tutanota-desktop.profile
```diff
@@ -23,9 +29,20 @@ noblacklist ${HOME}/.mozilla
 whitelist ${HOME}/.mozilla/firefox/profiles.ini
 read-only ${HOME}/.mozilla/firefox/profiles.ini
 
+machine-id
+nosound
+
 ?HAS_APPIMAGE: ignore private-dev
-private-etc alternatives,ca-certificates,crypto-policies,fonts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,machine-id,nsswitch.conf,pki,resolv.conf,ssl
+private-etc alternatives,ca-certificates,crypto-policies,fonts,gtk-2.0,gtk-3.0,ld.so.cache,ld.so.preload,nsswitch.conf,pki,resolv.conf,ssl
```
rusty-snake, line 36: Any comment on why change this?

etc/profile-m-z/tutanota-desktop.profile
```diff
@@ -9,8 +9,14 @@ include globals.local
 noblacklist ${HOME}/.config/tuta_integration
 noblacklist ${HOME}/.config/tutanota-desktop
 
+# breaks tray functionality and secret storage
```
rusty-snake, line 12:
```suggestion
```

rusty-snake (commented):
Save to disk fix and minor errors:
- Change comment
- Allow some D-Bus (To much?)
- Allow -mnt
- allow sh
- block sound
- change private-etc

kmk3 (dismissed):
To avoid duplication of efforts, I think it would be better to wait for #5298
before proceeding with this.

Then change tutanota-desktop.profile to redirect to electron-mail.profile
(while removing the duplicated entries from the former profile).

kmk3:
> To avoid duplication of efforts, I think it would be better to wait for #5298
> before proceeding with this.
> 
> Then change tutanota-desktop.profile to redirect to electron-mail.profile
> (while removing the duplicated entries from the former profile).

Nevermind, I see now that electron-mail.profile is its own thing/is not a
common profile; I'll comment about it on #5298.

etc/profile-m-z/tutanota-desktop.profile
```diff
@@ -9,23 +9,40 @@ include globals.local
 noblacklist ${HOME}/.config/tuta_integration
 noblacklist ${HOME}/.config/tutanota-desktop
 
+ignore dbus-user none
```
glitsj16, line 12: Suggestion: move this line and place it right above dbus-user filter

etc/profile-m-z/tutanota-desktop.profile
```diff
@@ -9,23 +9,40 @@ include globals.local
...
+ignore disable-mnt
```
glitsj16, line 13: Suggestion: move this line and place it right above machine-id

glitsj16 (commented):
Just two nits about placement and we're good to merge.

etc/profile-m-z/tutanota-desktop.profile
kmk3: > Suggestion: move this line and place it right above dbus-user filter

Shouldn't all ignores go below `include globals.local`?

glitsj16: > Shouldn't all ignores go below include globals.local?

Well, that's actually not very clear if one goes with the guidance of the profile.template. We do have a fair share of profiles that indeed place ignores higher up. And then there's others (like the firefox redirects for example) who place it below added dbus-user filter rules. Personally I like to keep the 'section' logic intact and place them accordingly. As long as the desired end goal is attained I don't hold strong opinions on this. To clarify:

foo.profile
```
[...]
include foo.local
[...]
[...]
```

foo.local
```
ignore protocol
protocol unix,inet
```

versus

foo.local
```
protocol unix,inet
ignore protocol
```

have a different impact on the list of allowed protocols in a sandbox.

Perhaps we can leave it here as is, un-confuse @pirate486743186 about this, and go over all profiles in a separate PR if streamlining ignores is desired? Anyway, I don't see my suggestions as something that would keep back merging this, which is why I didn't opt to 'request changes' for these nits. But I should have communicated that more explicitly.

kmk3: @glitsj16 [on Feb 26](https://github.com/netblue30/firejail/pull/5325/files#r1118021912):

> > Shouldn't all ignores go below include globals.local?
> 
> Well, that's actually not very clear if one goes with the guidance of the
> profile.template. We do have a fair share of profiles that indeed place
> ignores higher up. And then there's others (like the firefox redirects for
> example) who place it below added dbus-user filter rules. Personally I like
> to keep the 'section' logic intact and place them accordingly. As long as the
> desired end goal is attained I don't hold strong opinions on this.

I see, thanks for the details.

> Perhaps we can leave it here as is, un-confuse @pirate486743186 about this,
> and go over all profiles in a separate PR if streamlining ignores is desired?
> Anyway, I don't see my suggestions as something that would keep back merging
> this, which is why I didn't opt to 'request changes' for these nits. But I
> should have communicated that more explicitly.

Sure, let's leave it as is and maybe discuss it after this PR.

kmk3:
(Force-pushed to squash the commits)

glitsj16 (approved):
LGTM

kmk3 (approved):

netblue30:
all merged, thanks!

Issue #5329, opened 2022-08-21
bluez-alsa


imxade:
### Discussed in https://github.com/netblue30/firejail/discussions/5328

<div type='discussions-op-text'>

<sup>Originally posted by **imxade** August 20, 2022</sup>
On **NixOS**,
im using **bluez-alsa** for bluetooth headset

```
firejail --noprofile firefox
```
plays audio through default Bluetooth device

but 
```
firejail firefox
```
doesn't play audio when default device is a Bluetooth one.

i've whitelisted $HOME & $XDG_RUNTIME_DIR but no luck.

In case of **pulseaudio, firejailed firefox** plays fine through bluetooth.

how can i get it working with bluez-alsa ?

imxade:
> Try using PulseAudio or PipeWire instead of raw ALSA? PulseAudio has working firejail integration.

check the 2nd last line

jaens:
If it works with PulseAudio, what's the point of using ALSA?

imxade:
> granting sandboxed programs access to raw audio devices does not seem very security-minded either

Got it, but still is there any way to get it working ?

kmk3:
@imxade commented [on Aug 21](https://github.com/netblue30/firejail/issues/5329#issue-1345409752):

> On **NixOS**, im using **bluez-alsa** for bluetooth headset
> 
> ```
> firejail --noprofile firefox
> ```
> 
> plays audio through default Bluetooth device
> 
> but
> 
> ```
> firejail firefox
> ```
> 
> doesn't play audio when default device is a Bluetooth one.
> 
> i've whitelisted $HOME & $XDG_RUNTIME_DIR but no luck.
> 
> In case of **pulseaudio, firejailed firefox** plays fine through bluetooth.
> 
> how can i get it working with bluez-alsa ?

What is the path to the headset in /dev?

`nou2f` might be blocking it if it shows up in a path matching `/dev/hidraw*`
and Firefox tries to access it through that.

Does adding `browser-disable-u2f no` to /etc/firejail/firejail.config fix it?

If not, try commenting out lines in firefox.profile and firefox-common.profile
to find out what command is blocking it.

imxade:
@kmk3 
on commenting ```#dbus-system none``` bluetooth device is playing fine.

Is there any better way to achieve the same with touching profiles ?

PR #5343, opened 2022-08-28
Update firefox-common.profile

imxade:
enabled tracelog by default


etc/profile-a-l/firefox-common.profile
```diff
@@ -65,4 +65,5 @@ blacklist ${PATH}/wget2
 # 'dbus-user none' breaks various desktop integration features like global menus, native notifications,
 # Gnome connector, KDE connect and power management on KDE Plasma.
 dbus-user none
+# comment next line to use bluez-alsa.
```
rusty-snake, line 68:
```suggestion
# Add 'ignore dbus-system' to your firefox-common.local to use bluez-alsa.
```

rusty-snake, line 68: I didn't followed the bluzed-alsa discussion, but is this a Firefox specific problem or general for all programs (i.e. should go into the wiki page)?

etc/profile-a-l/firefox-common.profile
```diff
@@ -48,8 +48,8 @@ notv
 protocol unix,inet,inet6,netlink
 # The below seccomp configuration still permits chroot syscall. See https://github.com/netblue30/firejail/issues/2506 for possible workarounds.
 seccomp !chroot
-# Disable tracelog, it breaks or causes major issues with many firefox based browsers, see https://github.com/netblue30/firejail/issues/1930.
-#tracelog
+# tracelog, it breaks or causes major issues with older (< v100) firefox based browsers, see https://github.com/netblue30/firejail/issues/1930.
+tracelog
```
rusty-snake, line 52: Todo: Which profiles include firefox-common.

etc/profile-a-l/firefox-common.profile
```diff
@@ -34,7 +34,7 @@ include whitelist-var-common.inc
 
 apparmor
 caps.drop all
-# machine-id breaks pulse audio; add it to your firefox-common.local if sound is not required.
+# machine-id breaks pulse audio; add it to your firefox-common.local if pulse audio is not being used.
```
rusty-snake, line 37: - PulseAudio ?
- Do we expect readers of this comment to know what PA is?
- In which reasonable scenario is sound not used for browsers?

kmk3:
Relates to #5329.

etc/profile-a-l/firefox-common.profile
kmk3: Note that there are similar comments in different profiles:

```console
fd5659e43 2022-08-28 22:00:36 +0200 tuir.profile creation (#5330)
$ git grep '# machine-id ' -- etc
etc/profile-a-l/enpass.profile:# machine-id and nosound break audio notification functionality.
etc/profile-a-l/ephemeral.profile:# machine-id breaks pulse audio; it should work fine in setups where sound is not required.
etc/profile-a-l/firefox-common.profile:# machine-id breaks pulse audio; add it to your firefox-common.local if sound is not required.
etc/profile-m-z/thunderbird.profile:# machine-id breaks audio in browsers; enable or put it in your thunderbird.local when sound is not required
etc/profile-m-z/yelp.profile:# machine-id breaks sound - add the next line to your yelp.local if you don't need sound support.
```

Suggestion: Remove this change from this PR and fix/standardize all
relevant comments later in a new PR.

kmk3: @rusty-snake [on Aug 28](https://github.com/netblue30/firejail/pull/5343#discussion_r956771450):

> * In which reasonable scenario is sound not used for browsers?

I would guess that the point of the `s/sound/pulse audio/` change is that
multiple sound servers exist, but the comment only makes claims about
pulseaudio specifically.

Would be nice to know if `machine-id` affects alsa/sndio/pipewire as well.

rusty-snake: Only PA is broken by `machine-id`, not even pipewire-pulse.

PA is still the most common sound server (default in all major distros except Fedora). So for a user w/o much technical knowledge is PA=sound. Therefore it may be good to hint users that PA is for audio output/input. If you use alsa/jack/... you likely opted-in to it and know that you use it but if you just use the default install of Mint, do you know the name of some background deamon that just does its job?

etc/profile-a-l/firefox-common.profile
imxade: my bad,
even tho I've only tried it with firefox
it shouldn't be firefox specific

etc/profile-a-l/firefox-common.profile
imxade: As it's not enabled by default there should be no problem. I don't think someone will enable it without knowing "what PA is ?".

etc/profile-a-l/firefox-common.profile
rusty-snake: IMHO  https://github.com/netblue30/firejail/wiki/Troubleshooting#common-problems is then the better place.

etc/profile-a-l/firefox-common.profile
rusty-snake:
```
/etc/firejail/abrowser.profile
/etc/firejail/basilisk.profile
/etc/firejail/cachy-browser.profile
/etc/firejail/cliqz.profile
/etc/firejail/cyberfox.profile
/etc/firejail/firedragon.profile
/etc/firejail/firefox.profile
/etc/firejail/icecat.profile
/etc/firejail/librewolf.profile
/etc/firejail/palemoon.profile
/etc/firejail/thunderbird.profile
/etc/firejail/torbrowser.profile
/etc/firejail/waterfox.profile
```

Enabling it for all of them sounds risky to me. We should `ignore` it in untested profiles I guees (all except firefox/librewolf?).

imxade: it has been tested with tor.
should be ignored for the browsers based on older firefox.

rusty-snake (changes requested):
Marking as "Changes Requested" because of https://github.com/netblue30/firejail/pull/5343/files#r957556523

glitsj16:
@imxade Are you still working on this?

imxade:
@glitsj16 no

Issue #5338, opened 2022-08-23
No access to /tmp if blacklisted paths symlink to it


birdie-github:
firejail-0.9.66-3.fc36.x86_64

Relevant debug strings:

```
[birdie@localhost ~]$ firejail --debug
Mounting tmpfs on /run/firejail/mnt directory
631 630 0:24 / /var/tmp rw,nosuid,nodev,relatime master:30 - tmpfs tmpfs rw,seclabel
mountid=631 fsname=/ dir=/var/tmp fstype=tmpfs
Mounting read-only /var/tmp
632 631 0:24 / /var/tmp ro,nosuid,nodev,relatime master:30 - tmpfs tmpfs rw,seclabel
mountid=632 fsname=/ dir=/var/tmp fstype=tmpfs
635 634 0:24 / /var/tmp ro,nosuid,nodev,relatime master:30 - tmpfs tmpfs rw,seclabel
mountid=635 fsname=/ dir=/var/tmp fstype=tmpfs
Mounting noexec /var/tmp
636 635 0:24 / /var/tmp ro,nosuid,nodev,noexec,relatime master:30 - tmpfs tmpfs rw,seclabel
mountid=636 fsname=/ dir=/var/tmp fstype=tmpfs
Mounting tmpfs on /var/lock
Mounting tmpfs on /var/tmp
Mounting tmpfs on /var/log
Mounting tmpfs on /var/lib/nginx
Create the new utmp file
Mount the new utmp file
Relabeling /run/user as /run/user (system_u:object_r:user_tmp_t:s0)
Relabeling /run/user/1000 as /run/user/1000 (unconfined_u:object_r:user_tmp_t:s0)
Disable /tmp/ssh-XXXXXXsy8kyQ
Disable /tmp/.mozilla.ram (requested /home/birdie/.mozilla)
Disable /tmp (requested /home/birdie/.cache/mozilla)
Disable /tmp (requested /home/birdie/.cache/thunderbird)
1045 625 0:30 /pulse /run/firejail/mnt/pulse rw,nosuid,nodev,noexec - tmpfs tmpfs rw,seclabel,mode=755
mountid=1045 fsname=/pulse dir=/run/firejail/mnt/pulse fstype=tmpfs
1046 770 0:30 /pulse /home/birdie/.config/pulse rw,nosuid,nodev,noexec - tmpfs tmpfs rw,seclabel,mode=755
[...]

PR #5344, opened 2022-08-29
docs: note that blacklist/whitelist follow symlinks

kmk3:
Make it more explicit that they do and add an example for each command.

Relates to #5338.

Cc: @birdie-github @rusty-snake (from #5338)

### DIFF
--- a/src/man/firejail.txt
+++ b/src/man/firejail.txt
@@ -191,6 +191,13 @@ Blacklist directory or file. File globbing is supported, see \fBFILE GLOBBING\fR
 .br
 
 .br
+Symbolic link handling: Blacklisting a path that is a symbolic link will also
+blacklist the path that it points to.
+For example, if ~/foo is blacklisted and it points to /foo, then /foo will also
+be blacklisted.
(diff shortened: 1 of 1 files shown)


birdie-github:
> For example, if ~/foo is blacklisted and it points to /foo, then /foo will also be blacklisted.

-> 

`For example, if ~/foo is blacklisted and it points to /bar, then /bar will also be blacklisted.`

Please fix.

netblue30:
all merged, thanks!

kmk3:
@birdie-github commented [on Aug 29](https://github.com/netblue30/firejail/pull/5344#issuecomment-1229903967):

> > For example, if ~/foo is blacklisted and it points to /foo, then /foo will
> > also be blacklisted.
> 
> ->
> 
> `For example, if ~/foo is blacklisted and it points to /bar, then /bar will
> also be blacklisted.`
> 
> Please fix.

Done on commit 160ccd3d9 ("docs: change /foo to /bar in symlink handling
example", 2022-08-29).

PR #5347, opened 2022-08-29
Revert "Add Landlock support to Firejail"

kmk3:
This reverts commit 836ffe37ff891886f15243eacc70963368d57a3f.

This reverts commit 54cb3e741e972c754e595d56de0bca0792299f83, reversing
changes made to 97b1e02d5f4dca4261dc9928f8a5ebf8966682d7.

There were many issues and requests for changes raised in the pull
request (both code-wise and design-wise) and most of them are still
unresolved[1].

[1] https://github.com/netblue30/firejail/pull/5315

Cc: @ChrysoliteAzalea


rusty-snake (approved):

kmk3:
By the way, please wait for this pull request before merging anything else, as
the merge of #5315 is at the tip of master, so doing a revert right after would
make the result cleaner (and could possibly enable the entire revision range to
be ignored in places like .git-blame-ignore-revs).

netblue30:
Yes, lots of issues! I'll put a write-up shortly. We go with it as experimental for the next release, disabled by default at compile time. Meanwhile, you guys go in and fix whatever doesn't look right - I know both you and @rusty-snake  had some problems with it.

It is to early to get any consensus on any particular issue, so we just commit in whatever we think makes sense, and we reconcile them later. Probably we'll end up changing also the names of the commands - the code is not compiled by default anyway.

rusty-snake:
```diff
- --enable-landlock       Landlock self-restriction support
+ --enable-landlock       Landlock self-restriction support [EXPERIMENTAL]
```

ChrysoliteAzalea:
@netblue30 Thank you for invitation. I'm sorry for my absence, I was busy with finding a job.

I would also like to say, I doubt that merging the #5315 to master before it was ready (code style fixes, error message corrections, building on systems that don't support Landlock and other changes were planned) was the best idea.

> disabled by default at compile time. 

How does it work on systems that don't support Landlock (means it's not in the kernel, and there is no userspace support for it)?

I'd like to ask you to avoid merging pull requests with changes requested before proposing fixes, or at least contacting me -- I was working to introduce requested changes, and was a bit surprised to find out that #5315 was already merged.

I'd also like to ask to re-open the #5315 so I could introduce the proposed changes for review.

netblue30:
@ChrysoliteAzalea: Welcome to the club! Don''t bother reopening the old PR, commit the suggestions or any other changes directly in git. It will be marked as experimental in the next release, so the user knows what to expect.

At some point we will have to detect the kernel version at runtime and bring in the functionality. We've done this on several other features, it won't be a problem.

rusty-snake:
> we will have to detect the kernel version

We should use feature detection.

rusty-snake:
Did you mean a "runtime check" in `./configure`?

kmk3:
@netblue30 commented [on Aug 29](https://github.com/netblue30/firejail/pull/5347#issuecomment-1230562159):

> Yes, lots of issues! I'll put a write-up shortly.

> I know both you and @rusty-snake had some problems with it.

Why merge it as is then?

What is the hurry?

This adds technical debt and creates needless rework for anyone that may
eventually decide to fix any of the issues that had already been raised.

Disregarding the fact that a pull request is actively being discussed and
worked on and doing the merge regardless (especially without talking to anyone
involved) not only completely derails the discussions that were ongoing, but
also wastes the time of everyone who contributed, as a merged pull request
cannot be reopened, and the review threads cannot simply be "copied" or "moved"
to a new pull request in a straightforward manner, so all of them are
effectively left in limbo.

Additionally, even if fixed later, the merged code (and all of the issues in
it) remain part of the history in the repository.  This affects anyone that
uses `git log -S` and `git blame` in the future, which are tools not only used
for bisecting, but also for understanding how the existing code came to be.  It
is also unfair to the author, who was working on addressing the issues, which
would have lead to only the finished/amended commits to end up on master.

Also, note that fixing unmerged code is trivial (just amend/rebase and
force-push) compared to fixing existing code, which requires making a new
commit and writing a new commit message, where one would ideally at least
explain what the issue is and why it's being fixed.

All of that said, this pull request was created so that the WIP code can be
reverted, the relevant commits added to .git-blame-revs-ignore and then we can
start on a blank slate again, only merging the contribution when it is ready.

Now looking at the [current master][1], I see that [my comment][2] about
waiting for this pull request (#5347) was also disregarded, which unfortunately
makes the end result not as clean as it could have been, but the revert is
still an improvement over the status quo.

[1]: https://github.com/netblue30/firejail/compare/836ffe37ff891886f15243eacc70963368d57a3f...0594c5d3d0f1ddc4049cf2ed38676a1cdc8d6843
[2]: https://github.com/netblue30/firejail/pull/5347#issuecomment-1230472237

> It is to early to get any consensus on any particular issue

There appeared to be enough consensus on #5315 for at least the raised issues
to start being addressed, which indeed [seems to be what was happening][4].

> Probably we'll end up changing also the names of the commands

I was going to suggest on the pull request to only add the most basic support
first (such as configure-time support and the syscall wrappers) and leave the
rest (such as any new commands) for later.

I had already posted many review comments and I never would have expected a
pull request that was actively being worked on (and that had multiple requests
for changes) to be merged out of the blue, so I ended up leaving that
suggestion for later.

> Meanwhile, you guys go in and fix whatever doesn't look right

That is basically what was already being done on the pull request.

The main difference is that there was a thread for each issue in the pull
request, which were being discussed and fixed together with the author.

Why merge a pull request full of issues that weren't done being fixed and then
suggest that they be fixed later when the issues could have been fixed right at
the source?

> so we just commit in whatever we think makes sense, and we reconcile them
> later.

It would likely have been much easier to just checkout the source branch, make
every change myself that I thought made sense and commit them, compared to
writing a review comment for each issue.

The problem is that doing so precludes collaboration.

Firstly, I'm not necessarily 100% sure of every suggestion that I make and I
appreciate feedback from the author and from other contributors.

And secondly, working with the author helps them (and we) better understand the
codebase and improve their contributions, which should help improve their
future contributions (thus making future reviews and maintenance easier).

Also, note that by engaging in the discussions, the author shares the burden of
reviewing and committing the suggestions.  Considering that many pull requests
are made on every release and that whatever problems are not caught and dealt
with before merging them will result in yet more technical debt being mounted
upon the project, working with the authors directly seems easier and more
sustainable than a free-for-all, "anyone goes and commits whatever they feel
like on master and maybe someone else will fix the problems later", which
appears to be what is implied.

And indeed, considering @rusty-snake's [approval][3] and @ChrysoliteAzalea's
comments [here][4] and [here][5], discussing and improving the pull request
together seems to be what everyone involved wanted to be doing (and were in
fact doing on #5315).  So again, what is the point of going and killing off all
the discussions?

See also #4517.

[3]: https://github.com/netblue30/firejail/pull/5347#pullrequestreview-1088838068
[4]: https://github.com/netblue30/firejail/pull/5347#issuecomment-1231221280
[5]: https://github.com/netblue30/firejail/pull/5315#issuecomment-1231223352

kmk3:
Updated and force-pushed to fix conflict with commit 6e687c301 ("tracelog
disabled by default in /etc/firejail/firejail.config file", 2022-08-29), which
contains landlock-related changes.

kmk3:
Updated and force-pushed to also revert changes from a couple of commits that
were missed.  Removed the reference to #5315 from most commit messages to avoid
spamming the PR with timeline events.

netblue30:
Sorry guys, I have to reject this PR.

@kmk3: What you say is correct, and that's basically what we've been doing so far. It works fine for released features where it is critical to keep an eye on the history.

However, landlock has no history, it is just starting up. As new landlock features become available in the kernel, we will move a lot of the existing functionality to landlock. I am talking about our blacklist/whitelist implementation based on SUID and mount/bind. As soon as the kernel implements landlock blacklists, we phase out ours and move it to landlock. So, it is much better to put landlock in early, and fix all the integration problems we run into.

I've done some tests on Fedora (kernel 5.19), landlock is working fine as implemented by @ChrysoliteAzalea.  Probably we should drop the experimental flag and go live with it in Arch, Fedora and Gentoo - this way we get more testing from outside people.

kmk3:
@netblue30 commented [on Sep 4](https://github.com/netblue30/firejail/pull/5347#issuecomment-1236328910):

> Sorry guys, I have to reject this PR.
> 
> @kmk3: What you say is correct, and that's basically what we've been doing so
> far. It works fine for released features where it is critical to keep an eye
> on the history.
> 
> However, landlock has no history, it is just starting up.

I'm not sure what you mean by this.  To clarify, I previously mentioned
"history" in the sense of the git/repository history.  That is, the main issues
raised here are not specific to Landlock nor to the Landlock support itself.
In fact, I personally would really like to see Landlock being used in firejail
(as it is an unprivieged operation, as you allude to later in the comment) and
I'm reasonably sure that adding some form of Landlock support is something that
would have happened eventually anyway.

The first main issue is with the development process.

That is, with unilaterally merging an unfinished pull request that was actively
being discussed and worked on, which even the author [has asked][1] to avoid
doing after the fact.  There is probably not much that can be done about this
specific instance now, so I suppose that we can leave the process improvements
for a later discussion.

The second main issue is with the result of the merge.

To summarize the main reasons for doing the revert:

* Commits that were WIP and that were made in the author's name (and that they
  might not approve of) would not only be part of the git history, but also
  effectively be the official/initial Landlock commits, which the author might
  (understandably) not be very happy about
* WIP commits stay forever in the git history/repository, making it harder to
  understand during debugging and development (such as when using `git blame`)
* Reviewing and changing unmerged code is easier than code that is already on
  master (for example, there is no need to explain the changes being done and
  diff hunks that only add lines are easier to read and review)

This pull request is about fixing the above problems (which AFAIK can only be
fixed through a revert + adding the commits to .git-blame-ignore-revs), so that
we can get back to working on the issues raised on #5315.

Also, note that Landlock-related commits made after #5315 can later be
cherry-picked back.

I don't think that it would take a substantial amount of time overall for us to
finish working on #5315 and there is no public roadmap for the next release, so
I don't understand what would be the benefits of rushing the WIP code to
master.

Any clarifications would be appreciated.

[1]: https://github.com/netblue30/firejail/pull/5347#issuecomment-1231221280

kmk3:
(Rebased to master to fix conflicts)

netblue30:
OK, let's do it as you suggested, You run the show, git revert/rebase/forced push - whatever you think is necessary. We need a place to keep the discussion going, maybe the issue #5315 open by @rusty-snake?

kmk3:
@netblue30 commented [on Sep 5](https://github.com/netblue30/firejail/pull/5347#issuecomment-1237005711):

> OK, let's do it as you suggested, You run the show, git revert/rebase/forced
> push - whatever you think is necessary.

Alright, thanks!

> We need a place to keep the discussion going, maybe the issue #5315 open by
> @rusty-snake?

I assume you meant #5354, in which case sure.

I'll merge this, then put together a new `landlock_v2` branch with the existing
Landlock commits plus some fixes in new commits and post there when it's ready.

@ChrysoliteAzalea after that you can change it if needed, squash it, open a new
pull request and we go from there.  What do you think?

PR #5352, opened 2022-08-31
fix opening links in firefox youtube-viewers-common.profile

pirate486743186:
title...

### DIFF
--- a/etc/profile-m-z/youtube-viewers-common.profile
+++ b/etc/profile-m-z/youtube-viewers-common.profile
@@ -19,6 +19,13 @@ include allow-perl.inc
 include allow-python2.inc
 include allow-python3.inc
 
+# The lines below are needed to find the default Firefox profile name, to allow
+# opening links in an existing instance of Firefox (note that it still fails if
+# there isn't a Firefox instance running with the default profile; see #5352)
+noblacklist ${HOME}/.mozilla
+whitelist ${HOME}/.mozilla/firefox/profiles.ini
+read-only ${HOME}/.mozilla/firefox/profiles.ini
+
 include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
@@ -55,5 +62,8 @@ private-dev
 private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,gtk-2.0,gtk-3.0,host.conf,hostname,hosts,ld.so.cache,ld.so.preload,machine-id,mime.types,nsswitch.conf,passwd,pki,pulse,resolv.conf,ssl,X11,xdg
 private-tmp
 
-dbus-user none
+dbus-user filter
+# allow D-Bus communication with firefox for opening links
+dbus-user.talk org.mozilla.*
+
 dbus-system none


rusty-snake:
Did you tested this? What's your environment?

rusty-snake:
Which DE? WM? Xorg or Wayland? How is FF configured, to use X11? Which FF version?

If firefox runs as native Wayland client it uses D-Bus for it's IPC. Uner X11 it used some X11 magic for IPC in the past but IIRC this changed a few month ago.

mid-kid:
I found this trying to find a way to do this, thanks a lot for the PR.
However, in my case I use firefox esr, which needs `dbus-user.talk org.mozilla.firefox_esr.*`

glitsj16:
@mid-kid

> However, in my case I use firefox esr, which needs dbus-user.talk org.mozilla.firefox_esr.*

OFFTOPIC:
Your comment made me realize we should track down these different exposed D-Bus names for all supported firefox editions and add them to the relevant profiles. Thanks!

rusty-snake:
Is the Firefox_ESR Name use by officially build or is this a downstream patch? I never saw it before and IIRC every instance has it's own name so there is no reason to have a different prefix.

glitsj16:
> Is the Firefox_ESR Name use by officially build or is this a downstream patch?

I'm still looking for the 'proper' upstream source/documentation for these exposed D-Bus names relating to Mozilla browsers. No luck so far, but my search fu probably is too limited for the goal.

> I never saw it before and IIRC every instance has it's own name so there is no reason to have a different prefix.

A bit confused here. I also thought (still do for that matter) that each Firefox edition has its own prefix. I'm confused as to why you state there's no reason to use those. But, again, I might be overlooking things, or confusing these comments with #5360...

etc/profile-m-z/youtube-viewers-common.profile
```diff
@@ -55,5 +60,10 @@ private-dev
 private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,gtk-2.0,gtk-3.0,host.conf,hostname,hosts,ld.so.cache,ld.so.preload,machine-id,mime.types,nsswitch.conf,passwd,pki,pulse,resolv.conf,ssl,X11,xdg
 private-tmp
 
-dbus-user none
+dbus-user filter
+# allow D-Bus communication with firefox for opening links
+dbus-user.talk org.mozilla.Firefox.*
+dbus-user.talk org.mozilla.firefox.*
+dbus-user.talk org.mozilla.firefox_esr.*
```
rusty-snake, line 67: What about
```suggestion
dbus-user.talk org.mozilla.*
```
to cover all (developeredition, ...).

glitsj16: Elegant suggestion! It would sure bring in more sanity regarding future maintenance of the FF related profiles. I realize you can't see and/or respond to each and every thing that comes into the Firejail repo, but if you have some time I would appreciate a review of #5360, as this impacts that too. Regardless, and as a personal note, I would like to thank you for all the work you have and hopefully will be doing for Firejail, top notch :+1: .

glitsj16:
@pirate486743186 Yeah, that sounds very credible to me. Personally I use their Firefox 'unbranded' edition, which also carries yet another specific D-Bus naming scheme. But that's not in our official profiles last I checked so I left that out of this discussion. All in all a fine mess, to which there's now the elegant fix suggested by @rusty-snake... Thx for keeping this PR going.

etc/profile-m-z/youtube-viewers-common.profile
kmk3: >     dbus-user.talk org.mozilla.*

Wouldn't this also allow talking to something like Thunderbird?

Or to anything else that Mozilla currently (and eventually) publishes besides
Firefox.

rusty-snake: Yes

kmk3: @rusty-snake [on Sep 7](https://github.com/netblue30/firejail/pull/5352#discussion_r965036075):

> Yes

Isn't that too broad?

Why let a media player talk to an email client?

rusty-snake: - What does thunderbird expose via D-Bus?
- Why let it speak with an Browser? An Browser in even more sensitive IMHO.

Anyway, that's the details of the function if somebody want to create a call rule (which I'm still not happy to have them in profiles for maintaince reasons).

- Bus: session
- Bus Name: `org.mozilla.<firefox>.<instance>`
- Object Path: `/org/mozilla/firefox/Remote`
- Interface: `org.mozilla.firefox`
- Method: `OpenURL (url: ay) -> ()`

etc/profile-m-z/youtube-viewers-common.profile
```diff
@@ -19,6 +19,11 @@ include allow-perl.inc
 include allow-python2.inc
 include allow-python3.inc
 
+# These lines are needed to allow Firefox to open links
+noblacklist ${HOME}/.mozilla
+whitelist ${HOME}/.mozilla/firefox/profiles.ini
+read-only ${HOME}/.mozilla/firefox/profiles.ini
```
kmk3, line 25:
```suggestion
# These lines are needed on older Firefox versions (that use Xorg for IPC) to
# allow opening links on an existing instance of Firefox (note that it still
# fails if there are no running instances)
#noblacklist ${HOME}/.mozilla
#whitelist ${HOME}/.mozilla/firefox/profiles.ini
#read-only ${HOME}/.mozilla/firefox/profiles.ini
```

Changes:

* Explain the reasoning in more details
* Leave the code commented

Leave it commented because:

* It appears brittle/undocumented (as in, may break in future versions)
* It supposedly only affects older versions of Firefox
* It looks like it could result in Firefox creating a new profile (as the
  existing profile directories would not be visible)

The default Firefox profile is arguably rather lacking in security/privacy and
may not have the desired extensions (such as an adblocker) and network
configuration (proxies, etc).

etc/profile-m-z/youtube-viewers-common.profile
kmk3: Sorry for the late reply, I've been busy working on something else.

Feel free to keep the dbus part as is.

I think I'll just open a dedicated discussion about it later.

etc/profile-m-z/youtube-viewers-common.profile
rusty-snake: `profiles.ini` is required by Firefox to know to which instance it should talk (the hash of the profile name in `profiles.ini` is part of the D-Bus name).

rusty-snake: Could you reproduce something else than this?

Your Firefox profile cannot be loaded. It may be missing or inaccessible.

kmk3: @rusty-snake [on Sep 11](https://github.com/netblue30/firejail/pull/5352#discussion_r967797814):

> `profiles.ini` is required by Firefox to know to which instance it should
> talk (the hash of the profile name in `profiles.ini` is part of the D-Bus
> name).

I see; I thought it was only needed for the xorg IPC.

Does it depend on access to profiles.ini because it is used to find out which
one is the default profile?

What happens if there is only a non-default-profile instance of Firefox running
(without `--no-remote`)?

@rusty-snake [on Sep 11](https://github.com/netblue30/firejail/pull/5352#discussion_r967798603):

> > It looks like it could result in Firefox creating a new profile (as the
> > existing profile directories would not be visible
> 
> Could you reproduce something else than this?
> 
> Your Firefox profile cannot be loaded. It may be missing or
> inaccessible.

I have not tested this; my suggestion above was based on what I understood from
reading the following comments from @pirate486743186 (which only mention xorg
and nothing about dbus nor default profiles):

* <https://github.com/netblue30/firejail/pull/5352#issuecomment-1234217847>
* <https://github.com/netblue30/firejail/pull/5352#issuecomment-1234244758>

rusty-snake: > What happens if there is only a non-default-profile instance of Firefox running
(without --no-remote)?

The same as there is no firefox running.

kmk3: @rusty-snake [on Sep 11](https://github.com/netblue30/firejail/pull/5352#discussion_r967823024):

> > What happens if there is only a non-default-profile instance of Firefox
> > running (without --no-remote)?
> 
> The same as there is no firefox running.

Nice.

Updated suggestion:

```suggestion
# The lines below are needed to find the default Firefox profile name, to allow
# opening links in an existing instance of Firefox (note that it still fails if
# there isn't a Firefox instance running with the default profile; see #5352)
noblacklist ${HOME}/.mozilla
whitelist ${HOME}/.mozilla/firefox/profiles.ini
read-only ${HOME}/.mozilla/firefox/profiles.ini
```

kmk3 (commented):
Could you squash/fixup the commits and force-push?

Example using `git rebase`:

* <https://github.com/netblue30/firejail/pull/4779#issuecomment-993932535>

Also, feel free to amend the commits by default and force-push (especially for
more trivial changes), as GitHub adds a timeline event with a link to a
comparison of the commits before/after the force-push.

Other than that, LGTM.

rusty-snake (commented):
IMHO we need a better solution than supporting only one browser by adding the same boilerplate into a lot profiles.

Anyway LGTM.

PR #5353, opened 2022-08-31
Make /etc/ssh/ssh_config.d visible for ssh

slowpeek:
Currently ssh client cant see anything under `/etc/ssh/ssh_config.d`. It is evidently a bug.

### DIFF
--- a/etc/inc/allow-ssh.inc
+++ b/etc/inc/allow-ssh.inc
@@ -5,6 +5,7 @@ include allow-ssh.local
 noblacklist ${HOME}/.ssh
 noblacklist /etc/ssh
 noblacklist /etc/ssh/ssh_config
+noblacklist /etc/ssh/ssh_config.d
 noblacklist ${PATH}/ssh
 noblacklist /tmp/ssh-*
 # Arch Linux and derivatives


glitsj16 (approved):
LGTM

glitsj16:
Just curious, does /etc/ssh/ssh_config.d exists on Debian(-based) machines? Or do you have this dir on yet another Linux distro?

kmk3:
Relates to commit 3849e1201 ("allow-ssh.inc: allow /etc/ssh/ssh_config",
2021-01-09) and commit 2ec3f3a96 ("disable-common.inc: add missing openssh
paths", 2021-01-09) from PR #3885.

kmk3 (approved):
@slowpeek commented [on Aug 31](https://github.com/netblue30/firejail/pull/5353#issue-1357921593):

> Currently ssh client cant see anything under `/etc/ssh/ssh_config.d`. It is
> evidently a bug.

Considering that `/etc/ssh/ssh_config` exists and that `ssh_config(5)` supports
`Include`, `/etc/ssh/ssh_config.d` does seem like the most obvious path to
store extra configuration files, so I agree with `noblacklist`ing it.

But that path does not appear to be used by openssh itself nor by any other
package at least on Arch/Artix:

```console
$ pacman -F /etc/ssh/ssh_config
etc/ssh/ssh_config is owned by system/openssh 9.0p1-1
$ pacman -F /etc/ssh/ssh_config.d
$
$ zgrep -F etc/ssh/ssh_config /usr/share/man/man5/ssh_config.5.gz
.Pq Pa /etc/ssh/ssh_config
.Pa /etc/ssh/ssh_config
.It Pa /etc/ssh/ssh_config
```

So unless one or more of the following are true:

* There is a distro that patches openssh to create `/etc/ssh/ssh_config.d` by
  default and/or to make `/etc/ssh/ssh_config` include every file in that
  directory
* There is a package that creates files in `/etc/ssh/ssh_config.d`

Then you have to go out of your way to create and use that directory
specifically, which makes the issue presented sound rather niche compared to
"evidently a bug".

glitsj16:
@kmk3 Looking at the `Debian` man page for [ssh_config(5)](https://manpages.debian.org/unstable/openssh-client/ssh_config.5.en.html) I see: 
```
Note that the Debian openssh-client package sets several options as standard in /etc/ssh/ssh_config which are not the default in ssh(1):

    Include /etc/ssh/ssh_config.d/*.conf
```

So at least on Debian this dir exists, which is why I approved with LGTM in the first place and asked above if they might have it on a non-Debian distro too. I can see how the wording "evidently a bug" might be somewhat overkill, but IMO we should add it nonetheless. Just my opinion though...

kmk3:
@glitsj16 commented [on Sep 1](https://github.com/netblue30/firejail/pull/5353#issuecomment-1233686680):

> @kmk3 Looking at the `Debian` man page for
> [ssh_config(5)](https://manpages.debian.org/unstable/openssh-client/ssh_config.5.en.html)
> I see:
> 
> ```
> Note that the Debian openssh-client package sets several options as standard in /etc/ssh/ssh_config which are not the default in ssh(1):
> 
>     Include /etc/ssh/ssh_config.d/*.conf
> ```
> 
> So at least on Debian this dir exists, which is why I approved with LGTM in
> the first place and asked above if they might have it on a non-Debian distro
> too.

Thanks for the details; it makes a lot more sense now.

> I can see how the wording "evidently a bug" might be somewhat overkill, but
> IMO we should add it nonetheless. Just my opinion though...

Yes, I agree with adding it regardless; my point was that it would be nice to
have some basic evidence provided along with such claims, especially when they
are based on downstream patches.

glitsj16:
> Yes, I agree with adding it regardless; my point was that it would be nice to
have some basic evidence provided along with such claims, especially when they
are based on downstream patches.

No argument there :+1: .

rusty-snake (approved):

reinerh (approved):

PR #5360, opened 2022-09-05
Streamline Firefoxes D-Bus filtering

glitsj16:
The old CamelCased D-Bus names are no longer present in recent Firefox editions (checked via d-feet). Drop those unused names and add more specific D-Bus filter to our supported Firefox editions.
~~UPDATE: let's add/keep CamelCase D-Bus names for backward-compatibility.~~
UPDATE: implemented the elegant suggestion in [here](https://github.com/netblue30/firejail/pull/5352/files/c04ceb49c7001bb931a35667c831545cd64aeece#r963032151), which brings back some much needed sanity to the Firefox related profiles IMO.

### DIFF
--- a/etc/profile-a-l/firefox-beta.profile
+++ b/etc/profile-a-l/firefox-beta.profile
@@ -6,9 +6,5 @@ include firefox-beta.local
 # added by included profile
 #include globals.local
 
-# Edition-specific D-Bus filters
-dbus-user.own org.mozilla.Firefox_Beta.*
-dbus-user.own org.mozilla.firefox_beta.*
-
 # Redirect
 include firefox.profile
--- a/etc/profile-a-l/firefox-beta.profile
+++ b/etc/profile-a-l/firefox-beta.profile
@@ -6,5 +6,8 @@ include firefox-beta.local
 # added by included profile
 #include globals.local
 
+# Edition-specific D-Bus filter
+dbus-user.own org.mozilla.firefox_beta.*
+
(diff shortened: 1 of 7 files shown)


kmk3:
@glitsj16 commented [on Sep 5](https://github.com/netblue30/firejail/pull/5360#issue-1361347649):

> The old CamelCased D-Bus names are no longer present in recent Firefox
> editions (checked via d-feet).

Any idea more or less when they might have been removed?

> Drop those unused names and add more specific D-Bus filter to our supported
> Firefox editions.

Why not keep them for backwards compability?

glitsj16:
@kmk3

> Any idea more or less when they might have been removed?

Alas, I couldn't find clear documentation on this from Mozilla. I checked the source code, looked around on https://searfox.org and downloaded a few editions to check with d-feet.

> Why not keep them for backwards compatibility with older Firefox versions?

Fair point. Well, I mean I think it is in theory. Keeping around and using outdated web browser versions wouldn't be something I'd recommend/feel comfortable with. Regardless of whatever web browser it is. But keeping these outdated CamelCase strings around for a while longer at least isn't breaking anything, nor would that cause insecurities IMO. So I put them back in.

rusty-snake:
I'm not sure if these camelcase names (Firefox_Esr, ...) where ever used. They look very ugly and I would be suprised if the where choose by mozilla. The developer edition is special for instance and profile handling, so maybe only it need this version.

Issue #5190, opened 2022-06-10
Too many features


netblue30:
Done so far:

* removed cgroups code
* changed --disable-firetunnel into --enable-firetunnel in configure.ac
* disabled chroot by default in /etc/firejail/firejail.config
* shell none set as default
* removed --shell=
* private-lib disabled by defautl in /etc/firejail/firejail.config

### Discussed in https://github.com/netblue30/firejail/discussions/5183

<div type='discussions-op-text'>

<sup>Originally posted by **SkewedZeppelin** June  8, 2022</sup>
The past few versions have been adding more and more arguable unnecessary features further increasing attack surface of firejail.

The best part of firejail is the collection of ready to use profiles for programs, nothing else offers this.
What ever happened to the minimal LTS version?

Can these be #ifdef'ed out?
- [x] ids is --disable-ids
- [ ] --appimage
- [x] --bandwidth, --netstats, --nettrace are --disable-network
- [ ] --build?
- [ ] --cpu?
- [x] --tunnel is --disable-firetunnel

any others?</div>

<sup>Originaly posted by **rusty-snake** June  8, 2022</sup>
> Can these be #ifdef'ed out?

While for us #ifdef works good, the most users will get their firejail binary from their package-manager. Therefore we should make sure that there are options in firejail.config to disable those features too (--appimage has one).

> any others?

- [ ] --cgroup
- [ ] --shutdown
- [ ] --trace (subset of --build)
- [ ] --tracelog?

netblue30:
And that's exactly what we are going to do! I would say for most features we'll have compile time and run time configuration. Then we need to talk what we keep in the default configuration most people are using.

amano-kenji:
You can split privileged parts into a much smaller binary. Then, you can add more features in another binary.

rusty-snake:
#510 and a few other places

netblue30:
> You can split privileged parts into a much smaller binary. Then, you can add more features in another binary.

We've been doing it for a few years now. The files starting in "f" in /usr/lib/firejail directory (fseccomp, fcopy etc.) have been split out from SUID binary and are run in a temporary sandbox (see src/firejail/sbox.c). We even run external some programs this way - the largest one is ISC DHCP client (dhclient). I would say adding new fancy features in this moment is quite easy.

netblue30:
Is anybody using --tracelog or --cgroup? I would take them out, I mean completely gone!

rusty-snake:
> --cgroup

[...]

PR #5366, opened 2022-09-13
docs: man: Note that some commands can be disabled in firejail.config

glitsj16:


rusty-snake (commented):
LGTM

---

IMHO we can expand the firejail.config docu in the manpage even more.
- There's no central firejail.config section where it is explaind, what it does, where it is, ... just random notes like "can be configured in firejail.config".
- We could add a harden firejail with firejail.config by setting ...
- We could add notes about allow-tray, browser-allow-drm, ...
- We could add a note about restricted-network being yes on Debian.
- ...

src/man/firejail.txt
```diff
@@ -319,6 +319,7 @@ Print content of file from sandbox container, see FILE TRANSFER section for more
 Chroot the sandbox into a root filesystem. Unlike the regular filesystem container,
 the system directories are mounted read-write. If the sandbox is started as a
 regular user, nonewprivs and a default capabilities filter are enabled.
+Disabled by default in firejail.config.
```
kmk3, line 322: Note that the defaults may change, such as with `private-lib` in #5190.

So to avoid having to also update a man page when changing a default, just note
what is the relevant configuration option:

```suggestion
Note: This command may be disabled by using "chroot no" in firejail.config.
```

glitsj16: Indeed that's a better idea. As these are already disabled by default it makes more sense to detail how to enable them. Hence the rephrasing. Did the same for private-lib and tracelog. Thanks for the suggestion.

kmk3: @glitsj16 glitsj16 [on Sep 13](https://github.com/netblue30/firejail/pull/5366#discussion_r970075159):

> Thanks for the suggestion.

No problem.

> Indeed that's a better idea. As these are already disabled by default it
> makes more sense to detail how to enable them. Hence the rephrasing. Did the
> same for private-lib and tracelog.

The new version does look nice:

> ```
> Note: this command may be enabled in firejail.config with "chroot yes".
> ```

But "may be enabled" still implies that it is disabled by default, which
ideally would also be changed (for consistency) if/when the default changes in
firejail.config.

Note that this extra work affects not only us, but any downstream that happens
to change the defaults (such as with the Debian patch).  It seems trickier to
patch the man page (and to keep the patch working with man page updates)
compared to just running something like the following:

```sh
sed -i.bak 's/chroot no/chroot yes' etc/firejail.config
```

So how about just mentioning what the relevant option is in firejail.config?
Example:

```diff
-Note: this command may be enabled in firejail.config with "chroot yes".
+Note: Support for this command is controlled in firejail.config with the
+\fBchroot\fR option.
```

reinerh:
What about the firejail-profile manpage? Should this also get mentioned there, or is it sufficient in the firejail manpage?

kmk3 (commented):
I don't know if this PR is intended to be comprehensive, but note that there
are many more options in firejail.config that might affect existing commands.

Examples, including the options that are already in this PR (and formatted as
"firejail.config option: commands"):

* apparmor: --apparmor
* bind: --bind
* chroot: --chroot
* dbus: --dbus-user, --dbus-system
* disable-mnt: --disable-mnt
* file-transfer: --ls, --get, --put, --cat (maybe note it only once in "FILE
  TRANSFER"?)
* join: --join, --join-filesystem, --join-network, --join-or-start
* name-change: --name (already on man page, but in a different format)
* overlayfs: --overlay
* private-bin: --private-bin
* private-cache: --private-cache
* private-etc: --private-etc
* private-home: --private-home
* private-lib: --private-lib
* private-opt: --private-opt
* private-srv: --private-srv
* tracelog: --tracelog

There are also the following options, which I'm less sure about all the
commands that they might or might not affect:

* network
* restricted-network
* seccomp
* userns
* x11

src/man/firejail.txt
```diff
@@ -2085,6 +2087,8 @@ The files and directories in the list must be expressed as relative to
 the /lib directory.
 The idea is to build a new /lib in a temporary filesystem,
 with only the library files necessary to run the application.
+Note: Support for this command is controlled in firejail.config with the
+\fBprivate-lib\fR option.
 It could be as simple as:
```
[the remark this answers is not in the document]
glitsj16: Done. That's indeed a better place to add the notes.

glitsj16:
@reinerh

> What about the firejail-profile manpage? Should this also get mentioned there, or is it sufficient in the firejail manpage?

Like @kmk3 hinted on, this PR was never meant to be comprehensive. For now I think it gets the most exposure by placing the notes in the firejail manpage.

src/man/firejail.txt
```diff
@@ -2859,6 +2865,9 @@ $ firejail \-\-tree
 .br
   11970:netblue:transmission-gtk
 
+.br
+Note: Support for this command is controlled in firejail.config with the
++\fBtracelog\fR option.
```
kmk3, line 2870:
```suggestion
  11970:netblue:transmission-gtk
.br

.br
Note: Support for this command is controlled in firejail.config with the
\fBtracelog\fR option.
```

Remove the extraneous `+` at the beginning and use the double `.br` commands to
ensure that there is a blank newline in between.  Example:

```
Line 1.
.br

.br
Line 2.
```

Note: There are better ways to accomplish this, but this is what is usually
done in this man page.

Lastly, the `tracelog` note should be moved back to the relevant section (it's
in the `--tree` section).

reinerh (approved):

src/man/firejail.txt
```diff
@@ -2858,7 +2871,6 @@ $ firejail \-\-tree
 #endif
 .br
   11970:netblue:transmission-gtk
-
```
kmk3, line 2861: Suggestion: Undo this change as it's unrelated.

Other than that, LGTM.

glitsj16: Done!

kmk3 (approved):
LGTM.

rusty-snake:
FTR

> disable-mnt: --disable-mnt

disable-mnt in firejail.config is an enforcment setting (like force-nonewprivs) and does not affect disable-mnt.

> userns

maps to `noroot`. (https://github.com/netblue30/firejail/pull/5304#issuecomment-1212216757)

PR #5368, opened 2022-09-13
discord: fix issues with command-line usage

glitsj16:
This PR fixes #5367.

PR #5369, opened 2022-09-13
Test sort.py collate

kmk3:


glitsj16 (approved):
LGTM. I was a bit too late seeing this and assumed my sorting was broken.

kmk3:
Sorry, I meant to open this on my own repository for testing.

Even when clicking to open the PR from there, for some reason GitHub defaults
to diregarding the current page and always opening the PR on the upstream for
some reason...

kmk3:
So this PR doesn't actually work.

I looked at sort.py again after opening this and it seems that ignoring the
case is intentional (as it uses `casefold`):

```python
    items.sort(key=lambda s: s.casefold())
```

I would guess that the intent to make it easier to group similarly-named paths
on long `private-` lists.  Example:

```
private-bin awk,Bar,bar,cut,Foo,foo,grep
```

Though it does mean that it sees `Foo,foo` and `foo,Foo` as equivalent.

Either way, this is not just a simple bug as I had thought and I don't feel
like comparing the pros and cons of each sorting method at the moment, so I'm
closing this.

kmk3:
Relates to #5368.

rusty-snake:
IIRC I used casefold because case insensitive sorting was the current sorting used in the most profiles.

PR #5370, opened 2022-09-13
Remove deprecated syntax and modernize shell test scripts

a1346054:
I performed the minimum number of changes that still preserves the same old behavior, but fixes some errors and other minor issues.


test/compile/compile.sh
```diff
@@ -410,23 +409,23 @@ echo "**********************************************************"
...
+echo "${arr[10]}"
+echo "${arr[11]}"
+echo "${arr[12]}"
+echo "${arr[13]}"
+echo "${arr[14]}"
+echo "${arr[15]}"
+echo "${arr[16]}"
+echo "${arr[17]}"
+echo "${arr[18]}"
+echo "${arr[19]}"
```
rusty-snake, line 431: We could use a loop here.

test/filters/filters.sh
```diff
@@ -7,37 +7,37 @@ export MALLOC_CHECK_=3
...
-elif [ "$(uname -m)" = "i686" ]; then
-    echo "TESTING: memory-deny-write-execute (test/filters/memwrexe-32.exp)"
-    ./memwrexe-32.exp
+if [[ $(uname -m) == "x86_64" ]]; then
+	echo "TESTING: memory-deny-write-execute (test/filters/memwrexe.exp)"
+	./memwrexe.exp
+elif [[ $(uname -m) == "i686" ]]; then
+	echo "TESTING: memory-deny-write-execute (test/filters/memwrexe-32.exp)"
+	./memwrexe-32.exp
 else
```
rusty-snake, line 29: Note to me: Can this be refactored?

rusty-snake (approved):
LGTM

test/compile/compile.sh
a1346054: yes, it can be replaced with `for i in {1..19}; do echo "${arr[$i]}"; done` but I didn't want to make more changes than necessary

netblue30:
merged, thanks!

PR #5376, opened 2022-09-19
.git-blame-ignore-revs: change format and add Landlock v1

kmk3:
Put the committer date ("%cd") and commit message subject ("%s") as a
comment right after the commit hash ("%H").  The pipe ("|") is intended
to improve readability.

This format looks more similar to the output of `git log --oneline` and
should make it easier to visually parse long lists.  It also allows more
hashes to fit into the same amount of lines (at the cost of longer
lines).

Use committer date ("%cd") instead of author date ("%ad") as the former
tells when a commit actually landed on master.  This usually matches the
topological order, which should make it easier to verify that all of the
relevant commits are listed (and that there are no extraneous commits).

Use --date + "%cd" instead of just "%ci" because the former can be used
to ensure that the date is always printed in the same timezone (UTC).

Relates to #5315 #5347.

Cc: @ChrysoliteAzalea @netblue30 @rusty-snake (from #5347)

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,4 +1,4 @@
-# move whitelist/blacklist to allow/deny
-fe0f975f447d59977d90c3226cc8c623b31b20b3
 # Revert "move whitelist/blacklist to allow/deny"
 f43382f1e9707b4fd5e63c7bfe881912aa4ee994
+# move whitelist/blacklist to allow/deny
+fe0f975f447d59977d90c3226cc8c623b31b20b3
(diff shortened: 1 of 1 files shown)


kmk3:
Note: I wanted to get this out of the way first, but the `landlock_v2` branch
is basically ready.  I'll post it after this PR.

kmk3:
Note2: This works locally, but GitHub is showing the new format in red,
probably because its syntax highlighter isn't aware that .git-blame-ignore-revs
supports comments after the hash.

netblue30:
Let's try it out, thanks!

PR #5387, opened 2022-10-01
disable-common.inc: blacklist sudo/doas paths in /etc

kmk3:
Commands used to find the relevant paths in /etc:

    $ pacman -Qo /etc/* 2>/dev/null | grep sudo | LC_ALL=C sort
    /etc/pam.d/ is owned by sudo 1.9.14.p1-1
    /etc/sudo.conf is owned by sudo 1.9.14.p1-1
    /etc/sudo_logsrvd.conf is owned by sudo 1.9.14.p1-1
    /etc/sudoers is owned by sudo 1.9.14.p1-1
    /etc/sudoers.d/ is owned by sudo 1.9.14.p1-1

Environment: Artix Linux.

Also, add missing paths sudo/doas to etc/ids.config and jailcheck.

See also commit dbebd71db ("disable-common.inc: blacklist doas binary",
2022-10-05).

Relates to #5385.

Reported-by: Dieter Plaetinck <dieter@plaetinck.be>

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -416,6 +416,7 @@ blacklist /tmp/ssh-*
 # top secret
 blacklist /.fscrypt
 blacklist /etc/davfs2/secrets
+blacklist /etc/doas.conf
 blacklist /etc/group+
 blacklist /etc/group-
 blacklist /etc/gshadow
@@ -428,6 +429,8 @@ blacklist /etc/shadow+
 blacklist /etc/shadow-
 blacklist /etc/ssh
 blacklist /etc/ssh/*
+blacklist /etc/sudo*.conf
+blacklist /etc/sudoers*
 blacklist /home/.ecryptfs
 blacklist /home/.fscrypt
 blacklist ${HOME}/*.kdb
--- a/Makefile
+++ b/Makefile
@@ -362,7 +362,7 @@ scan-build: clean
 
 .PHONY: codespell
 codespell: clean
-	codespell --ignore-regex "UE|creat|shotcut|ether" src test
+	codespell --ignore-regex "UE|creat|doas|shotcut|ether" src test
 
 .PHONY: print-env
 print-env:
--- a/etc/ids.config
+++ b/etc/ids.config
@@ -139,6 +139,7 @@ ${HOME}/.local/share/autostart
 /etc/security
 /etc/selinux
 /etc/shadow*
+/etc/sudo*.conf
 /etc/sudoers*
 /etc/tripwire
 ${HOME}/.config/firejail
--- a/src/jailcheck/main.c
+++ b/src/jailcheck/main.c
@@ -120,6 +120,7 @@ int main(int argc, char **argv) {
 	// basic sysfiles
 	sysfiles_setup("/etc/shadow");
 	sysfiles_setup("/etc/gshadow");
+	sysfiles_setup("/usr/bin/doas");
 	sysfiles_setup("/usr/bin/mount");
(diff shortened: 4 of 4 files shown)


glitsj16 (approved):
Nice job. LGTM.

etc/inc/disable-common.inc
```diff
@@ -394,18 +394,24 @@ blacklist /tmp/ssh-*
 # top secret
 blacklist /.fscrypt
 blacklist /etc/davfs2/secrets
+blacklist /etc/doas.conf
 blacklist /etc/group+
 blacklist /etc/group-
 blacklist /etc/gshadow
 blacklist /etc/gshadow+
 blacklist /etc/gshadow-
+blacklist /etc/pam.d
```
rusty-snake, line 403: Will break sshd and maybe some other

kmk3: @rusty-snake [on Oct 2](https://github.com/netblue30/firejail/pull/5387#discussion_r985139410):

> Will break sshd and maybe some other

There is no profile for sshd.

I don't know much about PAM, but wouldn't this only affect login managers,
system daemons (which usually don't have profiles) or things that run outside
of firejail anyway?

rusty-snake: Some distros/packages/plugins/patches have a pam.d/sshd AFAIK. But I need to check later what else is in there.

rusty-snake: Uh, we don't have a sshd profile. I thought we have one.

kmk3: @rusty-snake [on Oct 2](https://github.com/netblue30/firejail/pull/5387#discussion_r985197303):

> Some distros/packages/plugins/patches have a pam.d/sshd AFAIK. But I need to
> check later what else is in there.

To be clear, I meant that there is no firejail profile for sshd.

From pam(8):

> Linux-PAM is a system of libraries that handle the authentication tasks of
> applications (services) on the system. The library provides a stable general
> interface (Application Programming Interface - API) that privilege granting
> programs (such as login(1) and su(1)) defer to to perform standard
> authentication tasks.

Is there any unprivileged application that uses PAM directly?

I think only system package managers would need to write to /etc/pam.d, and
they are already privileged anyway (and there are no firejail profiles for them
AFAIK).

kmk3:
(Force-pushed to also add /etc/pam.conf)

etc/inc/disable-common.inc
glitsj16: We never had a profile for sshd, only for ssh and ssh-agent. IMO that's all we should do, the sshd daemon is better left protected via systemd.

rusty-snake: > To be clear, I meant that there is no firejail profile for sshd.

Yes, I thought we had one.

> Is there any unprivileged application that uses PAM directly?

```console
$ ls /etc/pam.d 
atd          fingerprint-auth@       gdm-password   passwd          ppp        smartcard-auth@   sudo          vlock
config-util  gdm-autologin           gdm-smartcard  password-auth@  remote     sshd              sudo-i        xserver
crond        gdm-fingerprint         login          polkit-1        runuser    sssd-shadowutils  su-l
cups         gdm-launch-environment  other          postlogin@      runuser-l  su                system-auth@
```

Problems that I see are with awesome, blackbox, fluxbox, openbox, sway, tmux, ... since they can contain a (terminal)-session using su(do) or a screen-locker.

netblue30:
I would just merge it as is and get some testing going, anything else could go in a separate.

kmk3:
(Rebased to master and removed the pam-related paths)

PR #5398, opened 2022-10-03
docs: Add gist note to bug_report.md

rusty-snake:
We had some comments saying that the output exceeds the allowed length of a github comment (65534 characters IIRC).

This commit adds a comment to instruct users to create a secret gist in that case.


kmk3 (approved):
LGTM.

PR #5399, opened 2022-10-04
mpv: whitelist mpv-mpris in lib64

WhyNotHugo:
Might be a thing in some distros?

See: https://github.com/netblue30/firejail/pull/5386#issuecomment-1264478380


kmk3 (changes requested):
See #5400.

PR #5400, opened 2022-10-04
Revert "mpv: whitelist mpv-mpris (#5386)"

kmk3:
This reverts commit 393c5beff2686d7732221dadb6730917f24835a0.

Which broke mpv:

    $ mpv --version

Probably because mpv itself uses many libraries and it has plugins that
may depend on files in /usr/lib as well:

    $ pacman -Qlq mpv | grep /lib/ | grep -v '/$'
    /usr/lib/libmpv.so
    /usr/lib/libmpv.so.1
    /usr/lib/libmpv.so.1.109.0
    /usr/lib/pkgconfig/mpv.pc
    $ strings /usr/bin/mpv | grep '^lib.*\.so' | sort -u | wc -l
    53
    $ pacman -Qlq yt-dlp | grep /lib/ | grep -v '/$' |
      cut -f -4 -d / | sort -u
    /usr/lib/python3.10
    $ pacman -Q mpv yt-dlp
    mpv 1:0.34.1-5
    yt-dlp 2022.09.01-1

Environment: Artix Linux.

Also, private-lib is disabled by default in firejail.config (see #5190)
and mpv.profile does not use private-lib, so there should be no need to
whitelist anything in /usr/lib in the default profile.

Cc: @glitsj16 @rusty-snake @WhyNotHugo (from #5386)

### DIFF
--- a/etc/profile-m-z/mpv.profile
+++ b/etc/profile-m-z/mpv.profile
@@ -58,7 +58,6 @@ whitelist ${HOME}/.config/yt-dlp.conf
 whitelist ${HOME}/.netrc
 whitelist ${HOME}/yt-dlp.conf
 whitelist ${HOME}/yt-dlp.conf.txt
-whitelist /usr/lib/mpv-mpris
 whitelist /usr/share/lua
 whitelist /usr/share/lua*
 whitelist /usr/share/vulkan


glitsj16:
@kmk3 Oops, nice catch!

FWIW, I installed [mpv-mpris](https://archlinux.org/packages/community/x86_64/mpv-mpris/) and got it to work by adding

```
dbus-user filter
dbus-user.talk org.mpris.MediaPlayer2.*
ignore dbus-user none

```

Probably needs a separate PR if we want to support mpv-mpris.

glitsj16 (approved):
LGTM

WhyNotHugo (approved):
Yeah, I'm not sure why I needed this initially, but LGTM.

WhyNotHugo:
@glitsj16 That snippet did not work for me; things like `playerctl` still don´t pick up on mpv.

WhyNotHugo:
Yeah, I messed up my testing; the same patch that was installing the change for me was also disabling `mpv.profile` :facepalm:

glitsj16:
@WhyNotHugo Not sure what's going on at your end, I just took a lead from https://github.com/hoyon/mpv-mpris#d-bus-interfaces to get the exposed D-Bus path and added the snippet to my mpv.local. The mpv profile does have `private-bin` so you might need to add `playerctl` to it if that's what you use. Another incantation of the above might work for you:

```
dbus-user filter
dbus-user.own org.mpris.MediaPlayer2.*
ignore dbus-user none

```

WhyNotHugo:
Indeed, `dbus-user.own` is correct :+1:

netblue30:
fixed, thanks!

PR #5402, opened 2022-10-05
docs: Make appimage examples consistent with --appimage option short description

slowpeek:
`--appimage` option is mentioned twice in the man page: first with a short description, next with usage examples. The short description is:

```
firejail [OPTIONS] --appimage [appimage-file and arguments]
```

While the examples are of form `firejail --appimage <some extra options> someapp.appimage`.

Even though the option parser recognizes such order, it is better to keep it the same way in both places in the man page.


netblue30:
Merged, thanks for the fix.

Issue #5414, opened 2022-10-11
ktorrent: Cannot start application: No such file or directory


vendion:
### Description

Ktorrent 22.08.1 doesn't launch with the default firejail profile, instead a fatal python error is thrown.

### Steps to Reproduce

_Steps to reproduce the behavior_

1. Launch ktorrent via `LC_ALL=C firejail /usr/bin/ktorrent` (gave full path as I have firejail setup so apps are ran under it by default)
2. Ktorrent doesn't launch with lots of output to STDOUT/STDERR in the terminal

### Expected behavior

Ktorrent to open up.

### Actual behavior

Ktorrent errored out before even opening the main window.

### Behavior without a profile

Without a profile Ktorrent opens up just fine.

### Environment

- Distro: Arch Linux
- Firejail version: 0.9.70

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ ] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

<details>
<summary>Output of <code>LC_ALL=C firejail /path/to/program</code></summary>
<p>

```
Reading profile /etc/firejail/ktorrent.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-exec.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/disable-shell.inc
Warning: cannot create /home/vendion/.kde/share/config/ktorrentrc
Reading profile /etc/firejail/whitelist-common.inc
Reading profile /etc/firejail/whitelist-run-common.inc
Reading profile /etc/firejail/whitelist-var-common.inc
Parent pid 6660, child pid 6662
1 program installed in 4.39 ms
[...]

PR #5415, opened 2022-10-11
ktorrent.profile: fix mkfile without mkdir & comment legacy paths

kmk3:
firejail fails to create the following files:

* ~/.kde/share/config/ktorrentrc
* ~/.kde4/share/config/ktorrentrc

Because it does not create the preceding directories beforehand:

* ~/.kde/share/config
* ~/.kde4/share/config

Relates to #5414.


kmk3:
Force-push changes:

* Put the "fix" commit first
* Also consider ~/.kde paths as legacy paths
* Add missing private-bin paths

netblue30:
merged!

PR #5429, opened 2022-10-18
build: sort.py improvements

kmk3:
```console
* sort.py: enumerate lines from 1
* sort.py: pass the str.casefold function directly
* sort.py: simplify "checking" message code
* sort.py: rename raw_items to original_items
* sort.py: rename protocols to original_protocols
* sort.py: line-wrap and improve comments
* sort.py: print errors to stderr
* sort.py: print more standard error messages
```


contrib/sort.py
```diff
@@ -16,34 +20,38 @@
...
     $ sudo ./sort.py /etc/firejail/*.{profile,inc,local}
 
-Exit-Codes:
-  0: No Error; No Profile Fixed.
-  1: Error, one or more profiles were not processed correctly.
-  101: No Error; One or more profile were fixed.
+Exit Codes:
+  0: Success: No profiles needed fixing.
+  1: Error: One or more profiles could not be processed correctly.
+  101: Warning: One or more profile were fixed.
```
rusty-snake, line 25: I wouldn't say that this is a warning. More like a `Success,Information:`.

rusty-snake (commented):
LGTM

---

> PEP-8

Actually PEP-8 suggests 72 characters for docstrings[^1]. And PEP 257 has some recommendations for the first line of multi-line docstrings[^2].

Nothing to act on, just FTR.

[^1]: https://peps.python.org/pep-0008/#maximum-line-length
[^2]: https://peps.python.org/pep-0257/#multi-line-docstrings

contrib/sort.py
kmk3: @rusty-snake [on Oct 18](https://github.com/netblue30/firejail/pull/5429#discussion_r998576284):

> I wouldn't say that this is a warning. More like a `Success,Information:`.

Yeah, that makes more sense; changed to `Info:`.

rusty-snake (approved):

kmk3:
@rusty-snake reviewed [on Oct 18](https://github.com/netblue30/firejail/pull/5429#pullrequestreview-1146322308):

> > PEP-8
> 
> Actually PEP-8 suggests 72 characters for docstrings[1](#user-content-fn-1-dc8012c072a90287a85931699209850f).
> 
> And PEP 257 has some recommendations for the first line of multi-line docstrings[2](#user-content-fn-2-dc8012c072a90287a85931699209850f).
> 
> Nothing to act on, just FTR.
> 
> 1. https://peps.python.org/pep-0008/#maximum-line-length [leftwards_arrow_with_hook](#user-content-fnref-1-dc8012c072a90287a85931699209850f)
> 2. https://peps.python.org/pep-0257/#multi-line-docstrings [leftwards_arrow_with_hook](#user-content-fnref-2-dc8012c072a90287a85931699209850f)

Indeed; removed the reference to PEP-8.

I just wanted to make it fit in 79 or 80 columns.

Also, what do you think about making it print `__doc__` if no argument is
given?

And using `argv[0]` instead of `./sort.py` in the docstring?

Misc: Interesting, I didn't know about this footnotes functionality.

kmk3:
@rusty-snake reviewed [on Oct 18](https://github.com/netblue30/firejail/pull/5429#pullrequestreview-1146322308):

> And PEP 257 has some recommendations for the first line of multi-line docstrings[2](#user-content-fn-2-dc8012c072a90287a85931699209850f).
> 
> 2. https://peps.python.org/pep-0257/#multi-line-docstrings [leftwards_arrow_with_hook](#user-content-fnref-2-dc8012c072a90287a85931699209850f)

I saw docstrings using both forms in the docs (and in SO answers); I'm still
not sure when the first line should be blank or not.

Should it be not blank in the main docstring and blank in functions?

I think that it indeed makes sense for at least the main one to be non-blank,
to avoid a blank line in the usage output.

Though it doesn't look as good in the source and potentially makes the first
line a bit shorter than the other ones when printing, so if it's intended to be
used unprocessed as part of the output, I think that it would make sense to
leave it blank and escape the newline.

@rusty-snake commented [on Oct 18](https://github.com/netblue30/firejail/pull/5429#issuecomment-1282873131):

> > Also, what do you think about making it print **doc** if no argument is
> > given?
> 
> Sounds good.
> 
> It should also do this if only `--help` is given.

Since it currently does not have any options, IMO only printing it without
arguments keeps it nice and simple (as then all arguments mean the same thing
and there is no need to add an "Options" section).  But if you really want it I
can add it.

> > And using argv[0] instead of ./sort.py in the docstring?
> 
> Are there some rules about using f-strings for docstrings?

I don't know, but it works for me at least.  The main difference is that the
imports have to appear before before the docstring in this case (for
`sys.argv`).

> `argv[0]` tells you only the path to the file, be it relative or absolute, or
> the filename without path (some corner cases ignored). It does not tell you
> if the file is executable or if it was used with `python3 argv[0]`.

It always prints the filename used for running the script, even if it was
called indirectly.  Using `print(argv[0]); return`):

```console
$ ./contrib/sort.py
./contrib/sort.py
$ (cd contrib && ./sort.py)
./sort.py
$ python contrib/sort.py
contrib/sort.py
```

The same applies to shell scripts (and likely also to other scripting
languages).  Using `echo "$0"; return`:

```console
$ ./contrib/sort.sh
./contrib/sort.sh
$ (cd contrib && ./sort.sh)
./sort.sh
$ bash contrib/sort.sh
contrib/sort.sh
```

For example, the shell script `usage` equivalent would be something like this:

<details>
<summary>sort.sh</summary>
<p>

```sh
#!/bin/sh

usage() {
	cat <<EOF
Sort the arguments of the commands that support multiple arguments in profiles;
the following commands are supported:

    private-bin, private-etc, private-lib, caps.drop, caps.keep, seccomp.drop,
    seccomp.drop, protocol

Usage: $(basename "$0") [/path/to/profile ...]

Keep in mind that this will overwrite your profile(s).

Examples:
    $ $0 MyAwesomeProfile.profile
    $ $0 new_profile.profile second_new_profile.profile
    $ $0 ~/.config/firejail/*.{profile,inc,local}
    $ sudo $0 /etc/firejail/*.{profile,inc,local}

Exit Codes:
  0: Success: No profiles needed fixing.
  1: Error: One or more profiles could not be processed correctly.
  101: Info: One or more profile were fixed.
EOF
}

usage
```

</p>
</details>

Output:

<details>
<summary>output of sort.sh</summary>
<p>

```console
$ ./contrib/sort.sh
Sort the arguments of the commands that support multiple arguments in profiles;
the following commands are supported:

    private-bin, private-etc, private-lib, caps.drop, caps.keep, seccomp.drop,
    seccomp.drop, protocol

Usage: sort.sh [/path/to/profile ...]

Keep in mind that this will overwrite your profile(s).

Examples:
    $ ./contrib/sort.sh MyAwesomeProfile.profile
    $ ./contrib/sort.sh new_profile.profile second_new_profile.profile
    $ ./contrib/sort.sh ~/.config/firejail/*.{profile,inc,local}
    $ sudo ./contrib/sort.sh /etc/firejail/*.{profile,inc,local}

Exit Codes:
  0: Success: No profiles needed fixing.
  1: Error: One or more profiles could not be processed correctly.
  101: Info: One or more profile were fixed.
```

</p>
</details>

Anyway, the main advantage of doing this is that the help section remains
consistent even if the filename is changed.  For example, if someone renames
`sort.py` to `firejail-sort.py` and puts it somewhere in `$PATH`.

Since it's in a standalone script (and not a module in a larger system, for
example), I think that this makes sense in this case.

Anyway, I force-pushed with both changes; let me know what you think.

Not sure if it's a very "Pythonic" solution, but I can't think of any other way
of accomplishing this.

kmk3:
(Force-push)

Tried to make the main docstring a bit clearer; I think that it looks better
now when looking at it with pydoc:

```sh
pydoc contrib/sort.py
```

Also, added more details to some commit messages.

contrib/sort.py
```diff
@@ -84,22 +96,30 @@ def fix_profile(filename):
 
 
 def main(args):
+    if len(args) < 1:
+        print(__doc__, file=stderr)
+        return
```
rusty-snake, line 101: Add an exit code.

kmk3: @rusty-snake [on Oct 19](https://github.com/netblue30/firejail/pull/5429#discussion_r999637842):

> Add an exit code.

Done.

(Sorry for the delay, I'm having some infrastructure issues)

rusty-snake (approved):

Fred-Barclay (approved):
Looks good to me - I say merge it in @kmk3 !

glitsj16 (approved):
LGTM. Thanks for the reminder in https://github.com/netblue30/firejail/discussions/5479.

PR #5446, opened 2022-11-01
firejail/checkcfg: fix xephyr-extra-params size

mochaaP:
`xephyr-extra-params` is always invalid when parsing due to a size error.

### DIFF
--- a/src/firejail/checkcfg.c
+++ b/src/firejail/checkcfg.c
@@ -168,10 +168,10 @@ int checkcfg(int val) {
 			}
 
 			// Xephyr command extra parameters
-			else if (strncmp(ptr, "xephyr-extra-params ", 20) == 0) {
+			else if (strncmp(ptr, "xephyr-extra-params ", 21) == 0) {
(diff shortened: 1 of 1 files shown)


mochaaP:
cc @netblue30 can you review this? :)

Thanks.

rusty-snake (changes requested):
If your claim ~is~ would be¹ true, the following options are also invalid because they have the same `strncmp`+`strdup` pattern:

- `xvfb-extra-params`
- `xpra-extra-params`
- `seccomp-error-action`
- `whitelist-disable-topdir`

¹I don't think that we have multiple broken option for years without reports that.

mochaaP:
Hmm weird. `xephyr-extra-params` is in <https://github.com/netblue30/firejail/blob/master/src/firejail/checkcfg.c>, but not in <https://github.com/netblue30/firejail/blob/master/src/firejail/profile.c>. Is that excepted?

rusty-snake:
Yes.

profile.c parses profiles and
checkcfg.c parses firejail.config (firejails global config file).

PR #5451, opened 2022-11-04
docs: clarify that --appimage should appear before --profile

kmk3:
And fix the argument order in the examples to reflect that.

Background: The order in which these options appeared in the
documentation was inconsistent.  src/man/firejail.txt used --appimage
before --profile and src/man/firejail-profile.txt used --profile before
--appimage.  Then commit 44fefcac0 ("Make appimage examples consistent
with --appimage option short description", 2022-10-05) / PR #5402 was
made, which standardized on --profile before --appimage in both places.

But as mentioned by @rusty-snake[1], --appimage has be specified before
--profile in order for any `?HAS_APPIMAGE` conditionals inside of the
profile to evaluate to true.

So change the documentation to use and recommend the latter form.

Also, add --quiet to one example to make it clear that --appimage does
not have to be the first option (nor the last option before --profile).

[1] https://github.com/netblue30/firejail/pull/5402#issuecomment-1274889618

Cc: @slowpeek (from #5402)

### DIFF
--- a/src/man/firejail-profile.txt
+++ b/src/man/firejail-profile.txt
@@ -14,7 +14,7 @@ Using a specific profile:
 .br
 Example:
 .br
-$ firejail --profile=/etc/firejail/kdenlive.profile --appimage kdenlive.appimage
+$ firejail --appimage --profile=/etc/firejail/kdenlive.profile kdenlive.appimage
 .br
 
 .br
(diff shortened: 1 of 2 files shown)


rusty-snake (approved):

kmk3:
@rusty-snake Thanks for the review.

Misc: This PR kind of supersedes #5402, so I thought about adding #5402 only as
a PR reference in the RELNOTES to avoid confusion.  I'll go ahead and merge
this before I forget to do that.

Issue #5447, opened 2022-11-02
firefox: cannot communicate with keepassxc


neuroretransmit:
### Description

KeePassXC-Browser fails to communicate with KeePassXC (2.7.1-1, 2.7.4-1) using Firefox 106.0.3-1. Firefox 106.0.2-1 was working just fine.

Versions of KeePassXC tested : 2.7.1-1, 2.7.4-1

### Steps to Reproduce

Running either of the versions of KeePassXC listed above, run Firefox 106.0.3-1 (I'm on Arch Linux, I'd assume behavior is the same elsewhere).

1. Run in bash `LC_ALL=C firejail keepassxc`
2. Run in bash `LC_ALL=C firejail firefox`
3. Click on KeePassXC browser, then 'Reload' to receive a "Key exchange failed" message (this can also be done through KeePassXC-Browser's Settings->Connect - where nothing will happen). Debugging the plugin shows communication failures as well.
4. To show it is Firefox that is the problem, run in bash `LC_ALL=C firejail --noprofile firefox` after closing the previous instance - communication will succeed.

### Expected behavior

Successful key exchange/native-messaging-hosts transmission via keepassxc-proxy

### Actual behavior

Key exchange failure/no transmission of username/password.

### Behavior without a profile

KeePassXC is fine to run with a profile, Firefox is not. Using `--noprofile` on Firefox allows the communication from KeePassXC to KeePassXC-Browser

### Additional context

_Any other detail that may help to understand/debug the problem_

### Environment

- Arch Linux (6.0.6-hardened)
- Firejail version: 0.9.70

All KeePassXC-Browser relevant options enabled in `firefox.profile` (+ private-etc), `firefox-common.profile` (for private-etc), `firefox-common-addons.profile`, `keepassxc.profile`,

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
- [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.

### Log

<details>
[...]

PR #5452, opened 2022-11-05
Allow KeePassXC access to its socket path

WhyNotHugo:
KeePassXC now exposes its browser socket in a dedicated directory, so that this directory can be mounted inside the same sandbox as browser and browsers can talk to it.

This allows using the browser extension when KeePassXC and/or the browser are sandboxed.

See: https://github.com/keepassxreboot/keepassxc/issues/8018
Fixes: https://github.com/netblue30/firejail/issues/5447

### DIFF
--- a/etc/profile-a-l/keepassxc.profile
+++ b/etc/profile-a-l/keepassxc.profile
@@ -13,6 +13,7 @@ noblacklist ${HOME}/.config/keepassxc
 noblacklist ${HOME}/.config/KeePassXCrc
 noblacklist ${HOME}/.keepassxc
 noblacklist ${DOCUMENTS}
+noblacklist ${RUNUSER}/app
 
 # Allow browser profiles, required for browser integration.
 noblacklist ${HOME}/.config/BraveSoftware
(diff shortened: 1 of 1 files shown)


rusty-snake (changes requested):
This will break X11, D-Bus, ...

WhyNotHugo:
> This will break X11, D-Bus, ...

Can you finish the sentence? How? I'm not using X11, so can't really try that.

WhyNotHugo:
This only adds a new `whitelist` entry on top of the existing one. That is, a new directory is now exposed. I can't imagine how X11 would break because of that.

rusty-snake:
Because the Xauthority location of gdm, sddm, XWayland (mutter), ... are now hidden.

WhyNotHugo:
Right, as discussed in https://github.com/netblue30/firejail/pull/5453, this is not the right approach, but the original issue remains valid.

WhyNotHugo:
Oh, see https://github.com/netblue30/firejail/issues/5447

rusty-snake:
Actually the `noblacklist ${RUNUSRR]/app` should be done. We should also consider if it makes sense to blacklist app id programs start to require it.

WhyNotHugo:
Indeed, but https://github.com/netblue30/firejail/issues/5447 tracks this.

PR #5454, opened 2022-11-05
Grant KeePassXC access to the Wayland socket

WhyNotHugo:
KeePassXC is a graphical application, and won't start without this set.

### DIFF
--- a/etc/profile-a-l/keepassxc.profile
+++ b/etc/profile-a-l/keepassxc.profile
@@ -62,6 +62,7 @@ include disable-xdg.inc
 #whitelist ${HOME}/.config/KeePassXCrc
 #include whitelist-common.inc
 
+whitelist ${RUNUSER}/wayland-*
 whitelist /usr/share/keepassxc
 include whitelist-run-common.inc
 include whitelist-usr-share-common.inc


etc/profile-a-l/keepassxc.profile
```diff
@@ -62,6 +62,7 @@ include disable-xdg.inc
 #whitelist ${HOME}/.config/KeePassXCrc
 #include whitelist-common.inc
 
+whitelist ${RUNUSER}/wayland-*
```
WhyNotHugo, line 65: Actually this should be `whitelist ${RUNUSER}/$WAYLAND_DISPLAY`. Are environment variables usable here?

rusty-snake (changes requested):
This will break X11, D-Bus, ...

etc/profile-a-l/keepassxc.profile
rusty-snake: `$WAYLAND_DISPLAY` can be `/tmp/tomatoes/landway`.

WhyNotHugo: > To find the Unix socket to connect to, most implementations just do what libwayland does:
> 
>     If WAYLAND_SOCKET is set, interpret it as a file descriptor number on which the connection is already established, assuming that the parent process configured the connection for us.
>     If WAYLAND_DISPLAY is set, concat with XDG_RUNTIME_DIR to form the path to the Unix socket.
>     Assume the socket name is wayland-0 and concat with XDG_RUNTIME_DIR to form the path to the Unix socket.
>     Give up.

https://wayland-book.com/protocol-design/wire-protocol.html

I don't think it can be an absolute path.

WhyNotHugo: If you have suggestions on how to support _any_ value of `WAYLAND_DISPLAY`, I can update the PR, but this works for the default configuration on any compositor.

WhyNotHugo:
> This will break X11, D-Bus, ...

Why? How?

etc/profile-a-l/keepassxc.profile
rusty-snake: > If you have suggestions on how to support any value of WAYLAND_DISPLAY, I can update the PR, but this works for the default configuration on any compositor.

You would have to add a `${WAYLAND_DISPLAY}` macro.

---

https://wayland.freedesktop.org/docs/html/ch04.html#sect-Protocol-Wire-Format

> Beginning in Wayland 1.15, implementations can optionally support server socket endpoints located at arbitrary locations in the filesystem by setting WAYLAND_DISPLAY to the absolute path at which the server endpoint listens.

mara004:
I can confirm the submitter's report that KeePassXC fails to start with default firejail configuration.

rusty-snake:
If it breaks due to something other than Wayland, open a new bug report. If it breaks due to Wayland either you have a very strange setup or you modified the profile-chain to cause this.

mara004:
Using distribution-provided Firejail on Fedora 37:

```bash
$ firejail /usr/bin/keepassxc
Reading profile /etc/firejail/keepassxc.profile
Reading profile /etc/firejail/disable-shell.inc
```

-------

This is on Wayland. In the lights of what @WhyNotHugo reports, I'm confident it is the same issue.
My setup is neither strange nor modified in any way, this is just default Fedora.

rusty-snake:
FTR: Fedora 37 has 0.9.70 ATOW.

>  In the lights of what @WhyNotHugo reports

The fix in this PR can not fix a issue caused by the upstream keepassxc.profile (or I do not see the obvious).

> I'm confident it is the same issue.

So did you tried the fix? Did it worked?

etc/profile-a-l/keepassxc.profile
```diff
@@ -62,6 +62,7 @@ include disable-xdg.inc
...
 whitelist /usr/share/keepassxc
 include whitelist-run-common.inc
```
kmk3, line 67:
```suggestion
whitelist /usr/share/keepassxc
include whitelist-run-common.inc
include whitelist-runuser-common.inc
```

Does this fix the issue?

mara004:
> The fix in this PR can not fix a issue caused by the upstream keepassxc.profile (or I do not see the obvious).

It is highly unlikely that Fedora have patched the profile. They're just pulled from here, aren't they?

etc/profile-a-l/keepassxc.profile
mara004: Yes it does. Thanks!

mara004: Is there some way of putting this in a local config file? If I edit `/etc/firejail/keepassxc.profile` directly, the change may be overwritten on next update...

rusty-snake: https://github.com/netblue30/firejail/wiki/Creating-overrides

rusty-snake: > > Does this fix the issue?
>
> Yes it does. Thanks!

It still does not make sense because `keepassxc.profile` does not have any whitelisting for RUNUSER and should therefore default to blacklisting only. This can fix it only through a side-effect and if this is the case we should fix the real cause.

Also DO NOT MERGE any PR enabling whitelisting in RUNUSER for kpxc without careful consideration.

kmk3: @rusty-snake [on Dec 30](https://github.com/netblue30/firejail/pull/5454#discussion_r1059512589):

> > > Does this fix the issue?
> > 
> > Yes it does. Thanks!
> 
> It still does not make sense because `keepassxc.profile` does not have any
> whitelisting for RUNUSER and should therefore default to blacklisting only.
> This can fix it only through a side-effect and if this is the case we should
> fix the real cause.

Yeah, looking at it again, this issue shouldn't happen with the default
profiles.

From my testing with firejail-git, I cannot reproduce this; all paths in the
root of `${RUNUSER}` are present (even if some are blacklisted by default) with
the default keepassxc.profile:

```console
$ find "/run/user/$(id -u)" -maxdepth 1 | wc -l
10
$ firejail --quiet --ignore=private-bin \
  --ignore='include keepassxc.local' --ignore='include globals.local' \
  --profile=keepassxc \
  find "/run/user/$(id -u)" -maxdepth 1 | wc -l
10
```

Maybe firejail 0.9.70 doesn't automatically mount `${RUNUSER}` when
whitelisting /run.

Anyway, I posted the above suggestion because whitelist-runuser-common.inc
contains the following line:

```
whitelist ${RUNUSER}/wayland-?
```

Which is similar to the line that the PR adds:

```
whitelist ${RUNUSER}/wayland-*
```

> Also DO NOT MERGE any PR enabling whitelisting in RUNUSER for kpxc without
> careful consideration.

I'm personally in no hurry to merge this, especially before the root cause is
found.

From my testing, I see that including whitelist-runuser-common.inc makes less
paths visible in `${RUNUSER}`.

To clarify, are you thinking of any specific program/integration issue that
this could cause?

kmk3:
@mara004

Can you reproduce it with the current version of firejail-git (from master)?

Is the wayland socket present in `${RUNUSER}` with that version?  Example
commands to check:

```console
$ find "/run/user/$(id -u)" -maxdepth 1 | wc -l
10
$ firejail --quiet --ignore=private-bin \
  --ignore='include keepassxc.local' --ignore='include globals.local' \
  --profile=keepassxc \
  find "/run/user/$(id -u)" -maxdepth 1 | wc -l
10
```

etc/profile-a-l/keepassxc.profile
rusty-snake: > To clarify, are you thinking of any specific program/integration issue that
this could cause?

- Browser Integration
- ssh-agent
- ...

KPXC has support to provided secrets via a lot of ways. The most use sockets in RUNUSER. Some sockets depend on the distro or other conditions. While not all of them work out-of-the-box with the current profile, we should not make the situation worse than it is.

kmk3:
Note: The following PR includes whitelist-runuser-common.inc:

* #6391

So it should have access to the wayland socket now.

Marking this as a duplicate of #6391.

PR #5457, opened 2022-11-06
firefox-common.profile: allow semtimedop syscall

kmk3:
To prevent the audit logs from being flooded with this:

    audit: type=1326 [...] exe="/usr/lib/librewolf/librewolf" sig=0 arch=c000003e syscall=220 compat=0 [...] code=0x50000

### DIFF
--- a/etc/profile-a-l/firefox-common.profile
+++ b/etc/profile-a-l/firefox-common.profile
@@ -47,7 +47,7 @@ notv
 ?BROWSER_DISABLE_U2F: nou2f
 protocol unix,inet,inet6,netlink
 # The below seccomp configuration still permits chroot syscall. See https://github.com/netblue30/firejail/issues/2506 for possible workarounds.
-seccomp !chroot
+seccomp !chroot,!semtimedop
 # Disable tracelog, it breaks or causes major issues with many firefox based browsers, see https://github.com/netblue30/firejail/issues/1930.
 #tracelog
 


rusty-snake:
`semtimedop` is in group `@ipc` and not included in `@default`.
From where does the blocking come?

kmk3:
@rusty-snake commented [on Nov 6](https://github.com/netblue30/firejail/pull/5457#issuecomment-1304795225):

> `semtimedop` is in group `@ipc` and not included in `@default`.
> 
> From where does the blocking come?

What do you mean?

All I know is that after doing this change, the flooding seems to have stopped.

rusty-snake:
`semtimedop` shouldn't be blocked by this profile.

Do you have anything in your overrides?

kmk3:
@rusty-snake commented [on Nov 6](https://github.com/netblue30/firejail/pull/5457#issuecomment-1304796815):

> `semtimedop` shouldn't be blocked by this profile.
> 
> Do you have anything in your overrides? What does `--debug` show?

I see now that I have `ipc-namespace` in an include (it has probably been there
for a while, as I don't remember when I added it).

Could that be it?

glitsj16:
Not seeing these semtimedop entrees in audit logs with Firefox, which I have been running with ipc-namespace like forever. If it does affects librewolf, it makes more sense to add it to the librewolf profiles only, until we see it in firefox too IMO.

kmk3:
Thanks for the suggestions.

Web browsers are kind of annoying to debug, so I'd rather just close this and
keep the local override for now.

If I manage to reproduce it later without local modifications, I'll reopen.

PR #5459, opened 2022-11-07
Add python3 support/fixes to nicotine

glitsj16:
Add Python 3 support. Also fixes private-bin (reported on IRC).

### DIFF
--- a/etc/profile-m-z/nicotine.profile
+++ b/etc/profile-m-z/nicotine.profile
@@ -52,7 +52,7 @@ seccomp
 tracelog
 
 disable-mnt
-#private-bin bash,nicotine,python2*,python3*,sh
+#private-bin nicotine,python2*
 private-cache
 private-dev
 private-tmp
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/nicotine.profile
```diff
@@ -47,7 +52,7 @@ seccomp
 tracelog
 
 disable-mnt
-private-bin nicotine,python2*
+#private-bin bash,nicotine,python2*,python3*,sh
```
rusty-snake, line 55:
```suggestion
#private-bin bash,nicotine,python2*,sh
```

glitsj16: @rusty-snake During the help attempt on IRC I at one time suggested to try adding python3* to private-bin, besides including its allow file. Later on the user in question confirmed that private-bin kept breaking nicotine so I decided to comment it. Thanks for the suggestion, I'll take it out again.

kmk3 (approved):

Fred-Barclay (approved):

Issue #5462, opened 2022-11-09
Regression: custom apparmor profile support resulted in broken Firefox


KOLANICH:
### Description
7f3b6c19a0a87bfd240af7c0c9d61ae907668ce6 (#5274) has resulted in firefox being broken on Kubuntu 21.10 (impish).

`apparmor-profiles` and `apparmor-profiles-extra` are installed.

```bash
git revert 7f3b6c19a0a87bfd240af7c0c9d61ae907668ce6
```

and resolution of the conflict mitigate the issue.

### Steps to Reproduce

_Steps to reproduce the behavior_

1. build `firejail` from the latest git with apparmor support, install it
2. run `firefox` and see it crashed
3. revert the commit, build, install
4. enjoy working Firefox

### Expected behavior

* firefox working
* no regressions, Firefox as an essential software should be tested before merging PRs that have a chance to break it.

### Actual behavior

* the commit breaking Firefox has landed into `master`

### Behavior without a profile

_What changed calling `LC_ALL=C firejail --noprofile /path/to/program` in a terminal?_

```diff
--- nopr.log	2022-11-09 00:00:00.000000000 +0300
+++ crash.log	2022-11-09 00:00:00.000000000 +0300
@@ -1,4 +1,8 @@
+Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
 Parent pid AAAAA, child pid BBBBB
+Warning: An abstract unix socket for session D-BUS might still be available. Use --net or remove unix from --protocol set.
+Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
+Warning: cleaning all supplementary groups
 Child process initialized in nn.mm ms
 ATTENTION: default value of option mesa_glthread overridden by environment.
 libEGL warning: MESA-LOADER: failed to retrieve device information
@@ -8,9 +12,14 @@
 libGL error: MESA-LOADER: failed to open amdgpu: /usr/lib/dri/amdgpu_dri.so: cannot open shared object file: No such file or directory (search paths /usr/lib/x86_64-linux-gnu/dri:\$${ORIGIN}/dri:/usr/lib/dri, suffix _dri)
 libGL error: failed to load driver: amdgpu
 ATTENTION: default value of option mesa_glthread overridden by environment.
[...]

PR #5475, opened 2022-11-15
A temporary fix to the bug caused by apparmor profiles stacking.

KOLANICH:
A temporary workaround to #5462


netblue30:
Merged, thanks!

PR #5476, opened 2022-11-20
fix: PyCharm profiles

k4leg:


etc/profile-m-z/pycharm-community.profile
```diff
@@ -5,7 +5,13 @@ include pycharm-community.local
 # Persistent global definitions
 include globals.local
 
-noblacklist ${HOME}/.PyCharmCE*
+noblacklist ${HOME}/.PyCharm*
+# Persistent cache is needed for spell and grammar checkers, etc.
+noblacklist ${HOME}/.cache/JetBrains/PyCharm*
+noblacklist ${HOME}/.config/JetBrains/PyCharm*
```
rusty-snake, line 11: Never blacklisted.

rusty-snake (approved):

kmk3 (approved):

etc/profile-m-z/pycharm-community.profile
```diff
@@ -30,7 +36,6 @@ tracelog
 
 # private-etc alternatives,fonts,passwd - minimal required to run but will probably break
 # program!
-private-cache
 private-dev
```
Fred-Barclay, line 39: Just for the record - why did `private-cache` need removing?

Fred-Barclay (changes requested):

etc/profile-m-z/pycharm-community.profile
k4leg: Because PyCharm needs persistent cache for spell and grammar checkers, etc (i. e. if you download spelling dictionary with `private-cache`, on restart PyCharm you need to download spelling dictionary again).

kmk3: @k4leg [on Nov 23](https://github.com/netblue30/firejail/pull/5476#discussion_r1030686066):

> Because PyCharm needs persistent cache for spell and grammar checkers, etc
> (i. e. if you download spelling dictionary with `private-cache`, on restart
> PyCharm you need to download spelling dictionary again).

I see, that makes sense.

Just make sure to put these details in the commit message, so that the reason
for the change is clear.  If is unclear, eventually somebody might assume that
it was done by accident and undo it, for example.

kmk3:
@k4leg [on Nov 23](https://github.com/netblue30/firejail/pull/5476#discussion_r1030686066):

> Because PyCharm needs persistent cache for spell and grammar checkers, etc
> (i. e. if you download spelling dictionary with `private-cache`, on restart
> PyCharm you need to download spelling dictionary again).

Note that `$XDG_CACHE_HOME` (default: `~/.cache`) is generally intended to be
for more ephemeral things (like for caching website images) and may be mounted
in RAM (and so might not persist between reboots).

While spelling dictionaries are program data and so indeed users probably
expect them to be persistent between runs.  In which case I think that
`$XDG_STATE_HOME` (default: `~/.local/state`) would be more appropriate.

Please consider asking the developers to use the latter directory for this (and
posting the link to the request here if you do).

Details:

* xdg-basedir spec: <https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html>
* More docs: <https://wiki.archlinux.org/title/XDG_Base_Directory>

rusty-snake:
> $XDG_STATE_HOME

Spelling directories are data and not state therefore `$XDG_DATA_HOME` would be right.

Fred-Barclay (approved):

netblue30:
All set, thanks!

PR #5478, opened 2022-11-21
build: deduplicate makefiles

kmk3:
Main commits:

* makefiles: include config.mk directly
* makefiles: equalize object dependencies in program targets
* makefiles: rename H_FILE_LIST and C_FILE_LIST
* makefiles: expand HDRS, SRCS and OBJS immediately
* makefiles: move extra deps into new MOD vars
* makefiles: deduplicate main target name into new PROG var
* makefiles: deduplicate many makefiles into common.mk
* makefiles: deduplicate main target name into new SO var
* makefiles: deduplicate lib makefiles into so.mk
* makefiles: rename common.mk to prog.mk
* makefiles: organize CFLAGS

This is kind of a continuation of #5140 and #5219.


kmk3:
@reinerh commented on [Nov 2, 2021](https://github.com/netblue30/firejail/issues/4642#issuecomment-958007267):

> We currently also have a lot of redundancy in our `Makefile.in` files (and
> some are not even in sync with others).

This PR should help with this.

Note: Some other planned improvements will probably depend on adding some
duplication back, though it should not be nearly as much as in the current
state (that is, without this PR).

Note2: There is more that can be done with regards to deduplication, but this
PR is mostly focused on mechanical changes and on refactoring the code only.

kmk3:
Notes regarding extensibility/flexibility in the makefiles with this PR:

Any variable used on src/prog.mk or src/so.mk can be extended by the includer.

Any variable used on src/prog.mk or src/so.mk other than the following
variables can be completely overridden by the includer:

* HDRS
* SCRS
* OBJS
* CFLAGS
* LDFLAGS

Note: I think that it would be unlikely for this to be an issue, but it should
eventually be possible to override `CFLAGS` and `LDFLAGS` if they are later
moved from src/prog.mk and src/so.mk into config.mk.

For example, if a module in src/foo needs to define extra
compile-time/link-time flags, it can do so by adding them to `EXTRA_CFLAGS` /
`EXTRA_LDFLAGS` before including src/prog.mk.  If it produces additional/custom
files that should be cleaned up, it can set `TOCLEAN` and/or `TODISTCLEAN`:

src/foo/Makefile:

```make
ROOT = ../..
-include $(ROOT)/config.mk

PROG = foo
TARGET = $(PROG)

EXTRA_CFLAGS += -O1
EXTRA_LDFLAGS += -lm

TOCLEAN += foo
TODISTCLEAN += bar

include $(ROOT)/src/prog.mk
```

Also, if a module differs too much from what is used on src/prog.mk and
src/so.mk, instead of including them, the module's Makefile could include only
config.mk and manually define its own custom variables and targets, similarly
to what is currently done without this PR.

netblue30:
merged, thanks!

PR #5481, opened 2022-11-24
ci: sort and ignore more paths

kmk3:
Commits:

```console
* ci: sort items on paths-ignore lists
* ci: ignore github markdown templates in all workflows
* ci: ignore dependabot and gitlab-ci in all workflows
* ci: ignore codeql workflow file in other workflows
* ci: ignore firecfg.config in the codeql workflow
* ci: ignore man/vim paths in build-extra/codeql workflows
```

Sort the paths and then ignore unnecessary files (mainly ignore workflow files
in other workflows).

Relates to #5249.

Kind of relates to #5289 #5296.

### DIFF
--- a/.github/workflows/build-extra.yml
+++ b/.github/workflows/build-extra.yml
@@ -6,6 +6,8 @@ on:
     paths-ignore:
       - '.github/ISSUE_TEMPLATE/*'
       - 'etc/**'
+      - 'contrib/vim/**'
+      - 'src/man/*.txt'
       - .git-blame-ignore-revs
       - .github/dependabot.yml
       - .github/pull_request_template.md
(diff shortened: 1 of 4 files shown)


glitsj16 (approved):
LGTM

kmk3:
Force-push changes:

All directories appear to be using wildcards, so I put them also on the vim
path (`contrib/vim` -> `'contrib/vim/**'`).

Not sure if it's required or not, as they appear to always be used in the docs:

* <https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore>

But I also noticed the following and added it to the first commit message:

> Note that in the current workflow files, ignored directories always use
> wildcards, so this also ensures that ignored directories (along with all
> other paths with wildcards) are always listed before ignored files
> (similarly to `--group-directories-first` in GNU `ls`).

netblue30:
Cool!

PR #5496, opened 2022-12-01
kcalc.profile: fix mkfile without mkdir & comment legacy paths

kmk3:
kcalc.profile: fix mkfile without mkdir

firejail may fail to create the following files:

* ~/.kde/share/config/kcalcrc
* ~/.kde4/share/config/kcalcrc

Because it does not create the preceding directories beforehand:

* ~/.kde/share/config
* ~/.kde4/share/config

---

kcalc.profile: stop creating legacy KDE paths

Leave them commented.

With this commit, there are no more profiles creating paths in ~/.kde
nor in ~/.kde4:

    $ git grep -e '^mkdir .*\.kde' -e '^mkfile .*\.kde' -- etc
    $

Relates to #5415.

### DIFF
--- a/etc/profile-a-l/kcalc.profile
+++ b/etc/profile-a-l/kcalc.profile
@@ -16,6 +16,8 @@ include disable-programs.inc
 include disable-shell.inc
 include disable-xdg.inc
 
+mkdir ${HOME}/.kde/share/config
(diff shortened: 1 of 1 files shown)


glitsj16 (approved):
LGTM

PR #5488, opened 2022-11-28
build: fix formatting and misc in configure

kmk3:
Commits:

```console
$ git log --reverse --pretty='* %s' master..
* configure*: fix typo of "been"
* configure*: fix trailing comma in HAVE_CONTRIB_INSTALL
* configure*: fix indentation
* configure*: remove extraneous blank lines
* configure*: remove extraneous blank lines (warning)
* configure*: wrap long shell command output lines
* configure*: wrap long AS_HELP_STRING line (sanitizer)
* configure*: fix quotes/parens alignment (sanitizer)
* configure*: quote and line-wrap AC_CHECK_HEADER line
```

Relates to #4712.

PR #5504, opened 2022-12-04
build: actually set LDFLAGS/LIBS & stop overriding CFLAGS/LDFLAGS

kmk3:
Commits:

```console
* configure*: print CC and CFLAGS
* build: actually set LDFLAGS and LIBS in makefiles
* build: move library flags from EXTRA_LDFLAGS to LIBS
* makefiles: stop overriding CFLAGS/LDFLAGS
```

This is a follow-up to #5478 and #5488.


kmk3:
Reproducing the last commit message here as these details might not be commonly
known:

---

stop overriding CFLAGS/LDFLAGS

From the manual of GNU Automake (version 1.16.5)[1] [2]:

> 3.6 Variables reserved for the user
>
> Some `Makefile` variables are reserved by the GNU Coding Standards for
> the use of the "user"—the person building the package.  For instance,
> `CFLAGS` is one such variable.
>
>    Sometimes package developers are tempted to set user variables such
> as `CFLAGS` because it appears to make their job easier.  However, the
> package itself should never set a user variable, particularly not to
> include switches that are required for proper compilation of the
> package.  Since these variables are documented as being for the
> package builder, that person rightfully expects to be able to override
> any of these variables at build time.
>
>    To get around this problem, Automake introduces an
> automake-specific shadow variable for each user flag variable.
> (Shadow variables are not introduced for variables like `CC`, where
> they would make no sense.) The shadow variable is named by prepending
> `AM_` to the user variable's name.  For instance, the shadow variable
> for `YFLAGS` is `AM_YFLAGS`.  The package maintainer—that is, the
> author(s) of the `Makefile.am` and `configure.ac` files—may adjust
> these shadow variables however necessary.
>
>    Note Flag Variables Ordering::, for more discussion about these
> variables and how they interact with per-target variables.

See also the description of CFLAGS in the GNU Autoconf manual[3].

Note: We do not use automake (save for aclocal) nor generally follow the
GNU Coding Standards, but the concept still applies.  Also, the closest
analogous in the project to the `AM_` prefix would currently likely be
`EXTRA_`.

[1] https://www.gnu.org/software/automake/manual/1.16.5/html_node/User-Variables.html
[2] https://www.gnu.org/software/automake/manual/1.16.5/html_node/Flag-Variables-Ordering.html
[3] https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Preset-Output-Variables.html

glitsj16 (approved):
LGTM

PR #5516, opened 2022-12-08
New profile: tesseract

glitsj16:
An OCR program (CLI). Handy tool for creating PDF's and extracting text from PNG image(s).
https://github.com/tesseract-ocr/tesseract

### DIFF
--- a/etc/profile-m-z/tesseract.profile
+++ b/etc/profile-m-z/tesseract.profile
@@ -53,7 +53,7 @@ x11 none
 private-bin ambiguous_words,classifier_tester,cntraining,combine_lang_model,combine_tessdata,dawg2wordlist,lstmeval,lstmtraining,merge_unicharsets,mftraining,set_unicharset_properties,shapeclustering,tesseract,text2image,unicharset_extractor,wordlist2dawg
 private-cache
 private-dev
-private-etc fonts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload
+private-etc alternatives,fonts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload
 #private-lib libtesseract.so.*
 private-tmp
 
--- a/etc/profile-m-z/tesseract.profile
+++ b/etc/profile-m-z/tesseract.profile
@@ -8,6 +8,8 @@ include globals.local
 
 blacklist ${RUNUSER}
 
+noblacklist ${DOCUMENTS}
+noblacklist ${PICTURES}
 include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
--- a/etc/profile-m-z/tesseract.profile
+++ b/etc/profile-m-z/tesseract.profile
@@ -62,3 +62,4 @@ dbus-user none
 dbus-system none
 
 memory-deny-write-execute
+restrict-namespaces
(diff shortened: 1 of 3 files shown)


etc/profile-m-z/tesseract.profile
```diff
@@ -0,0 +1,63 @@
...
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+whitelist ${DOCUMENTS}
+whitelist ${DOWNLOADS}
+whitelist ${PICTURES}
```
rusty-snake, line 22: Documents and Pictures are blacklisted by disable-xdg.inc

etc/profile-m-z/tesseract.profile
```diff
@@ -0,0 +1,63 @@
...
+nogroups
+noinput
+nonewprivs
+noprinters
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
```
rusty-snake, line 47: Is unix used or can we use `seccomp socket`?

etc/profile-m-z/tesseract.profile
```diff
@@ -0,0 +1,63 @@
...
+private-cache
+private-dev
+private-etc fonts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload
+#private-lib libtesseract.so.*
+private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
```
rusty-snake, line 63: `restrict-namespaces`?

etc/profile-m-z/tesseract.profile
glitsj16: Nice catch! Fixed now.

etc/profile-m-z/tesseract.profile
glitsj16: I'll be testing this shortly. Thanks for the review and your top-notch remarks/suggestions.

etc/profile-m-z/tesseract.profile
glitsj16: Will test this ASAP. Please hold.

etc/profile-m-z/tesseract.profile
glitsj16: `seccomp socket` works just fine. Added!

etc/profile-m-z/tesseract.profile
glitsj16: `restrict-namespaces` works fine. Added!

OT: I think we can go-ahead with adding it to much more profiles.

rusty-snake (approved):

etc/profile-m-z/tesseract.profile
glitsj16: > Notable xceptions are firefox and chomium.

@rusty-snake Thanks for the response. I'll come back to this topic later. For me `restrict-namespaces` does work with Firefox. But my firejailing/configuration of that web browser is a bit euhm, `non-standard` (using full seccomp for example) for the lack of a better term. Anyway, I'll be doing more testing on this option.

rusty-snake: Correct, firefox can work with it, because it has compatibility code for systems where (user)namespaces aren't available. However this will lower security.

PR #5517, opened 2022-12-09
private-etc: always reference 'alternatives'

glitsj16:
CI does checks on `private-etc` to always have 'alternatives' for Debian and derivatives. This PR ensures it's there, even in comments.


kmk3 (approved):
LGTM.

PR #5521, opened 2022-12-10
clipit hardening

glitsj16:
Adding several options to harden clipit.profile.

### DIFF
--- a/etc/profile-a-l/clipit.profile
+++ b/etc/profile-a-l/clipit.profile
@@ -13,7 +13,9 @@ include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
+include disable-proc.inc
 include disable-programs.inc
+include disable-shell.inc
 include disable-xdg.inc
 
 mkdir ${HOME}/.config/clipit
@@ -21,6 +23,8 @@ mkdir ${HOME}/.local/share/clipit
 whitelist ${HOME}/.config/clipit
 whitelist ${HOME}/.local/share/clipit
 include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
@@ -34,16 +38,25 @@ nodvd
 nogroups
 noinput
 nonewprivs
+noprinters
 noroot
 nosound
 notv
 nou2f
 novideo
-protocol unix
+#protocol unix
 seccomp
+tracelog
 
 disable-mnt
+private-bin clipit
 private-cache
 private-dev
+private-lib
 private-tmp
 
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+read-only ${HOME}


etc/profile-a-l/clipit.profile
```diff
@@ -34,16 +38,25 @@ nodvd
...
 noinput
 nonewprivs
+noprinters
 noroot
 nosound
 notv
 nou2f
 novideo
-protocol unix
+#protocol unix
```
rusty-snake, line 47: Why? Does it break something?

etc/profile-a-l/clipit.profile
```diff
@@ -34,16 +38,25 @@ nodvd
...
+private-bin clipit
 private-cache
 private-dev
+private-lib
 private-tmp
 
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
```
rusty-snake, line 61: `restrict-namespaces`?

etc/profile-a-l/clipit.profile
```diff
@@ -34,16 +38,25 @@ nodvd
...
 notv
 nou2f
 novideo
-protocol unix
+#protocol unix
 seccomp
+tracelog
 
 disable-mnt
+private-bin clipit
```
rusty-snake, line 52: `xdotool`?

> Requirements:
>  - xdotool - for automatic paste functionality

etc/profile-a-l/clipit.profile
rusty-snake, line 61: Did you tested with wayland and gtk3?

etc/profile-a-l/clipit.profile
glitsj16: Duh, forgot to uncomment it, which I did while testing. Can confirm it doesn't break anything. Pushing a fix shortly.

etc/profile-a-l/clipit.profile
glitsj16: Added!

etc/profile-a-l/clipit.profile
```diff
@@ -34,16 +38,26 @@ nodvd
...
 novideo
 protocol unix
 seccomp
+tracelog
 
 disable-mnt
+private-bin clipit,xdotool
 private-cache
 private-dev
+private-lib
```
kmk3, line 55:
```suggestion
#private-lib
```

Wouldn't this break everything with the default firejail.config settings?

kmk3: @rusty-snake [on Dec 10](https://github.com/netblue30/firejail/pull/5521#discussion_r1045138911):

> Why?

```console
$ firejail --quiet --noprofile --private-lib echo hello
Error: private-lib feature is disabled in Firejail configuration file /etc/firejail/firejail.config
```

rusty-snake: This would mean if you harden firejail by disabling stuff in firejail.config would break half of your profiles? No of course not, if an command is disabled in firejail.config (or at compile time?) we ignore it in profiles to not cause unnecessary breakage. On the commandline however, where the user explicitly requested it, we error out.

kmk3: @rusty-snake [on Dec 10](https://github.com/netblue30/firejail/pull/5521#discussion_r1045139469):

> This would mean if you harden firejail by disabling stuff in firejail.config
> would break half of your profiles? No of course not, if an command is
> disabled in firejail.config (or at compile time?) we ignore it in profiles to
> not cause unnecessary breakage. On the commandline however, where the user
> explicitly requested it, we error out.

I see, thanks.  This discrepancy makes trying to debug things on the
command-line a bit more confusing, but indeed, I suppose that it makes sense
for usability.

Printing a warning when ignoring a disabled command could be useful (to make
the user aware that it is being ignored), but anyway, that's outside the scope
of this PR.

glitsj16: I am aware of only one downside of `private-lib`: it can slow down creation of the sandbox. Yes its disabled by default in /etc/firejail/firejail.config. And yes, these lines are ignored in profile files in that case. Wouldn't make much sense to comment the option IMO. Unless it breaks the app, obviously, which isn't happening here with clipit.

kmk3, line 55: I'm not familiar with the program, but I see that it calls GTK-related APIs
(and depends on gtk3 on Artix), so I don't really get how this works with an
empty `private-lib`.

```console
$ LC_ALL=C pacman -Sii clipit | grep -e ^Version -e ^Depends
Version         : 1.4.5-1
Depends On      : gtk3  libappindicator-gtk3  libayatana-appindicator
```

Doesn't it need access to GTK libraries?

```console
$ pacman -Q xdotool
xdotool 3.20211022.1-1
$ pacman -Qlq xdotool | grep -F .so
/usr/lib/libxdo.so
/usr/lib/libxdo.so.3
```

Could it need `libxdo.so.*`?

rusty-snake: > so I don't really get how this works with an
empty private-lib.

`private-lib` will include all libraries reported by `fldd` i.e. everything loaded as dependencies of the ELF and not via `dlopen`.

Of course this does not affect external programs (e.g. `xdotool`) making private-lib such an complicated thing.

glitsj16: > $ LC_ALL=C pacman -Sii clipit | grep -e ^Version -e ^Depends
Version         : 1.4.5-1
Depends On      : gtk3  libappindicator-gtk3  libayatana-appindicator

On Arch Linux the situation is slightly different. The package isn't in any of the official repo's, but in the [AUR](https://aur.archlinux.org/packages/clipit) and there are several reports of the latest version being broken. I've looked into this on several occasions and the breakage seems to be in the GUI part (the tray icon). If one starts clipit with the `-dn` flags (-d for daemonize and -n for not showing the tray icon) it works okayish. And in that case neither xdotool nor the tray icon are needed/used. Granted, all that isn't explicitly covered in the profile. I'm open to suggestions, although it might not be worth the trouble. Upstream seems as good as dead, which is (partly) why I wouldn't recommend clipit. There are modern, well-maintained alternatives around (like [gpaste](https://www.imagination-land.org/tags/GPaste.html) for example). But in this context my personal opinion isn't relevant.

If we had more supported clipboard managers in Firejail I would even go so far as to suggest retiring clipit.profile. Sadly that isn't the case.

kmk3 (approved):
LGTM.

PR #5538, opened 2022-12-20
audacity: support more config locations

glitsj16:


kmk3 (commented):
It seems to use more directories:

From <https://github.com/audacity/audacity/blob/b8d10528d4396e69e9bce8a750a5f9c9eaf1c1a0/libraries/lib-files/FileNames.h>:

> ```cpp
>    /** \brief Audacity user cache directory
>     *
>     * Where audacity keeps its cache squirreled away, by default ~/.cache/audacity/
>     * on Unix, Application Data/Audacity on windows system */
>    FILES_API FilePath CacheDir();
>    /** \brief Audacity user config directory
>     *
>     * Where audacity keeps its settigns squirreled away, by default ~/.config/audacity/
>     * on Unix, Application Data/Audacity on windows system */
>    FILES_API FilePath ConfigDir();
>    /** \brief Audacity user data directory
>     *
>     * Where audacity keeps its user data squirreled away, by default ~/.local/share/audacity/
>     * on Unix, Application Data/Audacity on windows system */
>    FILES_API FilePath DataDir();
>    /** \brief Audacity user state directory
>     *
>     * Where audacity keeps its user state squirreled away, by default ~/.local/state/audacity/
>     * on Unix, Application Data/Audacity on windows system */
>    FILES_API FilePath StateDir();
> ```

glitsj16:
@kmk3 Nice detective work! Added, thanks.

glitsj16:
Going over the audacity profile I just noticed something odd that is unrelated to this PR.
We have both `net none` and `protocol unix,inet`, which - combined - don't make sense. AFAIK Audacity only goes out to the internet to download 2 extra plugins. I'll dig deeper into this and open a separate PR if needed.

etc/inc/disable-programs.inc
```diff
@@ -1014,7 +1017,8 @@ blacklist ${HOME}/.local/share/wget
 blacklist ${HOME}/.local/share/wormux
 blacklist ${HOME}/.local/share/xplayer
 blacklist ${HOME}/.local/share/xreader
-blacklist ${HOME}/.local/share/zathura
+blacklist ${HOME}/.local/share/zathura*/
```
kmk3, line 1020:
```suggestion
blacklist ${HOME}/.local/share/zathura
```

glitsj16: GitHub syntax highlighting made me break disable-programs.inc. Thanks to your keen eyes and report this is now fixed :+1: .

kmk3: No problem :)

When I saw it I was confused, as it looked like it could be an unrelated fix
for something else, but since nothing was said about zathura and since
zathura.profile wasn't changed to match it, I concluded that it was likely a
typo.

kmk3 (approved):
LGTM.

Issue #5539, opened 2022-12-20
audacity: network access and sandbox violation report


glitsj16:
During another round of testing for work on #5538 I noticed a few things about our current profile.

The app scans the system for available plugins, including /lib/modules (which is a hard-coded no-no). That throws a `sandbox violation` which might confuse users. Adding the `allow-debuggers` option can take care of this and AFAICT shouldn't negatively affect the security of the sandbox. Due to our restrictive `private-bin audacity` there wouldn't actually be active support for debuggers, just silencing the blacklist violation.

A more difficult decision to make is whether we grant `network access` to Audacity. Currently we do not, although there's an inconsistency in our profile which we should fix. After testing I can only see the app going out on the internet to download 2 extra plugins:

- http://breakfastquay.com/rdf/lv2-rubberband#r3mono
- http://breakfastquay.com/rdf/lv2-rubberband#r3stereo

Besides using non-https URI's it's a bit of a pain to grant network access to an audio application for this sole purpose. There are plenty of built-in and installable (ladspa) filters/plugins already IMO. But that's an opinion nonetheless. Any thoughts on how we should proceed on this?

netblue30:
Last year there were some discussions regarding the new privacy policy:

https://lifehacker.com/is-audacity-really-spyware-1847230028
https://www.pcmag.com/news/audacity-is-being-called-spyware-after-privacy-policy-update

Let's leave it the way you put it in https://github.com/netblue30/firejail/commit/6e67801a4592494d3d24e696f6fe06acc7856d88. If anybody needs live updates they can enable it in the profile.

glitsj16:
@netblue30 Interesting info. Thanks for the heads-up!

kmk3:
> The app scans the system for available plugins, including /lib/modules (which
> is a hard-coded no-no). That throws a `sandbox violation` which might confuse
> users. Adding the `allow-debuggers` option can take care of this and AFAICT
> shouldn't negatively affect the security of the sandbox. Due to our
> restrictive `private-bin audacity` there wouldn't actually be active support
> for debuggers, just silencing the blacklist violation.

Is this still an issue?

> A more difficult decision to make is whether we grant `network access` to
> Audacity. Currently we do not, although there's an inconsistency in our
> profile which we should fix. After testing I can only see the app going out
> on the internet to download 2 extra plugins:

Networking is enabled by default in the profile now:

* #6321

glitsj16:
> Is this still an issue?

[...]

PR #5540, opened 2022-12-20
audacity: networking updates

glitsj16:
Fixes for #5539.

### DIFF
--- a/etc/profile-a-l/audacity.profile
+++ b/etc/profile-a-l/audacity.profile
@@ -6,6 +6,11 @@ include audacity.local
 # Persistent global definitions
 include globals.local
 
+# Add the below lines to your audacity.local if you need online plugins.
+#ignore net none
+#netfilter
+#protocol inet6
+
 noblacklist ${HOME}/.audacity-data
 noblacklist ${DOCUMENTS}
 noblacklist ${MUSIC}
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/audacity.profile
```diff
@@ -33,7 +39,7 @@ noroot
 notv
 nou2f
 novideo
-protocol unix,inet
```
rusty-snake, line 36: Only `inet` indicates audacity may run a local server.

rusty-snake: See #4659.

glitsj16: That flew way under my radar. Fixed, thanks!

etc/profile-a-l/audacity.profile
```diff
@@ -20,9 +20,15 @@ include disable-xdg.inc
...
 
+# Silence blacklist violation. See #5539.
+allow-debuggers
 ## Enabling App Armor appears to break some Fedora / Arch installs
 #apparmor
 caps.drop all
+# Add the below lines to your audacity.local if you need online plugins.
+#ignore net none
+#netfilter
+#protocol inet6
```
kmk3, line 31: Suggestion: Move these lines to the top of the profile (below
`include globals.local`), since they contain an `ignore` command and also to
make the list around them more readable.

kmk3 (approved):

PR #5543, opened 2022-12-20
seahorse refactoring

glitsj16:


etc/profile-m-z/seahorse.profile
```diff
@@ -26,14 +24,13 @@ include disable-xdg.inc
...
 #whitelist ${HOME}/.ssh
-whitelist /tmp/ssh-*
 whitelist /usr/share/gnupg
 whitelist /usr/share/gnupg2
 whitelist /usr/share/seahorse
 whitelist /usr/share/seahorse-nautilus
 whitelist ${RUNUSER}/gnupg
 whitelist ${RUNUSER}/keyring
-#include whitelist-common.inc
+include whitelist-common.inc
```
kmk3, line 33: What about the commented `whitelist ${HOME}` commands and the nearby comment?

> whitelisting in ${HOME} breaks file encryption feature of nautilus.

etc/profile-m-z/seahorse.profile
```diff
@@ -26,14 +24,13 @@ include disable-xdg.inc
 #mkdir ${HOME}/.ssh
 #whitelist ${HOME}/.gnupg
 #whitelist ${HOME}/.ssh
-whitelist /tmp/ssh-*
```
kmk3, line 29: This appears to be the path to a socket used by ssh-agent.

Can seahorse talk to ssh-agent by default without this?

See also commit 90f2d7369 ("allow-ssh.inc: allow access to ssh-agent(1)",
2021-01-11) / PR #3885.

etc/profile-m-z/seahorse.profile
glitsj16: Argh, leftover from my testing. Thanks!

etc/profile-m-z/seahorse.profile
glitsj16: noblacklist /tmp/ssh-* is already in the included allow-ssh.inc. No need to whitelist it here again explicitly.

kmk3: @glitsj16 [on Dec 20](https://github.com/netblue30/firejail/pull/5543/files#r1053838898):

> `noblacklist /tmp/ssh-*` is already in the included allow-ssh.inc. No need to
> whitelist it here again explicitly.

seahorse-tool.profile includes `private-tmp` and redirects to seahorse.profile.
Would seahorse-tool need access to ssh-agent?

Also, this is the only whitelist command for /tmp (which effectively enables
`private-tmp` on seahorse.profile), so how about keeping it and moving
`private-tmp` from seahorse-tool.profile to seahorse.profile?

kmk3: To be clear, my point is that removing this line would effectively disable
`private-tmp` on seahorse.profile.

glitsj16: Oh, now I see what you mean. Moved private-tmp to seahorse.profile after testing both. All good now I think. Thanks!

kmk3: @glitsj16 [on Dec 20](https://github.com/netblue30/firejail/pull/5543#discussion_r1053863459):

> Oh, now I see what you mean. Moved private-tmp to seahorse.profile after
> testing both. All good now I think. Thanks!

Nice, but the line is still being removed (so /tmp is likely empty now).

Can seahorse access ssh-agent without this path?

If so, where is the path to the ssh-agent socket on your system?

glitsj16: I see where the confusion was. On my Arch Linux box I run GNOME and that uses part of gnome-keyring as a ssh-agent. I've disabled that for another round of testing as per the Arch wiki and now - when putting back the `whitelist /tmp/ssh-*` line - I can see these files inside the sandbox. Thank you for your patience :+1: .

For reference:
- https://wiki.archlinux.org/title/SSH_keys#Start_ssh-agent_with_systemd_user
- https://wiki.archlinux.org/title/GNOME/Keyring#Disable_keyring_daemon_components

PR #5544, opened 2022-12-20
New profile: ssmtp

glitsj16:
Old software, but still useful as MTA.

### DIFF
--- a/etc/profile-m-z/ssmtp.profile
+++ b/etc/profile-m-z/ssmtp.profile
@@ -1,0 +1,70 @@
+# Firejail profile for ssmtp
+# Description: Extremely simple MTA to get mail off the system to a mailhub
+# This file is overwritten after every install/update
+quiet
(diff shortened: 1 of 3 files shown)


etc/profile-m-z/ssmtp.profile
```diff
@@ -0,0 +1,75 @@
...
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+seccomp.block-secondary
+tracelog
+
+disable-mnt
+# private works but we loose ${HOME}/dead.letter
+# which is useful to get notified on mail issues
```
kmk3, line 62:
```suggestion
# private works but then we lose ${HOME}/dead.letter
# which is useful to get notified on mail issues
```

glitsj16: Better, thanks!

kmk3 (approved):

PR #5546, opened 2022-12-21
spectre-meltdown-checker: fixes

glitsj16:
(Offtopic)
There's a typo in the latest release of the script that causes a leftover file (/devnull) being present in the filesystem. I've [opened a PR](https://github.com/speed47/spectre-meltdown-checker/pull/442) upstream to fix that. Out-of-scope for Firejail, although in theory we could use `private` to prevent it. But with a script that creates such a long and detailed output I opted not to do that.


kmk3 (approved):

speed47:
FWIW, just merged your upstream PR :)

Issue #5545, opened 2022-12-21
/proc/config.gz blacklisted twice


glitsj16:
Minor issue with blacklisting kernel configuration:

https://github.com/netblue30/firejail/blob/6e67801a4592494d3d24e696f6fe06acc7856d88/etc/inc/disable-common.inc#L624

https://github.com/netblue30/firejail/blob/6e67801a4592494d3d24e696f6fe06acc7856d88/etc/inc/disable-proc.inc#L11

Not sure which of these include files would be the better place for it. I've checked our profiles and there are _zero_ files that include disable-proc.inc without also including disable-common.inc, so it won't be a lot of work to remove it from one or the other. Or am I missing something here and it's supposed to be like this?

kmk3:
Since disable-common.inc is the more, well, common include I'd leave it only
there.

Alternatively, put something like this in disable-proc.inc:

```firejail
blacklist /proc/config.gz # also in disable-common.inc
```

To make it clear that the duplication is intentional.

rusty-snake:
>disable-common.inc is the more, well, common include 

Correct. But disable-proc should really disable proc and not depend on dc for full functionality.

I'd say dp must have it and we should keep it in dc as legacy blacklist for now.

OT: disable-proc can and should be used in much more profiles.

glitsj16:
> disable-proc should really disable proc and not depend on dc for full functionality

Agreed. Now we just made it a bit more visible via the comment. Might avoid future mishaps if someone tries to remove it from dc.

PR #5548, opened 2022-12-21
clarify that duplicated blacklisting of /proc/config.gz is intentional

glitsj16:
Relates to #5545.


rusty-snake:
I would flip the logic and add the comment to dc in the first place or to both.

glitsj16:
@rusty-snake I've added the comment to both files. Thanks for the input!

rusty-snake (approved):

kmk3 (approved):

PR #5551, opened 2022-12-21
Add new (redirect) profile: dirmngr

glitsj16:
Add support for dirmngr - GnuPG's network access daemon. Implemented as redirect to gpg-agent.profile.


glitsj16:
I'll come back to this later. Need more time to test and might need refactoring gpg, which is not a good idea so close to the upcoming 0.9.72 release. Closing.

PR #5553, opened 2022-12-23
Blacklist  google-drive-ocamlfuse config

slowpeek:
[google-drive-ocamlfuse](https://github.com/astrada/google-drive-ocamlfuse) is a _FUSE filesystem over Google Drive_. Its config stores sensitive access tokens, much like rclone's config. The latter is blacklisted in firejail by default, so let it be the same for the former.

According to the [docs](https://github.com/astrada/google-drive-ocamlfuse/blob/beta/doc/Configuration.md) there are two default pathes for the config: either `~/.gdfuse` or `~/.config/gdfuse`

### DIFF
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -116,6 +116,7 @@ blacklist ${HOME}/.cache/fossamail
 blacklist ${HOME}/.cache/fractal
 blacklist ${HOME}/.cache/freecol
 blacklist ${HOME}/.cache/gajim
+blacklist ${HOME}/.cache/gdfuse
 blacklist ${HOME}/.cache/geary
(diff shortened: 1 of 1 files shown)


slowpeek:
I've just figured out there is more to blacklist. google-drive-ocamlfuse can store its config in two ways:
- everything in one dir, by default `~/.gdfuse`
- scattered accross three XDG dirs:
  - `~/.local/share/gdfuse` this one is the most sensitive of the three, access tokens go here
  - `~/.cache/gdfuse`
  -  `~/.config/gdfuse`

rusty-snake (approved):

kmk3 (approved):

netblue30:
merged, thanks!

Issue #3701, opened 2020-10-26
Firefox native messaging regression in 0.9.62.4 -> 0.9.64rc1


cinnaboot:
When upgrading to firejail 0.9.64rc1, the firefox extension https://github.com/woodruffw/ff2mpv breaks. 
It makes use of https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging to talk to other native applications.

In firefox.local, I have
`# Python 3`
`noblacklist ${PATH}/python3*`
`noblacklist /usr/include/python3*`
`noblacklist /usr/lib/python3*`
`noblacklist /usr/lib64/python3*`
`noblacklist /usr/local/lib/python3*`
`noblacklist /usr/share/python3*`

After upgrading to 1.9.64rc1 (and in 1.9.64), I get the following in strace:
`[pid   261] execve("/home/user/.mozilla/ff2mpv.py", ["/home/user/.mozilla/ff2mpv.py", "/home/user/.mozilla/native-mes"..., "ff2mpv@yossarian.net"], 0x7f9a121b4c00 /* 86 vars */) = -1 EACCES (Permission denied)`

firejail --noprofile does work too

Do I need to add some more options for the new DBus filtering maybe?

Forgot to add, this is on Arch Linux

rusty-snake:
Add `ignore noexec ${HOME}` to your firefox.local.

In 0.9.62 was `ignore noexec ${HOME}` by default set even it shouldn't.

cinnaboot:
Thanks, that helped.
I ended up needing to add a bunch of stuff from mpv.profile to make it go also:

```
ignore noexec ${HOME}

# copy firejail config parts from mpv.profile
noblacklist ${HOME}/.config/mpv
noblacklist ${HOME}/.config/youtube-dl
noblacklist ${HOME}/.netrc

include allow-lua.inc
include allow-python3.inc
mkdir ${HOME}/.config/mpv
mkdir ${HOME}/.config/youtube-dl
mkfile ${HOME}/.netrc
whitelist ${HOME}/.config/mpv
whitelist ${HOME}/.config/youtube-dl
whitelist ${HOME}/.netrc
whitelist /usr/share/lua
whitelist /usr/share/lua*
whitelist /usr/share/vulkan

private-bin env,mpv,python*,waf,youtube-dl
```

PR #3619, opened 2020-09-01
Fixes smplayer

kortewegdevries:

PR #5556, opened 2022-12-25
Add profile for Chatterino

Dpeta:
Adds a profile for [Chatterino](https://github.com/chatterino/chatterino2), an IRC twitch chat client.

I tried to keep the Streamlink integration functional, so the relevant binaries are added to private-bin and the mpv/VLC config folders are whitelisted. The profile could be more restrictive without this, not sure if that's more desirable.

I'm also not certain how to get opening links to work properly without disabling options like private-bin, for login this isn't a major issue as the client just gives the link to copy-paste instead though.

### DIFF
--- a/etc/profile-a-l/chatterino.profile
+++ b/etc/profile-a-l/chatterino.profile
@@ -1,0 +1,116 @@
+# Firejail profile for Chatterino
+# Description: Chat client for https://twitch.tv
+# This file is overwritten after every install/update
+# Persistent local customizations
+include chatterino.local
+# Persistent global definitions
+include globals.local
+
+# Also allow access to mpv/vlc, they're usable via streamlink.
+noblacklist ${HOME}/.cache/vlc
+noblacklist ${HOME}/.config/aacs
+noblacklist ${HOME}/.config/mpv
+noblacklist ${HOME}/.config/pulse
+noblacklist ${HOME}/.config/vlc
+noblacklist ${HOME}/.local/share/chatterino
+noblacklist ${HOME}/.local/share/vlc
+# To upload images, whitelist/noblacklist their path in chatterino.local.
+#noblacklist ${HOME}/Pictures/
+# For custom notification sounds, whitelist/noblacklist their path in chatterino.local.
+#noblacklist ${HOME}/Music/
+
+# Allow Python for Streamlink integration (blacklisted by disable-interpreters.inc)
+include allow-python3.inc
+
+# Allow Lua for mpv (blacklisted by disable-interpreters.inc)
+include allow-lua.inc
+
+# disable-*.inc includes
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+# Also allow access to mpv/vlc, they're usable via streamlink.
+mkdir ${HOME}/.cache/vlc
+mkdir ${HOME}/.config/aacs
+mkdir ${HOME}/.config/mpv
+mkdir ${HOME}/.config/pulse
+mkdir ${HOME}/.config/vlc
+mkdir ${HOME}/.local/share/chatterino
+mkdir ${HOME}/.local/share/vlc
+whitelist ${HOME}/.cache/vlc
+whitelist ${HOME}/.config/aacs
+whitelist ${HOME}/.config/mpv
+whitelist ${HOME}/.config/pulse
+whitelist ${HOME}/.config/vlc
+whitelist ${HOME}/.local/share/chatterino
+whitelist ${HOME}/.local/share/vlc
+# To upload images, whitelist/noblacklist their path in chatterino.local.
+#whitelist ${HOME}/Pictures/
+# For custom notification sounds, whitelist/noblacklist their path in chatterino.local.
+#whitelist ${HOME}/Music/
+# whitelist-*.inc includes
+include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+# Streamlink+VLC doesn't seem to close properly with apparmor enabled.
+#apparmor
+caps.drop all
+netfilter
+nodvd
+nogroups
+nonewprivs
+noprinters
+noroot
+notv
+nou2f
+# Netlink is required for streamlink integration.
+protocol unix,inet,inet6,netlink
+# Seccomp may break browser integration.
+seccomp
+seccomp.block-secondary
+tracelog
+
+disable-mnt
+# Add more private-bin lines for browsers or video players to chatterino.local if wanted.
+private-bin chatterino,pgrep
+private-bin ffmpeg,python*,streamlink
+private-bin cvlc,nvlc,qvlc,rvlc,svlc,vlc
+private-bin env,mpv,python*,waf,youtube-dl,yt-dlp
+# private-cache may cause issues with mpv (see #2838)
+private-cache
+private-dev
+private-etc alsa,alternatives,asound.conf,ca-certificates,dbus-1,fonts,hostname,hosts,kde4rc,kde5rc,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,machine-id,nvidia,passwd,pulse,resolv.conf,rpc,services,ssl,Trolltech.conf,X11
+private-opt none
+private-srv none
+private-tmp
+
+dbus-user filter
+dbus-user.own com.chatterino.*
+# Session Bus Policy from flatpak
+dbus-user.talk com.canonical.AppMenu.Registrar
+dbus-user.talk org.kde.kconfig.notify
+dbus-user.talk org.kde.KGlobalSettings
+dbus-user.talk org.freedesktop.Flatpak
+# Allow notifications.
+dbus-user.talk org.freedesktop.Notifications
+# For media player integration.
+dbus-user.talk org.freedesktop.ScreenSaver
+?ALLOW_TRAY: dbus-user.talk org.kde.StatusNotifierWatcher
+dbus-user.talk org.mpris.MediaPlayer2.Player
+dbus-system none
+
+# Prevents browsers/players from lingering after Chatterino is closed.
+#deterministic-shutdown
+# Add to chatterino.local to force Qt to use its wayland QPA plugin.
+#env QT_QPA_PLATFORM=wayland
+# memory-deny-write-execute may break streamlink and browser integration.
+#memory-deny-write-execute
+restrict-namespaces
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -876,6 +876,7 @@ blacklist ${HOME}/.local/share/caja-python
 blacklist ${HOME}/.local/share/calligragemini
 blacklist ${HOME}/.local/share/cantata
 blacklist ${HOME}/.local/share/cdprojektred
+blacklist ${HOME}/.local/share/chatterino
 blacklist ${HOME}/.local/share/clipit
 blacklist ${HOME}/.local/share/com.github.johnfactotum.Foliate
 blacklist ${HOME}/.local/share/contacts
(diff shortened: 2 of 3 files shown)


etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,116 @@
...
+private-srv none
+private-tmp
+
+dbus-user filter
+dbus-user.own com.chatterino.*
+# Session Bus Policy from flatpak
+dbus-user.talk com.canonical.AppMenu.Registrar
+dbus-user.talk org.kde.kconfig.notify
+dbus-user.talk org.kde.KGlobalSettings
+dbus-user.talk org.freedesktop.Flatpak
```
rusty-snake, line 101:
```suggestion
```

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,116 @@
...
+# Allow notifications.
+dbus-user.talk org.freedesktop.Notifications
+# For media player integration.
+dbus-user.talk org.freedesktop.ScreenSaver
+?ALLOW_TRAY: dbus-user.talk org.kde.StatusNotifierWatcher
+dbus-user.talk org.mpris.MediaPlayer2.Player
```
rusty-snake, line 107: No `dbus-user.own org.mpris.MediraPlyer2.chatterino`?

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,116 @@
...
+dbus-system none
+
+# Prevents browsers/players from lingering after Chatterino is closed.
+#deterministic-shutdown
+# Add to chatterino.local to force Qt to use its wayland QPA plugin.
+#env QT_QPA_PLATFORM=wayland
```
rusty-snake, line 113:
```

File: etc/profile-a-l/chatterino.profile [hunk 4]
```diff
@@ -0,0 +1,116 @@
...
+# Add more private-bin lines for browsers or video players to chatterino.local if wanted.
+private-bin chatterino,pgrep
+private-bin ffmpeg,python*,streamlink
+private-bin cvlc,nvlc,qvlc,rvlc,svlc,vlc
+private-bin env,mpv,python*,waf,youtube-dl,yt-dlp
+# private-cache may cause issues with mpv (see #2838)
+private-cache
+private-dev
+private-etc alsa,alternatives,asound.conf,ca-certificates,dbus-1,fonts,hostname,hosts,kde4rc,kde5rc,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,machine-id,nvidia,passwd,pulse,resolv.conf,rpc,services,ssl,Trolltech.conf,X11
+private-opt none
```
rusty-snake, line 91:
```suggestion
```

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,116 @@
...
+seccomp
+seccomp.block-secondary
+tracelog
+
+disable-mnt
+# Add more private-bin lines for browsers or video players to chatterino.local if wanted.
+private-bin chatterino,pgrep
+private-bin ffmpeg,python*,streamlink
+private-bin cvlc,nvlc,qvlc,rvlc,svlc,vlc
+private-bin env,mpv,python*,waf,youtube-dl,yt-dlp
```
rusty-snake, line 86: Generally we have only one line.

Also `python*` is included twice.

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,116 @@
...
+mkdir ${HOME}/.local/share/vlc
+whitelist ${HOME}/.cache/vlc
+whitelist ${HOME}/.config/aacs
+whitelist ${HOME}/.config/mpv
+whitelist ${HOME}/.config/pulse
+whitelist ${HOME}/.config/vlc
+whitelist ${HOME}/.local/share/chatterino
+whitelist ${HOME}/.local/share/vlc
+# To upload images, whitelist/noblacklist their path in chatterino.local.
+#whitelist ${HOME}/Pictures/
```
rusty-snake, line 53:
```suggestion
#whitelist ${PICTURES}
```

rusty-snake, line 53:
```suggestion
#whitelist ${HOME}/Pictures/pic1.png
```

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,116 @@
...
+# For custom notification sounds, whitelist/noblacklist their path in chatterino.local.
+#whitelist ${HOME}/Music/
```
rusty-snake, line 55: Same

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,116 @@
...
+mkdir ${HOME}/.config/vlc
+mkdir ${HOME}/.local/share/chatterino
+mkdir ${HOME}/.local/share/vlc
+whitelist ${HOME}/.cache/vlc
+whitelist ${HOME}/.config/aacs
+whitelist ${HOME}/.config/mpv
+whitelist ${HOME}/.config/pulse
+whitelist ${HOME}/.config/vlc
+whitelist ${HOME}/.local/share/chatterino
+whitelist ${HOME}/.local/share/vlc
```
rusty-snake, line 51: All the media player can be read-only??

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,116 @@
...
+noblacklist ${HOME}/.config/aacs
+noblacklist ${HOME}/.config/mpv
+noblacklist ${HOME}/.config/pulse
+noblacklist ${HOME}/.config/vlc
+noblacklist ${HOME}/.local/share/chatterino
+noblacklist ${HOME}/.local/share/vlc
+# To upload images, whitelist/noblacklist their path in chatterino.local.
+#noblacklist ${HOME}/Pictures/
+# For custom notification sounds, whitelist/noblacklist their path in chatterino.local.
+#noblacklist ${HOME}/Music/
```
rusty-snake, line 20:
```suggestion
#noblacklist ${PICTUES}
# For custom notification sounds, whitelist/noblacklist their path in chatterino.local.
#noblacklist ${MUSIC}
```

rusty-snake (commented):
Not bad as a first profile contribution. Must be the coumentation :grin:. Some (opinion) nitpicks to look at.

etc/profile-a-l/chatterino.profile
```diff
@@ -34,25 +28,22 @@ include disable-proc.inc
...
+whitelist-ro ${HOME}/.config/mpv
+whitelist-ro ${HOME}/.config/pulse
+whitelist-ro ${HOME}/.config/vlc
+whitelist-ro ${HOME}/.local/share/vlc
 # To upload images, whitelist/noblacklist their path in chatterino.local.
-#whitelist ${HOME}/Pictures/pic1.png
+#whitelist ${PICTURES}/pic1.png
 # For custom notification sounds, whitelist/noblacklist their path in chatterino.local.
-#whitelist ${HOME}/Music/
+#whitelist ${MUSIC}/sound.ogg
```
rusty-snake, line 46: `${MUSCI|PICTURES}/abcd` does not work. Otherwiese LGTM.

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,97 @@
...
+include disable-proc.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+# Also allow read-only access to mpv/VLC, they're usable via streamlink.
+mkdir ${HOME}/.config/mpv
+mkdir ${HOME}/.config/pulse
+mkdir ${HOME}/.config/vlc
+mkdir ${HOME}/.local/share/chatterino
+mkdir ${HOME}/.local/share/vlc
```
kmk3, line 36: To avoid cluttering the home directory with potentially unnecessary paths, just
whitelist the paths of other programs instead of also creating them.

Doing this would probably only help if these programs were never used before
opening them through chatterino (which is unlikely).

Also, AFAIK the configuration file for at least mpv has to be created manually,
in which case creating ~/.config/mpv in the profile wouldn't help anyway.

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,97 @@
...
+# Netlink is required for streamlink integration.
+protocol unix,inet,inet6,netlink
+# Seccomp may break browser integration.
+seccomp
+seccomp.block-secondary
+tracelog
+
+disable-mnt
+# Add more private-bin lines for browsers or video players to chatterino.local if wanted.
+private-bin chatterino,cvlc,env,ffmpeg,mpv,nvlc,pgrep,python*,qvlc,rvlc,streamlink,svlc,vlc,waf
```
kmk3, line 70: `waf` as in the build system?  Why would it be needed here?

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,97 @@
...
+noblacklist ${HOME}/.config/pulse
+noblacklist ${HOME}/.config/vlc
+noblacklist ${HOME}/.local/share/chatterino
+noblacklist ${HOME}/.local/share/vlc
+
+# Allow Python for Streamlink integration (blacklisted by disable-interpreters.inc)
+include allow-python3.inc
+
+# Allow Lua for mpv (blacklisted by disable-interpreters.inc)
+include allow-lua.inc
```
kmk3, line 20: Sort alphabetically, like in the template (Lua, Python).

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,97 @@
...
+
+# disable-*.inc includes
```
kmk3, line 22:
```suggestion
```

While this line is part of the template, it is not present in any profile.

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,93 @@
...
+whitelist ${HOME}/.local/share/chatterino
+whitelist-ro ${HOME}/.config/mpv
+whitelist-ro ${HOME}/.config/pulse
+whitelist-ro ${HOME}/.config/vlc
+whitelist-ro ${HOME}/.local/share/vlc
+# To upload images, whitelist/noblacklist their path in chatterino.local.
+#whitelist ${PICTURES}
+# For custom notification sounds, whitelist/noblacklist their path in chatterino.local.
+#whitelist ${MUSIC}
+# whitelist-*.inc includes
```
kmk3, line 43:
```suggestion
```

etc/profile-a-l/chatterino.profile
```diff
@@ -0,0 +1,93 @@
...
+# VLC preferences will fail to save with read-only set.
+whitelist ${HOME}/.local/share/chatterino
+whitelist-ro ${HOME}/.config/mpv
+whitelist-ro ${HOME}/.config/pulse
+whitelist-ro ${HOME}/.config/vlc
+whitelist-ro ${HOME}/.local/share/vlc
+# To upload images, whitelist/noblacklist their path in chatterino.local.
+#whitelist ${PICTURES}
+# For custom notification sounds, whitelist/noblacklist their path in chatterino.local.
+#whitelist ${MUSIC}
```
kmk3, line 42: To avoid cluttering the whitelist section, put `ignore` commands and optional
commands on the top of the profile.  Example:

```
# [...]
include globals.local

# To upload images, whitelist/noblacklist their path in chatterino.local.
#whitelist ${PICTURES}
# For custom notification sounds, whitelist/noblacklist their path in chatterino.local.
#whitelist ${MUSIC}
```

etc/profile-a-l/chatterino.profile
Dpeta: Assumed mpv used it since it's present in the private-bin line in mpv.profile (since #3619?), seems to only be a build/make dependency though, I'll remove it.

glitsj16: I think this `waf` addition stems from https://github.com/netblue30/firejail/issues/3701. It turns up in https://github.com/netblue30/firejail/commit/f3585e53933c95d3be31bb53214145d9219ff3ea too and is still in those profiles too. I don't know if it is actually needed. Like you asked, it's indeed strange to have a build system like waf in there. In the context of this specific PR though, I assume the OP just copied what was already there to begin with. Which does seem to be the best way to proceed when not wanting to break things. We should try to determine this `waf` situation and tackle it separately if it turns out we can drop it. Just my two cents.

netblue30:
Let's go for it, thanks @Dpeta! Other fixes should go in new pull rq.

PR #5557, opened 2022-12-25
window manager profiles: fix browser/electron internal sandboxes

smitsohu:
Assuming that users expect a working browser on their desktops.

Fix browsers and electron apps running inside a window manager sandbox, most of the time.


glitsj16 (approved):
LGTM. Nice fixes!

PR #5563, opened 2023-01-03
New profiles: linuxqq/qq

glitsj16:
Add profiles for https://im.qq.com/linuxqq/.
See https://github.com/BeautyYuYanli/linuxqq-new-firejail/issues/1 for context.

### DIFF
--- a/etc/profile-a-l/linuxqq.profile
+++ b/etc/profile-a-l/linuxqq.profile
@@ -18,9 +18,6 @@ whitelist ${HOME}/.config/QQ
 whitelist ${HOME}/.mozilla/firefox/profiles.ini
 whitelist ${DESKTOP}
 
-# Add the next line to your linuxqq.local if your kernel allows unprivileged userns clone.
-#include chromium-common-hardened.inc.profile
-
 ignore apparmor
 noprinters
 
--- a/etc/profile-a-l/linuxqq.profile
+++ b/etc/profile-a-l/linuxqq.profile
@@ -9,13 +9,14 @@ include globals.local
 noblacklist ${HOME}/.config/QQ
 noblacklist ${HOME}/.mozilla
 
+include allow-bin-sh.inc
+
+include disable-shell.inc
+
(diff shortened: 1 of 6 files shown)


etc/profile-a-l/linuxqq.profile
```diff
@@ -0,0 +1,45 @@
...
+noblacklist ${HOME}/.config/QQ
+noblacklist ${HOME}/.mozilla
+
+mkdir ${HOME}/.config/QQ
+whitelist ${HOME}/.config/QQ
+whitelist ${HOME}/.mozilla/firefox/profiles.ini
+whitelist ${DESKTOP}
+
+include allow-bin-sh.inc
+include disable-shell.inc
```
rusty-snake, line 18:
```suggestion
include allow-bin-sh.inc

include disable-shell.inc

mkdir ${HOME}/.config/QQ
whitelist ${HOME}/.config/QQ
whitelist ${HOME}/.mozilla/firefox/profiles.ini
whitelist ${DESKTOP}
```

etc/profile-a-l/linuxqq.profile
```diff
@@ -0,0 +1,45 @@
...
+
+# Add the next line to your linuxqq.local if your kernel allows unprivileged userns clone.
+#include chromium-common-hardened.inc.profile
+
```
rusty-snake, line 22:
```suggestion
```

Already in electron.profile.

glitsj16: Correct. But if users do exactly what is suggested in the electron.profile comment and they add it to their `chromium-common.local` it will not be in the linuxqq sandbox profile chain right? That's why I added it here and changed to `linuxqq.local`. Would be different if it mentioned adding it in electron.local.

etc/profile-a-l/linuxqq.profile
glitsj16: Thanks! Suggestion implemented.

netblue30:
merged!

PR #5564, opened 2023-01-03
claws-mail and sylpheed D-Bus hardening

glitsj16:


netblue30:
merged!

kmk3 (commented):
> claws-mail and sylpheed D-Bus hardening

The changes seem to only allow more dbus stuff, so why "hardening"?

PR #5579, opened 2023-01-10
solves ansible/ansible#79702 and #1518 

marek22k:
…thub.com/netblue30/firejail/issues/1518 for ansible

Signed-off-by: Marek Küthe <m.k@mk16.de>

### DIFF
--- a/etc/inc/allow-ssh.inc
+++ b/etc/inc/allow-ssh.inc
@@ -14,3 +14,6 @@ noblacklist /usr/lib/ssh
 noblacklist /usr/lib/openssh
 # Fedora and derivatives
 noblacklist /usr/libexec/openssh
+
+# Ansible
+noblacklist ${HOME}/.ansible/cp
--- a/etc/inc/allow-ssh.inc
+++ b/etc/inc/allow-ssh.inc
@@ -2,6 +2,8 @@
 # Persistent customizations should go in a .local file.
 include allow-ssh.local
 
+# Ansible
+noblacklist ${HOME}/.ansible/cp
 noblacklist ${HOME}/.ssh
 noblacklist /etc/ssh
 noblacklist /etc/ssh/ssh_config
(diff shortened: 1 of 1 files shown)


etc/inc/allow-ssh.inc
```diff
@@ -2,6 +2,8 @@
 # Persistent customizations should go in a .local file.
 include allow-ssh.local
 
+# Ansible
+noblacklist ${HOME}/.ansible/cp
```
rusty-snake, line 6: Never blacklisted.

rusty-snake (changes requested):

etc/inc/allow-ssh.inc
marek22k: Could you expand on that a little bit? For example, why is `.ssh` inside, the same logic should apply to `.ssh`. Please note that this file is included by another one, which means that not all blacklists are in this file.

rusty-snake: No file contains a blacklist for `.ansible/cp`, therefore it does not make sense to nobalcklist it. It is a no-op.

`.ssh` is blacklisted (in contrast to `.ansible/cp`), see
https://github.com/netblue30/firejail/blob/a4eb84d5e8165586a59d5e8e30c771fb7f133e2f/etc/inc/disable-common.inc#L444

Having the blacklist and noblacklist in the same file would be nonsense anyway and should never be the case.

Where is the blacklist

rusty-snake (commented):
&nbsp;

etc/inc/allow-ssh.inc
rusty-snake: Maybe check `grep ansible ~/.config/firejail/*`.

marek22k: Ahh, I see. But somehow accessing this file with firejail seems to be problematic. Ansible with SSH does not work at least.

rusty-snake: You can use `--debug` to geht more information. Maybe a symlink? And use firemon to see which sandboxes are started.

PR #5589, opened 2023-01-14
fix: do not show apparmor CLI options in usage if app armor is not supported

nya1:
- If app armor is not supported do not show the apparmor CLI options in the usage info, needed because if any --apparmor* flag is used but apparmor is unsupported an invalid command line option error will be returned:
```
# (AppArmor support is disabled)
firejail --apparmor firefox
Error: invalid --apparmor command line option
```
- updated readme with the correct apparmor enable flag to be used with ./configure

### DIFF
--- a/README
+++ b/README
@@ -26,7 +26,7 @@ $ cd firejail
 $ ./configure && make && sudo make install-strip
 
 On Debian/Ubuntu you will need to install git and gcc compiler. AppArmor
-development libraries and pkg-config are required when using --apparmor
+development libraries and pkg-config are required when using --enable-apparmor
 ./configure option:
 
 $ sudo apt-get install git build-essential libapparmor-dev pkg-config gawk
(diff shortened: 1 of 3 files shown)


reinerh (approved):

kmk3 (commented):
The code LGTM.

---

Nitpicks:

> chore: update readme with flag to enable apparmor on ./configure 

> refactor: do not show apparmor CLI options if app armor is not supported 

I think that both of these would qualify as "docs" changes (as they only change
a markdown file and the output of --help); see the RELNOTES of 0.9.72 for the
prefixes that are usually used in this project:

* <https://github.com/netblue30/firejail/blob/master/RELNOTES>

Also, "refactor" seems misleading to me, as that actually changes the output of
firejail rather than being a code-only change.

Suggestion: Reword them to something like:

> docs: fix ./configure flag to enable apparmor on readme

> docs: remove apparmor options in --help if built without apparmor

(Using "built without apparmor" because "apparmor is not supported" may also
refer to the state of the user's system at runtime)

nya1:
Thank you for the tips, I will follow the styling hints on the next PRs

netblue30:
Merged, thanks for the patch.

Issue #5448, opened 2022-11-02
e-mail clients (like claws-mail) don't use correct timezone


bymoz089:
### Description

E-mail clients show wrong time of received mails and vcalendar. (experienced with claws-mail), due to using timezone `UTC` instead of system configured timezone.

### Steps to Reproduce

     > date +%Z
     CET

     > firejail --quiet --profile=email-common date +%Z
     UTC

### possible fix

adding `localtime` to line `private-etc` in file `email-common.profile`

new line in file `email-common.profile` should look like:

    private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,gcrypt,gnupg,groups,gtk-2.0,gtk-3.0,hostname,hosts,hosts.conf,ld.so.cache,ld.so.preload,machine-id,mailname,nsswitch.conf,passwd,pki,resolv.conf,selinux,ssl,xdg,localtime

### edit

above fix solves displayed time of received e-mails, but not vcalendar entries. claws-mail vcalendar plugin seems to use libical to calculate time.

glitsj16:
Nice find. The `private-etc` fix for the displayed time of received e-mails can be added right now. Can you ceate a PR for that part of the issue? If so, please keep in mind that the option is _alphabetically_ ordered, so tacking on `localtime` at the end of the string is not the way we do things...

I'm not sure there's anything firejail can do about the vcalendar plugin behavior, but we'll see about that later.

bymoz089:
See new PR (https://github.com/netblue30/firejail/pull/5594) to fix vcalendar issue.

PR #5594, opened 2023-01-16
add timezone access to make libical functional

bymoz089:
claws-mail vcalendar-plugin uses libical to get current timezone. Libical needs access to file `/etc/timezone` to work properly.

See (already closed but not yet completely solved) bug report: https://github.com/netblue30/firejail/issues/5448


netblue30:
Merged, thanks!

PR #5591, opened 2023-01-15
opt-in: hide blacklisted files in /etc 

smitsohu:
Closes #5010
Closes #5230

Issue #5230, opened 2022-07-04
/etc is unwritable on --chroot on debootstrap system


rayment:
### Description

Following the documentation at https://firejail.wordpress.com/documentation-2/basic-usage/ concerning the usage of firejail for creating an isolated debootstrap system, I am unable to create a user account and installing packages via. `apt` result in errors when attempting write to /etc.

### Steps to Reproduce

(as root)

1. `emerge -qv firejail`
2. `echo "force-nonewprivs yes" >> /etc/firejail/firejail.config`
3. `mkdir /jail`
4. `debootstrap --arch=amd64 stable /jail`
5. `LC_ALL=C TERM=xterm-color firejail --noprofile --chroot=/jail --shell=/bin/bash` (note this already deviates from the documentation as the provided command will actually fail without setting `--shell`)
6. `adduser foo`

### Expected behavior

```
$ adduser foo
Adding user `foo' ...
Adding new group `foo' (1000) ...
etc. etc.
```

### Actual behavior

```
$ adduser foo
Adding user `foo' ...
Adding new group `foo' (1000) ...
groupadd: failure while writing changes to /etc/group
adduser: `/sbin/groupadd -g 1000 foo' returned error code 10. Exiting.
```

### Additional context

I'm not sure if this is really a bug or merely a configuration error or a lack of concise documentation, but this occurs using the default firejail v0.9.68 as it comes on Gentoo with the only change in configuration being `force-nonewprivs yes` as suggested by the documentation.

Similar commands such as `useradd` or unrelated commands like calling `apt` are also failing:

```
$ useradd foo
useradd: failure while writing changes to /etc/passwd
```

```
$ apt install htop
...
ldconfig: Renaming of /etc/ld.so.cache~ to /etc/ld.so.cache failed: Device or resource busy
dkpg: error processing package libc-bin (--configure):
  installed libc-bin package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
  libc-bin
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

### Environment

- Linux distribution and version (e.g. "Ubuntu 20.04" or "Arch Linux")
```
Linux home 5.15.41-gentoo-x86_64 #1 SMP Thu Jun 30 20:08:43 UTC 2022 x86_64 AMD Ryzen 7 PRO 5850U with Radeon Graphics AuthenticAMD GNU/Linux
```

- Gentoo USE flags
```
[...]

Issue #5010, opened 2022-03-05
Firefox under Firejail can't make new connections after switching network connection methods


alexdelorenzo:
### Description
After using Firefox for a bit on WiFi and then experiencing a network failure, when I go to change my network connection method to Ethernet, I cannot open or refresh pages in Firefox. 

Connections will time out and I'll have to close the browser and open it again in order to load or refresh pages.

### Steps to Reproduce
Open Firefox with Firejail, let it run for a bit. After experiencing a network connection error, change your connection via a separate network device. Go back to Firefox and try to open "google.com". 

### Expected behavior
I should be able to resume using the browser after experiencing a network failure and/or network device change.

### Actual behavior
After network failure/network device change, trying to open new pages or refreshing tabs results in those tabs eventually timing out without loading. The browser must exit and start again for it to work.

### Behavior without a profile
This does not seem to be an error with Firejail-less Chromium or when I've tried to replicate the issue in Firefox without Firejail.

### Additional context


### Environment
- Manjaro unstable
- firejail version 0.9.68                                                                                                                                       


### Checklist

<!--
Note: Items are checked with an "x", like so:

- [x] This is a checked item.
-->

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ ] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

<details>
<summary>Output of <code>LC_ALL=C firejail /path/to/program</code></summary>
<p>

```
$ LC_ALL=C firejail firefox  --new-instance --profile "tmp" --private-window "https://imgur.com" 
Reading profile /etc/firejail/firefox.profile                                                                                                                 
Reading profile ~/.config/firejail/firefox.local                                                                                                     
Reading profile /etc/firejail/whitelist-usr-share-common.inc                                                                                                  
[...]

PR #5595, opened 2023-01-16
Rename etc-no-blacklisted to etc-hide-blacklisted

kmk3:
To avoid boolean confusion (`no-foo no` / `no-foo yes`) in
firejail.config:

    etc-no-blacklisted no
    etc-no-blacklisted yes

    git grep -Ilz -i 'etc.no.blacklisted' -- etc src |
      xargs -0 -I '{}' sh -c "printf '%s\n' \"\$(sed \
        -e 's/etc-no-blacklisted/etc-hide-blacklisted/' \
        -e 's/ETC_NO_BLACKLISTED/ETC_HIDE_BLACKLISTED/' \
        '{}')\" >'{}'"

Added on commit ded50200e ("opt-in: skip blacklisted files in
private-etc - #5010, #5230", 2023-01-15) / PR #5591.


glitsj16 (approved):
LGTM. Potentially less confusing calling the pref this way IMO.

PR #5596, opened 2023-01-16
zoom.profile: whitelist ~/.config/zoom.conf

fkrone:
With Zoom version 5.12.6 Zoom changed how they handle encrypting the local database. This change resulted in the new file zoom.conf being used. As it was not allowed by the default profile this could lead to users losing their chat history if they cannot be retrieved from the cloud (e.g. when e2e encryption is used).

### DIFF
--- a/etc/profile-m-z/zoom.profile
+++ b/etc/profile-m-z/zoom.profile
@@ -22,9 +22,11 @@ noblacklist ${HOME}/.zoom
 nowhitelist ${DOWNLOADS}
 
 mkdir ${HOME}/.cache/zoom
+mkfile ${HOME}/.config/zoom.conf
 mkfile ${HOME}/.config/zoomus.conf
 mkdir ${HOME}/.zoom
 whitelist ${HOME}/.cache/zoom
+whitelist ${HOME}/.config/zoom.conf
 whitelist ${HOME}/.config/zoomus.conf
 whitelist ${HOME}/.zoom
 
--- a/etc/profile-m-z/zoom.profile
+++ b/etc/profile-m-z/zoom.profile
@@ -16,6 +16,7 @@ ignore dbus-system none
 # If you use such a system, add 'ignore nogroups' to your zoom.local.
 #ignore nogroups
 
+noblacklist ${HOME}/.config/zoom.conf
 noblacklist ${HOME}/.config/zoomus.conf
 noblacklist ${HOME}/.zoom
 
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/zoom.profile
```diff
@@ -22,9 +22,11 @@ noblacklist ${HOME}/.zoom
 nowhitelist ${DOWNLOADS}
 
 mkdir ${HOME}/.cache/zoom
+mkfile ${HOME}/.config/zoom.conf
 mkfile ${HOME}/.config/zoomus.conf
 mkdir ${HOME}/.zoom
 whitelist ${HOME}/.cache/zoom
+whitelist ${HOME}/.config/zoom.conf
```
rusty-snake, line 29: No blacklist + noblacklist for it?

fkrone: Tried to keep it in sync with the rest of the profile without having the feeling of completely rewriting it (and maybe messing up in the process).

kmk3 (commented):
This line in the original commit was over 300 characters long:

```
With Zoom version 5.12.6 Zoom changed how they handle encrypting the local database. This change resulted in the new file zoom.conf being used. As it was not allowed by the default profile this could lead to users losing their chat history if they cannot be retrieved from the cloud (e.g. when e2e encryption is used).
```

And the commit message is intentionally not line-wrapped in the git log, which
makes the above line unreadable and also makes it stick out.

So I line-wrapped the commit message to the usual 72 characters per line and
force-pushed.

Please try to follow that in future commits.

Also, note that the GitHub web UI is unsuited for writing longer commit
messages due to the lack of line-wrapping (and due the lack of warning about
the 50/72 standard), so I would suggest using a proper text editor for that.

The next best thing would be to at least write one sentence per line to try to
mitigate the issue.

Anyway, see the following for details:

etc/profile-m-z/zoom.profile
kmk3: To clarify:

* Add `blacklist ${HOME}/.config/zoom.conf` to disable-programs.inc, to prevent
  other programs from accessing it
* Add `noblacklist ${HOME}/.config/zoom.conf` to zoom.profile, to allow zoom to
  access it

See also [profile.template](https://github.com/netblue30/firejail/blob/master/etc/templates/profile.template).

kmk3 (changes requested):
https://github.com/netblue30/firejail/pull/5596/files#r1071404037

kmk3 (approved):
LGTM.

rusty-snake (approved):

glitsj16 (approved):
LGTM

PR #5602, opened 2023-01-18
profanity: fix notifications

mhmdanas:


etc/profile-m-z/profanity.profile
```diff
@@ -46,7 +46,8 @@ private-dev
 private-etc alternatives,ca-certificates,crypto-policies,ld.so.cache,ld.so.preload,localtime,mime.types,nsswitch.conf,pki,resolv.conf,ssl
 private-tmp
 
-dbus-user none
+dbus-user filter
+dbus-user.talk org.freedesktop.Notifications
```
mhmdanas, line 50: Since maybe not everybody uses notifications, should this be commented out (similar to how it's done in `firefox.profile`)?

mhmdanas: Sorry, I didn't quite understand that. Are you saying that the PR is fine as-is, or should I comment these lines out instead?

rusty-snake: In my eyes it's ok to enable it by default.

glitsj16 (approved):
LGTM

PR #5604, opened 2023-01-18
abiword: fix libEGL warning

mhmdanas:
Here's the warning:

> libEGL warning: wayland-egl: could not open /dev/dri/renderD129 (No such file or directory)

Abiword actually seems to work perfectly fine even with this warning, but I'm submitting this PR just as a heads-up and because it's only a one-line deletion. If this change is undesired, please close this PR.

### DIFF
--- a/etc/profile-a-l/abiword.profile
+++ b/etc/profile-a-l/abiword.profile
@@ -24,7 +24,6 @@ apparmor
 caps.drop all
 machine-id
 net none
-no3d
 nodvd
 nogroups
 noinput


kmk3 (commented):
@mhmdanas [on Jan 18](https://github.com/netblue30/firejail/pull/5604#issue-1548252762):

> Here's the warning:
> 
> > libEGL warning: wayland-egl: could not open /dev/dri/renderD129 (No such file or directory)
> 
> Abiword actually seems to work perfectly fine even with this warning, but I'm
> submitting this PR just as a heads-up and because it's only a one-line
> deletion. If this change is undesired, please close this PR.

The warning is fine; I've seen similar warnings in other programs using `no3d`.

I think that unless `no3d` is causing issues, it should stay as is, as it
blocks access to the GPU.

Thanks for the heads up either way.

rusty-snake:
As @kmk3 said there are a lot programs with this warning (or similar one) w/o issues. Maybe all GTK and Qt programs?

Regarding the "only one line", one line can contain 50% of the security. Not in this case but it is a weak argument in general.

mhmdanas:
> Regarding the "only one line", one line can contain 50% of the security. Not in this case but it is a weak argument in general.

It was an argument for opening the PR directly instead of discussing this in an issue first (since in this case it's less effort to just open the PR directly and have it get rejected if the change is undesired). I was not implying that this PR should be merged in merely because it's a one-line deletion.

Issue #5601, opened 2023-01-18
qutebrowser won't open in tabs since last upgrade


aleprovencio:
### Description

Until upgrading `firejail` to this last release, I used to open links on external apps and they would open on a new `qutebrowser` tab.

### Steps to Reproduce

_Steps to reproduce the behavior_

1. Run a firejailed `qutebrowser`
2. Open a link from external app, ex: `kitty`'s `open_url_with_hints`

### Expected behavior

The link would open a new tab of the already running `qutebrowser` instance

### Actual behavior

A new `qutebrowser` instance is opened with the link

### Behavior without a profile

The same happens

### Additional context

I've had to make other adjustments on my `qutebrowser.local` as well after this upgrade, but couldn't figure this one out. I've tried things like `ignore dbus-system none` and `dbus-user.own org.qutebrowser.*` without success.

### Environment

- Arch Linux
- Firejail 0.9.72

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [x] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

<details>
<summary>Output of <code>LC_ALL=C firejail /path/to/program</code></summary>
<p>

```
 LC_ALL=C firejail qutebrowser
Reading profile /etc/firejail/qutebrowser.profile
Reading profile /home/aleprovencio/.config/firejail/qutebrowser.local
Reading profile /etc/firejail/allow-lua.inc
Reading profile /etc/firejail/allow-bin-sh.inc
Reading profile /etc/firejail/allow-python2.inc
Reading profile /etc/firejail/allow-python3.inc
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-programs.inc
[...]

PR #5618, opened 2023-01-25
fix qutebrowser profile

aleprovencio:
Addresses issue #5601

### DIFF
--- a/etc/profile-m-z/qutebrowser.profile
+++ b/etc/profile-m-z/qutebrowser.profile
@@ -27,11 +27,13 @@ include disable-shell.inc
 mkdir ${HOME}/.cache/qutebrowser
 mkdir ${HOME}/.config/qutebrowser
 mkdir ${HOME}/.local/share/qutebrowser
+mkdir ${RUNUSER}/qutebrowser
 whitelist ${DOWNLOADS}
 whitelist ${HOME}/.cache/qutebrowser
 whitelist ${HOME}/.config/qutebrowser
 whitelist ${HOME}/.local/share/qutebrowser
 whitelist /usr/share/qutebrowser
+whitelist ${RUNUSER}/qutebrowser
 include whitelist-common.inc
 include whitelist-run-common.inc
 include whitelist-runuser-common.inc


etc/profile-m-z/qutebrowser.profile
```diff
@@ -27,11 +27,13 @@ include disable-shell.inc
 mkdir ${HOME}/.cache/qutebrowser
 mkdir ${HOME}/.config/qutebrowser
 mkdir ${HOME}/.local/share/qutebrowser
+mkdir ${RUNUSER}/qutebrowser
 whitelist ${DOWNLOADS}
 whitelist ${HOME}/.cache/qutebrowser
 whitelist ${HOME}/.config/qutebrowser
 whitelist ${HOME}/.local/share/qutebrowser
 whitelist /usr/share/qutebrowser
+whitelist ${RUNUSER}/qutebrowser
```
kmk3, line 36:
```suggestion
whitelist ${RUNUSER}/qutebrowser
whitelist /usr/share/qutebrowser
```

Sort.

etc/profile-m-z/qutebrowser.profile
```diff
@@ -27,10 +27,12 @@ include disable-shell.inc
 mkdir ${HOME}/.cache/qutebrowser
 mkdir ${HOME}/.config/qutebrowser
 mkdir ${HOME}/.local/share/qutebrowser
+mkdir ${RUNUSER}/qutebrowser
 whitelist ${DOWNLOADS}
 whitelist ${HOME}/.cache/qutebrowser
 whitelist ${HOME}/.config/qutebrowser
 whitelist ${HOME}/.local/share/qutebrowser
+whitelist ${RUNUSER}/qutebrowser
```
kmk3, line 35: Usually a whitelisted path is also added to disable-programs.inc, to block
access to it by default:

```
blacklist ${RUNUSER}/qutebrowser
```

And then access to it would be granted only to programs that need it (like in
qutebrowser.profile):

```
noblacklist ${RUNUSER}/qutebrowser
```

But I'm not sure about sockets.

Is this used just by qutebrowser or also by the programs that try to open tabs?

If you do the above, can the external programs still open tabs?

aleprovencio: yes, doing the above will keep it working

kmk3 (approved):
LGTM.

rusty-snake (approved):

kmk3:
Merged, thanks!

PR #5502, opened 2022-12-03
Add basic gtksourceview language-spec

rusty-snake:
Tested with org.gnome.TextEditor.

The gtksourceview language-spec hasn't changed between gtksourceview 3, 4 and 5 AFAIK so it should also work on older systems if you copy/link the file in the right places.

PR #5627, opened 2023-01-28
build: auto-generate syntax files

kmk3:
Main changes:

* Use the commands from contrib/vim/syntax/firejail.vim to create
  makefile targets to generate syntax lists in contrib/syntax/lists
* Add contrib/syntax/files/example.in as an example of how to generate
  syntax files
* Generate and add the syntax lists, to make it easier to spot if they
  are properly updated when a new command is added or if their recipes
  also need changes
* Add "syntax" and "contrib" makefile targets
* Generate firejail.vim from firejail.vim.in
* Generate firejail-profile.lang from firejail-profile.lang.in
* Update the manual syntax file steps on the new command checklist on
  CONTRIBUTING.md to use `make syntax` instead
* Make contrib target by default

Relates to #2679 #5502 #5577 #5612.


glitsj16 (approved):
LGTM. Very nice work!

kmk3:
> LGTM. Very nice work!

Thanks!

netblue30:
merged!

Issue #5608, opened 2023-01-21
private-etc rework broke profiles without resolv.conf in private-etc


glitsj16:
After the private-etc rework from https://github.com/netblue30/firejail/commit/c2b6b6b1a348d70b776983051851e42ba66ab271 I noticed a few profiles are now in a broken state. When private-etc is enabled without including `resolv.conf` firejail exits with the below error:

```
Error mount: fs_etc.c:58 fs_resolvconf: No such file or directory
Error: proc xxxx cannot sync with peer: unexpected EOF
Peer yyyy unexpectedly exited with status 1
```

The fix is to add resolv.conf to private-etc. Relates to #5607.

rusty-snake:
8377c54b1d803e2d4a282751652d824376fb371d

PR #5629, opened 2023-01-28
atool: fix private-etc

glitsj16:
Related to https://github.com/netblue30/firejail/discussions/5610.

### DIFF
--- a/etc/profile-a-l/atool.profile
+++ b/etc/profile-a-l/atool.profile
@@ -13,7 +13,7 @@ include allow-perl.inc
 noroot
 
 # without login.defs atool complains and uses UID/GID 1000 by default
-private-etc alternatives,group,ld.so.cache,ld.so.preload,login.defs,passwd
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/atool.profile
```diff
@@ -13,7 +13,7 @@ include allow-perl.inc
 noroot
 
 # without login.defs atool complains and uses UID/GID 1000 by default
-private-etc alternatives,group,ld.so.cache,ld.so.preload,login.defs,passwd
+private-etc alternatives,group,ld.so.cache,ld.so.preload,login.defs,passwd,resolv.conf
```
kmk3, line 16: This profile includes archiver-common.inc, which has `net none`.

So what does this change fix?

netblue30:
merged!

etc/profile-a-l/atool.profile
glitsj16: It's the same issue as #5608. Maybe it doesn't make much sense to have private-etc in an archiver tool in the first place?

PR #5630, opened 2023-01-29
warzone2100: additional paths for Arch Linux

glitsj16:


etc/profile-m-z/warzone2100.profile
```diff
@@ -18,11 +19,13 @@ include disable-programs.inc
 
 mkdir ${HOME}/.warzone2100-3.1
 mkdir ${HOME}/.warzone2100-3.2
+whitelist ${HOME}/.local/share/warzone2100
```
kmk3, line 22: +`mkdir`

glitsj16: The original profile didn't have a `mkdir ${HOME}/.local/share/warzone2100-3.3.0` so I went with the same. But I agree it's better this way. Thanks for the review!

kmk3 (approved):

netblue30:
merged, thanks!

PR #5631, opened 2023-01-29
inkscape: additional hardening and settings saving via D-Bus

glitsj16:


etc/profile-a-l/inkscape.profile
```diff
@@ -28,8 +28,19 @@ include disable-interpreters.inc
 include disable-programs.inc
 include disable-xdg.inc
 
+mkdir ${HOME}/.cache/inkscape
+mkdir ${HOME}/.config/inkscape
+mkdir ${HOME}/.inkscape
```
kmk3, line 33: Are you sure that this path is required?

It looks like the legacy version of ~/.config/inkscape.

I tried running inkscape on Artix and only the other paths were created.

I think it would be good to avoid adding `mkdir` for legacy paths (to avoid
more clutter) unless it causes issues for someone.

kmk3: Sorry, I didn't see that this currently isn't a whitelisting profile.

Anyway, unless the path was changed a long time ago, feel free ignore this.

glitsj16: Fair point/question. I'm actually _not_ sure whether ${HOME}/.inkscape (the legacy path) is still being used on other distributions than Arch & derivatives. It just came up in the context of running the short test for the private-etc rework from #5610. I could dust of an old machine running Ubuntu 16.04 LTS but even that won't give any definitive answer I guess. Perhaps a Debian user can have a look at it. As this is still a draft, I can always wait a bit longer and drop the legacy path later, added it to my firejail notes.

> Sorry, I didn't see that this currently isn't a whitelisting profile.

Well, now I'm confused. My intention was to refactor this inkscape profile as a whitelisting one. What did I overlook?

kmk3: @glitsj16 [on Jan 29](https://github.com/netblue30/firejail/pull/5631#discussion_r1090045269):

> Fair point/question. I'm actually _not_ sure whether ${HOME}/.inkscape (the
> legacy path) is still being used on other distributions than Arch &
> derivatives. It just came up in the context of running the short test for the
> private-etc rework from #5610. I could dust of an old machine running Ubuntu
> 16.04 LTS but even that won't give any definitive answer I guess. Perhaps a
> Debian user can have a look at it. As this is still a draft, I can always
> wait a bit longer and drop the legacy path later, added it to my firejail
> notes.

I did a quick search and it was apparently added in this commit, which shipped
in version 1.0 (2020-05-01):

* <https://gitlab.com/inkscape/inkscape/-/commit/d8d0558e443aeac5414197c9b2e53aeac9db8c3b>
* <https://gitlab.com/inkscape/inkscape/-/tags/INKSCAPE_1_0>

It's not that old I suppose (it was released after Ubuntu 20.04), so I guess it
makes sense to create it.

> > Sorry, I didn't see that this currently isn't a whitelisting profile.
> 
> Well, now I'm confused. My intention was to refactor this inkscape profile as
> a whitelisting one. What did I overlook?

I meant that if this was already a whitelisting profile, there would be no
problem whitelisting only N instead of N+1 more paths in `${HOME}`, as it would
only make more paths available anyway (since whitelisting would already be in
effect).

But since it isn't, missing any paths when adding whitelisting to `${HOME}`
would result in them being written only to tmpfs, risking potential data loss.

glitsj16: Oh okay, I see what you mean now. Thanks for the detailed info!

kmk3 (approved):

netblue30:
Merged, thanks!

PR #5626, opened 2023-01-28
mutt.profile: add ~/.mutthistory & reduce amount of paths created

kmk3:
Commits:

```console
$ git log --reverse --pretty='* %s' master..
* mutt.profile: add ~/.mutthistory
* mutt.profile: stop creating editor/browser paths
* mutt.profile: stop creating config files for other programs
```

Relates to #3849 and #5619.

Cc: @bbhtt @hyder365 (from #3849 and #5619)

PR #5635, opened 2023-01-30
disable-common.inc: add more ro editor/browser paths

kmk3:
Move some paths from mutt.profile and neomutt.profile.

Added on commit 6b9bfad ("Fix python; add read-only to editors/cli
browsers;re-add cache directory", 2020-12-29) / PR #3849.

And also make ~/.config/nano read-only.

Misc: This is a follow-up to #5626.


netblue30:
merged, thanks!

PR #5637, opened 2023-02-03
build: mark most phony targets as such

kmk3:
To improve clarity and to prevent unnecessary filesystem lookups.

Overall, this appears to reduce the amount of implicit rule searches by
~4% for the default build and by ~12% for the "man" target (as an
example):

    $ git checkout master >/dev/null 2>&1
    b55cb6a80 2023-01-31 18:56:42 -0500 testing
    $ ./configure >/dev/null
    $ make clean >/dev/null && make --debug=i -j 4     | grep -F 'Trying implicit' | wc -l
    7101
    $ make clean >/dev/null && make --debug=i -j 4 man | grep -F 'Trying implicit' | wc -l
    1239
    # (with this commit applied)
    $ make clean >/dev/null && make --debug=i -j 4     | grep -F 'Trying implicit' | wc -l
    6793
    $ make clean >/dev/null && make --debug=i -j 4 man | grep -F 'Trying implicit' | wc -l
    1085

Environment: GNU make 4.4-1 on Artix Linux.

Note: The amount lines printed is the same on non-parallel builds (that
is, without `-j 4`).

See commit 2465f9248 ("makefiles: make all, clean and distclean PHONY",
2021-02-12) / PR #4024 for details.

Note: By "most phony targets" I mean all non-path targets except for the
testing targets, which were being changed recently (for example, the
"test-github" target) and so might still be under development.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ mydirs: $(MYDIRS)
 $(MYDIRS):
 	$(MAKE) -C $@
 
+.PHONY: filters
 filters: $(SECCOMP_FILTERS) $(SBOX_APPS_NON_DUMPABLE)
 seccomp: src/fseccomp/fseccomp src/fsec-optimize/fsec-optimize
 	src/fseccomp/fseccomp default seccomp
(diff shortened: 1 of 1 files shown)


netblue30:
merged!

PR #5641, opened 2023-02-05
private-etc fixes

glitsj16:
The private-etc rework is really showing its value in https://github.com/netblue30/firejail/commit/5d0822c52c9a5e631676899e9642911d9143dba8. Easier to spot `oddities` now, like prior typos that stayed undetected.

Relates to #5610.

### DIFF
--- a/etc/profile-m-z/whois.profile
+++ b/etc/profile-m-z/whois.profile
@@ -46,7 +46,7 @@ private
 private-bin bash,sh,whois
 private-cache
 private-dev
-private-etc jwhois.conf,services,whois.conf
+private-etc services,whois.conf
 private-lib gconv
 private-tmp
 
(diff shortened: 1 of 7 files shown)


netblue30:
all merged, thanks!

etc/profile-m-z/whois.profile
```diff
@@ -46,7 +46,7 @@ private
 private-bin bash,sh,whois
 private-cache
 private-dev
-private-etc jwhois.conf,services,whois.conf
+private-etc services,whois.conf
```
rusty-snake, line 49: That's not a typo, there is a whois implementation called jwhois.

glitsj16: Thanks for catching this. TIL :+1: . PR to resolve is in.

PR #5643, opened 2023-02-06
whois: re-fix private-etc

glitsj16:
Thanks @rusty-snake for [pointing out](https://github.com/netblue30/firejail/pull/5641#discussion_r1097637201) jwhois actually is a thing.

### DIFF
--- a/etc/profile-m-z/whois.profile
+++ b/etc/profile-m-z/whois.profile
@@ -46,7 +46,7 @@ private
 private-bin bash,sh,whois
 private-cache
 private-dev
-private-etc services,whois.conf
+private-etc jwhois.conf,services,whois.conf
 private-lib gconv
 private-tmp
 


kmk3 (changes requested):
If the changes are reverted as mentioned [here][1], this PR won't be needed.

[1]: https://github.com/netblue30/firejail/discussions/5610#discussioncomment-4883096

netblue30:
No problem, let's grab it as is. There could be more changes coming.

kmk3:
@netblue30 [on Feb 6](https://github.com/netblue30/firejail/pull/5643#issuecomment-1419528962):

> No problem, let's grab it as is. There could be more changes coming.

It is a problem actually: It makes it harder to revert only the relevant
changes, which is already very difficult due to some of the affected commits
changing both the profiles and the source code in the same commit and because
there is a merge of master into itself.

PR #5646, opened 2023-02-07
New profile: parsecd

NetSysFire:
Upstream website: https://parsec.app
It is partly open source, see https://github.com/chrisd1100/libmatoya

Directives that wont work:

- private-dev - it needs that for the passthrough of input devices
- noinput, novideo, noaudio, no3d - because it needs all that
- netfilter - would need custom rules because it connects to STUN stuff
- private-cache - because it does not need any cache and having just ~/.parsec is enough
- the no-exec include - it has executable .so stuff in ~/.parsec for whatever reason and not just in /usr/share/parsec

Did not try seccomp because I am not in the mood to debug segfaults due to missing syscalls, but it otherwise works perfectly with the given profile. The profile is probably missing comments, too.

### DIFF
--- a/etc/profile-m-z/parsecd.profile
+++ b/etc/profile-m-z/parsecd.profile
@@ -7,12 +7,25 @@ include parsecd.local
 include globals.local
 
 noblacklist ${HOME}/.parsec
+ignore noexec ${HOME}
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
 
 mkdir ${HOME}/.parsec
 whitelist ${HOME}/.parsec
 whitelist /usr/share/parsec
 include whitelist-common.inc
 include whitelist-usr-share-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-var-common.inc
 
 # Due to the nature of parsec, the following directives will not work:
 # - no3d
--- a/etc/profile-m-z/parsecd.profile
+++ b/etc/profile-m-z/parsecd.profile
@@ -1,0 +1,44 @@
+# Firejail profile for Parsec
+# Description: Remote desktop application focused on gaming and other 3D applications
+# This file is overwritten after every install/update
+# Persistent local customizations
+include parsecd.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.parsec
+
+mkdir ${HOME}/.parsec
+whitelist ${HOME}/.parsec
+whitelist /usr/share/parsec
+include whitelist-common.inc
+include whitelist-usr-share-common.inc
+
+# Due to the nature of parsec, the following directives will not work:
+# - no3d
+# - novideo
+# - nosound
+# - noinput (it does remote passthrough stuff for gamepads)
+# - private-dev (because of the above)
+apparmor
+caps.drop all
+nodvd
+nogroups
+nonewprivs
+notv
+nou2f
+noroot
+# Will fail to start with mty_evdev_create: 'udev_monitor_new_from_netlink' failed without netlink
+protocol unix,inet,inet6,netlink
+seccomp !tgkill
+seccomp.block-secondary
+
+# Will not start with zenity missing
+private-bin parsecd,zenity
+private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+restrict-namespaces
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -1077,6 +1077,7 @@ blacklist ${HOME}/.ostrichriders
 blacklist ${HOME}/.paradoxinteractive
 blacklist ${HOME}/.paradoxlauncher
 blacklist ${HOME}/.parallelrealities/blobwars
+blacklist ${HOME}/.parsec
 blacklist ${HOME}/.pcsxr
 blacklist ${HOME}/.penguin-command
 blacklist ${HOME}/.pine-crash


NetSysFire:
I think the Profile Checks is failing because of the nextcloud profile and not my addition

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,22 @@
+include parsecd.local
```
glitsj16, line 1: Please add the headers we use in all .profile files. There's a template for that [here](https://github.com/netblue30/firejail/blob/master/etc/templates/profile.template).

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,22 @@
...
+include globals.local
+
+whitelist ${HOME}/.parsec
```
glitsj16, line 4: To protect a program's configuration path, it's necessary to (1) blacklist it in `disable-programs.inc`, (2) noblacklist it in the program's dedicated profile and (3) then proceed with any whitelisting. Again, use the template for guidance here.

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,22 @@
...
+whitelist /usr/share/parsec
+include whitelist-common.inc
+include whitelist-usr-share-common.inc
+
+apparmor
```
glitsj16, line 9: It's fine and very well you've tested which options work and which don't, like you mention in the accompanying explanations. To keep such valuable info we indeed use comments. Please add those as much as you can.

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,22 @@
...
+nodvd
+nonewprivs
+nogroups
+notv
+nou2f
+noroot
+
+memory-deny-write-execute
```
glitsj16, line 17: Move this at the bottom cfr. the profile template.

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,22 @@
...
+include whitelist-usr-share-common.inc
+
+apparmor
+nodvd
+nonewprivs
+nogroups
+notv
+nou2f
+noroot
+
```
glitsj16, line 16: `seccomp` is a very powerful and important option for creating the sandbox. Even though I can feel your pain in having to go through the time-consuming work that's sometimes needed in this regard, it is what needs to be done :-) Take all the time you need for doing so and ask for help when something's not clear.

glitsj16:
> I think the Profile Checks is failing because of the nextcloud profile and not my addition

Don't worry about CI for now, that happens more often than not. It could be nextcloud, it could also be the big changes we're going through for the private-etc overhaul (cfr. #5610).

etc/profile-m-z/parsecd.profile
NetSysFire: I do not see the need to add a comment here, as I added this in the same way I added private-tmp and both work perfectly and there is nothing to note about this.

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
```
kmk3, line 1:
```suggestion
```

rusty-snake, line 1: firecfg?

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+# Firejail profile for Parsec
+# Description: Remote desktop application focused on gaming and other 3D applications
+# This file is overwritten after every install/update
+
```
rusty-snake, line 5:
```suggestion
```

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+# Persistent local customizations
+include parsecd.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.parsec
```
rusty-snake, line 11:
```suggestion
noblacklist ${HOME}/.parsec

```

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+whitelist ${HOME}/.parsec
```
rusty-snake, line 12: `mkdir` it

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+whitelist /usr/share/parsec
+include whitelist-common.inc
+include whitelist-usr-share-common.inc
```
rusty-snake, line 15: What about more `whitelist-*-common.inc`?

etc/profile-m-z/parsecd.profile
rusty-snake, line 11: What about `include disable-*.inc`?

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+
+# Due to the nature of parsec, the following directives will not work:
+# - no3d
+# - novideo
+# - noaudio
```
rusty-snake, line 20:
```suggestion
# - nosound
```

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+# - noinput (it does remote passthrough stuff for gamepads)
+# - private-dev (because of the above)
```
rusty-snake, line 22: Actually it should work if you plug it in before you start parsecd.

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+apparmor
```
rusty-snake, line 23: You said it needs executable `$HOME`, but here you set `apparmor`?

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+nodvd
```
rusty-snake, line 24: `caps.drop all`?

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+nonewprivs
+nogroups
```
rusty-snake, line 26:
```suggestion
nogroups
nonewprivs
```

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+notv
+nou2f
+noroot
```
rusty-snake, line 29: `protocol`?

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,38 @@
...
+
+# Will not start with zenity missing
+private-bin parsecd,zenity
+private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
```
rusty-snake, line 38: `restrict-namespaces`?

rusty-snake, line 38: Is it a GTK program?

rusty-snake (commented):
Already good, here are some comments from mine.

etc/profile-m-z/parsecd.profile
NetSysFire: No, paste error in github editor

etc/profile-m-z/parsecd.profile
NetSysFire: Why? It has no filesharing capability and does not need access to more paths

etc/profile-m-z/parsecd.profile
NetSysFire: Unsure, quite probably not. It does some custom things afaik.

etc/profile-m-z/parsecd.profile
NetSysFire: include disable-exec.inc will break it, but apparmor will not. I have no idea why. The  thing in question in ~/.parsec: `parsecd-150-86e.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=1a9684ced0e2b6198904b4fc070d8ba09db27fcf, not stripped`

etc/profile-m-z/parsecd.profile
NetSysFire: Parsec will detect the devices on demand (when you press any button on the gamepad) and I am fairly sure this will break it

etc/profile-m-z/parsecd.profile
rusty-snake: Do you use AA? Is it enabled in your kernel? And the firejail profile is loaded?

etc/profile-m-z/parsecd.profile
rusty-snake: Greate, then let's restict it

```
```

etc/profile-m-z/parsecd.profile
NetSysFire: There are only some minor json configs, binary user and device IDs and the .so thing I mentioned below in that dir. I do not think the blacklists will do much good here because there is simply nothing there. Do you have a specific thing in mind here?

rusty-snake: I don't talk about the noblacklist, just needed to put the comment somewhere.

```
ignore noexec ${HOME}

include disable-common.inc
include disable-devel.inc
include disable-exec.inc
include disable-proc.inc
include disable-programs.inc
include disable-shell.inc
include disable-xdg.inc
```

etc/profile-m-z/parsecd.profile
NetSysFire: Yes I do.

```
apparmor module is loaded.
63 profiles are loaded.
63 profiles are in enforce mode.
   /usr/local/bin/parsecd (1070703) firejail-default//&unconfined   /usr/local/bin/parsecd (1070703) firejail-default//&unconfined
```

rusty-snake: Interesting

glitsj16: I've seen and used this AA 'trickery' before in combination with Firejail. BUT I have enabled stuff in /etc/apparmor.d/local/firejail-default for doing so:

`
owner @{HOME}/some/path/** ix,
`

Perhaps @NetSysFire uses a similar setup?

NetSysFire: No. I never touched firejail-default.

etc/profile-m-z/parsecd.profile
NetSysFire: Hm, since it doesnt need those paths, can we just blacklist /var entirely?

etc/profile-m-z/parsecd.profile
NetSysFire: This went much smoother than anticipated, only one syscall was missing: 243, aka tgkill and parsec handled that very gracefully.

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,43 @@
...
+caps.drop all
+nodvd
+nogroups
+nonewprivs
+notv
+nou2f
+noroot
+# Will fail to start with mty_evdev_create: 'udev_monitor_new_from_netlink' failed without netlink
+protocol unix,inet,inet6,netlink
+seccomp !tgkill
```
rusty-snake, line 33:
```suggestion
seccomp !tgkill
seccomp.block-secondary
```

etc/profile-m-z/parsecd.profile
```diff
@@ -0,0 +1,43 @@
...
+
+# Will not start with zenity missing
+private-bin parsecd,zenity
+private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+restrict-namespaces1
```
rusty-snake, line 43:
```suggestion
restrict-namespaces
```

NetSysFire: Ouch that was a copy paste error lol

etc/profile-m-z/parsecd.profile
NetSysFire: I can apply this without testing because I already globally filter non-native syscalls with systemd:

```
[Manager]
SystemCallArchitectures=native
```

etc/profile-m-z/parsecd.profile
rusty-snake: Correction, it doesn't need those paths *on your system*. Some of those paths are dependent on distribution/configuration/hardware/...

etc/profile-m-z/parsecd.profile
rusty-snake: I meant should it be added to firecfg? 

Could have been clearer here but if you ask the same question for every second new profile you become lazy.

NetSysFire: I dont see a reason not to. It works perfectly.

etc/profile-m-z/parsecd.profile
NetSysFire: This is correct but I dont see parsec looking for those paths in the strace for example

etc/profile-m-z/parsecd.profile
NetSysFire: I noticed when ctrl+c'ing the application it complains of

```
mty_sleep: 'nanosleep' failed with errno 4
```

and in the journal:

```
Feb 11 17:59:16 archlinux kernel: audit: type=1300 audit(1676134756.097:498): arch=c000003e syscall=234 success=no exit=-13 a0=c a1=c a2=f a3=8 items=0 ppid=7627 pid=7639 auid=1000 uid=1000 gid=1000 euid=1000 suid=1000 fsuid=1000 egid=1000 sgid=1000 fsgid=1000 tty=pts3 ses=2 comm="parsecd" exe="/usr/local/bin/parsecd" subj=firejail-default//&unconfined key=(null)
Feb 11 17:59:16 archlinux kernel: audit: type=1400 audit(1676134756.097:498): apparmor="DENIED" operation="signal" profile="firejail-default" pid=7639 comm="parsecd" requested_mask="send" denied_mask="send" signal=term peer="firejail-default//&unconfined"
```

But i explicitely noblacklisted that syscall with !tgkill

rusty-snake: seccomp != AppArmor

Even if you all a syscall in the seccomp filter, AppArmor can still deny it because it is an independent of seccomp.

NetSysFire: How do I make my seccomp changes consistent with those AppArmor uses?

glitsj16 (approved):

glitsj16:
@kmk3 @rusty-snake Anything left to do here or can we proceed to bring this in?

NetSysFire:
The include disable-* thing and some more whitelisting is still missing. I actually tested this one but I was too lazy to push here because parsecd works perfectly and I was slacking off playing Elden Ring with it :p

kmk3:
@glitsj16 [on Feb 23](https://github.com/netblue30/firejail/pull/5646#issuecomment-1441690849):

> Anything left to do here or can we proceed to bring this in?

As @NetSysFire [mentioned][1], the disable/whitelist includes.

Relevant comments:

whitelist includes:

* <https://github.com/netblue30/firejail/pull/5646#discussion_r1099075600>

disable includes:

* <https://github.com/netblue30/firejail/pull/5646#discussion_r1099079907>

[1]: https://github.com/netblue30/firejail/pull/5646#issuecomment-1441865409

glitsj16:
Thanks for the update guys. Enjoy playing :+1:.

NetSysFire:
There, finally added the missing stuff I was too lazy to push.

One last thing is remaining though: `apparmor` and `!tgkill` in seccomp, see the last unresolved conversation. Any idea how to resolve that? It looks like it will work nonetheless but it is certainly not ideal.

netblue30:
all merged, thanks!

Issue #2877, opened 2019-07-29
firejail inside firejail broken


rusty-snake:
## preparation

 * install firejail from git
 * `sudo firecfg`

## syntax

`$`: regular user shell
`%`: shell inside firejail
`>`: comment
`$PRG$`: any program with a firecfg symlink

## Issue 1

_Outer firejail with `private-etc` without `passwd`._

```
$ firejail --private-etc="empty" --noprofile bash
% $PRG$
Error getpwuid: main.c:299 init_cfg: No such file or directory
% firejail echo
Error getpwuid: main.c:299 init_cfg: No such file or directory
```

## Issue 2

Let's fix 1 and add `passwd` to `private-etc`.

_Outer firejail with `private-etc` without `login.defs`._

> adding `firejail` to `private-etc` because of 3.

```
$ firejail --private-etc="passwd,firejail" --noprofile bash
% firejail echo "Hello, Wolrd!"
Error: cannot read UID_MIN and/or GID_MIN from /etc/login.defs, using 1000 by default
Warning: an existing sandbox was detected. echo will run without any additional sandboxing features
Hello, Wolrd!
> Ugly message, but works.
```

## Issue 3

```
$ firejail --private-etc="passwd,login.defs" --noprofile bash
% $PRG$
Error: Firejail configuration file /etc/firejail/firejail.config not found
% firejail echo "Hello, World!"
Error: Firejail configuration file /etc/firejail/firejail.config not found
```

Bad commit: 8bff773d @smitsohu 

### 3.1

```
$ firejail --private-etc="passwd,login.defs" --noprofile $PRG$
Error: Firejail configuration file /etc/firejail/firejail.config not found
```
Workaround: `private-bin`

```
$ firejail --private-etc="passwd,login.defs" --noprofile --shell=none --private-bin="$PRG$" $PRG$
```
or
```
$ firejail --private-etc="passwd,login.defs" --noprofile --private-bin="$PRG$,SHELL" $PRG$
```

netblue30:
I guess it was broken for some time now, thanks!

smitsohu:
> Bad commit: 8bff773 @smitsohu

Regarding this commit specifically, we maybe can just revert most of it. The proposed alternative (updating /etc/passwd) is the cleaner solution anyway.

kmk3:
> ```
> Error: Firejail configuration file /etc/firejail/firejail.config not found
> ```

The following PR might fix at least that error (as reported in #6966):

* #6970

PR #5648, opened 2023-02-08
profiles: hardening and self-ref fixes

glitsj16:
We have 2 profiles that refer to `firejail` in private-etc (ghostwriter and peek), and one (peek) additionally includes it private-bin. IMO there is never a legitimate reason for this. This PR removes firejail from both profiles.


glitsj16:
@rusty-snake I noticed the comment in ghostwriter.profile on #2877. If it's not too much trouble, can you take a look/review this from that perspective?

etc/profile-a-l/ghostwriter.profile
```diff
@@ -21,10 +21,17 @@ include disable-programs.inc
 include disable-shell.inc
 include disable-xdg.inc
 
+whitelist ${HOME}/.config/ghostwriter
+whitelist ${HOME}/.local/share/ghostwriter
+whitelist ${DOCUMENTS}
+whitelist ${DOWNLOADS}
+whitelist ${PICTURES}
 whitelist /usr/share/ghostwriter
+whitelist  /usr/share/mathjax
```
rusty-snake, line 30:
```suggestion
whitelist  usr/share/mathjax
```

etc/profile-a-l/ghostwriter.profile
```diff
@@ -47,14 +54,15 @@ seccomp !chroot
...
+private-bin affixcompress,analyze,chmorph,cmark,context,gettext,ghostwriter,hunspell,hunzip,hzip,latex,makealias,mktexfmt,munch,multimarkdown,pandoc,pdflatex,pdfroff,prince,unmunch,weasyprint,wkhtmltopdf,wordforms,wordlist2hunspell
 private-cache
 private-dev
 # passwd,login.defs,firejail are a temporary workaround for #2877 and can be removed once it is fixed
-private-etc @tls-ca,@x11,dbus-1,firejail,gconf,host.conf,mime.types,rpc,services,texlive
+private-etc @tls-ca,@x11,dbus-1,gconf,host.conf,mime.types,rpc,services,texlive
 private-tmp
 
 dbus-user filter
+dbus-user.talk ca.desrt.dconf
```
rusty-snake, line 65: It is a Qt program. Why alloe dconf?

glitsj16: Nice catch, thanks! Probably I mixed things up with the peek profile. Anyway, I took it out.

etc/profile-a-l/ghostwriter.profile
rusty-snake: Or I just had the cursor of by one when hitting backspace.

netblue30:
Merged, thanks!

PR #5649, opened 2023-02-08
qutebrowser: allow userscripts by default

### LINKED ISSUE #5639: Qutebrowser profile breaks userscripts
As described [here](https://github.com/qutebrowser/qutebrowser/issues/7575) , one of the recent changes to the Qutebrowser profile, most likely this pull request https://github.com/netblue30/firejail/pull/5389, breaks the invocation of qutebrowser userscripts. I have been able to circumvent this problem by setting 

```
ignore apparmor
ignore noexec ${HOME}
```
in qutebrowser.local . I believe userscripts should work by default. Perhaps there is a more fine-grained way of getting the functionality back?

glitsj16:
Fixes #5639.

### DIFF
--- a/etc/profile-m-z/qutebrowser.profile
+++ b/etc/profile-m-z/qutebrowser.profile
@@ -18,6 +18,8 @@ include allow-bin-sh.inc
 include allow-python2.inc
 include allow-python3.inc
 
+ignore noexec ${HOME}
+
 include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/qutebrowser.profile
```diff
@@ -18,6 +18,8 @@ include allow-bin-sh.inc
 include allow-python2.inc
 include allow-python3.inc
 
+ignore noexec ${HOME}
+
```
rusty-snake, line 22: The profile has apparmor?

glitsj16: Yes indeed, qutebrowser.profile has apparmor. Wasn't sure how OP will test what I suggested [here](https://github.com/netblue30/firejail/issues/5639#issuecomment-1419285365). But you're correct, we need to take out apparmor too probably.

kmk3 (approved):

PR #5651, opened 2023-02-08
profiles: gnome-logs: fix missing machine-id in private-etc

glitsj16:
Relates to #5610.

Side-note: although `private-etc machine-id` also fixes this, I opted to go with the 'new' groups syntax to avoid potential mishaps in future work on #5610 (e.g. etc-cleanup seeing machine-id again).

### DIFF
--- a/etc/profile-a-l/gnome-logs.profile
+++ b/etc/profile-a-l/gnome-logs.profile
@@ -39,7 +39,7 @@ disable-mnt
 private-bin gnome-logs
 private-cache
 private-dev
-private-etc
+private-etc machine-id
 private-lib gdk-pixbuf-2.*,gio,gvfs/libgvfscommon.so,libgconf-2.so.*,librsvg-2.so.*
 private-tmp
 writable-var-log


SkewedZeppelin:
don't all programs with dbus need machine-id?

glitsj16:
> don't all programs with dbus need machine-id?

If you mean programs that use the D-Bus filtering, no not that I'm aware of. Good and interesting question though. I saw your question just before popping out for a walk, and I'll have to check this more thoroughly to be 100% sure my asumptions are valid. But I ran some quick check op apps I use daily and those that have dbus-user filter _and_ don't have machine-id in private-etc seem to work just fine: artha, claws-mail, evince, firefox, freetube (others I didn't re-check yet but I will do so later).

I think gnome-logs is a bit of an odd-one-out. I straced it before and I think it needs machine-id because systemd uses that for (checking) its user access. It wasn't entirely broken without machine-id, but it shows a large red notification widget stating the current user doesn't have the proper privileges to read the logs. Yet it shows those nonetheless :)

Thanks for the hint!

etc/profile-a-l/gnome-logs.profile
```diff
@@ -39,7 +39,7 @@ disable-mnt
 private-bin gnome-logs
 private-cache
 private-dev
-private-etc
+private-etc @sound
```
kmk3, line 42:
```suggestion
private-etc machine-id
```

`@sound` seems rather confusing, as this has `nosound` and it only really needs
`machine-id` for the fix.

Either way, this likely conflicts with #5645 (which also fixes the issue), so
I'd wait for that instead.

glitsj16: I don't have any problem waiting for #5645 to fix this. If it would make that any easier, I can close here. Feel free to suggest what you think is best.

kmk3: @glitsj16 [on Feb 11](https://github.com/netblue30/firejail/pull/5651#discussion_r1103710300):

> I don't have any problem waiting for #5645 to fix this.

:+1:

> If it would make that any easier, I can close here. Feel free to suggest what
> you think is best.

I don't think it needs closing, but to note the conflict I think it would make
sense to mark it as a draft.

kmk3:
This was already added in:

* #5996

PR #5653, opened 2023-02-09
Blacklist sendgmail config

slowpeek:
sendgmail is a cli tool by Google that "uses Gmail in order to mimic sendmail for git send-email" as per its own description. In other words it is a basic sendmail replacement with OAuth2 support to send emails from Gmail accounts.

https://github.com/google/gmail-oauth2-tools/tree/master/go/sendgmail

Config files location depends on 'xdg' build tag. Without the tag it would be '~/.sendgmail.*'. With the tag it is either under $XDG_CONFIG_HOME if set or '~/.config/sendgmail' otherwise.


etc/inc/disable-programs.inc
```diff
@@ -1107,6 +1108,7 @@ blacklist ${HOME}/.sbt
 blacklist ${HOME}/.scorched3d
 blacklist ${HOME}/.scribus
 blacklist ${HOME}/.scribusrc
+blacklist ${HOME}/.sendgmail.*
```
kmk3, line 1111:
```suggestion
blacklist ${HOME}/.sendgmail.json
```

> https://github.com/google/gmail-oauth2-tools/tree/master/go/sendgmail

That page only mentions ~/.sendgmail.json in the root of the home diectory.

Is the wildcard intended to match anything else specifically?

If not, I think it would be better to be explicit.

slowpeek: There are per-account configs as well like `.sendgmail.some@email.json` created with `-setup` option. https://github.com/google/gmail-oauth2-tools/blob/e6adc9e1d2fb66344295449e73c71b9008fddd51/go/sendgmail/paths.go#L38

kmk3 (approved):
LGTM.

PR #5654, opened 2023-02-10
build: mkdeb.sh: pass all arguments to ./configure

kmk3:
Instead of using the first argument as the `EXTRA_VERSION` variable.
This should make the usage of mkdeb.sh less confusing, especially when
one is not trying to set the variable.

As for using `EXTRA_VERSION` (which is still optional with this commit),
make sure that it is set as an environment variable before caling
mkdeb.sh.  Example:

    env EXTRA_VERSION=-apparmor ./mkdeb.sh --enable-apparmor

See also commit 9a0fbbd71 ("mkdeb.sh.in: pass remaining arguments to
./configure", 2022-05-13) / PR #5154.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -341,7 +341,7 @@ deb: dist config.sh
 
 .PHONY: deb-apparmor
 deb-apparmor: dist config.sh
-	./mkdeb.sh -apparmor --enable-apparmor
+	env EXTRA_VERSION=-apparmor ./mkdeb.sh --enable-apparmor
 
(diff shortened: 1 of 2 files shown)


kmk3:
Misc: This branch was based on #5219 and I think the original idea was to post
it a bit after #5219.

Also, there is a branch that removes `deb-apparmor` which was built on top of
this one and is intended as the next PR.

netblue30:
All merged, thanks!

PR #5655, opened 2023-02-11
archivers: drop private-etc now that it's in archiver-common

glitsj16:
Commit https://github.com/netblue30/firejail/commit/29da82d08aab5120262f42032fed6dc7807e5b0d added `private-etc` to `archiver-common.profile`. To avoid doubled options this PR removes it from archiver profiles which already had it.

Relates to #5610.

### DIFF
--- a/etc/profile-a-l/atool.profile
+++ b/etc/profile-a-l/atool.profile
@@ -12,8 +12,6 @@ include allow-perl.inc
 
 noroot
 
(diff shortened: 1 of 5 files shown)


kmk3:
(The force-pushes in this and related PRs were done to squash the commits)

kmk3 (approved):

PR #5660, opened 2023-02-13
transmission-cli: allow web client

glitsj16:


etc/profile-m-z/transmission-cli.profile
```diff
@@ -10,5 +10,7 @@ include globals.local
 private-bin transmission-cli
 private-etc @tls-ca
 
+whitelist /usr/share/transmission
+
```
kmk3, line 14: Suggestion: Move it above `private-` commands as per profile.template.

glitsj16: Thx, moved it.

kmk3 (approved):

netblue30:
All merged, thanks!

PR #5663, opened 2023-02-14
profiles: disable-common: prevent access to LUKS keyfile

netcarver:
Ref: https://github.com/netblue30/firejail/discussions/5662


netblue30:
All merged, thanks!

PR #5664, opened 2023-02-14
Update copyright to 2023

davidfetter:
As title

### DIFF
--- a/README
+++ b/README
@@ -1246,4 +1246,4 @@ Zack Weinberg (https://github.com/zackw)
 zupatisc (https://github.com/zupatisc)
 	- patch-util fix
 
-Copyright (C) 2014-2022 Firejail Authors
+Copyright (C) 2014-2023 Firejail Authors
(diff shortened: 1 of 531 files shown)


glitsj16 (approved):
Nice work! Thanks.

kmk3 (approved):
LGTM.

kmk3:
I tried doing the same thing to verify the PR:

```sh
git grep -Ilz 'Copyright .*-2022 Firejail Authors' |
xargs -0 -I '{}' sh -c "printf '%s\n' \"\$(sed -E \
  's/(Copyright .*)-2022 (Firejail Authors)/\1-2023 \2/' '{}')\" >'{}'"
```

It only missed 3 files:

* src/firejail/selinux.c: It should probably have 2 separate copyright lines
* src/include/gcov_wrapper.h: The years are inconsistent; I'll fix it later
* src/man/preproc.awk: Arguably it should not be bumped automatically, as it
  has a different copyright author and it was not changed in 2 years

I think I'll add something similar to the code above as a script or as a make
target later.

By the way, how did you update the copyright?

davidfetter:
Nothing super clever. First pass with something along the lines of
```
rg -l0 '-2022 Firejail Authors' | xargs -0 sed -i 's/-2022 Firejail Authors/-2023 Firejail Authors/'
```
That got 529 files. I poked around a little bit for others, found two, clearly missed some.

PR #5668, opened 2023-02-17
build: deb: enable apparmor by default & remove deb-apparmor

kmk3:
The official .deb package is always built with apparmor support, so use
`--enable-apparmor` in mkdeb.sh and remove the "deb-apparmor" target in
order to reduce redundancy.

Note that custom configure options may be specified by calling
./mkdeb.sh directly.

For example, to build the .deb package without apparmor support, instead
of running `make deb`, the following commands can be used:

    make dist
    ./mkdeb.sh --disable-apparmor

Also, change the `build_apparmor` GitLab CI job into
`build_no_apparmor`, which is intended to check that building without
apparmor still works.

Note: This commit makes the resulting .deb package not have an
"-apparmor" suffix (see `EXTRA_VERSION` in mkdeb.sh), to avoid
redundancy (as having apparmor support becomes the default).

Misc: This is a follow-up to #5654.

Relates to #5154 #5176 #5547.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,7 +62,7 @@ jobs:
     - name: install dependencies
       run: sudo apt-get install gcc-12 libapparmor-dev libselinux1-dev expect xzdec
     - name: configure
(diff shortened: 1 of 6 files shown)


netblue30:
all in, thanks!

PR #5674, opened 2023-02-20
build: Fix whitespace and add .editorconfig

kmk3:
Using .editorconfig should make it easier to avoid whitespace errors, as long
as the editor used supports it (either natively or through a plugin).

See the editorconfig website for the editors that support it:

* https://editorconfig.org

Relates to #4572 #4575.

Cc: @a1346054 @reinerh (from #4572 #4575)


glitsj16 (approved):
LGTM. Very nice work!

kmk3:
@smitsohu

By the way, `indent_size = 8` only sets the default width of the tab character;
it should still be possible to set its visual width in the editor.

I think that in vim it would be something like `set sw=4 ts=4 noet` (not sure)
to keep outputting tabs while displaying them as 4 characters wide, for
example.

See also the `tab_width` option in editorconfig:

* <https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#tab_width>

@glitsj16 [on Feb 20](https://github.com/netblue30/firejail/pull/5674#pullrequestreview-1306353837):

> LGTM. Very nice work!

Thanks!

netblue30:
all merged!

test/profiles/test2.profile
```diff
@@ -1,6 +1,6 @@
 caps
 seccomp
-  private
-  include test.profile
-  include test.local
-  include test25.profile
+private
+include test.profile
+include test.local
+include test25.profile
```
rusty-snake, line 6: I think this is intentional to test if firejail can parse commands with leading spaces.

.gitattributes
```diff
@@ -1 +1,6 @@
+* text=auto eol=lf
+
+*.AppImage -text
+*.zip -text
```
rusty-snake, line 4: `test/filters/memwrexe` and other in this directory aren't text either.

.github/pull_request_template.md
```diff
@@ -1,17 +1,17 @@
...
-   | Arch/Fedora | `/usr/lib64/firejail/sort.py` |
-   | Debian/Ubuntu/Mint | `/usr/lib/x86_64-linux-gnu/firejail/sort.py` |
-   | local git clone | `contrib/sort.py` |
+| Distro | Path |
+| ------ | ---- |
+| Arch/Fedora | `/usr/lib64/firejail/sort.py` |
+| Debian/Ubuntu/Mint | `/usr/lib/x86_64-linux-gnu/firejail/sort.py` |
+| local git clone | `contrib/sort.py` |
 
-   Note also that the sort.py script exists only since firejail `0.9.61`.
```
rusty-snake, line 15: Why moving this out of the list element?

rusty-snake (commented):
Sorry for the late review. But I have a few things to note.

test/profiles/test2.profile
kmk3: > I think this is intentional to test if firejail can parse commands with
> leading spaces.

Good catch; fixed on commit bdb4f03ed ("test2.profile: restore leading spaces",
2023-02-25).

.gitattributes
kmk3: > `test/filters/memwrexe` and other in this directory aren't text either.

Please see the explanation on commit 77e9e4938 (".gitattributes: mark some
binary files as such", 2023-02-20).

.github/pull_request_template.md
kmk3: > Why moving this out of the list element?

Changes undone on commit be5e5f5cd ("pull_request_template.md: restore
indentation", 2023-02-25).

Issue #639, opened 2016-07-17
allow empty `--protocol=` list


vn971:
Could an empty protocol list be allowed for firejail?
Like this: `firejail --noprofile --protocol=`
I think it may make sense in some scenarions. If it's not too difficult, please add such an ability?

(Sorry, I don't feel myself confident with C, therefore I can only raise issues & make some tests.. No code changes since I'd probably leak something or worse.)

netblue30:
I'll bring a fix in this week, thanks!

Fred-Barclay:
I'm afraid I'm in the dark here (which is my fault) and I can't quite figure out what this would accomplish. Would `--protocol=` be the same as `--protocol=*` i.e. would it allow **all** protocols (unix, inet, inet6...)? In which case I would suggest either `--protocol` or (preferably) `--protocol=all` as the syntax. To me at any rate, `--protocol=` looks like it's missing an argument on the end. :stuck_out_tongue_winking_eye: 

Or would it deny access to any protocols? In which case, why use it? Wouldn't the absence of the protocol filter, in and of itself deny access to the filtered protocols?

Like I said, I don't mean to indicate that I think this is a bad idea in any way; I just can't quite figure out what it does.

vn971:
Hi @Fred-Barclay, by a CLI argument `--protocol=` I mean an empty set of allowed protocols. (Same as an empty set in Math, or `[]` as empty arrays in programming).

An empty protocol list means the application cannot use unix sockets, inet interfaces and so on. But it still has access to the filesystem, so it can be a perfectly fine CLI/terminal program.

Fred-Barclay:
Thanks @vn971 
Like you, I'm not very confident with C (though I'm practising!) so I'm not really able to dig into the source code to understand these things. :wink: 
What would be the difference between this (which seems to basically be `--protocol=none`, am I correct?) and simply not having a protocol filter at all? Wouldn't the lack of a protocol argument deny access to the unix sockets, inet interfaces, and so on?

vn971:
@Fred-Barclay I think it can be the whitelist/blacklist story that is confusing you. The `--protocol` directive is _whitelist_, not blacklist. And the absence of `--protocol` means to allow everything there is. If you think about it, it really is a little bit strange (at least semantically). `--protocol=` and the absence of a `--protocol` directive have the opposite meaning.
[...]

PR #5675, opened 2023-02-20
New profiles: qpdf and redirects

glitsj16:
qpdf (CLI) provides PDF metadata cleaning.

See [privacy-handbuch.de](https://www.privacy-handbuch.de/handbuch_43a.htm) for details. The site offers [pdf-meta-clean.sh](https://www.privacy-handbuch.de/download/pdf-meta-clean.sh), which works very well with firejailed qpdf.

### DIFF
--- a/etc/profile-m-z/qpdf.profile
+++ b/etc/profile-m-z/qpdf.profile
@@ -1,0 +1,68 @@
+# Firejail profile for qpdf
+# Description: A Content-Preserving PDF Transformation System
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include qpdf.local
+# Persistent global definitions
+include globals.local
+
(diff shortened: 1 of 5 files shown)


kmk3 (approved):

etc/profile-m-z/qpdf.profile
```diff
@@ -0,0 +1,68 @@
...
+nogroups
+noinput
+nonewprivs
+noprinters
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
```
rusty-snake, line 49: Used? Vor can we `seccomp socket`?

glitsj16: Nice catch, thanks. I always forget about #639.

Issue #5676, opened 2023-02-21
kitty + ssh: Error: too long arguments: argv[22] len (5056) >= MAX_ARG_LEN (4128)


StandingPadAnimations:
### Description
When using firejail with SSH, I receive the following error:
```
$ ssh pi@IP_ADDRESS -p 42069
Error: too long arguments: argv[22] len (5056) >= MAX_ARG_LEN (4128)
```
### Steps to Reproduce
1. perform `sudo firecfg`
2. Attempt to connect to server with SSH

### Expected behavior

For SSH to not complain

### Actual behavior

SSH not connecting due to too long arguments

### Behavior without a profile
SSH was able to work perfectly fine

### Environment

Arch Linux
Firejail 0.9.72

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [ x I have performed a short search for similar issues (to avoid opening a duplicate).
  - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ x I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

<details>
<summary>Output of <code>LC_ALL=C firejail /path/to/program</code></summary>
<p>

```
Reading profile /home/mahid/.config/firejail/ssh.profile
Reading profile /home/mahid/.config/firejail/common.inc
Warning: "shell none" command in the profile file is done by default; the command will be deprecated
Warning: private-lib feature is disabled in Firejail configuration file
Parent pid 13449, child pid 13451
Warning: skipping emp for private /opt
Private /opt installed in 0.12 ms
Warning: skipping emp for private /srv
Private /srv installed in 0.05 ms
1 program installed in 3.04 ms
Warning: An abstract unix socket for session D-BUS might still be available. Use --net or remove unix from --protocol set.
Private /etc installed in 9.43 ms
Private /usr/etc installed in 0.00 ms
Warning: not remounting /run/user/1000/doc
Warning: Cannot confine the application using AppArmor.
[...]

PR #4676, opened 2021-11-10
Make env/arg sanity check failure messages more useful

hlein:
This change doesn't alter any checks, but it gives more specific
errors when a sanity check of env vars or argv does not pass, which
can point to limits to raise or at least give us better detailed bug
reports.

Signed-off-by: Hank Leininger <hlein@korelogic.com>
Bug: https://github.com/netblue30/firejail/issues/3678
Bug: https://github.com/netblue30/firejail/issues/3851
Bug: https://github.com/netblue30/firejail/issues/4633

PR #5677, opened 2023-02-21
Print the argument when failing with "too long arguments"

kmk3:
Also, s/arguments/argument/ since the message refers to one specific
argument.

Relates to commit 0d06369a8 ("Make env/arg sanity check failure messages
more useful", 2021-11-10) / PR #4676.

Relates to #5676.

Cc: @hlein (from #4676)


netblue30:
all merged!

PR #5702, opened 2023-03-01
Drop hostname option from all profiles

glitsj16:
Now we have randomized [UTS namespaces support](https://github.com/netblue30/firejail/discussions/5597#discussioncomment-4996357) by default for every sandbox there's no longer a need to set hostname to a fixed value. This PR removes such fixed hostname entrees from all profiles that had it.


glitsj16:
@kmk3 I opened this PR as a `draft` because it could relate to #5645. If that's incorrect or not necessary, please let me know so I can change the status.

PR #5706, opened 2023-03-03
add lobster.profile

pirate486743186:
it's a shell script for streaming

### DIFF
--- a/etc/profile-a-l/lobster.profile
+++ b/etc/profile-a-l/lobster.profile
@@ -1,0 +1,41 @@
+# Firejail profile for lobster
+# Description: Shell script to watch Movies/Webseries/Shows from the terminal
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include lobster.local
+# Persistent global definitions
+# added by included profile
+#include globals.local
+
+noblacklist ${HOME}/.config/lobster
+noblacklist ${HOME}/.local/share/lobster
+
+# Allow /bin/sh (blacklisted by disable-shell.inc)
+include allow-bin-sh.inc
+
+include disable-proc.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/lobster
+mkdir ${HOME}/.local/share/lobster
+whitelist ${HOME}/.config/lobster
+whitelist ${HOME}/.local/share/lobster
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+
+#machine-id
+nodvd
+noprinters
+notv
+
+disable-mnt
+private-bin curl,cut,fzf,grep,head,lobster,mv,patch,rm,sed,sh,tail,tput,tr,uname
+#private-cache
+private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,machine-id,mime.types,nsswitch.conf,pango,pki,protocols,pulse,resolv.conf,rpc,services,ssl,X11,xdg
+private-tmp
+
+read-only ${HOME}/.config/mpv
+
+# Redirect
+include mpv.profile
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -517,6 +517,7 @@ blacklist ${HOME}/.config/leafpad
 blacklist ${HOME}/.config/libreoffice
 blacklist ${HOME}/.config/liferea
 blacklist ${HOME}/.config/linphone
+blacklist ${HOME}/.config/lobster
 blacklist ${HOME}/.config/lugaru
 blacklist ${HOME}/.config/lutris
 blacklist ${HOME}/.config/lximage-qt
@@ -952,6 +953,7 @@ blacklist ${HOME}/.local/share/kwrite
 blacklist ${HOME}/.local/share/kxmlgui5/*
 blacklist ${HOME}/.local/share/liferea
 blacklist ${HOME}/.local/share/linphone
+blacklist ${HOME}/.local/share/lobster
 blacklist ${HOME}/.local/share/local-mail
 blacklist ${HOME}/.local/share/lollypop
 blacklist ${HOME}/.local/share/love
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -483,6 +483,7 @@ linphone
 linuxqq
 lmms
 lobase
+lobster
 localc
 lodraw
 loffice


kmk3:
(Rebased to fix CI)

Just to confirm, is this <https://github.com/justchokingaround/lobster>?

etc/profile-a-l/lobster.profile
```diff
@@ -0,0 +1,70 @@
...
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/lobster
+mkdir ${HOME}/.local/share/lobster
+
```
glitsj16, line 30: Drop this line.

etc/profile-a-l/lobster.profile
```diff
@@ -0,0 +1,70 @@
...
+whitelist ${HOME}/.config/lobster
+whitelist ${HOME}/.config/mpv
+whitelist ${HOME}/.local/share/lobster
+
```
glitsj16, line 34: Drop this line.

etc/profile-a-l/lobster.profile
```diff
@@ -0,0 +1,70 @@
...
+protocol unix,inet,inet6
+seccomp
+seccomp.block-secondary
+tracelog
+
+disable-mnt
+private-bin curl,cut,fzf,grep,head,lobster,mpv,mv,patch,rm,sed,sh,tail,tput,tr,uname
+private-cache
+private-dev
+private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,machine-id,mime.types,nsswitch.conf,pango,pki,protocols,pulse,resolv.conf,rpc,services,ssl,X11,xdg
```
glitsj16, line 62: Our private-etc is being refactored (cfr. #5610). I suggest to keep this 'old' format commented and add the 'new' format just below. That would be `private-etc @network,@sound,@tls-ca,@x11,mime.types`.

etc/profile-a-l/lobster.profile
```diff
@@ -0,0 +1,70 @@
...
+private-tmp
+
+dbus-user none
+dbus-system none
+
+read-only ${HOME}/.config/mpv
+
+restrict-namespaces
```
glitsj16, line 70: If this is a CLI program, can we add memory-deny-write-execute?

glitsj16 (changes requested):
Some checks needed for potential hardening.

kmk3 (commented):
So this script is mostly a wrapper for mpv and yet the profile is more
restrictive than mpv.profile.

For example, it does not include paths which may be needed by mpv plugins:

```
include allow-lua.inc

whitelist /usr/share/lua
whitelist /usr/share/lua*
```

And it contains `machine-id`, which AFAIK may break pulseaudio.

Some of the extra restrictions make sense to me given the more focused
scenario, such as `nodvd` and `include disable-xdg.inc`.

How about keeping them and making this a redirect to mpv.profile?

To keep it in sync with mpv.profile and avoid duplication.

etc/profile-a-l/lobster.profile
kmk3: @glitsj16 [on Mar 3](https://github.com/netblue30/firejail/pull/5706#discussion_r1124236445):

> Our private-etc is being refactored (cfr. #5610). I suggest to keep this
> 'old' format commented and add the 'new' format just below. That would be
> `private-etc @network,@sound,@tls-ca,@x11,mime.types`.

That conversion can be trivially done automatically later if desired (see
below), while such a comment would then have to be manually edited/removed.
Also note that there is no way to convert back into the paths without losing
information (that is, it would potentially allow more paths than is needed).

So to me it seems better to leave it as is for now.

Example of the automatic conversion:

```console
$ ./src/etc-cleanup/etc-cleanup --replace etc/profile-a-l/lobster.profile

********************
file: etc/profile-a-l/lobster.profile

old: private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,machine-id,mime.types,nsswitch.conf,pango,pki,protocols,pulse,resolv.conf,rpc,services,ssl,X11,xdg

new: private-etc @tls-ca,@x11,host.conf,mime.types,rpc,services

$ git diff
diff --git a/etc/profile-a-l/lobster.profile b/etc/profile-a-l/lobster.profile
index 40d6d4206..7ced26eef 100644
--- a/etc/profile-a-l/lobster.profile
+++ b/etc/profile-a-l/lobster.profile
@@ -59,7 +59,7 @@ disable-mnt
 private-bin curl,cut,fzf,grep,head,lobster,mpv,mv,patch,rm,sed,sh,tail,tput,tr,uname
 private-cache
 private-dev
-private-etc alsa,alternatives,asound.conf,ca-certificates,crypto-policies,fonts,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,machine-id,mime.types,nsswitch.conf,pango,pki,protocols,pulse,resolv.conf,rpc,services,ssl,X11,xdg
+private-etc @tls-ca,@x11,host.conf,mime.types,rpc,services
 private-tmp

 dbus-user none
```

glitsj16: > So to me it seems better to leave it as is for now.

You are much better placed/informed than I am on this matter, so I'll keep this in mind on future reviews. Thx for clearing things up!

etc/profile-a-l/lobster.profile
glitsj16: I meant the empty line between the mkdir's and start of the whitelisting :)

glitsj16 (approved):
Approved, just a few nits.

glitsj16 (commented):
LGTM on principal. Can you try @kmk3's suggestion to refactor this as mpv redirect?

etc/profile-a-l/lobster.profile
```diff
@@ -0,0 +1,40 @@
+# Firejail profile for lobster
+# Description: Shell script to watch Movies/Webseries/Shows from the terminal
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include lobster.local
+# Persistent global definitions
```
glitsj16, line 7: Now that this is a redirect profile, it's important to `avoid including globals.local twice`. Here's what we use in other profiles for that purpose:
```
# added by included profile
#include globals.local
```

glitsj16 (changes requested):
Avoid including globals.local twice. Format to achieve that is in my comment.

glitsj16 (approved):
LGTM. Nice to have this as a redirect profile now, congrats :+1:.

netblue30:
all in, thanks!

PR #5710, opened 2023-03-04
add mov-cli.profile

pirate486743186:
browse and watch movies/shows in the terminal in python.
https://github.com/mov-cli/mov-cli

### DIFF
--- a/etc/profile-m-z/mov-cli.profile
+++ b/etc/profile-m-z/mov-cli.profile
@@ -1,0 +1,31 @@
+# Firejail profile for mov-cli
+# Description: Python script for watching movies and TV shows via the terminal
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/mov-cli.profile
```diff
@@ -0,0 +1,32 @@
...
+# Persistent local customizations
+include mov-cli.local
+# Persistent global definitions
+# added by included profile
+#include globals.local
+
+include disable-proc.inc
+include disable-xdg.inc
+
+whitelist ${DOWNLOADS}
```
rusty-snake, line 14:
```suggestion
```

Added via mpv.profile (whitelist-player-common)

glitsj16 (commented):
LGTM on its own. Hold off merging for a bit so people can think about #5712 for this one too...

PR #5715, opened 2023-03-06
refactor yt-dlp

pirate486743186:
move content from youtube-dl to yt-dlp
youtude-dl and gallery-dl redirect to yt-dlp

yt-dlp has become the main program now. The future of youtube-dl is uncertain.

### DIFF
--- a/etc/profile-m-z/youtube-dl.profile
+++ b/etc/profile-m-z/youtube-dl.profile
@@ -5,63 +5,17 @@ quiet
 # Persistent local customizations
 include youtube-dl.local
 # Persistent global definitions
-include globals.local
-
-# breaks when installed under ${HOME} via `pip install --user` (see #2833)
-ignore noexec ${HOME}
+# added by included profile
+#include globals.local
 
 noblacklist ${HOME}/.cache/youtube-dl
 noblacklist ${HOME}/.config/youtube-dl
-noblacklist ${HOME}/.netrc
-noblacklist ${MUSIC}
-noblacklist ${VIDEOS}
 
 # Allow python (blacklisted by disable-interpreters.inc)
 include allow-python2.inc
-include allow-python3.inc
 
-blacklist /tmp/.X11-unix
-blacklist ${RUNUSER}
+private-bin youtube-dl
+private-etc youtube-dl.conf
 
-include disable-common.inc
-include disable-devel.inc
-include disable-exec.inc
-include disable-interpreters.inc
-include disable-programs.inc
-include disable-shell.inc
-include disable-xdg.inc
-
-include whitelist-usr-share-common.inc
-include whitelist-var-common.inc
-
-apparmor
-caps.drop all
-ipc-namespace
-machine-id
-netfilter
-no3d
-nodvd
-nogroups
-noinput
-nonewprivs
-noroot
-nosound
-notv
-nou2f
-novideo
-protocol unix,inet,inet6
-seccomp
-seccomp.block-secondary
-tracelog
-
-private-bin env,ffmpeg,python*,youtube-dl
-private-cache
-private-dev
-private-etc @tls-ca,mime.types,youtube-dl.conf
-private-tmp
-
-dbus-user none
-dbus-system none
-
-#memory-deny-write-execute - breaks on Arch (see issue #1803)
-restrict-namespaces
+# Redirect
+include yt-dlp.profile
--- a/etc/profile-m-z/yt-dlp.profile
+++ b/etc/profile-m-z/yt-dlp.profile
@@ -5,17 +5,73 @@ quiet
 # Persistent local customizations
 include yt-dlp.local
 # Persistent global definitions
-# added by included profile
-#include globals.local
+include globals.local
+
+# If you installed via pip under ${HOME}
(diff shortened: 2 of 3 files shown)


etc/profile-m-z/youtube-dl.profile
```diff
@@ -5,63 +5,17 @@ quiet
...
+#include globals.local
 
 noblacklist ${HOME}/.cache/youtube-dl
 noblacklist ${HOME}/.config/youtube-dl
-noblacklist ${HOME}/.netrc
-noblacklist ${MUSIC}
-noblacklist ${VIDEOS}
 
 # Allow python (blacklisted by disable-interpreters.inc)
 include allow-python2.inc
```
rusty-snake, line 15: included after include of disable-interpreters.inc => no-op

rusty-snake (changes requested):
LGTM

The python2 include does not work.

etc/profile-m-z/yt-dlp.profile
```diff
@@ -5,17 +5,65 @@ quiet
 # Persistent local customizations
 include yt-dlp.local
 # Persistent global definitions
-# added by included profile
-#include globals.local
+include globals.local
+
+# breaks when installed under ${HOME} via `pip install --user` (see #2833)
```
glitsj16, line 10: Wouldn't it make more sense to keep `noexec ${HOME}` and add a new comment instead? Users who install it like this face apparmor issues anyway if they use that.

etc/profile-m-z/yt-dlp.profile
```diff
@@ -5,17 +5,65 @@ quiet
...
+private-tmp
 
-private-bin ffprobe,yt-dlp
-private-etc yt-dlp.conf
+dbus-user none
+dbus-system none
 
-# Redirect
-include youtube-dl.profile
+#memory-deny-write-execute - breaks on Arch (see issue #1803)
```
glitsj16, line 68: Is this still correct? I've been running with mdwe (on Arch Linux and Ubuntu LTS) in a local override for quite some time without apparent issues.

glitsj16 (commented):
LGTM. Just a few general questions that are not related to changing the redirect logic.

etc/profile-m-z/yt-dlp.profile
glitsj16: @rusty-snake has it in [kyst](https://github.com/rusty-snake/kyst/blob/main/firejail/youtube-dl.local) as well. Enabling it seems well supported, let's do it!

etc/profile-m-z/yt-dlp.profile
glitsj16: I'd go for:

```
# If you installed via pip under ${HOME} add `ignore noexec ${HOME}`
# to your yt-dlp.local. AppArmor needs to allow it too, as per examples
# in /etc/apparmor.d/local/firejail-default.
noexec ${HOME}
```

etc/profile-m-z/youtube-dl.profile
glitsj16: @pirate486743186 Looks correct to me.

etc/profile-m-z/yt-dlp.profile
glitsj16: > isn't "noexec ${HOME}" already active from an inc?

I messed up here. That's correct.

> from what you said, i have no idea what to put in "/etc/apparmor.d/local/firejail-default"

https://github.com/netblue30/firejail/blob/1f5f84f9596ba7d26c2f2e35dad7728e12afaec9/etc/apparmor/firejail-local#L3-L10

Users with yt-dlp installed like that could go for the third example:
`owner @HOME/.local/bin/** ix,`

NOTE: I'll fix these examples, the comma at the end is needed in all three examples.

etc/profile-m-z/youtube-dl.profile
rusty-snake: Your right, was comparing it in the wrong order. Shouldn't review late at night.

etc/profile-m-z/yt-dlp.profile
glitsj16: This is a bit more involved than usual due to the combination of a local install and AppArmor. The thing is, if users (1) install yt-dlp via pip under /home/foo _and_ (2) are using AppArmor they can harden the Firejail sandbox by (3) adding those rules to /etc/apparmor.d/local/firejail-default. In that case they benefit from keeping the `apparmor` option.

Not sure if all that is conveyed in the current comment. To be clear, I'm not at all suggesting your comment is incorrect. I do appreciate your effort to make it as precise as possible for (Ubuntu) users. Not an easy job though to explain how to configure AA in a few lines. Maybe we can refer to distro documentation?

glitsj16: > I rather not give them a "RTFM" comment.

No problem, I can dig that. Marking conversation as resolved.

etc/profile-m-z/yt-dlp.profile
glitsj16: > i'll add them all at once

Okay :) Marking conversation as resolved.

rusty-snake (approved):

netblue30:
merged, thanks!

PR #5711, opened 2023-03-05
mpv: improve and clarify comments

glitsj16:
~~Although mpv.profile includes `allow-lua.inc`, lua support was effectively broken due to not being referenced in `private-bin`. This PR fixes this.~~
Lua support is working despite of not being referenced in private-bin. Edit comments to clarify things accordingly.

PR #5728, opened 2023-03-11
profiles: refactor chromium/electron into blink-common

ghost:
creating `chromium-electron-common` common profile for
`chromium-common` and `electron-common`

Was very conservative, didn't change anything. It's an error if i did.
`whitelist-run-common.inc` seams like it could be added too, left it like a comment. That would change electron-common and the ~30 profiles that depend on it.
Don't request real changes for this commit.

Now you can have a single local file for both electron apps and chromium browsers.
Also, electron-common was a bit neglected. It will take more attention now.

### DIFF
--- a/etc/profile-a-l/blink-common-hardened.inc.profile
+++ b/etc/profile-a-l/blink-common-hardened.inc.profile
@@ -1,0 +1,11 @@
+# This file is overwritten during software install.
+# Persistent customizations should go in a .local file.
+include blink-common-hardened.inc.local
+
+caps.drop all
+nonewprivs
+noroot
+protocol unix,inet,inet6,netlink
+seccomp !chroot
+
+#restrict-namespaces
--- a/etc/profile-a-l/blink-common.profile
+++ b/etc/profile-a-l/blink-common.profile
@@ -1,0 +1,40 @@
+# Firejail profile for blink-common
+# Description: Common profile for Blink-based applications
+# This file is overwritten after every install/update
+# Persistent local customizations
+include blink-common.local
+# Persistent global definitions
+# added by caller profile
+#include globals.local
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+whitelist ${DOWNLOADS}
+include whitelist-common.inc
+#include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+# If your kernel allows the creation of user namespaces by unprivileged users
+# (for example, if running `unshare -U echo enabled` prints "enabled"), you
+# can add the next line to your blink-common.local.
+#include blink-common-hardened.inc.profile
+
+apparmor
+caps.keep sys_admin,sys_chroot
+netfilter
+nodvd
+nogroups
+noinput
+notv
+
+disable-mnt
+private-cache
+
+dbus-system none
--- a/etc/profile-a-l/chromium-common-hardened.inc.profile
+++ b/etc/profile-a-l/chromium-common-hardened.inc.profile
@@ -1,11 +1,10 @@
-# This file is overwritten during software install.
-# Persistent customizations should go in a .local file.
+# Firejail profile alias for blink-common-hardened.inc
+# This file is overwritten after every install/update
+# Persistent local customizations
 include chromium-common-hardened.inc.local
+# Persistent global definitions
+# added by caller profile
+#include globals.local
 
-caps.drop all
-nonewprivs
-noroot
-protocol unix,inet,inet6,netlink
-seccomp !chroot
-
-#restrict-namespaces
+# Redirect
+include blink-common-hardened.inc.profile
--- a/etc/profile-a-l/chromium-common.profile
+++ b/etc/profile-a-l/chromium-common.profile
@@ -17,42 +17,21 @@ noblacklist /usr/lib/chromium/chrome-sandbox
 # to have access to Gnome extensions (extensions.gnome.org) via browser connector
 #include allow-python3.inc
 
-include disable-common.inc
-include disable-devel.inc
-include disable-exec.inc
-include disable-interpreters.inc
-include disable-programs.inc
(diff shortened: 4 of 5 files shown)


kmk3 (changes requested):
> Was very conservative, didn't change anything.

+1 for this, it makes it easier to review.

Note: I'm not an expert in Chromium's internals; the descriptions below might
not be entirely correct.

> creating `chromium-electron-common` common profile for `chromium-common` and
> `electron-common`

> Now you can have a single local file for both electron apps and chromium
> browsers. Also, electron-common was a bit neglected. It will take more
> attention now.

I get the idea, but I find "chromium-electron-common" confusing.

Note that besides Electron there are other frameworks/runtimes that depend on
Blink (Chromium's layout engine), such as CEF (used by Steam and GOG AFAIK) and
QtWebEngine (used by Falkon and some KDE programs).

So it might make more sense to reference the common denominator (that is, the
layout engine / the Chromium "platform") in the profile name:

* chromium-engine-common.profile

Or:

* chromium-platform-common.profile

Misc: Besides Blink there is also V8, Chromium's JavaScript engine.  Both Blink
and nodejs use V8, but nodejs does not use Blink, for example.  So "Chromium
engine" is a bit ambiguous, but AFAICT Blink is the thing responsible for
rendering web pages and it is the module that contains/uses everything else
(such as V8) in order to achieve that.  While "Chromium platform" could maybe
be mistaken for something like ChromeOS.  Feel free to suggest better names.

Note also that there are already many "chromium-something" profiles:

* chromium-browser-privacy.profile (ungoogled chromium something from #3688)
* chromium-browser.profile (redirect to chromium.profile)
* chromium-common-hardened.inc.profile (misc + redirect)
* chromium-common.profile (common to chromium-based web browsers)
* chromium.profile (the web browser itself)
* electron-common.profile (common to electron-based programs)

They already seem hard to tell apart at a glance; I'm not sure if adding
another one would be worth the added complexity.  I'll try to expand on this
later.

kmk3:
@pirate486743186 [on Mar 12](https://github.com/netblue30/firejail/pull/5728#issuecomment-1465314939):

> electron apps are basically a local website inside a chromium based "web
> browser". This is why they are crossplatform and very heavy. It's not
> something reasonable like QtWebEngine or nodejs. They are controversial like
> snaps.
> 
> If you know this, i think that `chromium-electron-common` make sense. For
> other related frameworks, the differences are much larger.

Yes, Electron programs are usually one of more orders of magnitude heavier than
the alternatives.

But I think that profile-wise the difference between Electron programs and
Chromium-based web browsers (that is, Blink -> Chromium) could bigger than the
difference between Electron and QtWebEngine (both of which just use Blink).

Mainly because even if Electron programs weigh similarly to a normal Chromium
instance, each of those programs is usually written for a particular purpose
(and/or a particular website), while Chromium/Chrome try to do ever more things
(and render every website).

So it might be worthwhile to have one include for Blink-based frameworks and
another for Chromium-based browsers.  Especially considering that there is
already electron-common.profile for Electron-based programs.

For example, comparing the version of chromium-electron-common.profile from
this PR to falkon.profile, it seems to use all of the same commands except for
`caps.keep sys_admin,sys_chroot`.

With all that said, besides organization, the point is also about the naming
and the following seems less confusing:

* chromium-engine-common.profile (`$runtime-common.profile`)

Than this:

* chromium-electron-common.profile (`$runtime1-$runtime2-common.profile`)

Especially considering the many already existing Chromium-related profiles.

---

Relates to the following discussion:

* <https://github.com/netblue30/firejail/pull/5734#issuecomment-1468022029>

kmk3:
@pirate486743186 [on Mar 14](https://github.com/netblue30/firejail/pull/5728#issuecomment-1468825007):

> there's also otter-browser, they are probably a bit neglected too.

Indeed and otter-browser.profile also seems to use similar entries.

Even steam.profile (which uses CEF) does some similar things (like allowing
`chroot`), though it is quite different in purpose and in what it allows.

> what about `chromium-base-common`?

That also seems workable, but to minimize confusion with the existing
Chromium-related profiles, how about going with the exact runtime name?

* `blink-common.profile`

I didn't say it before because I was thinking about the WebKit -> Blink
renaming, though that was due to Google forking it and so it is unlikely to
happen again (as Blink is its own separate thing now).

etc/profile-a-l/blink-common.profile
```diff
@@ -0,0 +1,40 @@
+# Firejail profile for blink-common
+# Description: Common profile for blink applications
```
kmk3, line 2:
```suggestion
# Description: Common profile for Blink-based applications
```

To make it a bit clearer.

etc/profile-a-l/blink-common-hardened.inc.profile
```diff
@@ -0,0 +1,10 @@
+# Firejail profile alias for chromium-common-hardened.inc
+# This file is overwritten after every install/update
+# Persistent local customizations
+include blink-common-hardened.inc.local
+# Persistent global definitions
+# added by caller profile
+#include globals.local
+
+# Redirect
+include chromium-common-hardened.inc.profile
```
kmk3, line 10: This seems backwards, as chromium-common.profile redirects to
blink-common.profile:

* chromium-common.profile -> blink-common.profile
* blink-common-hardened.inc.profile -> chromium-common-hardened.inc.profile

I think it would be better to remove this file.

Alternatively, test Blink-based programs with
chromium-common-hardened.inc.profile and if that works:

* Move the entries from chromium-common-hardened.inc.profile to this file
* Make chromium-common-hardened.inc.profile redirect to this file instead

@rusty-snake Since you worked on #3688, thoughts on this?

kmk3:
Can you try squashing the commits with `fixup` as mentioned here?

* <https://github.com/netblue30/firejail/pull/5711#issuecomment-1455862970>

Other than that, LGTM.

kmk3 (approved):
LGTM.

kmk3:
By the way, I didn't get any notifications from the force-push; it still
appears as being from 5 days ago.

IIRC they did send notifications for that before, but that does not seem to be
the case anymore.

glitsj16 (approved):
LGTM

netblue30:
all in!

PR #5734, opened 2023-03-13
refresh engrampa.profile

pirate486743186:
Light refresh.

this is a bit tricky, because it uses a lot of cli tools to compress/decompress.

### DIFF
--- a/etc/profile-a-l/engrampa.profile
+++ b/etc/profile-a-l/engrampa.profile
@@ -10,18 +10,21 @@ include disable-common.inc
 include disable-devel.inc
 include disable-exec.inc
 include disable-interpreters.inc
+include disable-proc.inc
 include disable-programs.inc
 
 include whitelist-var-common.inc
 
 apparmor
 caps.drop all
+machine-id
 net none
 no3d
 nodvd
 nogroups
 noinput
 nonewprivs
+noprinters
 noroot
 nosound
 notv
@@ -29,6 +32,7 @@ nou2f
 novideo
 protocol unix
 seccomp
+seccomp.block-secondary
 tracelog
 
 # private-bin engrampa


etc/profile-a-l/engrampa.profile
```diff
@@ -10,25 +10,29 @@ include disable-common.inc
...
 nogroups
 noinput
 nonewprivs
+noprinters
 noroot
 nosound
 notv
 nou2f
 novideo
 protocol unix
```
glitsj16, line 33: see #639: check if it works without protocol and `seccomp socket`

glitsj16 (commented):
Check if we can use seccomp socket

glitsj16:
> It's too hard to test. It uses too many external programs.

Oh okay, I see what you mean now I think. Engrampa is a fork of file-roller. That could do with a similar refresh too IMO. Its long private-bin is too brittle and we should add basic D-Bus filtering like in engrampa's. Anyway, thanks for the work.

glitsj16 (approved):
LGTM

kmk3 (approved):

glitsj16:
> hmm, maybe make a file-roller-engrampa-common like eo-common
or gnome-mate-archiver-common
You have a better idea of a name?

Personally I'm not particularly fond of creating more and more common include files 'because we can'. Include logic is already quite complicated for newish users IMO. I mentioned file-roller because that has a very specific and long private-bin that might suit engrampa as well - being a fork of it. Nothing more than a personal opinion though. I'd never block a PR because of it.

kmk3:
@glitsj16 [on Mar 14](https://github.com/netblue30/firejail/pull/5734#issuecomment-1468022029):

> > hmm, maybe make a file-roller-engrampa-common like eo-common or
> > gnome-mate-archiver-common You have a better idea of a name?
> 
> Personally I'm not particularly fond of creating more and more common include
> files 'because we can'. Include logic is already quite complicated for newish
> users IMO.

Agreed.  It might be a good idea when multiple profiles all have many unusual
entries compared to other profiles and/or if the include would be used by many
profiles.  Otherwise, I think it's unlikely to be worth the extra complexity.

> I mentioned file-roller because that has a very specific and long private-bin
> that might suit engrampa as well - being a fork of it.

In this case, if the same `private-bin` line indeed works for both profiles
(and is used in both profiles), it could be helpful to add a comment above each
`private-bin` line saying to keep them in sync.

kmk3:
@pirate486743186 [on Mar 14](https://github.com/netblue30/firejail/pull/5734#issuecomment-1468360340):

> > helpful to add a comment above each `private-bin` line saying to keep them
> > in sync.
> 
> They are 1000+ profiles.... Something like this is not practical. That's one
> of the points of common profiles. I'm not saying to over do it either.

Forks like this are probably a small part of the 1000+ profiles.  I'd wager
that the majority of profiles is composed of a) programs without relatively
popular forks and b) programs that already have many derivatives and thus
already have a common profile anyway (such as web browsers).

I'm not saying to do it on every such profile either; documenting things in
profiles would just be one option between doing nothing vs creating common
profiles for every group of forked (or just related) programs.

PR #5708, opened 2023-03-03
Forbid control chars in names

layderv:
Forbid control characters in names and filenames. https://github.com/netblue30/firejail/pull/5613.

Syntax for hostnames from https://en.wikipedia.org/wiki/Hostname#Syntax

PR #5741, opened 2023-03-21
util.c: add and use ascii-only char functions

kmk3:
The "invalid_name" function claims to "allow strict ASCII letters and
numbers".

However, it uses isalnum(3) and isdigit(3), which may take the current
locale into account and thus return 1 for non-ASCII characters.

So add the following functions:

* ascii_isalnum
* ascii_isalpha
* ascii_isdigit
* ascii_islower
* ascii_isupper
* ascii_isxdigit

And use the applicable ones in "invalid_name" so that it actually uses
strictly ASCII in its comparisons.

Added on commit b4ffaa207 ("merges; more on cleaning up esc chars",
2023-02-14).

Relates to #5578.

Kind of relates to #5708.

### DIFF
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -525,7 +525,6 @@ int macro_id(const char *name);
 
 
 // util.c
-int invalid_name(const char *name);
 void errLogExit(char* fmt, ...) __attribute__((noreturn));
 void fwarning(char* fmt, ...);
 void fmessage(char* fmt, ...);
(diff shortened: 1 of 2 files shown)


kmk3:
Misc: There are also improvements for #5708 which depend on this PR.

kmk3:
(Rebased to master to fix CI checks)

netblue30:
Nice, thanks!

PR #5744, opened 2023-03-21
keepassxc: Allow offering the Secret Service

ThijsRay:
KeePassXC can offer a Secret Service (see https://specifications.freedesktop.org/secret-service/latest/) to store secrets for other programs

### DIFF
--- a/etc/profile-a-l/keepassxc.profile
+++ b/etc/profile-a-l/keepassxc.profile
@@ -93,6 +93,7 @@ private-etc
 private-tmp
 
 dbus-user filter
+dbus-user.own org.freedesktop.secrets
 dbus-user.own org.keepassxc.KeePassXC.*
 dbus-user.talk com.canonical.Unity
 dbus-user.talk org.freedesktop.ScreenSaver


etc/profile-a-l/keepassxc.profile
```diff
@@ -94,6 +94,8 @@ private-tmp
 
 dbus-user filter
 dbus-user.own org.keepassxc.KeePassXC.*
+dbus-user.own org.freedesktop.secrets
+dbus-user.own org.freedesktop.Secret.*
```
kmk3, line 98:
```suggestion
dbus-user.own org.freedesktop.Secret.*
dbus-user.own org.freedesktop.secrets
dbus-user.own org.keepassxc.KeePassXC.*
```

Sort.

etc/profile-a-l/keepassxc.profile
```diff
@@ -93,6 +93,8 @@ private-etc
 private-tmp
 
 dbus-user filter
+dbus-user.own org.freedesktop.Secret.*
```
glitsj16, line 96: Is this path actually exposed/needed? All our other profiles that support this API only use `dbus-user.talk org.freedesktop.secrets` (cfr. [this](https://specifications.freedesktop.org/secret-service/latest/ch12.html)). Note the difference with `dbus-user.own`...

glitsj16 (commented):
Please check if we can use the same format as in other profiles.

etc/profile-a-l/keepassxc.profile
ThijsRay: You're right! I've tested it without this line and it still works as intended. I've removed it.

etc/profile-a-l/keepassxc.profile
```diff
@@ -93,6 +93,7 @@ private-etc
 private-tmp
 
 dbus-user filter
+dbus-user.own org.freedesktop.secrets
```
glitsj16, line 96: One final remark. As hinted at in my previous comment there's a difference between dbus-user.`own` foo and dbus-user.`talk` foo. KeePassXC only needs to _talk_ to the Secret Service, so drop this line and add `dbus-user.talk org.freedesktop.secrets` at the bottom of this D-Bus options block (to keep options sorted alphabetically).

glitsj16 (changes requested):
Use dbus-user.talk instead of dbus-user.own...

etc/profile-a-l/keepassxc.profile
ThijsRay: The Secret Service functionality of KeePassXC doesn't work with `dbus-user.talk`. [KeePassXC offers the Secret Service](https://github.com/keepassxreboot/keepassxc/blob/a31c5ba006198f2fdadfd6c4f587cfed602c81c6/src/fdosecrets/dbus/DBusMgr.cpp#L406-L428). It therefore needs `dbus-user.own`. When other programs talk to `org.freedesktop.secrets`, then end up at KeePassXC, which manages the secrets for them.

glitsj16: Thanks for the detailed info. All good then as is!

glitsj16 (approved):
LGTM

kmk3 (approved):

kmk3:
@ThijsRay

Note that no line-wrapping is performed in the git log:

```console
$ git show --pretty='%s%n%n%b' -s 628d3b1c44
Allow KeePassXC to offer the Secret Service

KeePassXC can offer a Secret Service (see https://specifications.freedesktop.org/secret-service/latest/) to store secrets for other programs
```

I fixed it in this PR and force-pushed:

```console
$ git show --pretty='%s%n%n%b' -s 1bbb5a50cc
keepassxc: Allow offering the Secret Service

KeePassXC can offer a Secret Service to store secrets for other
programs.

See https://specifications.freedesktop.org/secret-service/latest/
```

On future PRs, please try to limit each line to at most 72 characters so that
it is readable in the terminal.

Note: This does not apply to lines that usually shouldn't be broken, such as
lines containing the output of programs or logs.

netblue30:
merged, thanks!

PR #5752, opened 2023-03-25
ssh-agent: add deterministic-shutdown

### LINKED ISSUE #5751: ssh-agent: eval hangs without `--deterministic-shutdown`
I've had `deterministic-shutdown` in my `ssh-agent.local` ever since the option was introduced. Never realised until accidentally bumping into this today that the current ssh-agent.profile is likely broken for most users.

The bad...

```console
$ eval "$(firejail /usr/bin/ssh-agent -s)"
[hangs indefinately]
```

The good...

```console
$ eval "$(firejail --deterministic-shutdown /usr/bin/ssh-agent -s)"
Agent pid 7
$
```

glitsj16:
Fixes #5751.

IMO the profile can be hardened considerably, but that's food for a follow-up PR.

### DIFF
--- a/etc/profile-m-z/ssh-agent.profile
+++ b/etc/profile-m-z/ssh-agent.profile
@@ -34,4 +34,5 @@ writable-run-user
 dbus-user none
 dbus-system none
 
+deterministic-shutdown
 restrict-namespaces


kmk3 (approved):

rusty-snake (changes requested):
See comments in https://github.com/netblue30/firejail/issues/5751 before mergeing.

glitsj16:
See #5751 for details on closing.

PR #5753, opened 2023-03-25
ssh-agent: hardening

glitsj16:
IMO there's nothing holding back making the (OpenSSH) authenticator agent profile very tight. Contrary to the SSH client its functionality is very limited and it doesn't need access to much of the file system.

### DIFF
--- a/etc/profile-m-z/ssh-agent.profile
+++ b/etc/profile-m-z/ssh-agent.profile
@@ -7,6 +7,7 @@ include ssh-agent.local
 # Persistent global definitions
 include globals.local
 
+blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}/wayland-*
 
 # Allow ssh (blacklisted by disable-common.inc)
@@ -42,12 +43,10 @@ noprinters
 noroot
 nosound
 notv
-nou2f
 novideo
 protocol unix,inet,inet6
 seccomp
 tracelog
-x11 none
 
 disable-mnt
 private-bin ssh-agent
--- a/etc/profile-m-z/ssh-agent.profile
+++ b/etc/profile-m-z/ssh-agent.profile
@@ -7,7 +7,6 @@ include ssh-agent.local
 # Persistent global definitions
 include globals.local
 
-blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}/wayland-*
 
 # Allow ssh (blacklisted by disable-common.inc)
--- a/etc/profile-m-z/ssh-agent.profile
+++ b/etc/profile-m-z/ssh-agent.profile
@@ -1,4 +1,5 @@
 # Firejail profile for ssh-agent
+# Description: OpenSSH authentication agent
 # This file is overwritten after every install/update
 quiet
 # Persistent local customizations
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/ssh-agent.profile
```diff
@@ -1,37 +1,62 @@
...
+noprinters
 noroot
+nosound
 notv
+nou2f
 novideo
 protocol unix,inet,inet6
 seccomp
 tracelog
+x11 none
```
rusty-snake, line 50: Will break w/o `net none`.

etc/profile-m-z/ssh-agent.profile
```diff
@@ -1,37 +1,62 @@
...
+include disable-interpreters.inc
+include disable-proc.inc
 include disable-programs.inc
+include disable-shell.inc
+include disable-X11.inc
+include disable-xdg.inc
 
+include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
```
rusty-snake, line 29: So it does not need to access `$XDG_RUNTIME_DIR/keyring/ssh` or other ssh sockets in there?

etc/profile-m-z/ssh-agent.profile
```diff
@@ -1,37 +1,62 @@
...
 no3d
 nodvd
+noinput
+nogroups
 nonewprivs
+noprinters
 noroot
+nosound
 notv
+nou2f
```
kmk3, line 45:
```suggestion
```

If anything, this would be the quintessential application that uses hardware
keys :p

kmk3 (commented):
Note that the following might be one reason that this is not currently hardened
too much:

From ssh-agent(1):

```
     command [arg ...]
             If a command (and optional arguments) is given, this is executed
             as a subprocess of the agent.  The agent exits automatically when
             the command given on the command line terminates.

     There are two main ways to get an agent set up.  The first is at the
     start of an X session, where all other windows or programs are started as
     children of the ssh-agent program.  The agent starts a command under
     which its environment variables are exported, for example ssh-agent xterm
     &.  When the command terminates, so does the agent.

     The second method is used for a login session.  When ssh-agent is
     started, it prints the shell commands required to set its environment
     variables, which in turn can be evaluated in the calling shell, for
     example eval `ssh-agent -s`.

     In both cases, ssh(1) looks at these environment variables and uses them
     to establish a connection to the agent.
```

That is, hardening ssh-agent could break running subcommands like in the
example:

```sh
ssh-agent xterm &
```

If that is a supported use-case, it might be better to only use firejail with
the subcommand.

Example:

```sh
/usr/bin/ssh-agent firejail /usr/bin/foo &
```

Or to remove ssh-agent from firecfg to ensure that the following works:

```sh
ssh-agent foo & # -> /usr/bin/ssh-agent firejail /usr/bin/foo &
```

etc/profile-m-z/ssh-agent.profile
glitsj16: Not according to my observations no. Other parts of OpenSSH need that kind of access, like ssh, ssh-add, ssh-copy-id, ssh-key{gen,scan}... The authentication agent itself is kept 'dumb' on purpose (which is a good thing security-wise). We don't have profiles (yet) for ssh-add and the others mentioned.

etc/profile-m-z/ssh-agent.profile
glitsj16: Fixed, thanks!

etc/profile-m-z/ssh-agent.profile
kmk3: @glitsj16 [on Mar 25](https://github.com/netblue30/firejail/pull/5753#discussion_r1148330269):

> Not according to my observations no. Other parts of OpenSSH need that kind of
> access, like ssh, ssh-add, ssh-copy-id, ssh-key{gen,scan}... The
> authentication agent itself is kept 'dumb' on purpose (which is a good thing
> security-wise). We don't have profiles (yet) for ssh-add and the others
> mentioned.

Sorry, now that you mention it, the hardware keys might only be needed by
ssh-add, not ssh-agent (I'm not sure).

glitsj16: I agree. But the way I read its man page (in this case more specifically the `-P` switch) it's actually ssh-add that needs u2f access, the agent only stores whatever is put in. Sadly I don't have access to hardware keys to properly test my assumptions. Does anyone on the collaborator team?

etc/profile-m-z/ssh-agent.profile
```diff
@@ -7,6 +7,7 @@ include ssh-agent.local
 # Persistent global definitions
 include globals.local
 
+blacklist /tmp/.X11-unix
```
rusty-snake, line 10: Done by `include disable-X11.inc`

glitsj16:
Converted to draft. There are several important issues raised in review that need more testing. Thanks @kmk3 and @rusty-snake for their initial observations, remarks and questions. IMO we need to de-mystify our OpenSSH profiles/use cases. Could take some time though.

etc/profile-m-z/ssh-agent.profile
glitsj16: Well that's the thing. I'm not sure either. Too many open-ended questions for comfort... Especially the points you refer to below, in combination with @rusty-snake's [hint](https://github.com/netblue30/firejail/issues/5751#issuecomment-1483760338) send me back to the drawing board :)

glitsj16:
See #5751 for details on closing.

PR #5707, opened 2023-03-03
add ani-cli.profile

pirate486743186:
shell script for streaming anime

Issue #1642, opened 2017-11-12
[Question] Does 'netfilter=filename' affect the system firewall, eg UFW


Irvinehimself:
As the title states: Does playing around with Firejail netfilters on a per application basis represent a safe way of learning about netfilter rules?

At the moment, I use the default UFW rules
```
 Default: deny (incoming), allow (outgoing), disabled (routed)
```
and am loathe to experiment with these settings until I have a better understanding of what I am doing

Thanks 
Irvine

Ps, If if you are interested, I have attached zipped profiles for:  bsdtar, cower, makepkg, ping and archaudit-report
[Profiles.zip](https://github.com/netblue30/firejail/files/1464407/Profiles.zip)

SpotComms:
I don't think Firejail ever directly interacts with UFW. And afaik UFW is just a fancy bash script to control iptables.

As for learning about iptables here are some nice writeups:
- https://www.digitalocean.com/community/tutorials/how-to-choose-an-effective-firewall-policy-to-secure-your-servers
- https://security.stackexchange.com/questions/4603/tips-for-a-secure-iptables-config-to-defend-from-attacks-client-side/4745#4745
- https://people.netfilter.org/hawk/presentations/devconf2014/iptables-ddos-mitigation_JesperBrouer.pdf (outdated! linux 4.4 features a lockless TCP listener)

netblue30:
> Does 'netfilter=filename' affect the system firewall, eg UFW

No, Firejail doesn't touch your system firewall. It installs a new one in the sandbox if you use --net to start another network namespace. Each network namespace (system or sandbox) has its own firewall.

I'll start bringing in your profiles, thanks.

netblue30:
all merged, thanks.

PR #5763, opened 2023-03-28
profiles: move read-only config entries to disable-common.inc

kmk3:
This should reduce duplication and increase hardening a bit.

Relates to #1642 #5706 #5707 #5710.

Cc: @pirate486743186 (from #5706 #5707 #5710)

### DIFF
--- a/etc/profile-a-l/firefox.profile
+++ b/etc/profile-a-l/firefox.profile
@@ -14,6 +14,9 @@ include globals.local
 # https://github.com/netblue30/firejail/wiki/Frequently-Asked-Questions#how-do-i-run-two-instances-of-firefox
 # https://github.com/netblue30/firejail/issues/4206#issuecomment-824806968
 
+# (Ignore entry from disable-common.inc)
+ignore read-only ${HOME}/.mozilla/firefox/profiles.ini
+
 noblacklist ${HOME}/.cache/mozilla
 noblacklist ${HOME}/.mozilla
 noblacklist ${RUNUSER}/*firefox*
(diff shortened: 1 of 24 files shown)


etc/profile-a-l/firefox.profile
```diff
@@ -66,5 +66,7 @@ dbus-user.own org.mpris.MediaPlayer2.firefox.*
 #ignore noroot
 ignore dbus-user none
 
+read-write ${HOME}/.mozilla/firefox/profiles.ini
```
rusty-snake, line 69: Overridden by later read-only.

rusty-snake (dismissed):

etc/profile-a-l/firefox.profile
kmk3: > Overridden by later read-only.

Good catch; fixed.

etc/templates/profile.template
```diff
@@ -221,6 +221,8 @@ include globals.local
 #dbus-user.talk org.freedesktop.Notifications
 #dbus-system none
 
+# Note: read-only entries should usually go on disable-common.inc (especially
```
glitsj16, line 224: Typo: ... go in disable-common.inc ...

glitsj16 (approved):
LGTM. A single typo change and we can bring this in IMO.

etc/templates/profile.template
kmk3: > Typo: ... go in disable-common.inc ...

Changed.

By the way, are you sure that there is a preferred form?

I think I've seen it used both ways, so I don't know.

glitsj16: > By the way, are you sure that there is a preferred form?

Sorry, I don't understand what preferred form you're referring to in this context.

glitsj16: Oh, I just thought it was an actual typo. If you'd like to keep the "on $file" form, please disregard my nitpick.

netblue30:
all merged!

PR #5780, opened 2023-04-07
New profile: url-eater

glitsj16:
Clipboard URL privacy daemon: url-eater.
Homepage: https://github.com/AgathaSorceress/url-eater.
Works on Xorg and Wayland.

### DIFF
--- a/etc/profile-m-z/url-eater.profile
+++ b/etc/profile-m-z/url-eater.profile
@@ -1,0 +1,58 @@
+# Firejail profile for url-eater
+# Description: Clean unnecessary parameters from URLs copied to clipboard
+# This file is overwritten after every install/update
+# Persistent local customizations
+include url-eater.local
+# Persistent global definitions
+include globals.local
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
+net none
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noprinters
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+seccomp.block-secondary
+tracelog
+
+disable-mnt
+private-bin url-eater
+private-cache
+private-dev
+private-etc url-eater.kdl
+private-lib
+#private-tmp # breaks on Arch
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+read-only ${HOME}
+restrict-namespaces
--- a/RELNOTES
+++ b/RELNOTES
@@ -29,7 +29,7 @@ firejail (0.9.73) baseline; urgency=low
   * docs: markdown formatting and misc improvements (#5757)
   * legal: selinux.c: Split Copyright notice & use same license as upstream
     (#5667)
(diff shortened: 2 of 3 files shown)


etc/profile-m-z/url-eater.profile
```diff
@@ -0,0 +1,59 @@
+# Firejail profile for url-eater
+# Description: Clean unnecessary parameters from URLs copied to clipboard
+# This file is overwritten after every install/update
+# Persistent local customizations
+include url-eater.local
+# Persistent global definitions
+include globals.local
+
+blacklist ${RUNUSER}/wayland-*
```
rusty-snake, line 9: Tested on Wayland?

etc/profile-m-z/url-eater.profile
```diff
@@ -0,0 +1,59 @@
...
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+tracelog
+
+disable-mnt
+private-bin url-eater
```
rusty-snake, line 47: Can we set `private`?

etc/profile-m-z/url-eater.profile
```diff
@@ -0,0 +1,59 @@
...
+noinput
+nonewprivs
+noprinters
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
```
rusty-snake, line 43: Seccimo.block-secondary

etc/profile-m-z/url-eater.profile
glitsj16: Yeah, didn't expect it would work when blacklisting that but it does. Tested several times in fact, rebooting/relogin, Xorg, Wayland, Sway and even in my old Compiz session the URL cleaning works just as advertised. Neat little app.

etc/profile-m-z/url-eater.profile
glitsj16: No its filtering functionality is broken when using `private`.

etc/profile-m-z/url-eater.profile
glitsj16: Ah yes, I keep forgetting that I have that in my globals.local. Will fix it asap. Thanks for the review!

etc/profile-m-z/url-eater.profile
rusty-snake: Did you make sure it did not fallback to XWayland?

glitsj16: Well, I used the example URL from the app's GitHub page in my Firefox (which definately runs on Wayland) and I copied that to several other apps, a mix of both Wayland and Xwayland apps. WFM, in both directions (copy from Xwayland app and paste in Wayland app and the other way around. I wouldn't mind taking the line out if you think that's safer, can always put it in my local's.

rusty-snake: I need to test the wayland clipboard interaction myself later.

glitsj16: No rush. I'll flip this PR into draft mode, schedule another round of testing and hopefully we can figure this all out before the next firejail release. I had 'clipboard privacy/security' on my radar lately and in that context I stumbled on to url-eater. Will try to dig through its source code to get a better understanding of how it works. Thanks for following-up on this topic!

kmk3:
(Offtopic)

FYI: The program appears to be under a non-free license:

* <https://github.com/AgathaSorceress/url-eater/blob/b728e7baaf5bb19f5115304402ea7ba66a9c24f3/LICENSE.md>

etc/profile-m-z/url-eater.profile
rusty-snake: My question is if you just block the wayland socket does it fallback to be an XWayland client and fail in Wayland-only or does it really work in Wayland w/o any (neither waland nor X11) socket.

kmk3:
(Squashed a few commits and then rebased to master)

kmk3 (approved):

PR #5795, opened 2023-04-20
ci: fix swapped name/email arguments in debian_ci

kmk3:
Currently, the author name is used as the author email in git and vice
versa.

Changes:

* Split the commands to make them easier to read
* Put the name command first to match the usual git usage
* Fix swapped name/email arguments
* Use the DEBFULLNAME / DEBEMAIL environment variables instead of
  GitLab-specific ones

Added on commit 4cf51b591 ("extend gitlab-ci configuration to run Debian
CI tests", 2019-06-21).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -83,7 +83,8 @@ debian_ci:
         - echo "deb-src http://deb.debian.org/debian sid main" >> /etc/apt/sources.list
         - echo "deb-src http://deb.debian.org/debian experimental main" >> /etc/apt/sources.list
         - apt-get update
-        - git config user.email "$GITLAB_USER_NAME" && git config user.name "$GITLAB_USER_EMAIL"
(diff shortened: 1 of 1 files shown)


kmk3:
@reinerh

Any idea if this could be causing any issues?

reinerh (approved):

reinerh:
Thanks, that looks good. I don't think that it will cause any problems.

kmk3:
> Thanks, that looks good.

Thanks.

> I don't think that it will cause any problems.

To be clear, I meant is there any problem if the already published packages
contain the wrong name/email?

Is this job used just for CI or is it used to generate the PPA packages?

reinerh:
> Is this job used just for CI or is it used to generate the PPA packages?

This job is just for CI.
PPA packages are uploaded as signed source packages to launchpad servers and are built there.

Issue #5098, opened 2022-04-07
Firejail disables pdf thumbnailing


MikeNavy:
Hi,
When I install Firejail, PDF thumbnailing is disabled; old PDF thumbnails stay there, but when a new PDF document is created / copied, there is no thumbnail created, iust a PDF icon.

**System**:
Linux Mint Mate 20.3 Una (based on Ubuntu 20.04 LTS Fossa, with Mate 1.26 desktop environment).
File Manager: caja
Thumbnailer: xplayer-thumbnailer "Thumbnailer for xreader".
**Firejail**:
Installed version "firejail_0.9.68-apparmor_1_amd64.deb" from [https://sourceforge.net/projects/firejail/files/firejail/firejail_0.9.68-apparmor_1_amd64.deb/download](url)

The problem occurs as soon as Firejail deb is installed, even without desktop integration (no use of "$ sudo firecfg").

**To reproduce the problem**:
- install Firejail,
- open file manager,
- copy / paste an existing PDF file,
- the copied file has no thumbnail, just a PDF icon.

**To correct the problem**:
- uninstall Firejail,
- open file manager,
- copy / paste an existing PDF file,
- the copied filed has a thumbnail.

It seems that Firejail prevents xplayer-thumbnailer to function.

MikeNavy:
Not sure if PDF thumbnails are made by xreader or by xplayer-thumbnailer; xplayer-thumbnailer is presented as "Thumbnailer for xreader", but it might be a typo instead of "Thumbnailer for xplayer".

rusty-snake:
> The problem occurs as soon as Firejail deb is installed, even without desktop integration (no use of "$ sudo firecfg").

Can you watch `sudo firemon` so wee which sandboxes are running/started when you test this.

MikeNavy:
Hi,
- I completely uninstalled Firejail.
- Then installed it again.
- I launched no sandbox, no "sudo firecfg".
- "sudo firemon" gives no answer, empty screen.
- Once again I copy / paste a PDF file, the copied file has a PDF icon, not a thumbnail.
- "sudo firemon" output still empty.

MikeNavy:
Hi,
I have finally found the root cause and solved the problem.

It seems that, in a former attempt to use Firejail, it had created symbolic links from "/usr/bin" to "/usr/local/bin". 
[...]

PR #5812, opened 2023-05-02
docs: add uninstall instructions to README.md

kmk3:
Clarify how to remove the symlinks created by firecfg.

Reported by @MikeNavy in #5098.


rusty-snake (approved):
LGTM

- The most users will not read it here either.
- We could provide a small bash command `for file in /usr/local/bin/*; do readlink == firejail -> rm $file; done`.

PR #5815, opened 2023-05-03
ci: run for every branch instead of just master

kmk3:
Having CI always run on WIP branches without having to open a PR
beforehand makes it easier to debug CI issues.

GitHub currently does not have any apparent limit for CI runs and there
are no project-specific secrets as far as I know, so it should be safe
to remove these restrictions.


rusty-snake (approved):

glitsj16 (approved):
LGTM

PR #5824, opened 2023-05-09
profiles: mcomix: allow exec in ${HOME}

ghost:
it's a python application, for some reason it want's exec on home or it can't find PIL
```
15:39:48 [MainThread] ERROR: Pillow 5.1.0 or higher is required.
15:39:48 [MainThread] ERROR: No version of the Pillow was found on your system.
```

I'm not sure what's going on, it doesn't really need exec on home. I have the impression that it used to work, then a new version of firejail broke it? It uses pickle for bookmarks and maybe that messes it up? I think it has it's own tests, to check if the right version of PIL is there.

### DIFF
--- a/etc/profile-m-z/mcomix.profile
+++ b/etc/profile-m-z/mcomix.profile
@@ -6,6 +6,8 @@ include mcomix.local
 # Persistent global definitions
 include globals.local
 
+ignore noexec ${HOME}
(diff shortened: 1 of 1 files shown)


rusty-snake:
Die you installed anything into $HOME via pip?

PR #5275, opened 2022-07-25
CI: bump ubuntu to 22.04 and use newer compilers / analyzers

reinerh:

PR #5829, opened 2023-05-14
modif: Improve --version/--help & print version on startup

kmk3:
It is not too uncommon for the firejail version to be missing when issues are
reported; this PR makes it more likely that any posted logs will contain the
program version.

Do so just for firejail and firecfg for now because they are the most common
user-facing programs.

Kind of relates to #5275.


glitsj16 (approved):
LGTM. Nice work!

netblue30:
all in, thanks!

PR #5847, opened 2023-06-06
update lobster profile

pirate486743186:
lobster went through a major rewrite.
ueberzugpp is a ueberzug alternative in C++, it's more like a library

.local/share/applications/lobster is just a symlink to /tmp, it doesn't need to be persistent. If it already exist it will be used.

### DIFF
--- a/etc/profile-a-l/lobster.profile
+++ b/etc/profile-a-l/lobster.profile
@@ -8,8 +8,14 @@ include lobster.local
 # added by included profile
 #include globals.local
 
+ignore private-dev
+
+noblacklist ${HOME}/.cache/ueberzugpp
 noblacklist ${HOME}/.config/lobster
+noblacklist ${HOME}/.config/ueberzugpp
+noblacklist ${HOME}/.local/share/applications/lobster
 noblacklist ${HOME}/.local/share/lobster
+noblacklist ${PATH}/openssl
 
 # Allow /bin/sh (blacklisted by disable-shell.inc)
 include allow-bin-sh.inc
@@ -19,7 +25,11 @@ include disable-xdg.inc
 
 mkdir ${HOME}/.config/lobster
 mkdir ${HOME}/.local/share/lobster
+whitelist ${DOWNLOADS}
+whitelist ${HOME}/.cache/ueberzugpp
 whitelist ${HOME}/.config/lobster
+whitelist ${HOME}/.config/ueberzugpp
+whitelist ${HOME}/.local/share/applications/lobster
 whitelist ${HOME}/.local/share/lobster
 include whitelist-run-common.inc
 include whitelist-runuser-common.inc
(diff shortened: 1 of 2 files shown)


glitsj16:
>  .local/share/applications/lobster is just a symlink to /tmp, it doesn't need to be persistent. If it already exist it will be used.

Have you checked if this breaks lobster's firecfg? Not sure, but we might have to drop it from firecfg.config it it does...

glitsj16 (commented):
LGTM. Just holding of approval for now until we have a better view on the mentioned firecfg situation.

etc/profile-a-l/lobster.profile
```diff
@@ -8,8 +8,14 @@ include lobster.local
 # added by included profile
 #include globals.local
 
+ignore private-dev
```
rusty-snake, line 11: Why?

etc/profile-a-l/lobster.profile
```diff
@@ -30,7 +40,7 @@ noprinters
 notv
 
 disable-mnt
-private-bin curl,cut,fzf,grep,head,lobster,mv,patch,rm,sed,sh,tail,tput,tr,uname
+private-bin base64,bash,cat,curl,cut,date,dirname,echo,ffmpeg,ffprobe,fzf,grep,head,hxunent,lobster,ls,mkdir,nano,nohup,openssl,patch,ps,rm,rofi,sed,sh,sleep,socat,tail,tee,tput,tr,ueberzugpp,uname,uuidgen,vim,wc
```
rusty-snake, line 43: Why `bash`?

etc/profile-a-l/lobster.profile
rusty-snake: A bit hacky. And very unsafe.

And does not work with Wayland I guess.

etc/profile-a-l/lobster.profile
rusty-snake: Ok, ueberzugpp. I only saw lobster which is a /bin/sh script.

glitsj16:
> rofi is a light weight app launcher. It expects to find apps in .local/share/applications . In practice people also use it as a light weight menu like fzf, with fake desktop apps. They are not real apps, they plug into itself...

@pirate486743186 Thanks for the explanation. All clear.

glitsj16 (approved):
LGTM

etc/profile-a-l/lobster.profile
```diff
@@ -30,7 +40,7 @@ noprinters
 notv
 
 disable-mnt
-private-bin curl,cut,fzf,grep,head,lobster,mv,patch,rm,sed,sh,tail,tput,tr,uname
+private-bin base64,bash,cat,curl,cut,date,dirname,echo,ffmpeg,ffprobe,find,fzf,grep,head,hxunent,lobster,ls,mkdir,mkfifo,nano,nohup,openssl,patch,pgrep,ps,rm,rofi,sed,sh,sleep,socat,tail,tee,tput,tr,ueberzugpp,uname,uuidgen,vim,wc
```
glitsj16, line 43: I'm seeing two additional commands in [lobster](https://github.com/justchokingaround/lobster/blob/main/lobster.sh) that are missing from private-bin:
- command
- ln

glitsj16 (changes requested):
private-bin additions

glitsj16 (approved):
LGTM

kmk3 (approved):

etc/profile-a-l/lobster.profile
```diff
@@ -8,8 +8,14 @@ include lobster.local
...
 #include globals.local
 
+ignore private-dev
+
+noblacklist ${HOME}/.cache/ueberzugpp
 noblacklist ${HOME}/.config/lobster
+noblacklist ${HOME}/.config/ueberzugpp
+noblacklist ${HOME}/.local/share/applications/lobster
 noblacklist ${HOME}/.local/share/lobster
+noblacklist ${PATH}/openssl
```
kmk3, line 18: Is this line actually needed?

It is only blacklisted in disable-devel.inc, which is not included in here.

Other than that, LGTM.

glitsj16: > It is only blacklisted in disable-devel.inc, which is not included in here.

It's included in mpv.profile, so we do need it here.

PR #5857, opened 2023-06-14
ci: standardize apt-get update/install & misc improvements

kmk3:
General changes:

* Use a single -q on update, as the output is not too long
* Use a single -q on install, to show all packages at once

GitLab-specific changes:

* Use `DEBIAN_FRONTEND=noninteractive` to reduce noise
* Use --no-install-recommends to avoid installing unnecessary packages
* Filter out uninteresting lines on install

Note: `DEBIAN_FRONTEND` does not appear to be needed in the default
GitHub runner container and not many packages are currently being
downloaded/installed in them, so do the above changes only jobs that use
custom Docker images.


glitsj16 (approved):
LGTM.

PR #5859, opened 2023-06-18
build: remove -mretpoline and NO_EXTRA_CFLAGS

kmk3:
The -mretpoline flag is not documented in the current versions of gcc
and clang and it is what causes scan-build to fail:

    $ ./configure CC=clang | grep retpoline
    checking whether C compiler accepts -mretpoline... yes
       EXTRA_CFLAGS:  -mretpoline -fstack-clash-protection -fstack-protector-strong
    $ scan-build --status-bugs make
    scan-build: Using '/usr/bin/clang-15' for static analysis
    make -C src/lib
    make[1]: Entering directory '/tmp/firejail/src/lib'
    /usr/bin/../lib/clang/ccc-analyzer [...] -mretpoline [...] -c common.c -o common.o
    gcc: error: unrecognized command-line option ‘-mretpoline’
    make[1]: *** [../../src/prog.mk:16: common.o] Error 1
    make[1]: Leaving directory '/tmp/firejail/src/lib'
    make: *** [Makefile:59: src/lib] Error 2
    scan-build: Analysis run complete.
    scan-build: Removing directory '/tmp/scan-build-[...]' because it contains no reports.
    scan-build: No bugs found.

Environment: clang 15.0.7-9 and gcc 13.1.1-1 on Artix Linux.

Note: NO_EXTRA_CFLAGS was added to work around this issue by causing all
of the flags in EXTRA_CFLAGS to be ignored.

Note2: -mretpoline was added on commit 4a99c8aa2 ("spectre support for
clang compiler", 2018-03-30) and NO_EXTRA_CFLAGS was added on commit
490918c35 ("fix make scan-build for debian 10 and arch", 2019-07-22).
See also commit 2c64d1fdd ("use AX_CHECK_COMPILE_FLAG to check for
spectre flags", 2019-06-21).

Closes #5509.

Kind of relates to #2661.

Cc: @netblue30 @reinerh

PR #5864, opened 2023-06-20
build: disable all built-in implicit make rules

kmk3:
To reduce unnecessary filesystem lookups.

Overall, this appears to reduce the amount of implicit rule searches by ~97.5%
for the default build and by ~99.3% for the "man" target (as an example):

    a8f01a383 2023-06-20 05:26:23 +0000 Merge pull request #5859 from kmk3/build-remove-retpoline
    $ ./configure >/dev/null
    $ make clean >/dev/null && make --debug=i -j 4     | grep -F 'Trying implicit' | wc -l
    6798
    $ make clean >/dev/null && make --debug=i -j 4 man | grep -F 'Trying implicit' | wc -l
    1085
    # (with the second commit applied)
    $ make clean >/dev/null && make --debug=i -j 4     | grep -F 'Trying implicit' | wc -l
    2535
    $ make clean >/dev/null && make --debug=i -j 4 man | grep -F 'Trying implicit' | wc -l
    42
    # (with the third commit applied)
    $ make clean >/dev/null && make --debug=i -j 4     | grep -F 'Trying implicit' | wc -l
    170
    $ make clean >/dev/null && make --debug=i -j 4 man | grep -F 'Trying implicit' | wc -l
    7

Environment: GNU make 4.4.1-2 on Artix Linux.

Note: According to make(1p) in POSIX.1-2017, "If .SUFFIXES does not have
any prerequisites, the list of known suffixes shall be cleared.", while
"The result of setting MAKEFLAGS in the Makefile is unspecified."

See also commit f48886f25 ("build: mark most phony targets as such",
2023-02-01) / PR #5637.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
 .SUFFIXES:
+MAKEFLAGS += -r
+
 ROOT = .
 -include config.mk
 
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+.SUFFIXES:
 ROOT = .
 -include config.mk
 
--- a/src/bash_completion/Makefile
+++ b/src/bash_completion/Makefile
@@ -1,3 +1,4 @@
+.SUFFIXES:
 ROOT = ../..
 -include $(ROOT)/config.mk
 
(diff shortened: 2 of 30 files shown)


glitsj16 (approved):
LGTM

PR #3526, opened 2020-07-19
Add vmware profile

Neo00001:

PR #5865, opened 2023-06-20
profiles: vmware improvements and new redirect

kmk3:
Add vmplayer.profile as a redirect to vmware.profile.

This is apparently the filename for the "VMWare Workstation Player" on
Linux Mint 20.3 (based on Ubuntu 20.04)[1].

Relates to #3526 #5861.

[1] https://github.com/netblue30/firejail/issues/5861#issuecomment-1598132860

Reported-by: @MikeNavy

### DIFF
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -895,6 +895,7 @@ vivaldi-beta
 vivaldi-snapshot
 vivaldi-stable
 vlc
+vmplayer
 vmware
 vmware-player
 vmware-workstation
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -895,10 +895,10 @@ vivaldi-beta
 vivaldi-snapshot
 vivaldi-stable
 vlc
-vmplayer
-vmware
-vmware-player
-vmware-workstation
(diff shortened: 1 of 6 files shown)


src/firecfg/firecfg.config
```diff
@@ -895,6 +895,7 @@ vivaldi-beta
 vivaldi-snapshot
 vivaldi-stable
 vlc
+vmplayer
```
glitsj16, line 898: If VMWare usage under Firejail shows the `kernel modules` problem like #5861 suggests, IMO it would make more sense to `disable` the related commands with a note. Suppose a package manager hook for firejail runs firecfg post-install. Some time later a user installs VMWare. When auto-sandboxed, its kernel modules never actually get installed. GUI-minded users won't have much indication of what's going on...

glitsj16 (commented):
I'm having doubts as to enable VMWare commands in firecfg when its kernel modules can't be installed properly cfr. #5861.

src/firecfg/firecfg.config
kmk3: Done.

glitsj16 (approved):
LGTM

PR #5876, opened 2023-07-01
feature: firecfg: add firecfg.d & add ignore command

### LINKED ISSUE #2097: firecfg: allow for ignoring specific apps

kmk3:
Main changes:

* firecfg: parse config files in /etc/firejail/firecfg.d

`/etc/firejail/firecfg.d/*.conf` files are parsed before
/etc/firejail/firecfg.config, so the former can ignore/override any item in the
latter.

* Add ignore command (`!PROGRAM`)

It prevents firecfg from creating a symlink for the given program.

Also, document the paths used and the config file syntax.

Suggested by @WhyNotHugo:

* <https://github.com/netblue30/firejail/issues/2097#issuecomment-1179160459>

Closes #2097.


kmk3:
A few considerations:

Instead of `!PROGRAM`, I thought about using `ignore PROGRAM` for better
readability and to be consistent with firejail, but `ignore` works slightly
differently in firejail:

* <https://github.com/netblue30/firejail/blob/33c75b89328df03ef3245c7ec6f30759f9619223/src/firejail/profile.c#L103-L119>

It only compares up to the first space, so it considers `ignore foo` and
`ignore foo bar` to be equivalent, for example.  That makes sense for ignoring
commands with arguments (like `ignore name` ignores `name foo`), but not so
much for paths.

To avoid confusion (if a program name has a space in it for whatever reason),
it seems like a good idea to use a different name/syntax.

Other than that, I'd say it's fairly straightforward.

src/firecfg/firecfg.h
```diff
@@ -37,6 +37,15 @@
...
 
+// config files
+#define FIRECFG_CFGFILE SYSCONFDIR "/firecfg.config"
+#define FIRECFG_CONF_GLOB SYSCONFDIR "/firecfg.d/*.conf"
+
+// programs
+#define FIREJAIL_EXEC PREFIX "/bin/firejail"
+#define FIREJAIL_WELCOME_SH LIBDIR "/firejail/firejail-welcome.sh"
+#define FZENITY_EXEC        LIBDIR "/firejail/fzenity"
+#define ZENITY_EXEC "/usr/bin/zenity"
```
glitsj16, line 48: Looks like this adds a dependency on `zenity` to firejail, correct? Not that I'd mind, but package maintainers might need to get informed on this. Usually adding something to release notes takes care of such a change. Just a FYI.

glitsj16 (commented):
We might need to make it explicit in release notes that this introduces a dependency on zenity to the firejail package.

kmk3:
> We might need to make it explicit in release notes that this introduces a
> dependency on zenity to the firejail package.

The zenity-related code is already there and is only executed if `--guide` is
used:

* <https://github.com/netblue30/firejail/blob/33c75b89328df03ef3245c7ec6f30759f9619223/src/firecfg/main.c#L449-L464>

This PR only moves such paths to constants; see commit 1f07de85c ("firecfg:
turn constant strings into constants", 2023-06-30).

glitsj16:
@kmk3 Thank you for clearing up this zenity situation in such a detailed manner.

netblue30:
Cool job! If you don't have anything coming, check it in!

PR #5885, opened 2023-07-05
fix lobster.profile

pirate486743186:
it needs to write in `.local/share/applications` on first run.
It has to be at the end of the profile to take effect.

### DIFF
--- a/etc/profile-a-l/lobster.profile
+++ b/etc/profile-a-l/lobster.profile
@@ -9,6 +9,7 @@ include lobster.local
 #include globals.local
 
 ignore private-dev
+ignore read-only ${HOME}/.local/share/applications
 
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/lobster.profile
```diff
@@ -47,3 +47,5 @@ private-tmp
 
 # Redirect
 include mpv.profile
+
+read-write ${HOME}/.local/share/applications
```
rusty-snake, line 51: Or just use `ignore read-only ...`. Then it does not have to be after the `include`.

rusty-snake (approved):

kmk3 (approved):

netblue30:
merged, thanks!

PR #5886, opened 2023-07-08
torbrowser-launcher: hardening

glitsj16:


etc/profile-m-z/torbrowser-launcher.profile
```diff
@@ -53,12 +53,14 @@ nou2f
...
 protocol unix,inet,inet6
 seccomp !chroot
+seccomp.block-secondary
 #tracelog - may cause issues, see #1930
 
 disable-mnt
 private-bin bash,cat,cp,cut,dirname,env,expr,file,gpg,grep,gxmessage,id,kdialog,ln,mkdir,mv,python*,rm,sed,sh,tail,tar,tclsh,test,tor-browser,tor-browser-en,torbrowser-launcher,update-desktop-database,xmessage,xz,zenity
 private-dev
 private-etc @tls-ca
+private-opt tor-browser
```
rusty-snake, line 63: Can cause slow startup.

Breaks if any other name is used (e.g. torbrowser-launcher).

etc/profile-m-z/torbrowser-launcher.profile
```diff
@@ -53,12 +53,14 @@ nou2f
...
 private-tmp
 
 dbus-user none
```
rusty-snake, line 66: Some namespaces can be restricted.

etc/profile-m-z/torbrowser-launcher.profile
```diff
@@ -33,9 +33,9 @@ whitelist ${HOME}/.local/share/torbrowser
 whitelist /opt/tor-browser
 whitelist /usr/share/torbrowser-launcher
 include whitelist-common.inc
-include whitelist-var-common.inc
 include whitelist-runuser-common.inc
 include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
```
rusty-snake, line 38: Maybe we can include whitelist run.

etc/profile-m-z/torbrowser-launcher.profile
```diff
@@ -33,9 +33,9 @@ whitelist ${HOME}/.local/share/torbrowser
 whitelist /opt/tor-browser
 whitelist /usr/share/torbrowser-launcher
```
rusty-snake, line 34: Maybe we can include disable process.

rusty-snake (commented):
More hardening possible if you want. Here and on mullvad.

etc/profile-m-z/torbrowser-launcher.profile
glitsj16: Better revert it in that case. Thanks for the info!

etc/profile-m-z/torbrowser-launcher.profile
glitsj16: Tested, we can add it.

etc/profile-m-z/torbrowser-launcher.profile
glitsj16: Doesn't seem to cause any issues. Let's add it!

glitsj16:
@kmk3 @rusty-snake If you find the time, can you have another look at this? I think it's okay now. Same for #5887.

etc/profile-m-z/torbrowser-launcher.profile
```diff
@@ -53,15 +55,17 @@ nou2f
...
 private-dev
 private-etc @tls-ca
+#private-opt tor-browser - can cause slow startup
 private-tmp
 
 dbus-user none
 dbus-system none
 
-#restrict-namespaces
+restrict-namespaces
```
rusty-snake, line 71: Weakens torbrowsers sandbox?

kmk3 (approved):

PR #5887, opened 2023-07-08
New profile: mullvad-browser

glitsj16:
Add new profile for [mullvad-browser](https://mullvad.net/en/download/browser/linux). Requested [here](https://github.com/netblue30/firejail/issues/1139#issuecomment-1500052260) and suggested [here](https://github.com/netblue30/firejail/issues/1139#issuecomment-1500125403).

### DIFF
--- a/etc/profile-m-z/mullvad-browser.profile
+++ b/etc/profile-m-z/mullvad-browser.profile
@@ -1,0 +1,97 @@
+# Firejail profile for mullvad-browser
+# Description: Privacy-focused web browser developed in a collaboration between Mullvad VPN and the Tor Project
+# This file is overwritten after every install/update
+# Persistent local customizations
+include mullvad-browser.local
+# Persistent global definitions
+include globals.local
+
+# IMPORTANT ##########################################
+# The mullvad-browser can be downloaded from the official website
+# and installed manually or via the AUR for Arch Linux (derivatives).
+# The latter installs the browser under /opt/mullvad-browser, while
+# the former can be installed under ${HOME} just about anywhere.
+# If you decide to install it under ${HOME} this profile assumes to find
+# the browser files under ${HOME}/.local/share/mullvad-browser.
+# When you divert from that location you will need to make the needed
+# path adjustments yourself in the below instructions.
+####################################################
+
+# If you installed under ${HOME}, put the below line in your
+# mullvad-browser.local
+# Note: The relevant rule in /etc/apparmor.d/local/firejail-default will
+# need to be uncommented for the 'apparmor' option to work as expected.
+#ignore noexec ${HOME}
+
+noblacklist ${HOME}/.cache/mullvad/mullvadbrowser
+noblacklist ${HOME}/.config/mullvad-browser-flags.conf
+noblacklist ${HOME}/.local/share/mullvad-browser
+noblacklist ${HOME}/.mullvad/mullvadbrowser
+
+# Allow python 3 (blacklisted by disable-interpreters.inc)
+include allow-python3.inc
+
+blacklist /srv
+blacklist /sys/class/net
+blacklist /usr/libexec
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.cache/mullvad/mullvadbrowser
+mkdir ${HOME}/.local/share/mullvad-browser
+mkdir ${HOME}/.mullvad/mullvadbrowser
+mkfile ${HOME}/.config/mullvad-browser-flags.conf
+whitelist ${DOWNLOADS}
+whitelist ${HOME}/.cache/mullvad/mullvadbrowser
+whitelist ${HOME}/.config/mullvad-browser-flags.conf
+whitelist ${HOME}/.local/share/mullvad-browser
+whitelist ${HOME}/.mullvad/mullvadbrowser
+whitelist /opt/mullvad-browser
+include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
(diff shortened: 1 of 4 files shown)


etc/profile-m-z/mullvad-browser.profile
```diff
@@ -0,0 +1,76 @@
...
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-xdg.inc
+
+mkfile ${HOME}/.config/mullvad-browser-flags.conf
+# If you installed under ${HOME}, put the below lines in your mullvad-browser.local
+#mkdir ${HOME}/.local/share/mullvad-browser
+#whitelist ${HOME}/.local/share/mullvad-browser
```
kmk3, line 33: Can be moved below `#noblacklist ${HOME}/.local/share/mullvad-browser` to avoid
duplicated comments.

kmk3 (commented):
The only config path used is ~/.config/mullvad-browser-flags.conf.

Where is the user profile stored?  Is there no user profile directory?

etc/profile-m-z/mullvad-browser.profile
glitsj16: Fixed, thanks!

glitsj16:
@kmk3

> The only config path used is ~/.config/mullvad-browser-flags.conf.
> Where is the user profile stored? Is there no user profile directory?

Mullvad Browser stores everything under one directory, including its cache and profile dir. The [official download](https://mullvad.net/en/download/browser/linux) installs under ${HOME}/.local/share/mullvad-browser. The packaged versions in the [AUR](https://aur.archlinux.org/packages?O=0&SeB=nd&K=mullvad-browser&outdated=&SB=p&SO=d&PP=50&submit=Go) move everything into /opt/mullvad-browser. Both those paths are whitelisted. I didn't want to explicitly blacklist /opt/mullvad-browser in disable-programs.inc, nothing from /opt is in there. Do you think we should?

kmk3:
> > The only config path used is ~/.config/mullvad-browser-flags.conf.  Where
> > is the user profile stored? Is there no user profile directory?
> 
> Mullvad Browser stores everything under one directory, including its cache
> and profile dir. The [official
> download](https://mullvad.net/en/download/browser/linux) installs under
> ${HOME}/.local/share/mullvad-browser. The packaged versions in the
> [AUR](https://aur.archlinux.org/packages?O=0&SeB=nd&K=mullvad-browser&outdated=&SB=p&SO=d&PP=50&submit=Go)
> move everything into /opt/mullvad-browser.

Are you sure?

From <https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mullvad-browser-bin&id=4aabf2f8085b4712b9cf32eec09a9040f2184ef1#n47>:

```sh
# copy files from archive
install -dm0755 "$pkgdir/opt/mullvad-browser/"
cp --archive Browser/* "$pkgdir/opt/mullvad-browser/"

# ask it to create profiles in ~
touch "$pkgdir/opt/mullvad-browser/system-install"
```

Also, how would it handle multiple users in that case?  Does it create a
separate /home directory in /opt?  Does the install only work for a single
user?  Putting user data outside of the user home seems like a bad idea in any
case.

> Both those paths are whitelisted.

I'm not sure if you're only talking about /opt, but currently only the .conf
file is whitelisted in `${HOME}`; there is no data directory.

> I didn't want to explicitly blacklist
> /opt/mullvad-browser in disable-programs.inc, nothing from /opt is in there.
> Do you think we should?

Well, disable-programs is mostly just configuration/data paths, but if it
indeed can put user data in /opt then blacklisting makes sense to me.

glitsj16:
@kmk3 Good questions. I'll build those AUR packages again tomorrow and do more extensive testing guided by your remarks/questions. Thanks!

glitsj16:
@kmk3 UPDATE: there were indeed `very significant` changes to installation paths and user settings locations compared to when I initially created my version of mullvad-browser.profile! Thanks for (gently) pushing me to review this situation. I've tried to be as verbose in the newly introduced comments as I thought fit to clearly state the differences for both a system-wide installation under /opt and a local installation under ${HOME}. Feel free to make suggestions.

Some additional remarks for reviewers. My assumption of an install under ${HOME}/.local/share/mullvad-browser is based on what we do in `torbrowser-launcher.profile` and is tested intensively (it's where I have my copy of mullvad-browser installed since it became available). The blacklisting of `${HOME}/UpdateInfo` in `disable-programs.inc` _without_ noblacklisting that path anywhere else needs some clarification. That specific location is only used when installing under /opt. More correctly, when not installing under ${HOME}. Upgrading such a mullvad-browser installation from within the browser will not work anyway - obviously due to the user not having privileges to install files under /opt.

Nothing of the above is set in stone, so feel free to make suggestions!

glitsj16:
@kmk3 @rusty-snake If you find the time, can you have another look at this? I think it's okay now. Same for #5886.

etc/profile-m-z/mullvad-browser.profile
```diff
@@ -0,0 +1,103 @@
...
+#ignore noexec ${HOME}
+#noblacklist ${HOME}/.local/share/mullvad-browser
+#mkdir ${HOME}/.local/share/mullvad-browser
+#whitelist ${HOME}/.local/share/mullvad-browser
+#ignore noblacklist ${HOME}/.cache/mullvad/mullvadbrowser
+#ignore noblacklist ${HOME}/.mullvad/mullvadbrowser
+#ignore mkdir ${HOME}/.cache/mullvad/mullvadbrowser
+#ignore mkdir ${HOME}/.mullvad/mullvadbrowser
+#nowhitelist ${HOME}/.cache/mullvad/mullvadbrowser
+#nowhitelist /opt/mullvad-browser
```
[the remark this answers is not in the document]
glitsj16: That's way more elegant! Thanks for the suggestions.

kmk3 (approved):

PR #5898, opened 2023-07-13
build: simplify code related to man pages

kmk3:
Simplify the main targets and use wildcards instead of repeating the
filenames manually.

Also, restore the `man` target and building only when `HAVE_MAN` is
enabled.

Note: Make automatically removes intermediate files (.1 and .5), so only
the .gz files have to be cleaned.

Commands used to rename the man pages:

    cd src/man
    git mv firecfg.txt firecfg.1.in
    git mv firejail-login.txt firejail-login.5.in
    git mv firejail-profile.txt firejail-profile.5.in
    git mv firejail-users.txt firejail-users.5.in
    git mv firejail.txt firejail.1.in
    git mv firemon.txt firemon.1.in
    git mv jailcheck.txt jailcheck.1.in

This is kind of a follow-up to commit 9e206b7f2 ("rework src/man
Makefile", 2023-07-07).


netblue30:
merging, thanks!

PR #5905, opened 2023-07-17
profiles: fix misc in kmail and transmission-qt & add kontact.profile

Sadoon-AlBader:
This removes "nosound" from kmail.profile to allow notification sounds, and also adds a "dbus-user.talk org.freedesktop.Notifications" to allow notifications to start.
Also adds kontact.profile based on kmail.profile, have been using it and works well.
Also edits transmission-qt.profile to allow systray, commented out as is usually the case with profiles, but better to at least let the user know that it might be necessary to add to .local if they need systray.

EDIT: if the maintainers also prefer to keep "nosound" for kmail, it might make sense to also add it to kontact.profile since I didn't add it there to start, I think it makes sense to either remove it or add a comment letting the user know that sound notifications won't work unless they "ignore nosound"


etc/profile-a-l/kontact.profile
```diff
@@ -0,0 +1,72 @@
+# Firejail profile for kontact
```
rusty-snake, line 1: Firecfg?

etc/profile-m-z/transmission-qt.profile
```diff
@@ -15,6 +15,9 @@ private-bin transmission-qt
 #dbus-user.own com.transmissionbt.Transmission.*
 #dbus-user.talk org.freedesktop.Notifications
 
+# If you need system tray support, add this line to your transmission-qt.local
```
rusty-snake, line 18: I would just enabled it with the condition.

etc/profile-a-l/kmail.profile
```diff
@@ -63,4 +62,7 @@ private-dev
 # writable-run-user is needed for signing and encrypting emails
 writable-run-user
 
+# enable desktop notifications
+dbus-user.talk org.freedesktop.Notifications
```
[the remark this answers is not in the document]
Sadoon-AlBader: In this case should the user be prompted to add both that line and `dbus-user filter` to their .local?

etc/profile-a-l/kontact.profile
Sadoon-AlBader: I'll add that, thanks for reminding me, added to my system but forgot to do so here.

etc/profile-m-z/transmission-qt.profile
Sadoon-AlBader: Will fix, thanks.

Sadoon-AlBader:
@rusty-snake All your suggestions are now implemented, thanks again :)

kmk3 (changes requested):
The new profile is nearly identical to kmail.profile; it should probably be a
redirect to the latter.

See redirect_alias-profile.template.

etc/profile-a-l/kmail.profile
```diff
@@ -63,4 +62,8 @@ private-dev
 # writable-run-user is needed for signing and encrypting emails
 writable-run-user
 
+# to enable desktop notifications, add these commented lines to your kmail.local file
+#dbus-user.talk org.freedesktop.Notifications
+#dbus-user filter
```
rusty-snake, line 67:
```suggestion
#dbus-user filter
#dbus-user.talk org.freedesktop.Notifications
```

etc/profile-a-l/kmail.profile
```diff
@@ -63,4 +62,8 @@ private-dev
 # writable-run-user is needed for signing and encrypting emails
 writable-run-user
 
+# to enable desktop notifications, add these commented lines to your kmail.local file
```
rusty-snake, line 65:
```suggestion
# To harden kmail.profile, add the following lines to your kmail.local file
```

rusty-snake (commented):
Now I had time for the full review.

PR #5910, opened 2023-07-19
profiles: remove /usr/share/vulkan already whitelisted by wusc

glitsj16:
No need for whitelisting /usr/share/vulkan twice in profiles that include wusc. We already do it there:

https://github.com/netblue30/firejail/blob/9863f982ca148f687a56690e2aeed10014d05e59/etc/inc/whitelist-usr-share-common.inc#L70

### DIFF
--- a/etc/profile-a-l/firefox-common-addons.profile
+++ b/etc/profile-a-l/firefox-common-addons.profile
@@ -74,7 +74,6 @@ whitelist ${HOME}/.zotero
 whitelist ${HOME}/dwhelper
 whitelist /usr/share/lua
 whitelist /usr/share/lua*
-whitelist /usr/share/vulkan
 
 # GNOME Shell integration (chrome-gnome-shell) needs dbus and python
 noblacklist ${HOME}/.local/share/gnome-shell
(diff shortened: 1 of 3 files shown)


kmk3 (approved):

etc/profile-a-l/firefox-common-addons.profile
```diff
@@ -74,7 +74,6 @@ whitelist ${HOME}/.zotero
 whitelist ${HOME}/dwhelper
 whitelist /usr/share/lua
 whitelist /usr/share/lua*
-whitelist /usr/share/vulkan
```
glitsj16, line 77: @kmk3

IMO we should keep `whitelist /usr/share/vulkan` here. The include logic is a bit more complicated than usual. Only `firefox.profile` includes `wusc`, most of the other profiles for Firefox-based web browsers do NOT:

- abrowser
- basilisk
- cliqz
- cyberfox
- firedragon
- icecat
- palemoon
- waterfox

We could add `wusc` in the above profiles to fix things, but we'd have to test it on all of these web browsers to make sure. Not easily done (time-wise).

glitsj16: No worries, that can happen. It caught my eye because I'm working on a branch to streamline/fix/unclutter this include logic in our Firefox-based web browser profiles. It's a bit of a mess IMO currently and mistakes like this can easily happen to anyone. Thanks for fixing!

PR #5913, opened 2023-07-23
thunderbird: D-Bus hardening

glitsj16:
The `Enigmail` add-on was replaced with built-in support for OpenPGP in Thunderbird 78. Development has [stopped](https://enigmail.net/index.php/en/home/news/70-2019-10-08-future-openpgp-support-in-thunderbird) in 2019. IMO it's time to drop allowing full access to the user and system bus and introduce selective filtering.

### DIFF
--- a/etc/profile-m-z/thunderbird.profile
+++ b/etc/profile-m-z/thunderbird.profile
@@ -8,9 +8,17 @@ include globals.local
 
 ignore include whitelist-runuser-common.inc
 
-# writable-run-user and dbus are needed by enigmail
+# TB stopped supporting enigmail in 2020 (v78) - let's harden D-Bus
+# https://support.mozilla.org/en-US/kb/openpgp-thunderbird-howto-and-faq
 ignore dbus-user none
-ignore dbus-system none
+dbus-user filter
+dbus-user.own org.mozilla.thunderbird.*
+dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.freedesktop.Notifications
+# allow D-Bus communication with firefox for opening links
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/thunderbird.profile
```diff
@@ -8,9 +8,17 @@ include globals.local
...
-ignore dbus-system none
+dbus-user filter
+dbus-user.own org.mozilla.thunderbird.*
+dbus-user.talk ca.desrt.dconf
+dbus-user.talk org.freedesktop.Notifications
+# allow D-Bus communication with firefox for opening links
+dbus-user.talk org.mozilla.*
+# e2ee email needs writable-run-user
+# https://support.mozilla.org/en-US/kb/introduction-to-e2e-encryption
 writable-run-user
```
rusty-snake, line 22: What about wrutable-run-user?

glitsj16: Not sure what you mean. I've tested the latest TB v115. End-to-end encryption is supported but users need to enable it and TB can create keys for each of their accounts. Encrypting and signing email does need `writable-run-user`. That didn't change by dropping enigmail support and integrating OpenPGP and S/MIME into TB. The same goes for claws-mail for example, which is why our email-common.profile has writable-run-user too.

Is there anything else you wanted to point out regarding wru?

rusty-snake: Was just asking because of the old comment.

glitsj16: Ow okay. No problem.

PR #5917, opened 2023-07-24
mpv: 0.36.0 requires system DBus access

polyzen:
Maybe could also use `dbus-system.talk org.freedesktop.DBus org.freedesktop.RealtimeKit1`. Not really sure how this works.

### DIFF
--- a/etc/profile-m-z/mpv.profile
+++ b/etc/profile-m-z/mpv.profile
@@ -86,6 +86,6 @@ private-bin env,mpv,python*,waf,youtube-dl,yt-dlp
 private-dev
 
 dbus-user none
-dbus-system none
+dbus-system filter
 
 restrict-namespaces


etc/profile-m-z/mpv.profile
```diff
@@ -86,6 +86,6 @@ private-bin env,mpv,python*,waf,youtube-dl,yt-dlp
 private-dev
 
 dbus-user none
-dbus-system none
+dbus-system filter
```
glitsj16, line 89: Applications very rarely need access to the system bus. How did you determine the newly released mpv 0.36.0 needs this? In the mean time I'll install the mpv package and do some testing. Thanks for opening the PR, we should definately check what's going on.

glitsj16 (commented):
Let's check what's going on with mpv 0.36.0 regarding D-Bus. Allowing access to the system bus isn't something we like to implement unless there's no other option(s).

etc/profile-m-z/mpv.profile
polyzen: Without it:
```
dbus[60]: arguments to dbus_message_new_method_call() were incorrect, assertion "path != NULL" failed in file dbus-message.c line 1373.
This is normally a bug in some application using the D-Bus library.

```

glitsj16: Cannot reproduce on my Arch Linux box. There have been a few changes to the mpv.profile after releasing 0.9.72 though. If you're on that version, check [this commit](https://github.com/netblue30/firejail/commit/2c5119d751cc78e52922544a52a0332f59304639) and try changing the seccomp line accordingly. Or try to reproduce on a [firejail-git](https://github.com/netblue30/firejail/wiki/Using-firejail-from-git) install.

Also, please open an `issue` on this problem you're facing. Properly fill in the issue template and provide as much details as you can, including a full log running with the --debug flag. I do realize this is a bit of a pain, but without eyes on your exact setup it's hard to pinpoint what's happening on your side.

polyzen:
On July 24, 2023 10:22:06 PM EDT, glitsj16 ***@***.***> wrote:
> @glitsj16 commented on this pull request.
> 
> 
> 
> > @@ -86,6 +86,6 @@ private-bin env,mpv,python*,waf,youtube-dl,yt-dlp
>  private-dev
>  
>  dbus-user none
> -dbus-system none
> +dbus-system filter
> 
> Cannot reproduce on my Arch Linux box. There have been a few changes to the mpv.profile after releasing 0.9.72 though. If you're on that version, check [this commit](https://github.com/netblue30/firejail/commit/2c5119d751cc78e52922544a52a0332f59304639) and try changing the seccomp line accordingly. Or try to reproduce on a [firejail-git](https://github.com/netblue30/firejail/wiki/Using-firejail-from-git) install.
> 
> Also, please open an `issue` on this problem you're facing. Properly fill in the issue template and provide as much details as you can, including a full log running with the --debug flag. I do realize this is a bit of a pain, but without eyes on your exact setup it's hard to pinpoint what's happening on your side.
> 

Ah right I forgot to test the latest commit.

--
Best,
Daniel <https://danielcapella.com>

rusty-snake:
Dropping the words "plug-ins" / "user-scripts".

kmk3:
> Dropping the words "plug-ins" / "user-scripts".

What do you mean?

rusty-snake:
@ Anyone who can reproduce (is on Arch with mpv 0.36.0). Assuming `firejail --noprofile --dbus-system=none /usr/bin/mpv` fails. What's the behaviour of
- `firejail --noprofile --dbus-system=none --blacklist=/run/dbus /usr/bin/mpv`
- `firejail --noprofile --dbus-system=none --whitelist=/run/media /usr/bin/mpv`

kmk3:
> @ Anyone who can reproduce (is on Arch with mpv 0.36.0). Assuming `firejail
> --noprofile --dbus-system=none /usr/bin/mpv` fails. What's the behaviour of
> 
> * `firejail --noprofile --dbus-system=none --blacklist=/run/dbus /usr/bin/mpv`
> * `firejail --noprofile --dbus-system=none --whitelist=/run/media /usr/bin/mpv`

Same error in both cases.

rusty-snake:
So it has a hard-dependency on D-Bus (system bus) without usefull error message. And the package metadata miss it too?

kmk3:
> So it has a hard-dependency on D-Bus (system bus) without usefull error
> message. And the package metadata miss it too?

Yes.

pipewire claims that this is just a bugfix release:

> ```
> # PipeWire 0.3.75 (2023-07-21)
> 
> This is a bugfix release that is API and ABI compatible with previous
> 0.3.x releases.
> ```

So the hard-dependency is probably a regression.

kmk3:
@polyzen Given that pipewire fixed the bug (as mentioned in #5921) and that
there is a workaround, can this be closed?

PR #5927, opened 2023-07-28
build: fix some shellcheck issues & use config.sh in more scripts

kmk3:
This removes the need to manually pass variables such as `$(TARNAME)`
and `$(VERSION)` to shell scripts in the root Makefile.

Relates to #5140.

### DIFF
--- a/platform/rpm/mkrpm.sh
+++ b/platform/rpm/mkrpm.sh
@@ -8,18 +8,18 @@
 # Builds rpms in a temporary directory then places the result in the
 # current working directory.
 
-name=$1
+name="$1"
 # Strip any trailing prefix from the version like -rc1 etc
-version=$(echo "$2" | sed 's/\-.*//g')
-config_opt=$3
+version="$(printf '%s\n' "$2" | sed 's/\-.*//g')"
+config_opt="$3"
 
-if [[ ! -f platform/rpm/${name}.spec ]]; then
-    echo error: spec file not found for name \"${name}\"
+if [[ ! -f "platform/rpm/${name}.spec" ]]; then
+    printf 'error: spec file not found for name %s\n' "${name}" >&2
     exit 1
 fi
 
 if [[ -z "${version}" ]]; then
-    echo error: version must be given
+    printf 'error: version must be given\n' >&2
     exit 1
 fi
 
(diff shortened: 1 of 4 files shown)


platform/rpm/mkrpm.sh
```diff
@@ -3,23 +3,26 @@
...
 
-name=$1
+# shellcheck source=config.sh
+. "$(dirname "$0")/../../config.sh" || exit 1
+
+name="$TARNAME"
 # Strip any trailing prefix from the version like -rc1 etc
-version=$(echo "$2" | sed 's/\-.*//g')
-config_opt=$3
+version="$(printf '%s\n' "$VERSION" | sed 's/\-.*//g')"
```
rusty-snake, line 16: Just wondering what's the benefit of `printf` over plain `echo` here is?

rusty-snake (approved):

kmk3:
@rusty-snake [on Jul 28](https://github.com/netblue30/firejail/pull/5927#pullrequestreview-1552553493):

> Just wondering what's the benefit of `printf` over plain `echo` here is?

Escape sequences (and certain flags) may be treated differently by different
`echo` implementations, so `printf` is better for portability, especially if
external input is involved.  In this specific case it's likely not an issue,
but it's simpler to just use `printf` basically everywhere rather than having
to check each `echo` invocation for potential issues.

From `echo(1p)`:

```
OPERANDS
       The following operands shall be supported:

       string    A string to be written to standard output. If the first
                 operand is -n, or if any of the operands contain a
                 <backslash> character, the results are implementation-
                 defined.
[...]

APPLICATION USAGE
       It is not possible to use echo portably across all POSIX systems unless
       both -n (as the first argument) and escape sequences are omitted.

       The printf utility can be used portably to emulate any of the
       traditional behaviors of the echo utility as follows (assuming that IFS
       has its standard value or is unset):
[...]

RATIONALE
       The echo utility has not been made obsolescent because of its extremely
       widespread use in historical applications. Conforming applications that
       wish to do prompting without <newline> characters or that could
       possibly be expecting to echo a -n, should use the printf utility
       derived from the Ninth Edition system.

       As specified, echo writes its arguments in the simplest of ways. The
       two different historical versions of echo vary in fatally incompatible
       ways.

       The BSD echo checks the first argument for the string -n which causes
       it to suppress the <newline> that would otherwise follow the final
       argument in the output.

       The System V echo does not support any options, but allows escape
       sequences within its operands, as described for XSI implementations in
       the OPERANDS section.

       The echo utility does not support Utility Syntax Guideline 10 because
       historical applications depend on echo to echo all of its arguments,
       except for the -n option in the BSD version.
```

rusty-snake:
Learned, Thanks.

Have to say that a `#!/bin/bash` script for a program that fundamentally relies on Linux API will hardly never see an other implementation.

kmk3:
@rusty-snake [on Jul 28](https://github.com/netblue30/firejail/pull/5927#issuecomment-1656349854):

> Learned, Thanks.

No problem.

> Have to say that a `#!/bin/bash` script for a program that fundamentally
> relies on Linux API will hardly never see an other implementation.

Note that even in bash, with `echo` you need to make sure that the arguments do
not start with `-` nor contain escape sequences, while with `printf` you can
just use one or more `%s` in the format string and put the variables afterwards
(which is arguably easier to check for correctness in most cases).

It also helps with consistency with `/bin/sh` scripts, which may point to a
shell with an incompatible `echo` implementation.

I'm not sure what is the relation of "a `#!/bin/bash` script" and "a program
that fundamentally relies on Linux API" (if you're referring to firejail), as
firejail can be used in distributions where bash is not the default shell (such
as Alpine) and considering that the script in question does not make
significant use of bash-specific features.

PR #5934, opened 2023-07-31
New profile: reader

glitsj16:
https://github.com/mrusme/reader

### DIFF
--- a/etc/profile-m-z/reader.profile
+++ b/etc/profile-m-z/reader.profile
@@ -1,0 +1,64 @@
+# Firejail profile for reader
+# Description: Better readability of web pages on the CLI
+# This file is overwritten after every install/update
+# Persistent local customizations
+include reader.local
+# Persistent global definitions
+include globals.local
+
+blacklist ${RUNUSER}/wayland-*
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/reader.profile
```diff
@@ -0,0 +1,64 @@
+# Firejail profile for reader
+# Description: Better readability of web pages on the CLI
+# This file is overwritten after every install/update
+# Persistent local customizations
+include reader.local
+# Persistent global definitions
+include globals.local
+
+blacklist ${RUNUSER}/wayland-*
```
rusty-snake, line 9: Can we blacklist entire RUNUSER?

etc/profile-m-z/reader.profile
```diff
@@ -0,0 +1,64 @@
...
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-X11.inc
```
rusty-snake, line 18:
```suggestion
```

`x11 none`

etc/profile-m-z/reader.profile
glitsj16: Indeed we can. Thanks for the suggestion!

etc/profile-m-z/reader.profile
glitsj16: Done. Thanks!

PR #5942, opened 2023-08-05
build: firecfg.config sorting improvements

kmk3:
Remove the space after `#` for commented code and use `#` instead of `-`
for comments at the end of the line.

Currently the CI check does not consider certain special characters
(such as `-`) when sorting due to `sort -d`.

So remove `-d`, sort firecfg using `LC_ALL=C` and enforce that order.

Also add `sort -u` to check for duplicates.

This also allows the CI check to ignore normal comments (lines starting
with `# `) anywhere in the file.

Relates to #4643.


rusty-snake (approved):

glitsj16 (approved):
LGTM. I've added the comment format to my notes to avoid 'old' habits :)

PR #5943, opened 2023-08-06
New profile: journal viewer

glitsj16:
journal-viewer
Visualize systemd logs
https://github.com/mingue/journal-viewer

### DIFF
--- a/etc/profile-a-l/journal-viewer.profile
+++ b/etc/profile-a-l/journal-viewer.profile
@@ -1,0 +1,68 @@
+# Firejail profile for journal-viewer
+# Description: Visualize systemd logs
+# This file is overwritten after every install/update
+# Persistent local customizations
+include journal-viewer.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.cache/journal-viewer
+noblacklist ${HOME}/.local/share/com.vmingueza.journal-viewer
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/journal-viewer.profile
```diff
@@ -0,0 +1,66 @@
...
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+seccomp.block-secondary
+tracelog
+
+disable-mnt
+private
```
kmk3, line 54: Are you sure that whitelisting in `${HOME}` works with `private`?

glitsj16: Great catch! Originally designed the profile without thinking of adding `private`. That dawned on me much later. I've taken out the ${HOME} whitelisting, also including whitelist-common.inc, which does the same thing.

With that regained awareness I've checked our profiles. Several that do use `private` also include `whitelist-common.inc`. I'm preparing a separate PR for that.

Thank you mucho!

kmk3: > Great catch!

Thanks.

> I've taken out the ${HOME} whitelisting, also including whitelist-common.inc,
> which does the same thing.

Since this program uses paths in `${HOME}`, wouldn't it make more sense to keep
the whitelisting and drop `private`?

Note that the current `noblacklist` commands probably don't do anything either
when `private` is used.

glitsj16: > Since this program uses paths in ${HOME}, wouldn't it make more sense to keep
the whitelisting and drop private?

Yeah, you're right. Was overdoing it there...

kmk3 (approved):

Issue #903, opened 2016-11-09
mkdir in profile does not respect --private=<directory>


pyamsoft:
Latest stable firejail: 0.9.44

The command line option `--private=<directory>` states that firejail uses the given directory as a new home. 
```
--private=directory
      Use directory as user home.

      Example:
      $ firejail --private=/home/netblue/firefox-home firefox
```

This works fine, and all whitelist and blacklist commands work as expected with respect to the private directory. However, using mkdir in a profile incorrectly creates the directory in the real home.

You can observe this in the shipped Dropbox profile for example, which makes calls to mkdir in its profile:
```
firejail --private=/home/pyamsoft/dropboxtesting dropbox
```

You can observe similar errors in the shipped firefox profile. 
```
firejail --private=/home/pyamsoft/firefoxtesting firefox
```

You will see that a Dropbox folder will be created in the actual home directory, instead of the private home at ~/dropboxtesting. You will also see the firefox directories like ~/.mozilla being created in the real home. I would expect that mkdir will respect the private home directory just as whitelist and blacklist do, so I believe this behavior to be a bug.

valoq:
I can confirm the behaviour. Seems like a bug to me.

netblue30:
Fixed, thanks for the bug.

okcomputerik:
Although marked as fixed, I see the same behavior running firejail 0.9.44.8 with the dropbox profile. I'm I missing something here? Thanks!

netblue30:
I'll look into it, thanks.

sunho0301:
Are there any updates on this issue?? I'm having the same problem.
Thanks!

chiraag-nataraj:
Is this still an issue? If so, let's figure out what's going on :)

chiraag-nataraj:
Yes, this is still an issue. The question becomes the following:
If we write `mkdir ${HOME}/<somedir>`, do we mean the _real_ `${HOME}` or what the _sandbox_ will see as `${HOME}`?

chiraag-nataraj:
@netblue30 Is this intended behavior? If we use `${HOME}` in, say, `mkdir`, do we want `${HOME}` to be resolved _after_ things like `--private` have been resolved or do we want to resolve it _before_ those predicates? That will change the behavior as I understand it, and either one is fine as long as we document it.

chiraag-nataraj:
Sorry to dredge up so many old issues, but @netblue30, is this intended behavior? If so, we can close. Otherwise, we should fix this.

[...]

PR #5946, opened 2023-08-08
profiles: improvements to profiles using private

glitsj16:
As I was [recently](https://github.com/netblue30/firejail/pull/5943#pullrequestreview-1564187005) reminded by @kmk3 to #903 I checked existing profiles for similar issues. This PR fixes whitelisting in ${HOME} and streamlines the reference to #903.

### DIFF
--- a/etc/profile-a-l/daisy.profile
+++ b/etc/profile-a-l/daisy.profile
@@ -15,7 +15,7 @@ include disable-interpreters.inc
 include disable-proc.inc
 include disable-programs.inc
 include disable-shell.inc
-#include disable-X11.inc - x11 none
+#include disable-X11.inc # x11 none
 include disable-xdg.inc
 
 include whitelist-common.inc
@@ -47,7 +47,6 @@ tracelog
 x11 none
 
 disable-mnt
-private
 private-bin daisy
 private-cache
 private-dev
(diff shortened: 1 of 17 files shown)


rusty-snake:
TBH we should fix #903 instead of adding references to profiles.

glitsj16:
@kmk3 I totally agree that it would be better to fix #903. It's an outstanding bug for quite a while now and IMO it won't be fixed anytime soon. Anyway, IMO having these comments would make it easier to track which profiles need changes whenever a fix arrives.
Just my $0.2.

kmk3 (commented):
As mentioned [here][1], I think it would make more sense to remove `private`
for profiles that use paths in `${HOME}`.

Also, I'd use `#` instead of `-` for comments at the end of the line, to avoid
a syntax error if uncommenting the line (see also #5942).

For the other profiles, if including whitelist-common.inc helps a given
profile, why not drop `private`?  And if it doesn't, why not simply remove it?

They seem to be made for different use cases AFAICT (need to access some vs no
files in the real `${HOME}`).

[1]: https://github.com/netblue30/firejail/pull/5943/files#r1287696407

glitsj16:
@kmk3 Thanks for the argumentation. Looks better now.

kmk3 (approved):

kmk3:
Note: I'm not sure what the intended behavior would be (copy vs bind-mount) if
whitelisting worked with `private`, so I'll leave that for #903.

rusty-snake:
@kmk3 #903 is about mkdir. The latest decision on whitelist+private was a2b81da0f 4909fa7ef.

PR #5947, opened 2023-08-08
New profile: clac

glitsj16:
clac
Simple command-line calculator
https://github.com/alex-laycalvert/clac

Note: the app has no configuration.

### DIFF
--- a/etc/profile-a-l/clac.profile
+++ b/etc/profile-a-l/clac.profile
@@ -1,0 +1,63 @@
+# Firejail profile for clac
+# Description: Simple command-line calculator
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include clac.local
+# Persistent global definitions
+include globals.local
+
+blacklist ${RUNUSER}
+
+include disable-common.inc
+include disable-devel.inc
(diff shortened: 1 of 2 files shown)


rusty-snake (changes requested):
The profile name / binary name conflicts with other programs. `calc` in fedora is https://isthe.com/chongo/tech/comp/calc/ for example.

glitsj16:
@rusty-snake This app (and its binary) is called `clac`, not `calc`. How does that imply a name clash? Or does Fedora have anything already named `clac`?

glitsj16:
@rusty-snake The name is indeed a bit awkward, no argument there. Thanks for the review and for pointing this out. If you'd prefer not including this profile, I'll drop the PR. It was something I had in my custom profiles, it isn't a widely used app and I can return it to my custom set of 'fringe app' profiles without any trouble :)

I'll leave this open for a while without merging so we can think about how we'd prefer to deal with these kind of apps.

rusty-snake:
No, I've no problem with a clac profile. A ambiguous calc profile (with firecfg) would be a probelm.

PR #5950, opened 2023-08-11
safeguard single line comments

glitsj16:
Using `-` on single line comments can break easily when users just uncomment a line. It's advised to use `#` instead.

Relates to https://github.com/netblue30/firejail/pull/5942

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -33,7 +33,7 @@ blacklist-nolog ${HOME}/.viminfo
 blacklist-nolog /tmp/clipmenu*
 
 # X11 session autostart
-# blacklist ${HOME}/.xpra - this will kill --x11=xpra cmdline option for all programs
+#blacklist ${HOME}/.xpra # this will kill --x11=xpra cmdline option for all programs
 blacklist ${HOME}/.Xsession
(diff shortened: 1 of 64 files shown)


rusty-snake:
TBH uncommenting lines can break easily if you update firejail.

rusty-snake:
And :+1: for `#COMMAND` `# COMMENT` difference.

kmk3:
@glitsj16

Ah you beat me to it :p

I also started working on this with a script (I probably should have commented
about it on a related PR), though it's only for the `# ` to `#` and fixing the
end-of-line comments (mostly no changes to existing normal comments).

It currently changes about ~220 files.

Do you mind if we try the automated way first and then continue with this PR
for whatever is missing?

glitsj16:
> And 👍 for #COMMAND # COMMENT difference.

@rusty-snake Thanks. There's bound to be plenty of places left were we could streamline this. Perhaps we can add an item in the template so it's a bit more explicit for anyone who contributes and cares to actually follow the template.

> Do you mind if we try the automated way first and then continue with this PR
for whatever is missing?

@kmk3 Absolute not, feel free to proceed with your scripted version :+1: .

glitsj16:
Let's close this now that #5987 takes over.

PR #5957, opened 2023-08-13
fcopy: Use lstat when copy directory.

### LINKED ISSUE #5378: Yet another --private-etc symlink issue, with /etc/alternatives/
### Description

Depending on how I set --private-etc, `/etc/alternatives/cc` is sometimes a symlink, sometimes a copy:

### Steps to Reproduce

If I whitelist `/etc/alternatives/cc` alone, I get a copy of `cc` instead of its symlink:
```text
$ ls -lah /etc/alternatives/cc
lrwxrwxrwx 1 root root 12 Oct 14  2019 /etc/alternatives/cc -> /usr/bin/gcc
$ firejail --quiet --private-etc=alternatives/cc ls -lah /etc/alternatives/cc
-rwxr-xr-x 1 65534 65534 1.3M Sep 23 12:22 /etc/alternatives/cc
```

Which is bad (cc is not happy to be here). On the other hand, if I just whielist the whole `alternatives` directory:

```text
$ firejail --quiet --private-etc=alternatives ls -lah /etc/alternatives/cc
Error fcopy: size limit of 500 MB reached
lrwxrwxrwx 1 65534 65534 32 Sep 23 12:22 /etc/alternatives/cc -> /usr/bin/x86_64-linux-gnu-gcc-12
```

this time `/etc/alternatives/cc` is a symlink (and `cc` is happy). But `fcopy` whines about a 500MB limit being reached, which is probably wrong as /etc/alternatives only contains symlinks:

```
$ du -sh /etc/
4.0K	/etc/
```

### Expected behavior

I know that symlink is a security issue magnet, so I may miss something obvious, but as there's already a way to have the symlink I quickly though "what about the other way also giving a symlink ?".

### Environment

- Debian testing (bookworm)
- Firejail 0.9.70

### Checklist

- [ ] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
  - [ ] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ ] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

gerasiov:
When copying directories use lstat when reading info about source files.

Closes #5378.

### DIFF
--- a/src/fcopy/main.c
+++ b/src/fcopy/main.c
@@ -277,7 +277,7 @@ static int fs_copydir(const char *infname, const struct stat *st, int ftype, str
 
 	// don't copy it if we already have the file
 	struct stat s;
-	if (stat(outfname, &s) == 0) {
+	if (lstat(outfname, &s) == 0) {
 		if (first)
 			first = 0;
 		else if (!arg_quiet)
@@ -286,7 +286,7 @@ static int fs_copydir(const char *infname, const struct stat *st, int ftype, str
 	}
 
 	// extract mode and ownership
-	if (stat(infname, &s) != 0)
+	if (lstat(infname, &s) != 0)
 		goto out;
 
 	uid_t uid = s.st_uid;


glitsj16 (approved):
LGTM

src/fcopy/main.c
```diff
@@ -286,7 +286,7 @@ static int fs_copydir(const char *infname, const struct stat *st, int ftype, str
 	}
 
 	// extract mode and ownership
-	if (stat(infname, &s) != 0)
+	if (lstat(infname, &s) != 0)
```
kmk3, line 289: Good catch!

Shouldn't this be done for `outfname` as well?

gerasiov: This does not make any changes in behavior, but your right, it'll be more accurate.

kmk3 (approved):

kmk3 (commented):
@gerasiov

Actually, I just realized that the issue is probably because it is both a
directory and a symlink and the directory test appears before the symlink test.

Does it work with only the following changes?

```diff
diff --git a/src/fcopy/main.c b/src/fcopy/main.c
index a56e8a91b..72fb8e01f 100644
--- a/src/fcopy/main.c
+++ b/src/fcopy/main.c
@@ -503,12 +503,12 @@ int main(int argc, char **argv) {
                exit(1);
        }

-       if (S_ISDIR(s.st_mode))
+       if (S_ISLNK(s.st_mode))
+               duplicate_link(src, dest, &s);
+       else if (S_ISDIR(s.st_mode))
                duplicate_dir(src, dest, &s);
        else if (S_ISREG(s.st_mode))
                duplicate_file(src, dest, &s);
-       else if (S_ISLNK(s.st_mode))
-               duplicate_link(src, dest, &s);
        else {
                fprintf(stderr, "Error fcopy: src %s is an unsupported type of file\n", src);
                exit(1);
```

gerasiov:
In my case the problem was with nvidia driver libs, like
`/etc/alternatives/nvidia--libnvidia-ml.so.1-x86_64-linux-gnu -> /usr/lib/x86_64-linux-gnu/nvidia/current/libnvidia-ml.so.525.125.06`
there are bunch of them.
They're symlinks, but not directories.

kmk3:
> In my case the problem was with nvidia driver libs, like
> `/etc/alternatives/nvidia--libnvidia-ml.so.1-x86_64-linux-gnu ->
> /usr/lib/x86_64-linux-gnu/nvidia/current/libnvidia-ml.so.525.125.06` there
> are bunch of them. They're symlinks, but not directories.

Ah that makes sense now, `S_ISDIR(s.st_mode)` -> `duplicate_dir` ->
`fs_copydir` is called because `/etc/alternatives` is a directory, but then the
paths in it are files.

I see that there is some code in `main()` that chooses between stat and lstat
based on the `arg_follow_links` variable:

```C
	// copy files
	if ((arg_follow_link ? stat : lstat)(src, &s) == -1) {
		fprintf(stderr, "Error fcopy: src %s: %s\n", src, strerror(errno));
		exit(1);
	}
```

Though I'm not sure if it is indeed only intended to apply to the top-level
path or to every symlink, in which case it might make sense to reuse the same
construct in `fs_copydir`.

Cc: @netblue30 @smitsohu

PR #5968, opened 2023-08-23
profiles: move fakeroot blacklisting to disable-devel.inc

glitsj16:
As of https://github.com/netblue30/firejail/commit/96beb3358c430a5e470ce02fd64ffc3f7fc23706 `fakeroot` is blacklisted in disable-common.inc. Without that, makepkg will break, cfr. https://github.com/netblue30/firejail/commit/96beb3358c430a5e470ce02fd64ffc3f7fc23706#r125237349.


kmk3 (commented):
`fakeroot` is probably only used for building things and blocking it may break
other build-related tools as well, so I'd just move the blacklist command to
disable-devel.inc instead.

kmk3 (approved):

PR #5993, opened 2023-09-07
modif: keep pipewire group unless nosound is used

### LINKED ISSUE #5992: Leave `pipewire` group along with `audio` group if logind is absent.
### Description

Currently, if `noroot` and `nogroups` are present without `nosound` and logind is absent, `audio` group remains, but `pipewire` groups doesn't remain. `pipewire` group is important for giving realtime priority to pipewire clients.

kmk3:
This group is apparently used on Gentoo[1].

Currently only the "audio" supplementary group is kept.

Fixes #5992.

See also commit f32938669 ("Keep vglusers group unless no3d is used
(virtualgl)", 2022-01-07) / PR #4851.

[1] https://wiki.gentoo.org/wiki/PipeWire

Reported-by: @amano-kenji

### DIFF
--- a/RELNOTES
+++ b/RELNOTES
@@ -21,6 +21,7 @@ firejail (0.9.73) baseline; urgency=low
   * modif: Improve --version/--help & print version on startup (#5829)
   * modif: improve errExit error messages (#5871)
   * modif: drop deprecated 'shell' option references (#5894)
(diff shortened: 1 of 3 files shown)


glitsj16 (approved):
LGTM

PR #5994, opened 2023-09-07
telegram.profile: allow ~/.local/share/telegram-desktop

mr-tron:
New TelegramWebApps uses another directory for saving local storage.

### DIFF
--- a/etc/profile-m-z/telegram.profile
+++ b/etc/profile-m-z/telegram.profile
@@ -7,6 +7,7 @@ include globals.local
 
 noblacklist ${HOME}/.TelegramDesktop
 noblacklist ${HOME}/.local/share/TelegramDesktop
+noblacklist ${HOME}/.local/share/telegram-desktop
 
 # Allow opening hyperlinks
 include allow-bin-sh.inc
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/telegram.profile
```diff
@@ -7,6 +7,7 @@ include globals.local
 
 noblacklist ${HOME}/.TelegramDesktop
 noblacklist ${HOME}/.local/share/TelegramDesktop
+noblacklist ${HOME}/.local/share/telegram-desktop
```
glitsj16, line 10: This needs a `blacklist ${HOME}/.local/share/telegram-desktop` counterpart in `disable-programs.inc` to protect the path from leaking into other sandboxes. Respecting alphabetical sorting it should go right after `blacklist ${HOME}/.local/share/swell-foop`.

glitsj16 (changes requested):
Needs additional blacklisting of the new path in disable-programs.inc. Otherwise fine. Thank you for the PR!

mr-tron:
added

glitsj16 (approved):
LGTM

kmk3 (approved):

PR #6012, opened 2023-09-17
steam.profile: Allow Factorio

archaon616:
Add directories to config so Factorio runs correctly.

### DIFF
--- a/etc/profile-m-z/steam.profile
+++ b/etc/profile-m-z/steam.profile
@@ -12,6 +12,7 @@ noblacklist ${HOME}/.config/MangoHud
 noblacklist ${HOME}/.config/ModTheSpire
 noblacklist ${HOME}/.config/RogueLegacy
 noblacklist ${HOME}/.config/RogueLegacyStorageContainer
+noblacklist ${HOME}/.factorio
 noblacklist ${HOME}/.killingfloor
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/steam.profile
```diff
@@ -12,6 +12,7 @@ noblacklist ${HOME}/.config/MangoHud
 noblacklist ${HOME}/.config/ModTheSpire
 noblacklist ${HOME}/.config/RogueLegacy
 noblacklist ${HOME}/.config/RogueLegacyStorageContainer
+noblacklist ${HOME}/.factorio
```
glitsj16, line 15: Please add the counterpart `blacklist ${HOME}/.factorio` to `disable-programs.inc`. Noblacklisting otherwise won't be doing anything meaningful.

glitsj16 (changes requested):
Thanks for the PR. A single request: add blacklist path to disable-programs.inc.

archaon616:
Thanks for the feedback. Change added as requested.

etc/profile-m-z/steam.profile
archaon616: Thanks for the feedback. Change added as requested.

glitsj16 (approved):
LGTM

PR #6019, opened 2023-09-24
profiles: dpkg fix

glitsj16:
The two profiles with dpkg in private-bin didn't override `blacklist ${PATH}/dpkg*` from disable-common.inc. Let's fix that.

### DIFF
--- a/etc/profile-a-l/file-roller.profile
+++ b/etc/profile-a-l/file-roller.profile
@@ -6,7 +6,6 @@ include file-roller.local
 # Persistent global definitions
 include globals.local
 
-# Allow dpkg (blacklisted by disable-common.inc)
 noblacklist ${PATH}/dpkg*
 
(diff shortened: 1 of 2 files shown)


etc/profile-a-l/file-roller.profile
```diff
@@ -6,6 +6,9 @@ include file-roller.local
 # Persistent global definitions
 include globals.local
 
+# Allow dpkg (blacklisted by disable-common.inc)
```
kmk3, line 9:
```suggestion
```

For consistency; such comments are usually only used when including allow-.inc
files.

kmk3 (approved):

kmk3:
Related commit:

* f37cd57cd ("disable all /bin/dpkg* programs in disable-common.inc",
  2023-08-20)

PR #6021, opened 2023-09-25
profiles: exchange private-opt with a whitelist

glitsj16:
private-opt breaks file-copy-limit, use a whitelist instead of draining RAM
cfr. https://github.com/netblue30/firejail/discussions/5307

This PR streamlines the comment and does the relevant whitelisting in /opt.

### DIFF
--- a/etc/profile-m-z/youtube.profile
+++ b/etc/profile-m-z/youtube.profile
@@ -15,10 +15,10 @@ include disable-shell.inc
 
 mkdir ${HOME}/.config/Youtube
 whitelist ${HOME}/.config/Youtube
+whitelist /opt/Youtube
 
 private-bin electron,electron[0-9],electron[0-9][0-9],youtube
 private-etc @tls-ca,@x11,bumblebee,host.conf,mime.types
-private-opt Youtube
 
 # Redirect
 include electron-common.profile
--- a/etc/profile-a-l/bitwarden.profile
+++ b/etc/profile-a-l/bitwarden.profile
@@ -17,6 +17,7 @@ include disable-shell.inc
 
 mkdir ${HOME}/.config/Bitwarden
 whitelist ${HOME}/.config/Bitwarden
+whitelist /opt/Bitwarden
 
 machine-id
 no3d
@@ -24,7 +25,6 @@ nosound
 
 ?HAS_APPIMAGE: ignore private-dev
 private-etc @tls-ca
-private-opt Bitwarden
 
 # Redirect
 include electron-common.profile
--- a/etc/profile-a-l/discord-canary.profile
+++ b/etc/profile-a-l/discord-canary.profile
@@ -9,9 +9,10 @@ noblacklist ${HOME}/.config/discordcanary
 
 mkdir ${HOME}/.config/discordcanary
 whitelist ${HOME}/.config/discordcanary
+whitelist /opt/DiscordCanary
+whitelist /opt/discord-canary
 
 private-bin discord-canary,DiscordCanary
-private-opt discord-canary,DiscordCanary
 
 # Redirect
 include discord-common.profile
--- a/etc/profile-a-l/discord-ptb.profile
+++ b/etc/profile-a-l/discord-ptb.profile
@@ -9,9 +9,10 @@ noblacklist ${HOME}/.config/discordptb
 
 mkdir ${HOME}/.config/discordptb
 whitelist ${HOME}/.config/discordptb
+whitelist /opt/DiscordPTB
+whitelist /opt/discord
 
 private-bin discord-ptb,DiscordPTB
-private-opt discord-ptb,DiscordPTB
 
 # Redirect
 include discord-common.profile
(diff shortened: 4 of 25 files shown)


etc/profile-m-z/youtube.profile
```diff
@@ -18,7 +18,10 @@ whitelist ${HOME}/.config/Youtube
 
 private-bin electron,electron[0-9],electron[0-9][0-9],youtube
 private-etc @tls-ca,@x11,bumblebee,host.conf,mime.types
-private-opt Youtube
+# private-opt breaks file-copy-limit, use a whitelist instead of draining RAM
+# https://github.com/netblue30/firejail/discussions/5307
+#private-opt Youtube
+whitelist /opt/Youtube
```
kmk3, line 24:
```suggestion
```

I think it would be better to put this information in firejail(1) instead of
repeating it everywhere for `private-opt` (the docs can be left for later; I
could suggest or add them later).

Suggestion: Delete `private-opt` and the comment and move the `whitelist`
command to the usual place.

kmk3 (changes requested):
(Adding a NAK mostly due to the duplication)

Misc: It seems like a good idea to avoid `private-opt` for very heavy programs
(like electron-based ones), but I'm not sure about never using it on upstream
profiles (I might expand on this later).

etc/profile-m-z/youtube.profile
glitsj16: The suggested changes look very reasonable and shouldn't take long to implement. Seeing your misc. remarks, I'll freeze this for a while so we can have a proper discussion on the private-opt topic in general.

glitsj16:
> Misc: It seems like a good idea to avoid private-opt for very heavy programs
(like electron-based ones), but I'm not sure about never using it on upstream
profiles (I might expand on this later).

@kmk3 Thanks for your review. I had similar doubts about discarding private-opt across the board. True, for less sizable applications the RAM overhead looks tolerable. I'm just not sure where we should draw the line. IMO it's rather likely that users will have way more than one sandbox running in parallel, and things start to add up. I should have elaborated on my reasoning in this context. When I compared the sandbox gains of having private-opt on top of whitelisting the relevant path(s) under /opt I just couldn't see all that much extra hardening to be honest. Looking forward to your insights/arguments.

etc/profile-m-z/youtube.profile
glitsj16: @kmk3 Did some extra research on the size of the non-electron apps. I implemented your suggestions, including a warning note to firejail(1). Kept private-opt for enpass (~85MB), mate-dictionary (<20MB), minecraft-launcher (~1.6MB) and ppsspp (~44MB). The only app I couldn't check: xmr-stak (FTBFS on my Arch Linux box). I returned it to the functional state it had prior to this PR.

HTH

kmk3:
Sorry for the delay, I couldn't get to this earlier.

@glitsj16 [on Sep 26](https://github.com/netblue30/firejail/pull/6021#issuecomment-1734703051):

> Thanks for your review. I had similar doubts about discarding private-opt
> across the board. True, for less sizable applications the RAM overhead looks
> tolerable. I'm just not sure where we should draw the line.  IMO it's rather
> likely that users will have way more than one sandbox running in parallel,
> and things start to add up. I should have elaborated on my reasoning in this
> context. When I compared the sandbox gains of having private-opt on top of
> whitelisting the relevant path(s) under /opt I just couldn't see all that
> much extra hardening to be honest.

:+1:

> Looking forward to your insights/arguments.

Not much insight, just this:

Pros:

* Programs in /opt tend to be heavy, so less RAM would be wasted
* Always using `whitelist /opt/dir` upstream would be simpler

Cons:

* Programs in /opt tend to be complex and proprietary, so a bit of extra
  security might be useful

I mostly just wanted to give it some consideration, but indeed it's hard to
draw a line and I'm also not sure how much it really matters.

@glitsj16 [on Sep 27](https://github.com/netblue30/firejail/pull/6021#discussion_r1338165262):

> Did some extra research on the size of the non-electron apps. I implemented
> your suggestions, including a warning note to firejail(1). Kept private-opt
> for enpass (~85MB), mate-dictionary (<20MB), minecraft-launcher (~1.6MB) and
> ppsspp (~44MB). The only app I couldn't check: xmr-stak (FTBFS on my Arch
> Linux box). I returned it to the functional state it had prior to this PR.

Sounds good to me.

kmk3:
Main force-push changes:

* Removed whitelist when there's private-opt for consistency
* Edited/added more to the docs

It currently looks good to me.

Thoughts on the changes?

glitsj16:
@kmk3

> Sorry for the delay, I couldn't get to this earlier.

That's quite allright, no worries.

> Thoughts on the changes?

I'm okay with your changes. Like you mentioned, there will always be pros & cons to private-opt and users should be aware of the RAM-ifications. Not much more we can do here IMO.

Thanks again for your efforts and opinions!

PR #6026, opened 2023-09-26
ci: allow running workflows manually

kmk3:
Add `on.workflow_dispatch`.

See:

* https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatch
* https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

### DIFF
--- a/.github/workflows/build-extra.yml
+++ b/.github/workflows/build-extra.yml
@@ -3,6 +3,7 @@
 name: Build-extra
 
 on:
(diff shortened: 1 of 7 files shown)


glitsj16 (approved):
LGTM.

netblue30:
merged, thanks!

PR #6028, opened 2023-09-29
New profile: brz

glitsj16:
https://www.breezy-vcs.org/
https://github.com/breezy-team/breezy

Implemented as git redirect

### DIFF
--- a/etc/profile-a-l/brz.profile
+++ b/etc/profile-a-l/brz.profile
@@ -1,0 +1,14 @@
+# Firejail profile for brz
+# Description: Distributed VCS with support for Bazaar and Git file formats
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include brz.local
+# Persistent global definitions
+# added by included profile
+#include globals.local
+
+noblacklist ${HOME}/.config/breezy
+
+# Redirect
+include git.profile
--- a/etc/profile-a-l/bzr.profile
+++ b/etc/profile-a-l/bzr.profile
@@ -1,0 +1,10 @@
+# Firejail profile alias for bzr
+# This file is overwritten after every install/update
+# Persistent local customizations
+include bzr.local
+# Persistent global definitions
+# added by included profile
+#include globals.local
+
+# Redirect
+include brz.profile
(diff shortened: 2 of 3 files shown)


etc/profile-a-l/brz.profile
```diff
@@ -0,0 +1,14 @@
...
+# Description: A decentralized revision control system with support for Bazaar and Git file formats
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include brz.local
+# Persistent global definitions
+# added by included profile
+#include globals.local
+
+noblacklist ${HOME}/.config/breezy
```
kmk3, line 11: Not blacklisted.

etc/profile-a-l/bzr.profile
```diff
@@ -0,0 +1,10 @@
+# Firejail profile alias for bzr
+# This file is overwritten after every install/update
+# Persistent local customizations
+include bzr.local
+# Persistent global definitions
+# added by included profile
+#include globals.local
+
+# Redirect
+include brz.profile
```
kmk3, line 10: `bzr` is apparently the same program name used by bazaar itself, so it doesn't
seem like a good idea to redirect to the profile of a different program.

I'm not familiar with bazaar, so I'd probably just remove this profile.

etc/profile-a-l/brz.profile
glitsj16: Duh, forgot to add it. Done!

etc/profile-a-l/bzr.profile
glitsj16: > bzr is apparently the same program name used by bazaar itself, so it doesn't
seem like a good idea to redirect to the profile of a different program.

That was the point of the recent breezy introduction on Arch Linux. breezy is the [drop-in replacement](https://gitlab.archlinux.org/archlinux/packaging/packages/pacman/-/commit/c68a4e6602e3488fa093a18d35202c76a730faf6) for bazaar since pacman 6.0.2-8. I've tested it, no breakage whatsoever. And it doesn't get auto-enabled via firecfg. IMO we can keep it.

kmk3 (approved):

Issue #6034, opened 2023-10-05
Nicotine won't start if Fcitx rules are set in the .local file


glu8716:
### Description

According to the [FAQ page](https://github.com/netblue30/firejail/wiki/Frequently-Asked-Questions#how-can-i-enable-fcitx) if the dbus policy of the profile is set to none, in order to have Fcitx working you need to add
```
dbus-user filter
dbus-user.talk org.freedesktop.portal.Fcitx
ignore dbus-user none
```
to the .local file. And that's what I did for Nicotine.
```
cat /home/debu/.config/firejail/nicotine.local 
dbus-user filter
dbus-user.talk org.freedesktop.portal.Fcitx
ignore dbus-user none
```
However by doing so the program doesn't start.

### Steps to Reproduce

1. Add Fcitx rules to the `nicotine.local` file
2. Start Nicotine

### Expected behavior

Program should start and Fcitx should work.

### Actual behavior

Program doesn't start.

### Behavior without a profile

The programs starts fine and Fcitx works too.

### Additional context

I see that the program is complaining about the dbus socket location (see logs). My dbus socket file is located in the /tmp folder as I found out in [another issue](https://github.com/netblue30/firejail/issues/6031#issuecomment-1745644600) I've opened recently. This is how Artix deals with it, apparently. I don't know if this could be related.

### Environment

- Linux distribution and version: Artix
- Firejail version: 0.9.72

### Checklist

- [X] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [ ] I can reproduce the issue without custom modifications (e.g. globals.local).
- [X] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [X] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [X] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [ ] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ ] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

<details>
<summary>Output of <code> firejail /usr/bin/nicotine</code></summary>
<p>

```
Reading profile /etc/firejail/nicotine.profile
Reading profile /home/debu/.config/firejail/nicotine.local
[...]

PR #6036, opened 2023-10-06
nicotine: support Fcitx and dconf via dbus-user filter

glu8716:
From https://github.com/netblue30/firejail/issues/6034. Nicotine+ needs `dbus-user.own org.nicotine_plus.Nicotine` or it will crash if started with options for Fcitx. And `dbus-user.talk ca.desrt.dconf` is to support dconf as stated [here]( https://github.com/netblue30/firejail/issues/6034#issuecomment-1750542403).

### DIFF
--- a/etc/profile-m-z/nicotine.profile
+++ b/etc/profile-m-z/nicotine.profile
@@ -58,6 +58,8 @@ private-dev
 private-tmp
 
 dbus-user none
+dbus-user.own org.nicotine_plus.Nicotine
+dbus-user.talk ca.desrt.dconf
 dbus-system none
 
 restrict-namespaces
--- a/etc/profile-m-z/nicotine.profile
+++ b/etc/profile-m-z/nicotine.profile
@@ -57,7 +57,7 @@ private-cache
 private-dev
 private-tmp
 
-dbus-user none
+dbus-user filter
 dbus-user.own org.nicotine_plus.Nicotine
 dbus-user.talk ca.desrt.dconf
 dbus-system none


etc/profile-m-z/nicotine.profile
```diff
@@ -58,6 +58,8 @@ private-dev
 private-tmp
 
 dbus-user none
```
glitsj16, line 60: We need to replace `dbus-user none` with `dbus-user filter` for the filtering below to work.

etc/profile-m-z/nicotine.profile
```diff
@@ -58,6 +58,8 @@ private-dev
...
+dbus-user.own org.nicotine_plus.Nicotine
+dbus-user.talk ca.desrt.dconf
```
glitsj16, line 62: Add `dbus-user.talk org.freedesktop.portal.Fcitx` below this line for the Fcitx fix.

glitsj16 (changes requested):
Just needs a replacement and the Fcitx dbus filter.

etc/profile-m-z/nicotine.profile
glu8716: I haven't thought of adding the Fcitx part. Isn't it something that should be optional? At least I don't see it being in other profiles.

etc/profile-m-z/nicotine.profile
glu8716: Done

etc/profile-m-z/nicotine.profile
glitsj16: We do have https://github.com/netblue30/firejail/blob/2a8621cd940262397b048881d7fc34e5ba9644aa/etc/profile-a-l/linuxqq.profile#L32

But you're correct. Let's keep it optional :+1: . I'll fix linuxqq later. Thanks for the PR!

PR #6039, opened 2023-10-06
New profile: termshark

glitsj16:
termshark

Terminal UI for tshark, inspired by Wireshark
https://github.com/gcla/termshark

### DIFF
--- a/etc/profile-m-z/termshark.profile
+++ b/etc/profile-m-z/termshark.profile
@@ -1,0 +1,12 @@
+# Firejail profile for termshark
+# Description: Terminal UI for tshark, inspired by Wireshark
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include termshark.local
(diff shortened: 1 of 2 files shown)


kmk3 (approved):

etc/profile-m-z/termshark.profile
```diff
@@ -0,0 +1,12 @@
...
+# Description: Terminal UI for tshark, inspired by Wireshark
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include termshark.local
+# Persistent global definitions
+# added by included profile
+#include globals.local
+
+# Redirect
```
rusty-snake, line 11: We could block x11/Wayland stuff right?

glitsj16: Yeah. In fact we can block ${RUNUSER} completely. Thanks for the pointer!

PR #6049, opened 2023-10-11
disable-common.inc: add more suid programs

kmk3:
Programs:

    $ pacman -Qo fusermount3 groupmems mount.cifs wall write
    /usr/bin/fusermount3 is owned by fuse3 3.16.1-1
    /usr/bin/groupmems is owned by shadow 4.14.0-4
    /usr/bin/mount.cifs is owned by cifs-utils 7.0-3
    /usr/bin/wall is owned by util-linux 2.39.2-1
    /usr/bin/write is owned by util-linux 2.39.2-1


etc/inc/disable-common.inc
```diff
@@ -512,71 +513,73 @@ blacklist ${PATH}/crontab
 blacklist ${PATH}/doas
 blacklist ${PATH}/evtest
 blacklist ${PATH}/expiry
-blacklist ${PATH}/fusermount
+blacklist ${PATH}/fping
+blacklist ${PATH}/fping6
+blacklist ${PATH}/fusermount*
```
glitsj16, line 518: Realizing it's convenient to use `globbing` IMO we should make an exception here and use separate entrees for fusermount and fusermount3.

etc/inc/disable-common.inc
```diff
@@ -512,71 +513,73 @@ blacklist ${PATH}/crontab
...
 blacklist ${PATH}/gksudo
 blacklist ${PATH}/gpasswd
+blacklist ${PATH}/groupmems
+blacklist ${PATH}/hostname
+#blacklist ${PATH}/ip # breaks --ip=dhcp
 blacklist ${PATH}/kdesudo
 blacklist ${PATH}/ksu
 blacklist ${PATH}/mount
-blacklist ${PATH}/mount.ecryptfs_private
+blacklist ${PATH}/mount.*
```
glitsj16, line 528: Cfr. above, IMO this is technically incorrect. I mean, there are several ${PATH}/mount.foo binaries that aren't SUID/SGID (). There might be good reasons to blacklist those too, but in that case - IMO - we should group them under a differently named section.

glitsj16 (commented):
Nice work. I just have some doubts about globbing in a few cases.

etc/inc/disable-common.inc
glitsj16: Fair enough. My intention wasn't to question blacklisting of SUID binaries, simply nitpicking about the comment/section heading (if one can call it that) :)

glitsj16 (approved):
LGTM

etc/inc/disable-common.inc
glitsj16: > That is, I don't think that there is an easy way to list the paths of all SUID
binaries for every package on Arch, for example.

Indeed, none that I know of (not only on Arch). Would be a very nice feature if Firejail could check for SUID binaries in the sandbox and blacklist them dynamically instead of keeping track via another file list!

PR #6054, opened 2023-10-16
steam.profile: Allow Baba Is You

Frostbyte4664:
Baba Is You settings and save data are stored in `~/.local/share/Baba_Is_You`.


etc/profile-m-z/steam.profile
```diff
@@ -17,6 +17,7 @@ noblacklist ${HOME}/.killingfloor
 noblacklist ${HOME}/.klei
 noblacklist ${HOME}/.local/share/3909/PapersPlease
 noblacklist ${HOME}/.local/share/aspyr-media
+noblacklist ${HOME}/.local/share/Baba_Is_You
```
kmk3, line 20: This is not blacklisted.

kmk3 (changes requested):

kmk3 (approved):

PR #6060, opened 2023-10-22
profiles: support more msmtp configuration paths

glitsj16:
Since version 1.8.6 msmtp supports per-user configuration at either
~/.msmtprc (already supported by firejail) or
`$XDG_CONFIG_HOME/msmtp/config`. System-wide support can be placed at
/etc/msmtprc.

This adds the missing paths to the relevant .inc and .profile files.

Note that `blacklist ${HOME}/.msmtprc` is present on both
disable-common.inc and disable-programs.inc, so the new paths are added
to both files.

References:

https://wiki.archlinux.org/title/Msmtp#Basic_setup
https://marlam.de/msmtp/msmtp.html#Configuration-files

### DIFF
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -559,6 +559,7 @@ blacklist ${HOME}/.config/mpDris2
 blacklist ${HOME}/.config/mpd
 blacklist ${HOME}/.config/mps-youtube
 blacklist ${HOME}/.config/mpv
+blacklist ${HOME}/.config/msmtp
 blacklist ${HOME}/.config/mullvad-browser-flags.conf
 blacklist ${HOME}/.config/mupen64plus
 blacklist ${HOME}/.config/mutt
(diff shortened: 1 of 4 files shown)


etc/inc/disable-programs.inc
kmk3, line 1: Remove these entries, as they are duplicated from dc.

kmk3: I see now that this line is present on both, so nevermind:

```
blacklist ${HOME}/.msmtprc
```

I'll fix the duplication after this PR.

kmk3 (approved):

etc/inc/disable-programs.inc
glitsj16: Indeed, I noticed the duplication too. But it could be on purpose, not sure of that though. IIRC there are other overlaps between dc and dp. I'll dig around and try to find more details. Thanks for the heads-up!

kmk3: > Indeed, I noticed the duplication too. But it could be on purpose, not sure
> of that though. IIRC there are other overlaps between dc and dp. I'll dig
> around and try to find more details. Thanks for the heads-up!

No problem.

Indeed it could be on purpose, but judging by which entries are duplicated (and
how few of them there are) it looks to me like it was done by accident:

```
blacklist ${HOME}/.VirtualBox
blacklist ${HOME}/.config/VirtualBox
blacklist ${HOME}/.config/msmtp
blacklist ${HOME}/.local/share/gnome-boxes
blacklist ${HOME}/.msmtprc
blacklist /etc/msmtprc
```

If that's the case, I think I'll add a CI check for this later.

PR #6070, opened 2023-10-27
build: sort.py: use case-sensitive sorting

kmk3:
To match how things are sorted elsewhere, such as with `noblacklist` /
`whitelist` lines (vertically) in profiles and in
ci/check/profiles/sort-disable-programs.sh and src/etc-cleanup/main.c.

This makes the order in `private-etc` always be groups (`@group`), then
uppercase paths, then lowercase paths.  Example from
etc/profile-m-z/softmaker-common.profile:

    private-etc @tls-ca,SoftMaker,fstab

Note that this does not affect a significant amount of profiles; most
changes are in `private-bin` / `private-lib` lines and in `private-etc`
lines for newer profiles that do not use groups.  This is partly due to
commit 5d0822c52 ("private-etc: big profile changes", 2023-02-05)
replacing `X11` with `@x11` in `private-etc` lines and then commit
0f996ea4d ("private-etc: groups modified", 2023-02-05) removing
`Trolltech.conf` from `private-etc` lines and using case-sensitive
sorting in them.

Relates to #5610.

### DIFF
--- a/contrib/sort.py
+++ b/contrib/sort.py
@@ -38,7 +38,7 @@ Exit Codes:
 
 def sort_alphabetical(original_items):
     items = original_items.split(",")
-    items.sort(key=str.casefold)
+    items.sort()
     return ",".join(items)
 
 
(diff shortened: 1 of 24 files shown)


glitsj16 (approved):
Makes more sense sorting things this way. Nice job!

netblue30:
All in!

Issue #5982, opened 2023-08-28
`firejail --ls` reports wrong file sizes for large files


reinerh:
I noticed that the `--ls` command prints negative file sizes for large files (the file has a size of several GB).
```
$ firejail --ls=... /
...
-rw------- root     root      -214748364 swap.bin
...
```

The problem seems to be that it's using a (32bit) int, while `st_size` is actually of type `off_t` (which is on x86_64 a `long int`).
https://github.com/netblue30/firejail/blob/dd5ae0d8f41cb7f054506c8cf373c8b609429d1b/src/firejail/ls.c#L155-L158

PR #6086, opened 2023-11-13
Fix displaying of large file sizes.

chestnykh:
The most generic way is to use `intmax_t`
because we dont't know what is the "parent" type of `off_t`.
This fixes https://github.com/netblue30/firejail/issues/5982 .


kmk3 (approved):
Good catch.

glitsj16 (approved):

Issue #6006, opened 2023-09-15
Xauth binary is hardcoded to /usr/bin/xauth


hashelq:
### Description
https://github.com/netblue30/firejail/blob/eb5c97197b699dbb8ba69e798c86e5e97c36e17e/src/firejail/x11.c#L1180-L1186

### Expected behavior
To find the executable in the path, or, at least, have an option to override the path to the binary.

### Actual behavior
unix-way

### Additional context
Despite the fact in most environments you can find binaries in /usr/bin, there are cases in which you cannot, like in NixOS, containers, small os images, etc.

### Environment
NixOS (/usr/bin, as well as /bin are not used for apps like xauth)

### Checklist

- [x] I have performed a short search for similar issues (to avoid opening a duplicate).

glitsj16:
Thank you for reporting this. Similar hardcoded reference to /usr/bin/xauth:

https://github.com/netblue30/firejail/blob/eb5c97197b699dbb8ba69e798c86e5e97c36e17e/src/firejail/fs_lib2.c#L168-L170

@netblue30 Can you have a look at this please?

PR #6087, opened 2023-11-13
Lookup xauth in PATH.

chestnykh:
Don't use hardcoded `/usr/bin/xauth`,
iterate over directories inside PATH instead.

This fixes https://github.com/netblue30/firejail/issues/6006

### DIFF
--- a/src/firejail/x11.c
+++ b/src/firejail/x11.c
@@ -1164,7 +1164,6 @@ void x11_start(int argc, char **argv) {
 }
 #endif
 
-
 void x11_xorg(void) {
 #ifdef HAVE_X11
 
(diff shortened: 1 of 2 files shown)


kmk3 (approved):

kmk3 (changes requested):
I see now that there already is a function for doing this: `find_in_path()` in
`fs_lib.c`:

* <https://github.com/netblue30/firejail/blob/b689b69f6c3b8a8ba633d6300cef6a19972d53dc/src/firejail/fs_lib.c#L55>

Please try using that instead of adding a new one.

chestnykh:
> I see now that there already is a function for doing this: `find_in_path()` in `fs_lib.c`:
> 
>     * https://github.com/netblue30/firejail/blob/b689b69f6c3b8a8ba633d6300cef6a19972d53dc/src/firejail/fs_lib.c#L55
> 
> 
> Please try using that instead of adding a new one.

done

kmk3 (approved):
Looks better now, thanks.

Note: There is still a TOCTOU issue with the checks due to this allowing a
potentially user-writable xauth binary, though fixing the issue properly
requires refactoring elsewhere, so I might try that after this PR.

src/firejail/x11.c
```diff
@@ -1175,32 +1174,41 @@ void x11_xorg(void) {
...
 	}
-	// copy /usr/bin/xauth in the sandbox and set mode to 0711
+	// copy xauth in the sandbox and set mode to 0711
 	// users are not able to trace the running xauth this way
 	if (arg_debug)
-		printf("Copying /usr/bin/xauth to %s\n", RUN_XAUTH_FILE);
-	if (copy_file("/usr/bin/xauth", RUN_XAUTH_FILE, 0, 0, 0711)) {
-		fprintf(stderr, "Error: cannot copy /usr/bin/xauth executable\n");
+		printf("Copying %s to %s\n", xauth_bin, RUN_XAUTH_FILE);
+	if (copy_file(xauth_bin, RUN_XAUTH_FILE, 0, 0, 0711)) {
```
kmk3, line 1205: One last thing, please try this:

```suggestion
	if (copy_file_from_user_to_root(xauth_bin, RUN_XAUTH_FILE, 0, 0, 0711)) {
```

Which might prevent arbitrary reads.

chestnykh: Done, but `copy_file_from_user_to_root` doesn't return any value, so just call it

netblue30:
Merged, thanks!

netblue30:
Merged, thanks!

PR #6099, opened 2023-11-23
steam.profile: noblacklist ~/.config/unity3d

nutta-git:
unity3d folder missing noblacklist flag

[question] If a folder is whitelisted, then is it necessary to `noblacklist` it?

### DIFF
--- a/etc/profile-m-z/steam.profile
+++ b/etc/profile-m-z/steam.profile
@@ -12,6 +12,7 @@ noblacklist ${HOME}/.config/MangoHud
 noblacklist ${HOME}/.config/ModTheSpire
 noblacklist ${HOME}/.config/RogueLegacy
 noblacklist ${HOME}/.config/RogueLegacyStorageContainer
+noblacklist ${HOME}/.config/unity3d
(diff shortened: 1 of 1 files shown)


glitsj16:
> [question] If a folder is whitelisted, then is it necessary to noblacklist it?

No. A `noblacklist foo` is only required when there's a `blacklist foo` (usually in one of the included disable-xxx.inc files).

etc/profile-m-z/steam.profile
```diff
@@ -12,6 +12,7 @@ noblacklist ${HOME}/.config/MangoHud
 noblacklist ${HOME}/.config/ModTheSpire
 noblacklist ${HOME}/.config/RogueLegacy
 noblacklist ${HOME}/.config/RogueLegacyStorageContainer
+noblacklist ${HOME}/.config/unity3d
```
glitsj16, line 15: ${HOME}/.config/unity3d is not blacklisted anywhere, hence this line is not needed.

glitsj16 (commented):
PR does not fix anything we're aware of.

PR #6118, opened 2023-12-07
minecraft-launcher.profile: allow keyring access

NetSysFire:
It turns out it is using it to save logins and will error out with just a generic "something went wrong, so we can not save your login".

I copied the directives from email-common.inc. Unsure which specific dbus topic thingy is actually used here but it works and I think it is fine to leave all 3.

Dear maintainers, if the ordering of the directives, my arch nemesis, is wrong, please apply your suggestion or edit without asking.

### DIFF
--- a/etc/profile-m-z/minecraft-launcher.profile
+++ b/etc/profile-m-z/minecraft-launcher.profile
@@ -56,6 +56,7 @@ private-etc @tls-ca,@x11,host.conf,java*,mime.types,services,timezone
 private-opt minecraft-launcher
 private-tmp
 
+dbus-user filter
 dbus-user.talk org.freedesktop.secrets
 dbus-user.talk org.gnome.keyring.*
 dbus-user.talk org.gnome.seahorse.*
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/minecraft-launcher.profile
```diff
@@ -54,7 +56,9 @@ private-etc @tls-ca,@x11,host.conf,java*,mime.types,services,timezone
 private-opt minecraft-launcher
 private-tmp
 
-dbus-user none
+dbus-user.talk org.freedesktop.secrets
```
glitsj16, line 59: Without `dbus-user filter` the 3 `dbus-user.talk` rules don't get applied. So in fact the complete session bus is now allowed in the sandbox. Can you confirm that adding `dbus-user filter` above this line does allow keyring access?

NetSysFire: I'll definitely check tomorrow

NetSysFire: It works.

glitsj16 (approved):

NetSysFire:
I wonder if adding an `allow-keyring.inc` would make sense since multiple, although maybe not enough, profiles could make use of that.

glitsj16:
> I wonder if adding an allow-keyring.inc would make sense since multiple, although maybe not enough, profiles could make use of that.

Great idea. We have other include files that aren't used by a huge amount of profiles. Feel free to implement!

netblue30:
All merged, thanks!

PR #6125, opened 2023-12-11
landlock: move commands into profile and add landlock.enforce

kmk3:
Changes:

* Move commands from --landlock and --landlock.proc= into
  etc/inc/landlock-common.inc
* Remove --landlock and --landlock.proc=
* Add --landlock.enforce

Instead of hard-coding the default commands (and having a separate
command just for /proc), move them into a dedicated profile to make it
easier for users to interact with the entries (view, copy, add ignore
entries, etc).

Only enforce the Landlock commands if --landlock.enforce is supplied.
This allows safely adding Landlock commands to (upstream) profiles while
keeping their enforcement opt-in.  It also makes it simpler to
effectively disable all Landlock commands, by using
`--ignore=landlock.enforce`.

Relates to #6078.

### DIFF
--- a/contrib/syntax/lists/profile_commands_arg0.list
+++ b/contrib/syntax/lists/profile_commands_arg0.list
@@ -12,7 +12,7 @@ keep-config-pulse
 keep-dev-shm
 keep-shell-rc
 keep-var-tmp
-landlock
+landlock.enforce
 machine-id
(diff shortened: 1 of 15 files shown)


glitsj16 (approved):

netblue30:
All merged, thanks!

PR #3422, opened 2020-05-18
Add configure options when building rpm

sfc-gh-hyu:
It seems configure options are hard-coded when building the RPM. I wish to use different config options with RPM, so I added a third argument to `platform/rpm/mkrpm.sh`, which can change the configure options in firejail.spec file.

PR #6126, opened 2023-12-13
build: mkrpm.sh: append instead of override configure args

kmk3:
For consistency with mkdeb.sh.

Note: The default arguments and support for argument overriding was
added to to mkrpm.sh on commit 3d97332fd ("Add configure options when
building rpm (#3422)", 2020-05-19).

The support for appending arguments was added to mkdeb.sh on commit
9a0fbbd71 ("mkdeb.sh.in: pass remaining arguments to ./configure",
2022-05-13) / PR #5154.

Cc: @rusty-snake @sfc-gh-hyu (from #3422)


kmk3:
By the way, is `--disable-userns --disable-contrib-install` used when building
the Fedora rpm?

PR #6128, opened 2023-12-14
mpv: whitelist /usr/share/mpv

pirate486743186:
use case, you install scripts in  `/usr/share/mpv` but they remain inactive. You then symlnk them to `/etc/mpv` to activate them if you want.

Question, isn't `whitelist /usr/share/lua` redundant because of `whitelist /usr/share/lua*` ?


kmk3 (approved):

kmk3:
Added some missing paths and force-pushed.

> Question, isn't `whitelist /usr/share/lua` redundant because of `whitelist
> /usr/share/lua*` ?

Good catch; I'll fix that after this PR.

netblue30:
Merged!

PR #6129, opened 2023-12-14
Revert "Lookup xauth in PATH."

### LINKED ISSUE #6113: firejail: fs_lib.c:56: find_in_path: Assertion \`geteuid() != 0' failed
Several apps show breakage when running via firejail from git master. Here's an example using gsettings, but I've seen the same bug on several other profiles. Bisecting points to https://github.com/netblue30/firejail/commit/407c05ebefe23e725f858b6170b3e52659e044a2.

```sh
$ firejail --quiet /usr/bin/gsettings get org.gnome.shell disable-user-extensions
firejail: fs_lib.c:56: find_in_path: Assertion \`geteuid() != 0' failed
Error: proc 64096 cannot sync with peer: unexpected EOF
Peer 64097 unexpectedly killed (Segmentation fault)
```

 All's well again when I build firejail while reverting this commit:

```sh
$ firejail --quiet /usr/bin/gsettings get org.gnome.shell disable-user-extensions
false
```

Can anyone reproduce?

kmk3:
This reverts commit 407c05ebefe23e725f858b6170b3e52659e044a2.

If --private-lib is used (and firejail is configured with
--enable-private-lib), the following error occurs:

    $ firejail --quiet --noprofile --private-lib true
    firejail: fs_lib.c:56: find_in_path: Assertion `geteuid() != 0' failed.
    Error: proc 10000 cannot sync with peer: unexpected EOF
    Peer 10001 unexpectedly killed (Segmentation fault)

Given that it causes an uid assertion failure, the logic appears to not
be correct and the current behavior may be unsafe, so for now revert
that commit until the issue is properly addressed.

Relates to #6006 #6087.

Fixes #6113.

Cc: @chestnykh @hashelq

### DIFF
--- a/src/firejail/fs_lib2.c
+++ b/src/firejail/fs_lib2.c
@@ -166,12 +166,8 @@ void fslib_install_firejail(void) {
 		fslib_mount_libs(RUN_MNT_DIR "/dhclient", 1); // parse as user
 
 	// bring in xauth libraries
-
-	char *xauth_bin = find_in_path("xauth");
 	if (arg_x11_xorg)
-		fslib_mount_libs(xauth_bin, 1); // parse as user
-
-	free(xauth_bin);
+		fslib_mount_libs("/usr/bin/xauth", 1); // parse as user
 
 	fmessage("Firejail libraries installed in %0.2f ms\n", timetrace_end());
 }
--- a/src/firejail/x11.c
+++ b/src/firejail/x11.c
@@ -1164,6 +1164,7 @@ void x11_start(int argc, char **argv) {
 }
 #endif
 
+
 void x11_xorg(void) {
 #ifdef HAVE_X11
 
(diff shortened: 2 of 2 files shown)


glitsj16 (approved):

glitsj16:
@kmk3 Thanks for reverting this. I assume my (unreviewed) patch mentioned in https://github.com/netblue30/firejail/issues/6113#issuecomment-1840062834 flew under the radar. Not a problem, yet I would be interested to hear opinions on it for learning purposes.

glitsj16:
> The idea to use EUID_USER seems good (especially considering that $PATH may
be set by the user), though I think it would make more sense to always use it
(along with if (called_as_root) etc) instead of special-casing private-lib.

That makes sense. Avoiding special-casing any option would be the more secure route to deal with this. Thanks for taking the time to elaborate!

netblue30:
Merged!

PR #6149, opened 2024-01-04
Exit if firejail is not SUID

EvgeniyRogov:


reinerh:
Why?

EvgeniyRogov:
Otherwise we get 2 errors:
"Error: Firejail needs to be SUID.
Assuming firejail is installed in /usr/bin, execute the following command as root:
  chmod u+s /usr/bin/firejail
Error: cannot create /run/firejail/profile/<pid>".
I think there is no need to try build /run/firejail directory without suid.

PR #6154, opened 2024-01-12
profiles: add profiles for gtk youtube viewers symlinks

pirate486743186:
Convenience symlinks for youtube-viewers-gtk


glitsj16 (approved):

kmk3 (commented):
Why is there both lbry-viewer-gtk and gtk-lbry-viewer?

What are their paths?

Is that the same file but named differently depending on the distribution?

Issue #6133, opened 2023-12-17
ARP probe failing due to gratuitous arp reply


brianvanderburg2:
### Description

Setup: A bridge is configured with a VLAN subinterface from the main interface bound to it.  T When using firejail, the bridge is used as the interface with an IP range to choose from.  

enp9s0 (main interface)
bm-home (vlan subinterface for home/internet traffic)
bhome (bridge, bm-home is placed into this bridge.)
no IPs are assigned to the bridges or VLAN/main interfaces

Firejail is launched to create a veth pair to put into bhome, then processes launched under the firejail proccess have an IP/are able to access the internet. The bridge is used to allow other processes/virtual machines/macvlans/etc to also be able to access the internet when desired. This has worked perfectly fine until recently.

Profile:

```
net bhome
iprange 192.168.1.30,192.168.1.39
defaultgw 192.168.1.1
netmask 255.255.255.0
dns 8.8.8.8
dns 8.8.4.4
```

Recently, I get an error indicating all the IPs are in use. I checked the wireshark capture and i notice:

Send arp probe
I get a gratuitous arp from a different address on the network
Send arp probe for next ip
same
Send arp probe for next ip
same

When I don't get the gratuitous arp response, then it assignes that IP address  Only one device in our network has that response so I asked them to turn it off, and when they do, it works with no problems.

![image](https://github.com/netblue30/firejail/assets/2142408/be020e88-d209-4b6e-a1e1-260eb70fa628)
![image](https://github.com/netblue30/firejail/assets/2142408/f6291780-dc0e-4c54-907a-124e927034dd)

### Steps to Reproduce

Unsure.  It seems specific to this situation

### Expected behavior

After getting no ARP responses matching the IPs of the ARP probe, it should use the IP for the jail

### Actual behavior

It performs an ARP probe of two random IPs, then the entire specified range, and fails with the error "Error: cannot assign an IP address; it looks like all of them are in use."

### Behavior without a profile

This works, however doesn't use a new network namespace with veth interface connected to the bridge and assigned an iP, so doesn't affect/reflect on this issue

### Environment

Debian Bullesye
Firejail 0.9.72

### Checklist

- [ x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [ ] I can reproduce the issue without custom modifications (e.g. globals.local). (Using the profile I use is a custom modification)
[...]

PR #6157, opened 2024-01-16
Fix duplicate address detection (ARP)

brianvanderburg2:
The original code compare the target IP of an ARP reply packet with the srcaddr/sender IP of the request packet.  Instead it should compare the source IP of the ARP reply with the destaddr/target IP of the request packet.  In addition this adds the check for ARP requests with the same IP address in case other systems are probing for the same IP at the same time as per recommendation in RFC5227:

image

### DIFF
--- a/src/firejail/arp.c
+++ b/src/firejail/arp.c
@@ -230,15 +230,26 @@ int arp_check(const char *dev, uint32_t destaddr) {
 			if (framerx[12] != (ETH_P_ARP / 256) || framerx[13] != (ETH_P_ARP % 256))
 				continue;
 			memcpy(&hdr, framerx + 14, sizeof(ArpHdr));
-			if (hdr.opcode == htons(1))
-				continue;
-			if (hdr.opcode == htons(2)) {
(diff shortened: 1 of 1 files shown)


brianvanderburg2:
This is related to #6133

kmk3 (commented):
Hello, the commit message body is currently empty; it does not contain any
explanation for what the commit does, which makes it unclear when viewing it
locally.

```
The current code compares the target IP of an ARP reply packet with the
srcaddr/sender IP of the request packet. Instead it should compare the
source IP of the ARP reply with the destaddr/target IP of the request
packet. In addition, this adds the check for ARP requests with the same
IP address in case other systems are probing for the same IP at the same
time as per the recommendation in RFC 5227.

This is related to #6133.
```

I formatted your original explanation above, can you add it to the commit
message body and force-push?

kmk3:
Thanks, it looks good now.

Note: I'm unfamiliar with the ARP code, so I'll avoid reviewing it for now.
Hopefully someone who is can review it.

PR #6180, opened 2024-01-30
Profile for RawTherapee

haplo:
This is a profile for the [RawTherapee](https://rawtherapee.com/) RAW image processing software.

I have been using this profile for a week and RawTherapee has been working fine. However I'm pretty new to firejail so I'd like someone with more experience to review it.


etc/profile-m-z/rawtherapee.profile
```diff
@@ -0,0 +1,40 @@
+# Firejail profile for rawtherapee
+# Description: Free cross-platform raw image processing program
+# This file is overwritten after every install/update
+
```
glitsj16, line 4: This line can be dropped.

etc/profile-m-z/rawtherapee.profile
```diff
@@ -0,0 +1,40 @@
...
+# Persistent local customizations
+include rawtherapee.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.cache/RawTherapee
```
glitsj16, line 10: Firejail's logic assumes a `noblacklist foo` is warranted when a `blacklist foo` exists somewhere in our other included files. So you'll need to add `blacklist ${HOME}/.cache/RawTherapee` and `blacklist ${HOME}/.config/RawTherapee` to `disable-programs.inc` to protect those paths from being accessible in other application sandboxes.

glitsj16 (changes requested):
Looking good. Just add the blacklist counterparts for rawtherapee's cache and config paths please.

etc/profile-m-z/rawtherapee.profile
haplo: Done. :+1:

haplo:
Thank you for the review @glitsj16, I pushed the changes.

glitsj16 (approved):

PR #6181, opened 2024-01-30
Profile for Electron Cash

haplo:
This is a profile for the [Electron Cash](https://electroncash.org/) Bitcoin Cash SPV client.

### DIFF
--- a/etc/profile-a-l/electron-cash.profile
+++ b/etc/profile-a-l/electron-cash.profile
@@ -1,0 +1,54 @@
+# Firejail profile for electron-cash
+# Description: Lightweight Bitcoin Cash wallet
+# This file is overwritten after every install/update
+# Persistent local customizations
+include electron-cash.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.electron-cash
+
+# Allow python (blacklisted by disable-interpreters.inc)
+include allow-python3.inc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.electron-cash
+whitelist ${HOME}/.electron-cash
+include whitelist-common.inc
+include whitelist-var-common.inc
+
+caps.drop all
+ipc-namespace
+netfilter
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+
+disable-mnt
+private-bin electron-cash,python*
+private-cache
+?HAS_APPIMAGE: ignore private-dev
+private-dev
+private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,ld.so.cache,ld.so.preload,machine-id,pki,resolv.conf,ssl
+private-tmp
+
+# dbus-user none
+# dbus-system none
+
+restrict-namespaces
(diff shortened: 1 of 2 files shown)


etc/profile-a-l/electron-cash.profile
```diff
@@ -0,0 +1,54 @@
+# Firejail profile for electron-cash
+# Description: Lightweight Bitcoin Cash wallet
+# This file is overwritten after every install/update
+# Persistent local customizations
+include electron-cash.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.electron-cash
```
glitsj16, line 9: Firejail's logic assumes a `noblacklist foo` is warranted when a `blacklist foo` exists somewhere in our other included files. So you'll need to add `blacklist ${HOME}/.electron-cash` to `disable-programs.inc` to protect that path from being accessible in other application sandboxes.

etc/profile-a-l/electron-cash.profile
```diff
@@ -0,0 +1,54 @@
...
+novideo
+protocol unix,inet,inet6
+seccomp
+
+disable-mnt
+private-bin electron-cash,python*
+private-cache
+?HAS_APPIMAGE: ignore private-dev
+private-dev
+private-etc alternatives,ca-certificates,crypto-policies,dconf,fonts,ld.so.cache,ld.so.preload,machine-id,pki,resolv.conf,ssl
```
glitsj16, line 48: There's a new private-etc format now: https://github.com/netblue30/firejail/discussions/5610. Please use that for new profiles. In this case the line should look like:
`private-etc @network,@tls-ca,@x11`.

glitsj16 (changes requested):
Thanks for the profile! Some smallish changes are needed, check the comments please.

etc/profile-a-l/electron-cash.profile
haplo: That makes sense, I added the `blacklist` to `disable-programs.inc`.

etc/profile-a-l/electron-cash.profile
haplo: I tested this change and electron-cash fails to find `/etc/ssl/certs/ca-certificates.crt`.

Do I need master version of firejail?

```
$ firejail --version
firejail version 0.9.72

Compile time support:
- AppArmor support is enabled
- AppImage support is enabled
- chroot support is enabled
- firetunnel support is disabled
- IDS support is disabled
- output logging is enabled
- SELinux support is disabled
```

glitsj16: > Do I need master version of firejail?

Correct. This private-etc refactoring came in after 0.9.72. It's always good to use the git build when developing new profiles. Not strictly required, but saves some time and hair-pulling for code refactorings.

glitsj16 (approved):
Thanks for the changes!

etc/profile-a-l/electron-cash.profile
haplo: `firejail-git` is in Arch Linux's AUR so it'd be easy for me to install, I just don't want it to mess with my existing system. Is there any guide for developers?

I'm thinking setting up a VM might be best, although that would probably hide some hardware-access related issues. A Vagrantfile that sets everything up for development would be great IMO.

glitsj16: > Is there any guide for developers?

There are a few entrees on the [wiki](https://github.com/netblue30/firejail/wiki). Especially the ones under `firejail for contributors`.

Vagrant support is always welcome. I haven't seen anything from the regular contributors on that though. You could open a feature request for it.

PR #6182, opened 2024-01-31
crawl.profile: allow lua

luca0N:
Add common Lua include to crawl.profile (Dungeon Crawl Stone Soup) to allow Lua libraries, as both the ncurses and tiles executables are dynamically linked to Lua.


glitsj16 (approved):
LGTM

kmk3 (approved):

PR #6183, opened 2024-01-31
geeqie.profile: allow lua interpreter

haplo:
Recent versions of [geeqie](https://www.geeqie.org/) use a LUA interpreter, like the one in Arch Linux (2.2).

Without this fix it fails with:

```
/usr/bin/geeqie: error while loading shared libraries: liblua.so.5.4: cannot open shared object file: Permission denied
```

### DIFF
--- a/etc/profile-a-l/geeqie.profile
+++ b/etc/profile-a-l/geeqie.profile
@@ -10,6 +10,9 @@ noblacklist ${HOME}/.cache/geeqie
 noblacklist ${HOME}/.config/geeqie
 noblacklist ${HOME}/.local/share/geeqie
 
+# Allow lua (blacklisted by disable-interpreters.inc)
+include allow-lua.inc
+
 # Allow perl (blacklisted by disable-interpreters.inc)
 include allow-perl.inc
 


etc/profile-a-l/geeqie.profile
```diff
@@ -12,6 +12,8 @@ noblacklist ${HOME}/.local/share/geeqie
 
 # Allow perl (blacklisted by disable-interpreters.inc)
 include allow-perl.inc
+# Allow lua (blacklisted by disable-interpreters.inc)
```
glitsj16, line 15: Nitpick. We tend to sort things alphabetically. Please put the lua block above the perl one.

glitsj16 (commented):
LGTM. Just a nitpick about alphabetical ordering...

etc/profile-a-l/geeqie.profile
haplo: No worries, I'm a nitpicker myself. :)

haplo:
@glitsj16 Pushed new version.

glitsj16 (approved):

etc/profile-a-l/geeqie.profile
```diff
@@ -10,6 +10,8 @@ noblacklist ${HOME}/.cache/geeqie
 noblacklist ${HOME}/.config/geeqie
 noblacklist ${HOME}/.local/share/geeqie
 
+# Allow lua (blacklisted by disable-interpreters.inc)
+include allow-lua.inc
```
kmk3, line 14:
```suggestion
include allow-lua.inc

```

Add missing blank line.

See [profile.template](https://github.com/netblue30/firejail/blob/760f50f78ad13664d7a32b4577381c0341ab2d4a/etc/templates/profile.template#L76).

kmk3 (commented):
It would be helpful to have the explanation in the commit message itself so
that it is available in the git log.

Example (formatted to make it easier to read as plain text):

```
geeqie.profile: allow lua interpreter

Recent versions of geeqie[1] use a Lua interpreter, like the one in Arch
Linux (2.2).

Without this fix it fails with:

    /usr/bin/geeqie: error while loading shared libraries: liblua.so.5.4: [...]

[1] https://www.geeqie.org/
```

haplo:
> It would be helpful to have the explanation in the commit message itself so
> that it is available in the git log.

You are completely right! I reworded the commit and included the missing blank line.

kmk3 (approved):

PR #6187, opened 2024-02-02
landlock: split .special into .makeipc and .makedev

kmk3:
As discussed with @topimiettinen[1], it is unlikely that an unprivileged
process would need to directly create block or character devices.  Also,
`landlock.special` is not very descriptive of what it allows.

So split `landlock.special` into:

* `landlock.makeipc`: allow creating named pipes and sockets (which are
  usually used for inter-process communication)
* `landlock.makedev`: allow creating block and character devices

Misc: The `makedev` name is based on `nodev` from mount(8), which makes
mount not interpret block and character devices.  `ipc` was suggested by
@rusty-snake[2].

Relates to #6078.

[1] https://github.com/netblue30/firejail/pull/6078#pullrequestreview-1740569786
[2] https://github.com/netblue30/firejail/pull/6187#issuecomment-1924107294

### DIFF
--- a/contrib/syntax/lists/profile_commands_arg1.list
+++ b/contrib/syntax/lists/profile_commands_arg1.list
@@ -30,8 +30,9 @@ iprange
 join-or-start
 keep-fd
 landlock.execute
+landlock.makedev
+landlock.makeipc
 landlock.read
-landlock.special
 landlock.write
 mac
 mkdir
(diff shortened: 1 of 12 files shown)


rusty-snake:
Good to see this design choice getting corrected.

`landlock.special` is now a `landlock.ipc`. Do we want to rename it too (in a followup PR)?

kmk3:
> Good to see this design choice getting corrected.
> 
> `landlock.special` is now a `landlock.ipc`. Do we want to rename it too (in a
> followup PR)?

Good idea, renamed them to `landlock.makeipc` and `landlock.makedev`.

topimiettinen (approved):

Issue #6191, opened 2024-02-04
lutris: gamescope doesnt work with firejail 


nutta-git:
### Description

Using [gamescope](https://wiki.archlinux.org/title/Gamescope) within lutris crashes games on launch. 

### Steps to Reproduce

_Steps to reproduce the behavior_

- ln -s /usr/bin/firejail /usr/local/bin/lutris
- cp /etc/firejail/lutris.profile ~/.config/firejail/lutris.profile
- firecfg --fix
- Launch Lutris
- Right Click any game
- Click Configure
- Navigate to System options Tab
- Navigate to Gamescope 
![gamescope ](https://github.com/netblue30/firejail/assets/59726608/f0d3bda5-abca-40b9-a1cb-a0f4361850a2)

### Expected behavior

Games should run without crashing 

### Actual behavior

Games crash on launch

### Behavior without a profile

Games run through gamescope as intended 

### Additional context

![lutristerminallog](https://github.com/netblue30/firejail/assets/59726608/cee65339-2403-4ca4-9f7c-b9354b32593b)

### Environment

- Arch Linux
- firejail version 0.9.72

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [ ] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ ] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

<details>
<summary>Output of <code>LC_ALL=C firejail /path/to/program</code></summary>
<p>

```
[kevin@archlinux ~]$ lutris 
Reading profile /home/kevin/.config/firejail/lutris.profile
Reading profile /etc/firejail/allow-python2.inc
Reading profile /etc/firejail/allow-python3.inc
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-exec.inc
[...]

PR #6192, opened 2024-02-05
profiles: lutris: remove noroot & private tmp

nutta-git:
fixes: lutris: gamescope doesnt work with firejail. 
https://github.com/netblue30/firejail/issues/6191

### DIFF
--- a/etc/profile-a-l/lutris.profile
+++ b/etc/profile-a-l/lutris.profile
@@ -66,7 +66,6 @@ netfilter
 nodvd
 nogroups
 nonewprivs
-ignore noroot
 notv
 nou2f
 novideo
@@ -76,7 +75,6 @@ seccomp.32 !modify_ldt
 
 # Add the next line to your lutris.local if you do not need controller support.
 #private-dev
-ignore private-tmp
 
 dbus-user filter
 dbus-user.own net.lutris.Lutris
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/lutris.profile
```diff
@@ -76,7 +76,7 @@ seccomp.32 !modify_ldt
 
 # Add the next line to your lutris.local if you do not need controller support.
 #private-dev
-private-tmp
+ignore private-tmp
```
rusty-snake, line 79: They are never set. Just remove them.

nutta-git: is this what you meant?

etc/profile-a-l/lutris.profile
kmk3, line 1: As mentioned [here][1], the issue from #6191 seems to be a bug in
gamescope/xwayland, so instead of weakening the profile, it might be better to
just add a comment until it's fixed:

```diff
 # Persistent global definitions
 include globals.local
+
+# If you use gamescope, add the following to lutris.local to work around a bug
+# (see #6191):
+#ignore noroot
+#ignore private-tmp
```

[1]: https://github.com/netblue30/firejail/issues/6191#issuecomment-1937566865

netblue30:
Merging in the suggestion from kmk3 above
`````
+# If you use gamescope, add the following to lutris.local to work around a bug
+# (see #6191):
+#ignore noroot
+#ignore private-tmp
`````

Thanks for the bug!

PR #6201, opened 2024-02-07
gnome-keyring: harden and add gnome-keyring-daemon.profile

glitsj16:
We were missing a profile for gnome-keyring-daemon. Let's bring that in.
Also slightly hardened gnome-keyring.profile.

### DIFF
--- a/etc/profile-a-l/gnome-keyring-daemon.profile
+++ b/etc/profile-a-l/gnome-keyring-daemon.profile
@@ -1,0 +1,62 @@
+# Firejail profile for gnome-keyring-daemon
+# Description: Stores passwords and encryption keys
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include gnome-keyring-daemon.local
+# Persistent global definitions
+include globals.local
+
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}/wayland-*
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-programs.inc
+#include disable-X11.inc # x11 none
+include disable-xdg.inc
+
+whitelist ${RUNUSER}/gnupg
+whitelist ${RUNUSER}/keyring
+whitelist /usr/share/gnupg
+whitelist /usr/share/gnupg2
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
+netfilter
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+seccomp.block-secondary
+tracelog
+x11 none
+
+disable-mnt
+#private-bin gnome-keyrin*,secret-tool
+private-cache
+private-dev
+#private-lib alternatives,gnome-keyring,libsecret-1.so.*,pkcs11,security
+private-tmp
+
+#dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+restrict-namespaces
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/gnome-keyring-daemon.profile
```diff
@@ -0,0 +1,11 @@
...
+# Description: Stores passwords and encryption keys
+# This file is overwritten after every install/update
+# Persistent local customizations
+include gnome-keyring-daemon.local
+# Persistent global definitions
+# added by included profile
+#include globals.local
+
+# Redirect
+include gnome-keyring.profile
```
kmk3, line 11: From gnome-keyring.profile:

> ```
> mkdir ${HOME}/.gnupg
> whitelist ${HOME}/.gnupg
> whitelist ${DOWNLOADS}
> ```

I assume that gnome-keyring-daemon is the Gnome equivalent of ssh-agent.

In that case, does the daemon itself really need direct access to the user's
keys?  Isn't there something analogous to ssh-add?

kmk3: If the access is not needed, it might be better to reverse the redirect:

* Make the base profile be gnome-keyring-daemon
* Make gnome-keyring allow the extra paths and redirect to gnome-keyring-daemon

glitsj16: All valid points. Pushed the changes to reflect your review. Thanks!

kmk3 (approved):

kmk3:
Rebased to make the changes clearer.

Please make sure to use a normal merge rather than squash to keep the hardening
and refactor commits separate.

glitsj16:
@kmk3

> Please make sure to use a normal merge rather than squash to keep the hardening
and refactor commits separate.

Didn't realize the default I have on GH is still 'squash'. Changing that now!

kmk3:
@glitsj16 [on Feb 8](https://github.com/netblue30/firejail/pull/6201#issuecomment-1934151251):

> > Please make sure to use a normal merge rather than squash to keep the hardening
> > and refactor commits separate.
> 
> Didn't realize the default I have on GH is still 'squash'. Changing that now!

It depends, I'd avoid squashing if it looks like the author intended for the
commits to be separate, especially if there are multiple logical changes.

And I'd be inclined to organize the branch and/or squash it when there are
amend/fixup commits.

Also, when there is a single commit/single logical change it seems helpful to
squash to avoid the extra noise of a merge commit.

So it depends on whether the committer already intentionally split them and
whether squashing would make the end result more or less clear.

kmk3:
@glitsj16 [on Feb 8](https://github.com/netblue30/firejail/pull/6201#issuecomment-1934151251):

> Heh, the good old `it depends...` situation. Thanks again for explaining! I
> do realize my contributions via GH aren't always 'clean' (understating it).

It's alright, usually the changes are related and localized enough that
squashing is workable.

In the case of profiles, it's mostly when moving code/renaming files that
keeping such a change in its own separate commit makes any other changes much
clearer.

> On a side-note I wish to convey my appreciation for the recent landlock work
> you've been doing. I can start to grasp it all now and am impressed by the
> speed of it. Interesting times ahead! Have fun.

Glad to hear it!  I'm not sure how recent you mean, but the refactoring in
#6078 was indeed a bit mind-bending, as there was a lot of cleanup to do, the
changes had different authors and some changes were instructive enough to
warrant ensuring that they had their own commits.

For example, the following commits show how to add a linked list for a new
path-related command and how to process it all at once at a specific time
(instead of applying each command at parse time):

* b94cc754a ("landlock: apply rules in sandbox before app start", 2023-10-26)
* 520508d5b ("landlock: avoid parsing landlock commands twice", 2023-11-02)

Which is what made it easy to make the application of all landlock commands
conditional in #6125 (in [sandbox.c][1]):

* 760f50f78 ("landlock: move commands into profile and add landlock.enforce",
  2023-11-17)

In #6200 I was just lucky that my first guess was correct :)

[1]: https://github.com/netblue30/firejail/commit/760f50f78ad13664d7a32b4577381c0341ab2d4a#diff-767ac3de885e9c994fed6eb2a0f8234fd8d8611a5f817d2b0a37b50f4101b321L533-R523

PR #6203, opened 2024-02-08
landlock: sort options in man pages

glitsj16:


kmk3 (commented):
I thought about doing this too (especially for the docs), though note that the
same order is used everywhere in the repository (other than auto-generated
files), which makes it easier to keep organized.

For example, see the changes in #6187.

The current order is the usual `rwx` plus some other writes in the middle.

glitsj16:
@kmk3

> I thought about doing this too (especially for the docs), though note that the
same order is used everywhere in the repository (other than auto-generated
files), which makes it easier to keep organized.

In that case I'll just close this PR. No problem.

Thanks for the info!

PR #6216, opened 2024-02-19
multimc: instances not running, because of missing permissions

powerjungle:
When starting an instance, in the logs, a failed attempt to load the lwjgl library is shown and the game doesn't run.
The library is in the /tmp directory. The reason for this appears to be, in the lwjgl source code, the shared library loading function, extracts in the temporary directory and continues from there. This is fixed by whitelisting.

The reason for adding "ignore noexec /tmp" as well, is that without it, the game can't run, even if the directory is whitelisted. It seems the library needs to be loaded from /tmp.

A second error for a failed attempt to access /home/user/.cache/JNA is also shown in the logs. This is also fixed by whitelisting.

The errors which inspired the commit:
```
Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.UnsatisfiedLinkError: /home/user/.cache/JNA/temp/jna205883382989158645.tmp: /home/user/.cache/JNA/temp/jna205883382989158645.tmp: failed to map segment from shared object [in thread "main"]
Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.UnsatisfiedLinkError: /tmp/lwjgl_user/3.3.2+13/x64/liblwjgl.so: /tmp/lwjgl_user/3.3.2+13/x64/liblwjgl.so: failed to map segment from shared object [in thread "Render thread"]
```

Links to the source code that should be responsible for the /tmp error:

https://github.com/LWJGL/lwjgl3/blob/56d09c39ff26ad49c74f6be6884d60ef3f597384/modules/lwjgl/core/src/main/java/org/lwjgl/system/SharedLibraryLoader.java#L51

https://github.com/LWJGL/lwjgl3/blob/56d09c39ff26ad49c74f6be6884d60ef3f597384/modules/lwjgl/core/src/main/java/org/lwjgl/system/SharedLibraryLoader.java#L137

### DIFF
--- a/etc/profile-m-z/multimc5.profile
+++ b/etc/profile-m-z/multimc5.profile
@@ -8,10 +8,15 @@ include globals.local
 noblacklist ${HOME}/.local/share/multimc
 noblacklist ${HOME}/.local/share/multimc5
 noblacklist ${HOME}/.multimc5
+noblacklist ${HOME}/.cache/JNA
+noblacklist /tmp/lwjgl_*
 
 # Ignore noexec on ${HOME} as MultiMC installs LWJGL native
 # libraries in ${HOME}/.local/share/multimc
 ignore noexec ${HOME}
+
+# Ignore noexec on /tmp as LWJGL extracts libraries to /tmp
+ignore noexec /tmp
 
 # Allow java (blacklisted by disable-devel.inc)
 include allow-java.inc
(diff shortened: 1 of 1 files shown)


topimiettinen:
The extremely predictable path in `/tmp` is most likely a security vulnerability [(CWE-377](https://cwe.mitre.org/data/definitions/377.html)), especially on a multi-user system.

With Firejail, `private-tmp` gives protection.

powerjungle:
> With Firejail, `private-tmp` gives protection.

The original profile has it in and I haven't changed that. Thanks, for mentioning, since I would have forgotten to check.

netblue30:
merged, thanks!

PR #6211, opened 2024-02-16
nextcloud: D-Bus filtering changes

glitsj16:
Relates to https://github.com/netblue30/firejail/discussions/6209.

Additional context:
- created wiki entree on [how to create local D-Bus overrides](https://github.com/netblue30/firejail/wiki/Frequently-Asked-Questions#how-do-i-sandbox-systemd-started-applications)
- edited the [firecfg FAQ](https://github.com/netblue30/firejail/wiki/Frequently-Asked-Questions#how-do-i-undo-firecfg) to make it clear there's native support via firecfg.d now

PR #6218, opened 2024-02-21
profiles: drop paths already in wusc

glitsj16:
Detected a few profiles that include wusc yet still reference paths that are present in that file. This PR drops the 'double' whitelisting from the relevant profiles.


kmk3 (approved):

kmk3:
(Offtopic)

Any thoughts on the following comment?

* <https://github.com/netblue30/firejail/pull/6211#issuecomment-1952716459>

Issue #3595, opened 2020-08-19
Cannot firejail ledger-live-desktop appimage


eloyesp:
I'm trying to run ledger-live without network access but the application do not start (it stop immediately after start) without any error message.

It is an appimage, so I've always tested with the `--appimage` option.

The application does run when running without firejail, but always fails with it. I've tested with `--noprofile` and it fails anyway.

**Environment**

 - I'm testing on Trisquel 9 (based on ubuntu 18.4)
 - Firejail version (output of `firejail --version`) exclusive or used git commit (`git rev-parse HEAD`)

```
firejail version 0.9.52

Compile time support:
        - AppArmor support is enabled
        - AppImage support is enabled
        - bind support is enabled
        - chroot support is enabled
        - file and directory whitelisting support is enabled
        - file transfer support is enabled
        - git install support is disabled
        - networking support is enabled
        - overlayfs support is enabled
        - private-home support is enabled
        - seccomp-bpf support is enabled
        - user namespace support is enabled
        - X11 sandboxing support is enabled
```
 - What other programs interact with the affected program for the functionality? NO
 - Are these listed in the profile? NO

**Additional context**

I don't know how to relax firejail default options (no dropping capabilities, disable seccomp, etc) on the command line, and could not find any way to use a profile with an appimage.

**Checklist**
 - [X] The upstream profile (and redirect profile if exists) have no changes fixing it.
 - [X] The upstream profile exists (`find / -name 'firejail' 2>/dev/null`/`fd firejail` to locate profiles ie in `/usr/local/etc/firejail/PROGRAM.profile`)
 - [X] Programs needed for interaction are listed.
 - [X] Error was checked in search engine and on issue list without success.

<details><summary> debug output </summary>

```
OUTPUT OF `firejail --debug --appimage ./ledger-live-desktop-2.10.0-linux-x86_64.AppImage`
total 0
lrwx------ 1 username username 64 Aug 19 18:06 0 -> /dev/null
l-wx------ 1 username username 64 Aug 19 18:06 1 -> pipe:[3630483]
lrwx------ 1 username username 64 Aug 19 18:06 2 -> /dev/pts/2
lrwx------ 1 username username 64 Aug 19 18:06 20 -> socket:[26133]
lr-x------ 1 username username 64 Aug 19 18:06 3 -> /proc/22939/fd
lrwx------ 1 username username 64 Aug 19 18:06 37 -> /dev/dri/card0
Autoselecting /bin/bash as shell
[...]

PR #5987, opened 2023-09-06
profiles: fix commented code and eol comments

kmk3:
Main changes:

* Remove the space after `#` for commented code lines to distinguish
  them from normal comments
* Use `#` instead of `-` for comments at the end of the line so that
  commented code lines work after being uncommented

Commands used to search and replace:

    arg0="$(cat contrib/syntax/lists/profile_commands_arg0.list |
      LC_ALL=C sort -u | tr '\n' '|' | sed -e 's/|$//' -e 's/\./\\./g')"
    arg1="$(cat contrib/syntax/lists/profile_commands_arg1.list |
      LC_ALL=C sort -u | tr '\n' '|' | sed -e 's/|$//' -e 's/\./\\./g')"
    git ls-files -z -- etc/inc etc/profile* | xargs -0 -I '{}' \
      sh -c "printf '%s\n' \"\$(sed -E \
        -e 's/^# ($arg0)( [#-]-? .*)?\$/#\\1\\2/' \
        -e 's/^# ($arg1)( [^ ]*)?( [#-]-? .*)?\$/#\\1\\2\\3/' \
        -e 's/^# (whitelist \\$)/#\\1/' \
        -e 's/^(#[^ ].+) --? /\\1 # /' \
        '{}')\" >'{}'"

Commands used to check for leftover entries:

    arg0="$(cat contrib/syntax/lists/profile_commands_arg0.list |
      LC_ALL=C sort -u | tr '\n' '|' | sed -e 's/|$//' -e 's/\./\\./g')"
    arg1="$(cat contrib/syntax/lists/profile_commands_arg1.list |
      LC_ALL=C sort -u | tr '\n' '|' | sed -e 's/|$//' -e 's/\./\\./g')"
    git grep -E "^# ($arg0|$arg1)( +|$)" -- etc/inc etc/profile*

See also commit 30f9ad908 ("build: improve comments in firecfg.config",
2023-08-05) / PR #5942.

PR #6219, opened 2024-02-21
Profile for Ledger Live desktop app

haplo:
This is a profile for the [Ledger Live wallet](https://www.ledger.com/ledger-live).

I based it off the profile offered in #3595, but it needed several tweaks for it to work.

I tested the binary installation in /opt (I use the [ledger-live-bin package in AUR](https://aur.archlinux.org/packages/ledger-live-bin)). Ledger Live is also offered as an AppImage, but I didn't test that case.

I own a Ledger USB wallet and this profile is working for me.

### DIFF
--- a/etc/profile-a-l/ledger-live-desktop.profile
+++ b/etc/profile-a-l/ledger-live-desktop.profile
@@ -1,0 +1,61 @@
+# Firejail profile for Ledger Live desktop app
+# Description: Cryptocurrency wallet by the makers of Ledger hardware wallets
+# This file is overwritten after every install/update
+# Persistent local customizations
+include ledger-live-desktop.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/Ledger Live
+
+# Added by disable-exec.inc, breaks hardware wallet manager
+ignore noexec /tmp
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/Ledger Live
+whitelist ${HOME}/.config/Ledger Live
+whitelist ${DOWNLOADS}
+whitelist /opt/ledger-live
+include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
+netfilter
+nodvd
+nogroups
+nonewprivs
+noprinters
+noroot
+nosound
+notv
+novideo
+protocol unix,inet,inet6,netlink
+seccomp !chroot
+tracelog
+
+disable-mnt
+private-cache
+# enabling private-dev blocks USB hardware wallets, if you don't need access to
+# USB devices you can add private-dev to your ledger-live-desktop.local
+#private-dev
+private-etc @network,@tls-ca,@x11,host.conf,rpc
+private-lib
+private-tmp
+
+# app attempts to connect to dbus but seems to work fine when blocked
+dbus-user none
+dbus-system none


etc/profile-a-l/ledger-live-desktop.profile
```diff
@@ -0,0 +1,45 @@
...
+novideo
+protocol unix,inet,inet6,netlink
+seccomp !chroot
+tracelog
+
+disable-mnt
+private-cache
+# enabling private-dev blocks USB hardware wallets
+# private-dev
+private-etc @network,@tls-ca,@x11
```
haplo, line 39: I' not running firejail from master so my actual profile uses:

```
private-etc alternatives,ca-certificates,crypto-policies,host.conf,nsswitch.conf,pki,resolv.conf,rpc,selinux,ssl
```

Networking breaks without these options.

etc/profile-a-l/ledger-live-desktop.profile
```diff
@@ -0,0 +1,45 @@
...
+nosound
+novideo
+protocol unix,inet,inet6,netlink
+seccomp !chroot
+tracelog
+
+disable-mnt
+private-cache
+# enabling private-dev blocks USB hardware wallets
+# private-dev
```
haplo, line 38: I understand not having `private-dev` is a big security hole, but it's the only way I found to get the USB wallet to work. Is there a way to be more fine-grained?

etc/profile-a-l/ledger-live-desktop.profile
glitsj16: Good you referenced your private-etc in 0.9.72-format. It translates as private-etc @network,@tls-ca,@x11,host.conf,rpc so you'll need to append `,host.conf,rpc` here for master.

etc/profile-a-l/ledger-live-desktop.profile
```diff
@@ -0,0 +1,45 @@
...
+include globals.local
+
+noblacklist ${HOME}/.config/Ledger Live
+
+mkdir ${HOME}/.config/Ledger Live
+whitelist ${HOME}/.config/Ledger Live
+whitelist ${DOWNLOADS}
+whitelist /opt/ledger-live
+include whitelist-common.inc
+include whitelist-usr-share-common.inc
```
glitsj16, line 15: For alphabetical ordering, move wusc below wrc please.
Can we also add `whitelist run-common.inc`?

etc/profile-a-l/ledger-live-desktop.profile
```diff
@@ -0,0 +1,45 @@
...
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
+netfilter
+nodvd
+nogroups
+nonewprivs
```
glitsj16, line 26: I'm assuming the profile won't break when adding `noprinters`, but I can't test that.

etc/profile-a-l/ledger-live-desktop.profile
```diff
@@ -0,0 +1,45 @@
...
+noroot
+notv
```
glitsj16, line 28: Alphabetical ordering nitpick. Move this below nosound please.

etc/profile-a-l/ledger-live-desktop.profile
glitsj16: The breakage probably is due to your use of the AppImage. We have a conditional for that use case, so the same profile can support both:
`?HAS_APPIMAGE: ignore private-dev`
private-dev

etc/profile-a-l/ledger-live-desktop.profile
```diff
@@ -0,0 +1,45 @@
...
+private-cache
+# enabling private-dev blocks USB hardware wallets
+# private-dev
+private-etc @network,@tls-ca,@x11
+private-lib
+private-tmp
+
+# app attempts to connect to dbus but seems to work fine when blocked
+dbus-user none
+dbus-system none
```
glitsj16, line 45: Can we add `restrict-namespaces`?

etc/profile-a-l/ledger-live-desktop.profile
```diff
@@ -0,0 +1,45 @@
+# Firejail profile for Ledger Live
+# This file is overwritten after every install/update
+# Persistent local customizations
+include ledger-live-desktop.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/Ledger Live
+
```
glitsj16, line 9: You forgot all the `disable-*.inc` includes... Follow the template and double-check if the app still has its regular functionalities please.

```
[...]
# remove disable-write-mnt.inc if you set disable-mnt
include disable-common.inc
include disable-devel.inc
include disable-exec.inc
include disable-proc.inc
include disable-programs.inc
include disable-shell.inc
include disable-write-mnt.inc
include disable-X11.inc
include disable-xdg.inc
[...]
```

glitsj16 (changes requested):
The profile is missing some vitals. See individual comments.

etc/profile-a-l/ledger-live-desktop.profile
haplo: I'm using the _/opt/ledger-live_ installation, not the appimage. What I meant is that I haven't tested the appimage release.

Enabling `private-dev` breaks the hardware wallet integration, even after removing `nou2f`.

etc/profile-a-l/ledger-live-desktop.profile
haplo: Sorry for this oversight, initially I was including `electron.profile`, but eventually removed it and forgot to add the disables.

I have added them now, will comment on the issues I had in separate threads.

etc/profile-a-l/ledger-live-desktop.profile
```diff
@@ -0,0 +1,61 @@
...
+# Added by disable-exec.inc, breaks hardware wallet manager
+ignore noexec /tmp
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
```
haplo, line 18: This causes these error logs, but the app seems to work fine:

```
[15:0222/105618.284611:ERROR:file_path_watcher_inotify.cc(822)] Failed to read /proc/sys/fs/inotify/max_user_watches
[1:0222/105618.286875:ERROR:file_path_watcher_inotify.cc(822)] Failed to read /proc/sys/fs/inotify/max_user_watches
<----------CUT----------->
[12:0222/105618.665653:ERROR:file_path_watcher_inotify.cc(822)] Failed to read /proc/sys/fs/inotify/max_user_watches
```

etc/profile-a-l/ledger-live-desktop.profile
haplo: `disable-X11.inc` causes the app not to start with this error log:

```
[12:0222/101944.253536:ERROR:ozone_platform_x11.cc(240)] Missing X server or $DISPLAY
```

etc/profile-a-l/ledger-live-desktop.profile
haplo: I added `noprinters` and it seems to work fine.

haplo:
Thank you very much for the review @glitsj16. I believe I have addressed all your comments. The updated profile is working for me.

glitsj16 (approved):
Thanks for the changes. LGTM.

etc/profile-a-l/ledger-live-desktop.profile
```diff
@@ -0,0 +1,61 @@
...
+novideo
+protocol unix,inet,inet6,netlink
+seccomp !chroot
+tracelog
+
+disable-mnt
+private-cache
+# enabling private-dev blocks USB hardware wallets, if you don't need access to
+# USB devices you can add private-dev to your ledger-live-desktop.local
+# private-dev
```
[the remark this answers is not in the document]
haplo: > For clarity, use `#` for code and `# ` for normal comments (see #5987).

Good point, I fixed it and will keep it in mind going forward.

netblue30:
merged, thanks!

PR #6222, opened 2024-02-24
build: allow overriding certain tools & sync targets with CI

kmk3:
Allow overriding the following tools at configure-time and build-time:

* codespell
* cppcheck
* gawk
* scan-build

For example, instead of hardcoding `gawk`, enable overriding it at
configure-time with:

    ./configure GAWK=/path/to/gawk

To override it for a single `make` invocation:

    make GAWK=/path/to/gawk

Also, add default values for the programs that are not found (rather
than leaving the variables empty), to make error messages clearer when
trying to run them:

    $ make CPPCHECK= cppcheck-old
    [...]
    force --error-exitcode=1 --enable=warning,performance .
    make: force: No such file or directory
    $ make CPPCHECK=cppcheck cppcheck-old
    [...]
    cppcheck --force --error-exitcode=1 --enable=warning,performance .
    make: cppcheck: No such file or directory

### DIFF
--- a/.github/workflows/check-c.yml
+++ b/.github/workflows/check-c.yml
@@ -67,11 +67,11 @@ jobs:
       run: ./ci/printenv.sh
     - name: configure
       run: >
-        CC=clang-14 ./configure --enable-fatal-warnings --enable-apparmor
-        --enable-selinux
+        ./configure CC=clang-14 SCAN_BUILD=scan-build-14
+        --enable-fatal-warnings --enable-apparmor --enable-selinux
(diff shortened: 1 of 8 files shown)


kmk3:
Cc: @reinerh @rusty-snake

Hello, you might find this interesting since you also worked on these checks.

This PR enables overrides for certain tools, which should allow more
deduplication of code between the build and CI and make it easier to run the
same CI steps locally.

reinerh (approved):

glitsj16 (approved):

PR #6228, opened 2024-02-28
landlock: use "landlock.fs." prefix in filesystem commands

kmk3:
Since Landlock ABI v4 it is possible to restrict actions related to the
network and potentially more areas will be added in the future.

So use `landlock.fs.` as the prefix in the current filesystem-related
commands (and later `landlock.net.` for the network-related commands) to
keep them organized and to match what is used in the kernel.

Examples of filesystem and network access flags:

* `LANDLOCK_ACCESS_FS_EXECUTE`: Execute a file.
* `LANDLOCK_ACCESS_FS_READ_DIR`: Open a directory or list its content.
* `LANDLOCK_ACCESS_NET_BIND_TCP`: Bind a TCP socket to a local port.
* `LANDLOCK_ACCESS_NET_CONNECT_TCP`: Connect an active TCP socket to a
  remote port.

Relates to #6078.

### DIFF
--- a/contrib/syntax/lists/profile_commands_arg1.list
+++ b/contrib/syntax/lists/profile_commands_arg1.list
@@ -29,11 +29,11 @@ ip6
 iprange
 join-or-start
 keep-fd
-landlock.execute
-landlock.makedev
-landlock.makeipc
-landlock.read
-landlock.write
+landlock.fs.execute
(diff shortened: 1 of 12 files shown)


glitsj16 (approved):
Very nice work!

Issue #6224, opened 2024-02-27
build: libtrace(log) warnings on Alpine


kmk3:
On commit f7e2c2573 ("build: ensure fnettrace prints to stdout", 2024-02-12):

From `build_src_package`:

<https://gitlab.com/Firejail/firejail_ci/-/jobs/6193214960>:

<details>
<summary>Build log</summary>
<p>

```console
$ apk upgrade
OK: 7 MiB in 15 packages
$ apk add build-base linux-headers gawk
(1/24) Installing libgcc (13.2.1_git20231014-r0)
(2/24) Installing jansson (2.14-r4)
(3/24) Installing libstdc++ (13.2.1_git20231014-r0)
(4/24) Installing zstd-libs (1.5.5-r8)
(5/24) Installing binutils (2.41-r0)
(6/24) Installing libmagic (5.45-r1)
(7/24) Installing file (5.45-r1)
(8/24) Installing libgomp (13.2.1_git20231014-r0)
(9/24) Installing libatomic (13.2.1_git20231014-r0)
(10/24) Installing gmp (6.3.0-r0)
(11/24) Installing isl26 (0.26-r1)
(12/24) Installing mpfr4 (4.2.1-r0)
(13/24) Installing mpc1 (1.3.1-r1)
(14/24) Installing gcc (13.2.1_git20231014-r0)
(15/24) Installing libstdc++-dev (13.2.1_git20231014-r0)
(16/24) Installing musl-dev (1.2.4_git20230717-r4)
(17/24) Installing libc-dev (0.7.2-r5)
(18/24) Installing g++ (13.2.1_git20231014-r0)
(19/24) Installing make (4.4.1-r2)
(20/24) Installing fortify-headers (1.1-r3)
(21/24) Installing patch (2.7.6-r10)
(22/24) Installing build-base (0.5-r3)
(23/24) Installing gawk (5.3.0-r0)
(24/24) Installing linux-headers (6.5-r0)
Executing busybox-1.36.1-r15.trigger
OK: 233 MiB in 39 packages
[...]
$ ./configure --prefix=/usr || (cat config.log; exit 1)
[...]
$ make
[...]
make[1]: Entering directory '/builds/Firejail/firejail_ci/src/libtrace'
[...]

PR #5960, opened 2023-08-16
ci: whitelist paths, reorganize workflows & speed-up tests

kmk3:
Main changes:

* Run workflows based on path whitelists rather than blacklists
* Reorganize jobs and workflows, mostly based on the paths used
* Run make in parallel where applicable
* docs: add missing CI badges to README.md
* Split tests into multiple jobs

Considering the most recent runs, this reduces the total amount of time
it takes to run the tests from about 9-10 minutes to about 3 minutes.

PR #5148, opened 2022-05-17
ci: print version after install & fix apparmor support on build_apparmor

kmk3:
The "build_apparmor" job was added on commit 342e71cd8 ("Add
deb-apparmor build to Gitlab CI", 2019-01-26).  It would call
`./mkdeb-apparmor.sh`, which would run `./configure --enable-apparmor`
directly, adding `-lapparmor` to `EXTRA_LDFLAGS` and thus passing it to
the linker.

Later, commit 87e7b3139 ("Configure Debian package with AA and SELinux
options", 2020-05-13) / PR #3414 merged mkdeb.sh and mkdeb-apparmor.sh
into mkdeb.sh.in, which does not always pass `--enable-apparmor` to
./configure directly.  Instead, it adds `--enable-apparmor` depending on
whether the `$HAVE_APPARMOR` environment variable is set, which would be
done by a previous run of ./configure with `--enable-apparmor`.  Since
on "build_apparmor" ./configure is not run the first time with
`--enable-apparmor`, neither is it on the second time and thus
`-lapparmor` is never passed to the linker.  This commit adds
`--enable-apparmor` to the first ./configure run on the ci job, so that
it gets passed to the one being executed on mkdeb.sh as well.

PR #6230, opened 2024-02-29
build: reduce hardcoding and inconsistencies

kmk3:
Main changes:

* build: reduce TARNAME hardcoding
* build: standardize parallel make arguments
* build: standardize ./configure arguments

Relates to #5148 #5960 #6224.

This is a follow-up to #6222.


netblue30:
merged, let's try it!

PR #6235, opened 2024-03-01
New profile: editorconfiger

glitsj16:
editorconfiger

Description: Plain tool to validate and compare .editorconfig files.

https://github.com/aegoroff/editorconfiger
https://aur.archlinux.org/packages/editorconfiger
https://aur.archlinux.org/packages/editorconfiger-bin

### DIFF
--- a/etc/profile-a-l/editorconfiger.profile
+++ b/etc/profile-a-l/editorconfiger.profile
@@ -1,0 +1,56 @@
+# Firejail profile for editorconfiger
+# Description: Plain tool to validate and compare .editorconfig files
+# This file is overwritten after every install/update
+# Persistent local customizations
+include editorconfiger.local
+# Persistent global definitions
+include globals.local
+
+blacklist /tmp/.X11-unix
+blacklist /usr/libexec
+blacklist ${RUNUSER}
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
+net none
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noprinters
+noroot
+nosound
+notv
+nou2f
+novideo
+seccomp socket
+seccomp.block-secondary
+tracelog
+x11 none
+
+#disable-mnt
+private-bin editorconfiger
+private-cache
+private-dev
+private-etc .editorconfig
+private-lib
+private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+read-only ${HOME}
+restrict-namespaces
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -225,6 +225,7 @@ ebook-edit
 ebook-meta
 ebook-polish
 ebook-viewer
+editorconfiger
 electron-mail
 electrum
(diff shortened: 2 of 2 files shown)


etc/profile-a-l/editorconfiger.profile
```diff
@@ -0,0 +1,56 @@
...
+private-dev
+private-etc .editorconfig
+private-lib
+private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+read-only ${HOME}
```
kmk3, line 55:
```suggestion
```

From <https://github.com/aegoroff/editorconfiger>:

```
Usage: editorconfiger [COMMAND]

Commands:
  vf          Validate single .editorconfig file
  vd          Validate all found .editorconfig files in a directory and all its children
  c           Compare two .editorconfig files
  completion  Generate the autocompletion script for the specified shell
  help        Print this message or the help of the given subcommand(s)
```

Apparently it can generate files.

glitsj16: > Apparently it can generate files.

Only those shell autocompletions need rw in their specific locations under ${HOME}. I've added those.

kmk3: > > Apparently it can generate files.
> 
> Only those shell autocompletions need rw in their specific locations under
> ${HOME}. I've added those.

Does the program write to those specific directories?  I couldn't find them
mentioned in the repository.

Honestly I don't really understand what the shell would be autocompleting in
the CLI in the case of .editorconfig files and it's not immediately clear in
the source code.

Is it the content of the parsed .editorconfig files?

In this case, wouldn't they be generated at the current directory?

If it's for the program itself I'd think that the project would just generate
the completion files at build time and be packaged with them rather than
generating them at runtime.

glitsj16: > Does the program write to those specific directories? I couldn't find them
mentioned in the repository.

No, editorconfiger just writes to `stdout`. Users would have to redirect it to the appropriate path(s) themselves.

```sh
$ editorconfiger help generator
Generate the autocompletion script for the specified shell

Usage: editorconfiger completion <generator>

Arguments:
  <generator>  [possible values: bash, elvish, fish, powershell, zsh]

Options:
  -h, --help  Print help
```

Here's the code:

https://github.com/aegoroff/editorconfiger/blob/8c0ea442f879c01bcbb1dda81ea25ad8d44f99e1/src/bin/editorconfiger.rs#L57-L63

> If it's for the program itself I'd think that the project would just generate
the completion files at build time and be packaged with them rather than
generating them at runtime.

Agreed. But the AUR packager doesn't. Which is why I decided to simply read-only ${HOME} in the first place :-)

kmk3: @glitsj16 [on Mar 15](https://github.com/netblue30/firejail/pull/6235#discussion_r1525654837):

> > Does the program write to those specific directories? I couldn't find them
> > mentioned in the repository.
> 
> No, editorconfiger just writes to `stdout`. Users would have to redirect it
> to the appropriate path(s) themselves.
> 
> [...]
> 
> > If it's for the program itself I'd think that the project would just generate
> > the completion files at build time and be packaged with them rather than
> > generating them at runtime.
> 
> Agreed. But the AUR packager doesn't. Which is why I decided to simply
> read-only ${HOME} in the first place :-)

Thanks for the details, it all makes sense now.

Using just `read-only ${HOME}` seems right; I removed the completion
directories.

kmk3 (approved):

PR #6237, opened 2024-03-01
New profile: green-recorder

glitsj16:
green-recorder

Simple screen recorder for Linux desktop, supports Wayland & Xorg
https://github.com/dvershinin/green-recorder
https://aur.archlinux.org/packages/green-recorder
https://aur.archlinux.org/packages/green-recorder-git

### DIFF
--- a/etc/profile-a-l/green-recoder.profile
+++ b/etc/profile-a-l/green-recoder.profile
@@ -1,0 +1,72 @@
+# Firejail profile for green-recorder
+# Description: A simple screen recorder for Linux desktop (supports Wayland & Xorg)
+# This file is overwritten after every install/update
+# Persistent local customizations
+include green-recorder.local
+# Persistent global definitions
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/green-recoder.profile
```diff
@@ -0,0 +1,73 @@
...
+seccomp.block-secondary
+tracelog
+
+disable-mnt
+private-bin awk,bash,convert,ffmpeg,green-recorder,grep,mv,pactl,ps,python*,sh,sleep,xdg-open,xdpyinfo,xwininfo
+private-cache
+private-dev
+private-etc @x11
+private-tmp
+writable-run-user
```
kmk3, line 65: Why is this needed?

It's currently mostly only used in profiles for email clients and things that
may use gpg.

kmk3 (approved):

PR #6244, opened 2024-03-02
New profile: localsend_app

glitsj16:
localsend

Description: An open source cross-platform alternative to AirDrop.

https://github.com/localsend/localsend

### DIFF
--- a/etc/profile-a-l/localsend_app.profile
+++ b/etc/profile-a-l/localsend_app.profile
@@ -1,0 +1,65 @@
+# Firejail profile for localsend_app
+# Description: An open source cross-platform alternative to AirDrop
+# This file is overwritten after every install/update
+# Persistent local customizations
+include localsend_app.local
+# Persistent global definitions
+include globals.local
+
+blacklist /usr/libexec
+
+noblacklist ${HOME}/.local/share/localsend_app
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+#include disable-xdg.inc
+
+#mkdir ${HOME}/.local/share/localsend_app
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/localsend_app.profile
```diff
@@ -0,0 +1,68 @@
...
+# Description: An open source cross-platform alternative to AirDrop
+# This file is overwritten after every install/update
+# Persistent local customizations
+include localsend_app.local
+# Persistent global definitions
+include globals.local
+
+blacklist /usr/libexec
+
+noblacklist ${HOME}/.config/autostart
```
kmk3, line 11:
```suggestion
```

This would allow trivially escaping the sandbox.

I think it would only make sense to allow this path on the profile of a program
that is dedicated to managing startup programs.

glitsj16: Nice catch. Totally makes sense, thanks for the review!

etc/profile-a-l/localsend_app.profile
```diff
@@ -0,0 +1,64 @@
...
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.local/share/localsend_app
+whitelist ${HOME}/.local/share/localsend_app
+whitelist /usr/share/localsend_app
+include whitelist-common.inc
```
kmk3, line 25:
```suggestion
#include disable-xdg.inc

#mkdir ${HOME}/.local/share/localsend_app
#whitelist ${HOME}/.local/share/localsend_app
#whitelist ${DOWNLOADS}
#whitelist /usr/share/localsend_app
#include whitelist-common.inc
```

The PR enables the profile by default on firecfg and the program appears to be
intended to be used on arbitrary user files.

kmk3 (approved):

PR #6246, opened 2024-03-02
New profile: metadata-cleaner

glitsj16:
metadata-cleaner

Description: Python GTK3 application to view and clean metadata in
files, using mat2.

https://gitlab.com/rmnvgr/metadata-cleaner

### DIFF
--- a/etc/profile-m-z/metadata-cleaner.profile
+++ b/etc/profile-m-z/metadata-cleaner.profile
@@ -1,0 +1,65 @@
+# Firejail profile for metadata-cleaner
+# Description: Python GTK application to view and clean metadata in files, using mat2
+# This file is overwritten after every install/update
+# Persistent local customizations
+include metadata-cleaner.local
+# Persistent global definitions
+include globals.local
+
+blacklist /usr/libexec
+
+# Allow perl (blacklisted by disable-interpreters.inc)
+include allow-perl.inc
+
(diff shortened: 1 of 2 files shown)


etc/profile-m-z/metadata-cleaner.profile
```diff
@@ -0,0 +1,71 @@
...
+include disable-shell.inc
+
+whitelist ${DESKTOP}
+whitelist ${DOCUMENTS}
+whitelist ${DOWNLOADS}
+whitelist ${PICTURES}
+whitelist ${VIDEOS}
+whitelist /usr/share/metadata-remover
+whitelist /usr/share/perl-image-exiftool
+include whitelist-common.inc
```
kmk3, line 31:
```suggestion
whitelist /usr/share/metadata-remover
whitelist /usr/share/perl-image-exiftool
```

The PR enables the profile by default on firecfg and the program appears to be
intended to be used on arbitrary user files.

glitsj16: FYI: I'll keep things like this in mind for future profile work. Personally keeping a rather strict policy to move files I want to clean into ~/Downloads and return them to their original locations afterwards (via simple shell scripts). I lost sight of the impact of such a workflow on several profiles I recently PR'ed. Very much appreciate your detailed reviewing.

kmk3 (approved):

PR #6247, opened 2024-03-02
New profile: mimetype

glitsj16:
mimetype

Description: Determines the file type.

https://metacpan.org/release/File-MimeInfo
https://archlinux.org/packages/extra/any/perl-file-mimeinfo/

### DIFF
--- a/etc/profile-m-z/mimetype.profile
+++ b/etc/profile-m-z/mimetype.profile
@@ -1,0 +1,48 @@
+# Firejail profile for mimetype
+# Description: Determines the file type
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include mimetype.local
+# Persistent global definitions
+include globals.local
+
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}/wayland-*
+
+include disable-exec.inc
+include disable-proc.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
+net none
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noprinters
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+seccomp.block-secondary
+tracelog
+x11 none
+
+private-dev
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+read-only ${HOME}
+read-only ${RUNUSER}
+read-only /tmp
+
+restrict-namespaces


kmk3 (commented):
This appears to be a common CLI utility that is intended to be used on
arbitrary files and I think it's likely that other programs may depend on it.

In which case, it might make sense to avoid blacklisting paths and using
`private-` commands to avoid potentially breaking common use cases and/or other
tools, especially since the PR enables the profile by default.

Also, note that it reads `~/.local/share/mime`, so using `private` would break
that.

Does it work with the following?

```
read-only /
```

If not, how about the following?

```
read-only ${HOME}
read-only ${RUNUSER}
read-only /tmp
```

glitsj16:
Totally forgot about private, I removed it.

read-only /
Warning: cannot create a new user namespace, going forward without it...

So I went for your suggested
read-only ${HOME}
read-only ${RUNUSER}
read-only /tmp

Those work fine.

etc/profile-m-z/mimetype.profile
```diff
@@ -0,0 +1,67 @@
...
+# Description: Determines the file type
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include mimetype.local
+# Persistent global definitions
+include globals.local
+
+blacklist /tmp/.X11-unix
+blacklist ${RUNUSER}
```
kmk3, line 11:
```suggestion
#blacklist ${RUNUSER}
```

[As mentioned][1], the program appears to be intended to be used on arbitrary
files on the system and other programs may depend on it.

Also, since it uses `quiet`, this may lead to silent breakage of unrelated
programs that call it.

[1]: https://github.com/netblue30/firejail/pull/6247#pullrequestreview-1912649150

etc/profile-m-z/mimetype.profile
```diff
@@ -0,0 +1,67 @@
...
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+
+include whitelist-usr-share-common.inc
```
kmk3, line 24:
```suggestion
#include disable-common.inc
#include disable-devel.inc
include disable-exec.inc
#include disable-interpreters.inc
include disable-proc.inc
#include disable-programs.inc
#include disable-shell.inc

#include whitelist-usr-share-common.inc
```

Same here.

etc/profile-m-z/mimetype.profile
```diff
@@ -0,0 +1,67 @@
...
+# To support mimetype in private-bin on Arch Linux (and derivatives), symlink
+# /usr/bin/vendor_perl/mimetype to /usr/bin/mimetype and add the below to your
+# mimetype.local.
+# Non-Arch Linux users can safely add the below to their mimetype.local for
+# extra hardening.
+#private-bin mimetype,perl
+private-cache
+private-dev
+private-etc mime.types
+private-lib perl5
```
kmk3, line 57:
```suggestion
#private-cache
private-dev
#private-etc mime.types
#private-lib perl5
```

Same here.

src/firecfg/firecfg.config
```diff
@@ -554,6 +554,7 @@ microsoft-edge-beta
 microsoft-edge-dev
 microsoft-edge-stable
 midori
+mimetype
```
kmk3, line 557:
```suggestion
```

Same here.

It seems to contain both a general-purpose user-facing tool and a library, both
of which are intended to be used on arbitrary files, so enabling it by default
seems risky.

Also, it seems like it could potentially affect a lot of unrelated programs:

```
$ LC_ALL=C pacman -Sii extra/perl-file-mimeinfo |
  grep -e '^Required By' -e '^Optional For'
Required By     : auto-multiple-choice
Optional For    : xdg-utils
$ LC_ALL=C pacman -Sii extra/xdg-utils
Repository      : extra
Name            : xdg-utils
Version         : 1.2.1-1
Description     : Command line tools that assist applications with a variety of desktop integration tasks
[...]
Required By     : ario  blender  catfish  chromium  cuyo  deepin-daemon  deepin-printer  deluge-gtk  fff  filezilla  flatpak
                  gambas3-runtime  glob2  gramps  hplip  httrack  jre8-openjdk  keepass  libreoffice-fresh  libreoffice-still  linssid
                  minetest  mumble  nextcloud-client  openconnect  openra  qmmp  qt5-base  qt6-base  retext  rmlint-shredder
                  rofi-pass  shutter  subtitleeditor  surge-xt-clap  surge-xt-standalone  surge-xt-vst3  sxiv  synapse  terminator
                  wire-desktop  wireshark-qt  wiznote  xboard  xpdf
Optional For    : arch-wiki-lite  cliphist  cups  discord  electron22  electron23  electron27  electron28  electron29  foot  gscan2pdf
                  intel-oneapi-basekit  nnn  weechat-matrix  wl-clipboard  xarchiver  xarchiver-gtk2
```

etc/profile-m-z/mimetype.profile
glitsj16: I've implemented your suggestions, but added `blacklist ${RUNUSER}/wayland-*`.

etc/profile-m-z/mimetype.profile
glitsj16: Done!

src/firecfg/firecfg.config
glitsj16: Removed it from firecfg.

kmk3 (approved):

PR #6248, opened 2024-03-02
New profile: pacseek

glitsj16:
pacseek

Description: A TUI for searching and installing Arch Linux packages.

https://github.com/moson-mo/pacseek

### DIFF
--- a/etc/profile-m-z/pacseek.profile
+++ b/etc/profile-m-z/pacseek.profile
@@ -1,0 +1,88 @@
+# Firejail profile for pacseek
+# Description: TUI for searching and installing Arch Linux packages
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include pacseek.local
+# Persistent global definitions
+#include globals.local
+
+blacklist /tmp/.X11-unix
+blacklist /usr/libexec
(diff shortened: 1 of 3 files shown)


etc/profile-m-z/pacseek.profile
```diff
@@ -0,0 +1,88 @@
...
+private-dev
+private-lib
+private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+read-only ${HOME}
+read-write ${HOME}/.config/pacseek
```
kmk3, line 87:
```suggestion
```

From <https://github.com/moson-mo/pacseek>:

> Package installation / removal is done with an AUR helper.
> 
> In the default configuration, yay is being used.

yay downloads packages to ~/.cache/yay.

Does this profile work with yay?

Did you test the install function?

glitsj16: yay is not the problem. But installing isn't handled correctly, not with `noprofile.profile` either. Originally I created it for browsing/searching packages. That part works fine but that's not why users would install it. So I'm dropping the PR. Apologies for the noise and your time to review.

PR #6249, opened 2024-03-02
New profile: qt5ct

glitsj16:

PR #6250, opened 2024-03-02
New profile: qt6ct

glitsj16:


etc/profile-m-z/qt6ct.profile
```diff
@@ -0,0 +1,69 @@
...
+# Persistent local customizations
+include qt6ct.local
+# Persistent global definitions
+ignore dns
+include globals.local
+
+blacklist /usr/libexec
+
+noblacklist ${HOME}/.config/qt6ct
+noblacklist ${HOME}/.local/share/qt6ct
```
kmk3, line 13: Not blacklisted.

kmk3: Sorry, I meant add the paths to disable-programs.inc rather than remove.

kmk3 (commented):
@glitsj16

Note: This profile is nearly identical to the one in #6249 (other than the
program name and paths using it), so most review comments are likely to be
identical or very similar.

In such cases in future PRs please use the same PR for all similar profiles to
make reviews easier.

glitsj16:
> Sorry, I meant add the paths to disable-programs.inc rather than remove.

Oh, I misunderstood that. Additionally I got confused by there `not` being a counter-part blacklist ${HOME}/.config/qt{5,6}ct anywhere to the below:

https://github.com/netblue30/firejail/blob/df257a814245c8457f1334a01752117aca04d088/etc/inc/whitelist-1793-workaround.inc#L25-L26

So indeed, it would make sense to add those. I'll take care of it.

> In such cases in future PRs please use the same PR for all similar profiles to
make reviews easier.

Will keep it in mind!

etc/profile-m-z/qt6ct.profile
kmk3: These lines should be restored.

kmk3: > These lines should be restored.

Without removing the `blacklist` ones, as both are needed.

Anyway, I fixed and force-pushed.

etc/profile-m-z/qt6ct.profile
```diff
@@ -0,0 +1,69 @@
+# Firejail profile for qt6ct
+# Description: Qt6 Configuration Utility
+# This file is overwritten after every install/update
+# Persistent local customizations
+include qt6ct.local
+# Persistent global definitions
+ignore dns
```
kmk3, line 7: I just noticed this, why is it here?

There is no `dns` command in any profile.

Does this tool need internet access?

kmk3 (approved):

PR #6255, opened 2024-03-02
New profile: tqemu

glitsj16:
tqemu

Description: QEMU frontend without libvirt.

https://github.com/thanoulis/tqemu

### DIFF
--- a/etc/profile-m-z/tqemu.profile
+++ b/etc/profile-m-z/tqemu.profile
@@ -1,0 +1,26 @@
+# Firejail profile for tqemu
+# Description: QEMU frontend without libvirt
+# This file is overwritten after every install/update
+# Persistent local customizations
+include tqemu.local
+# Persistent global definitions
+include globals.local
+
+include disable-common.inc
+include disable-programs.inc
+
+# For host-only network sys_admin is needed.
+# See https://github.com/netblue30/firejail/issues/2868#issuecomment-518647630
+caps.keep net_raw,sys_nice
+#caps.keep net_raw,sys_admin
+netfilter
+nodvd
+notv
+tracelog
+
+private-cache
+private-tmp
+
+noexec /tmp
+# breaks app
+#restrict-namespaces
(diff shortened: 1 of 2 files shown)


kmk3 (commented):
Why is the profile for a qemu wrapper in some cases more restrictive than the
profile for qemu itself (`qemu-system-x86_64.profile`)?

Suggestions:

* Move the contents of `qemu-system-x86_64.profile` into `qemu-common.profile`
  and make the former (and `qemu-launcher.profile`) redirect to the latter
* Make `tqemu.profile` redirect to `qemu-common.profile`
* Harden `qemu-common.profile` (preferably in another PR as it may cause
  breakage for qemu itself)

glitsj16:
> Why is the profile for a qemu wrapper in some cases more restrictive than the
profile for qemu itself (qemu-system-x86_64.profile)?

AFAIK tqemu is the only such wrapper that doesn't rely on libvirt. It's more limited in its functionality, but that also makes it very straightforward to sandbox.

I'll need more time to refactor/double-check the qemu wrappers as you suggested. Thanks!

kmk3:
> > Why is the profile for a qemu wrapper in some cases more restrictive than
> > the profile for qemu itself (qemu-system-x86_64.profile)?
> 
> AFAIK tqemu is the only such wrapper that doesn't rely on libvirt. It's more
> limited in its functionality, but that also makes it very straightforward to
> sandbox.

That's good to know, but qemu itself does not use libvirt (it's essentially the
other way around).

My point was that in general it doesn't make much sense for a wrapper (tqemu)
to have a more restrictive profile than the thing that it is wrapping (qemu).

> I'll need more time to refactor/double-check the qemu wrappers as you
> suggested. Thanks!

On second though, for simplicity, I'd just copy and paste most of
`qemu-system-x86_64.profile` into `tqemu.profile` for now (so that they are
almost equal) and leave the refactoring/hardening for another PR.

glitsj16:
> On second though, for simplicity, I'd just copy and paste most of
qemu-system-x86_64.profile into tqemu.profile for now...

Done!

kmk3 (approved):

PR #6258, opened 2024-03-02
ssh: whitelist gcr-ssh-agent unix socket

michelesr:
Since gnome-keyring 1.46 the ssh-agent functionality has been removed and gcr-ssh-agent is the recommended alternative.

Source:
  - https://gitlab.gnome.org/GNOME/gcr/-/merge_requests/67
  - https://wiki.archlinux.org/title/GNOME/Keyring#SSH_keys

NOTE: @glitsj16 suggested to put the whitelist statement inside `whitelist-runuser-common.inc`, but since `whitelist-runuser-common.inc` is in this file, I think it makes more sense to leave it right after it.

### DIFF
--- a/etc/profile-m-z/ssh.profile
+++ b/etc/profile-m-z/ssh.profile
@@ -18,9 +18,9 @@ include disable-common.inc
 include disable-exec.inc
 include disable-programs.inc
 
+whitelist ${RUNUSER}/gcr/ssh
 whitelist ${RUNUSER}/gnupg/S.gpg-agent.ssh
 whitelist ${RUNUSER}/keyring/ssh
-whitelist ${RUNUSER}/gcr/ssh
 include whitelist-usr-share-common.inc
 include whitelist-runuser-common.inc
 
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/ssh.profile
```diff
@@ -20,6 +20,7 @@ include disable-programs.inc
 
 whitelist ${RUNUSER}/gnupg/S.gpg-agent.ssh
 whitelist ${RUNUSER}/keyring/ssh
+whitelist ${RUNUSER}/gcr/ssh
```
kmk3, line 23:
```suggestion
whitelist ${RUNUSER}/gcr/ssh
whitelist ${RUNUSER}/gnupg/S.gpg-agent.ssh
whitelist ${RUNUSER}/keyring/ssh
```

Sort.

michelesr: Thanks for this, I ran the sort.py script from `contrib/` but I didn't do anything for some reason.

kmk3: > Thanks for this, I ran the sort.py script from `contrib/` but I didn't do
> anything for some reason.

No problem; that script only sorts comma-separated items within a given line
for specific commands.  The lines themselves are currently sorted manually.

kmk3 (approved):

PR #6268, opened 2024-03-08
docs: firecfg: note different .desktop naming schemes

kmk3:
Based on the discussion at #5063.

Misc: The `\&` is used to escape the dot in `.desktop` (see roff(7)).

This amends commit a9c851ee4 ("firecfg: use ignorelist also for .desktop
files", 2024-01-08) / PR #6153.


glitsj16 (approved):
Fine clarification!

PR #6286, opened 2024-03-23
profiles: replace x11 socket blacklist with disable-X11.inc

kmk3:
Replace all occurrences of `blacklist /tmp/.X11-unix` with
`include disable-X11.inc`, which blacklists more X11-related files.

    $ git grep -Ilz '^blacklist /tmp/.X11-unix' -- \
      etc/profile*/*.profile | xargs -0 perl -0 -pi -e '\
        s/\nblacklist \/tmp\/.X11-unix\n/\n/; \
        s/(\ninclude disable-xdg.inc\n)/\ninclude disable-X11.inc$1/; \
        s/(\ninclude disable-[^Xx\n]+\n)(\n|# )/$1include disable-X11.inc\n$2/'

Note: The following files were also edited manually:

* etc/profile-a-l/erd.profile
* etc/profile-a-l/links-common.profile
* etc/profile-m-z/termshark.profile
* etc/profile-m-z/tmux.profile
* etc/profile-m-z/tshark.profile

Relates to #4462 #4854 #5544.

### DIFF
--- a/etc/profile-a-l/agetpkg.profile
+++ b/etc/profile-a-l/agetpkg.profile
@@ -7,7 +7,6 @@ include agetpkg.local
 # Persistent global definitions
 include globals.local
 
-blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}/wayland-*
 
 # Allow python (blacklisted by disable-interpreters.inc)
@@ -20,6 +19,7 @@ include disable-exec.inc
 include disable-interpreters.inc
 include disable-programs.inc
 include disable-shell.inc
+include disable-X11.inc
 include disable-xdg.inc
 
 whitelist ${DOWNLOADS}
--- a/etc/profile-a-l/alpine.profile
+++ b/etc/profile-a-l/alpine.profile
@@ -30,7 +30,6 @@ noblacklist ${HOME}/.pinercex
 noblacklist ${HOME}/.signature
 noblacklist ${HOME}/mail
 
-blacklist /tmp/.X11-unix
 blacklist ${RUNUSER}/wayland-*
 
 include disable-common.inc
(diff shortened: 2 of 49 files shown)


rusty-snake (approved):

kmk3:
Note: I did not manage to call `include disable-X11.inc` from the code itself
without errors, so I left just the main profile search/replace for this PR.

glitsj16 (approved):

etc/profile-a-l/gconf-editor.profile
```diff
@@ -7,13 +7,13 @@ include gconf-editor.local
 # added by included profile
 #include globals.local
 
-blacklist /tmp/.X11-unix
```
glitsj16, line 10: This never should have been in the profile before this PR, my mistake originally. It's a GUI app and it obviously breaks when including disable-X11.inc.

Not sure how best to proceed to fix it.

glitsj16 (commented):
See gconf-editor.profile comment.

etc/profile-a-l/gconf-editor.profile
kmk3: > This never should have been in the profile before this PR, my mistake
> originally. It's a GUI app and it obviously breaks when including
> disable-X11.inc.
> 
> Not sure how best to proceed to fix it.

I'll remove it after this PR.

glitsj16: Thanks!

kmk3: > I'll remove it after this PR.

Fixed it on master and rebased the PR to make the result a bit less confusing.

Anyway, good catch.

PR #6293, opened 2024-03-25
New profile: gh (GitHub CLI)

glitsj16:
Description: GitHub's official command-line tool.

https://github.com/cli/cli

### DIFF
--- a/etc/profile-a-l/gh.profile
+++ b/etc/profile-a-l/gh.profile
@@ -1,0 +1,14 @@
+# Firejail profile for gh
+# Description: GitHub's official command-line tool
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include gh.local
+# Persistent global definitions
+# added by included profile
+#include globals.local
+
+noblacklist ${HOME}/.config/gh
+
+# Redirect
+include git.profile
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -462,6 +462,7 @@ blacklist ${HOME}/.config/geany
 blacklist ${HOME}/.config/geary
 blacklist ${HOME}/.config/gedit
 blacklist ${HOME}/.config/geeqie
+blacklist ${HOME}/.config/gh
 blacklist ${HOME}/.config/ghb
 blacklist ${HOME}/.config/ghostwriter
 blacklist ${HOME}/.config/git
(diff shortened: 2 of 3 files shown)


etc/profile-a-l/gh.profile
```diff
@@ -0,0 +1,97 @@
...
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+seccomp.block-secondary
+tracelog
+x11 none
```
kmk3, line 81:
```suggestion
```

There is no `net none`, so this would likely break the program for most X11
users.

etc/profile-a-l/gh.profile
```diff
@@ -0,0 +1,97 @@
...
+whitelist ${HOME}/.emacs
+whitelist ${HOME}/.emacs.d
+whitelist ${HOME}/.git-credential-cache
+whitelist ${HOME}/.git-credentials
+whitelist ${HOME}/.gitconfig
+whitelist ${HOME}/.gnupg
+whitelist ${HOME}/.nanorc
+whitelist ${HOME}/.vim
+whitelist ${HOME}/.viminfo
+include whitelist-common.inc
```
kmk3, line 56:
```suggestion
```

As mentioned in other recent PRs, this tool is intended for development and so
whitelisting in `${HOME}` cannot be done by default without breaking most
setups.

etc/profile-a-l/gh.profile
```diff
@@ -0,0 +1,97 @@
...
+private-etc @network,@tls-ca,rpc,services
+private-tmp
+
+dbus-user none
+dbus-system none
+
+memory-deny-write-execute
+read-only ${HOME}
+read-write ${HOME}/.config/gh
+read-write ${HOME}/.config/git
```
kmk3, line 96:
```suggestion
```

Same here.

etc/profile-a-l/gh.profile
kmk3, line 1: This profile is similar to git.profile, it uses git and it's also intended for
development, so why not make it overall a redirect to git.profile?

glitsj16: That makes sense. I'll do some testing with a refactored git redirect profile and edit the PR accordingly. That should take care of your other comments too. Thanks for the review!

glitsj16: Done!

kmk3 (approved):

PR #6302, opened 2024-04-05
docs: warn about limitations of landlock

kmk3:
And mark it as experimental.

Relates to #6078.

### DIFF
--- a/src/man/firejail-profile.5.in
+++ b/src/man/firejail-profile.5.in
@@ -509,30 +509,30 @@ Blacklist all Linux capabilities.
 Whitelist given Linux capabilities.
 #ifdef HAVE_LANDLOCK
 .TP
-\fBlandlock.enforce
+\fBlandlock.enforce (experimental)
 Enforce the Landlock ruleset.
 .PP
 Without it, the other Landlock commands have no effect.
 .TP
-\fBlandlock.fs.read path
+\fBlandlock.fs.read path (experimental)
 Create a Landlock ruleset (if it doesn't already exist) and add a read access
 rule for path.
 .TP
-\fBlandlock.fs.write path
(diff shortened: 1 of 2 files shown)


glitsj16 (approved):

glitsj16:
@kmk3 Although the Firejail/Landlock status is made much clearer in this PR it might still be nice to note that without having `landlock` in the `lsm=x,y,z` kernel parameter (cfr. apparmor) Firejail will ignore landlock commands.

Example bootloader lsm param for good Firejail support: `lsm=landlock,lockdown,yama,integrity,apparmor,bpf`

kmk3:
> Although the Firejail/Landlock status is made much clearer in this PR it
> might still be nice to note that without having `landlock` in the `lsm=x,y,z`
> kernel parameter (cfr. apparmor) Firejail will ignore landlock commands.
> 
> Example bootloader lsm param for good Firejail support:
> `lsm=landlock,lockdown,yama,integrity,apparmor,bpf`

Landlock should work by default without needing to enable it.

Is there any distribution that disables it?

Example to check if it works:

```console
$ firejail --quiet --noprofile true
$ firejail --quiet --noprofile --landlock.enforce --landlock.fs.read=/foo true
```

glitsj16:
> Is there any distribution that disables it?

Not that I know of. I wasn't clear enough though, let me try to explain what I mean. If a user enables `AppArmor` as default security model on every boot via the `lsm` kernel parameter and `landlock` isn't part of that param, it does get disabled.

kmk3:
> If a user enables `AppArmor` as default security model on every boot via the
> `lsm` kernel parameter and `landlock` isn't part of that param, it does get
> disabled.

I see it now; added an item about it.

netblue30:
all set, thanks!

PR #6305, opened 2024-04-10
landlock: amend empty functions and comments

kmk3:
Changes:

* Always declare public landlock functions, regardless of
  `HAVE_LANDLOCK`
* Make the other public landlock functions (besides `ll_add_profile`)
  also be empty when `HAVE_LANDLOCK` is not defined
* Clarify related comments

This amends commit 8259f66e1 ("landlock fix for old kernel versions",
2024-04-06).

For clarity, landlock-common.inc is included by default.profile and the
issue that the aforementioned commit fixes is that if profile.c is built
without the part that parses landlock commands (that is, when
`HAVE_LANDLOCK` is not defined), using default.profile would cause
firejail to abort due to "invalid lines".

Note that the issue would only occur when firejail is built with an
older kernel (or with --disable-landlock), not when simply running on an
older kernel.

See also commit b02a7a337 ("landlock: remove empty functions",
2023-12-07).

Relates to #6078.

### DIFF
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -966,10 +966,8 @@ void run_ids(int argc, char **argv);
 void oom_set(const char *oom_string);
 
 // landlock.c
-#ifdef HAVE_LANDLOCK
 int ll_get_fd(void);
 int ll_restrict(uint32_t flags);
 void ll_add_profile(int type, const char *data);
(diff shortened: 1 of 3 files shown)


glitsj16 (approved):

netblue30 (approved):

netblue30:
Looks fine, merging in!

Issue #4535, opened 2021-09-12
RUNUSER should default to $XDG_RUNTIME_DIR


crocket:
Right now, RUNUSER is `/run/user/user-id`. But, pam_rundir.so sets XDG_RUNTIME_DIR to `/run/users/user-id`.

Thus, RUNUSER fails to capture XDG_RUNTIME_DIR.

RUNUSER should only be set to `/run/user/user-id` if there is no XDG_RUNTIME_DIR.

kmk3:
@crocket commented [on Sep 12](https://github.com/netblue30/firejail/issues/4535#issue-994044649):

> Right now, RUNUSER is `/run/user/user-id`. But, pam_rundir.so sets
> XDG_RUNTIME_DIR to `/run/users/user-id`.
> 
> Thus, RUNUSER fails to capture XDG_RUNTIME_DIR.
> 
> RUNUSER should only be set to `/run/user/user-id` if there is no
> XDG_RUNTIME_DIR.

Using `XDG_RUNTIME_DIR` sounds good to me.

But on what distro does that happen?

I only found a reference to `XDG_RUNTIME_DIR` on pam-related man pages on
Arch's `pam_systemd(8)`, which claims to use the default path:

* <https://man.archlinux.org/man/pam_systemd.8>

> On login, this module — in conjunction with systemd-logind.service — ensures
> the following:
> 
> 1.If it does not exist yet, the user runtime directory /run/user/$UID is
> either created or mounted as new "tmpfs" file system with quota applied, and
> its ownership changed to the user that is logging in.

crocket:
I set up `pam_rundir` manually because I use openrc instead of systemd.

rusty-snake:
> I think people should stick to standards instead of hardcoding

In general yes, but with firejail it is a bit special since it must not trust a environment variable set by the users. There must be some sanitization.

crocket:
I can't make pam_rundir change its XDG_RUNTIME_DIR. Neither can I make pam_systemd change its XDG_RUNTIME_DIR.

I think it's best to let PAM modules decide XDG_RUNTIME_DIR and make firejail set RUNUSER to XDG_RUNTIME_DIR when XDG_RUNTIME_DIR is available.

crocket:
I maintain gentoo linux package for pam_rundir, and I discovered that I could compile pam_rundir with `/run/user` as its XDG_RUNTIME_DIR.

[...]

PR #6307, opened 2024-04-12
Construct /run/firejail while holding flock

spiiroin:
There are reports of firejail sandboxed applications occasionally taking long time (12 seconds) to start up. When this happens, it affects all sandboxed applications until the device is rebooted.

The reason for the slowdown seems to be a timing hazard in the way remounts under /run/firejail are handled. This gets triggered when multiple firejail processes are launched in parallel as part of user session bring up and results in some, dozens, hundreds, or even thousands of stray /run/firejail/xxx mounts. The amount of mount points then affects every mount operation that is done during sandbox filesystem construction.

To stop this from happening, arrange it so that only one firejail process at time is inspecting and/or modifying mountpoints under /run/firejail by doing:
  1) Create /run/firejail directory using atomic operations
  2) Create and obtain lock for /run/firejail/firejail-run.lock
  3) Setup files, directories and mounts under /run/firejail
  4) Release /run/firejail/firejail-run.lock


spiiroin:
From Saifish OS patch here: https://github.com/sailfishos/firejail/pull/18

glitsj16 (approved):
LGTM. Thanks for the patch.

topimiettinen:
Systemd is using temporary directories with random namse (not predictable and unlike to collide), like `systemd-private-f10ec674d0684f66b78990ff81e3153f-auditd.service-jKO6ug/`. Perhaps this approach (under `/run/firejail/`) would be better approach than locking?

src/firejail/preproc.c
```diff
@@ -21,12 +21,61 @@
...
+	if (lockfd != -1) {
+		*lockfd_ptr = -1;
+		if (arg_debug)
+			fprintf(stderr, "pid=%d: unlock %s\n", (int)getpid(), path);
+		if (flock(lockfd, LOCK_UN) == -1)
+			fprintf(stderr, "Error: cannot unlock %s: %m\n", path);
+		if (close(lockfd) == -1)
+			fprintf(stderr, "Error: cannot close %s: %m\n", path);
+	}
+}
```
kmk3, line 64:
```suggestion
static void preproc_lock_file(const char *path, int *lockfd_ptr) {
	assert(path != NULL);
	assert(lockfd_ptr != NULL);

	// Already locked
	if (*lockfd_ptr != -1)
		return;

	long pid = (long)getpid();
	if (arg_debug)
		fprintf(stderr, "pid=%ld: locking %s ...\n", pid, path);

	int lockfd = open(path, O_WRONLY | O_CREAT | O_CLOEXEC, S_IRUSR | S_IWUSR);
	if (lockfd == -1) {
		fprintf(stderr, "Error: cannot create a lockfile at %s\n", path);
		errExit("open");
	}

	if (fchown(lockfd, 0, 0) == -1) {
		fprintf(stderr, "Error: cannot chown root:root %s\n", path);
		errExit("fchown");
	}

	if (flock(lockfd, LOCK_EX) == -1) {
		fprintf(stderr, "Error: cannot lock %s\n", path);
		errExit("flock");
	}

	*lockfd_ptr = lockfd;
	if (arg_debug)
		fprintf(stderr, "pid=%ld: locked %s\n", pid, path);
}

static void preproc_unlock_file(const char *path, int *lockfd_ptr) {
	assert(path != NULL);
	assert(lockfd_ptr != NULL);

	int lockfd = *lockfd_ptr;
	// Already unlocked
	if (lockfd == -1)
		return;

	long pid = (long)getpid();
	if (arg_debug)
		fprintf(stderr, "pid=%ld: unlocking %s ...\n", pid, path);

	if (flock(lockfd, LOCK_UN) == -1) {
		fprintf(stderr, "Error: cannot unlock %s\n", path);
		errExit("flock");
	}

	if (close(lockfd) == -1) {
		fprintf(stderr, "Error: cannot close %s\n", path);
		errExit("close");
	}

	*lockfd_ptr = -1;
	if (arg_debug)
		fprintf(stderr, "pid=%ld: unlocked %s\n", pid, path);
}
```

Main changes:

* Fix formatting
* Reduce indentation/return early
* Add asserts
* Avoid non-standard `%m`
* Make messages more consistent
* Abort if fchown fails
* Abort on most/all errors

Since all flock operations are synchronous/blocking in this code, is there a
reason not to abort if they fail?

kmk3 (changes requested):

src/firejail/chroot.c
```diff
@@ -273,7 +273,11 @@ void fs_chroot(const char *rootdir) {
 		errExit("mounting /proc");
 
 	// create all other /run/firejail files and directories
-	preproc_build_firejail_dir();
+	int lockfd_directory = -1;
+	preproc_build_firejail_dir_unlocked();
+	preproc_lock_firejail_dir(&lockfd_directory);
+	preproc_build_firejail_dir_locked();
+	preproc_unlock_firejail_dir(&lockfd_directory);
```
kmk3, line 280:
```suggestion
	preproc_lock_firejail_dir(&lockfd_directory);
	preproc_build_firejail_dir();
	preproc_unlock_firejail_dir(&lockfd_directory);
```

Instead of special-casing the creation of the base directory, why not put the
lock directly under /run?

```diff
diff --git a/src/include/rundefs.h b/src/include/rundefs.h
index 00eaa47c6..eede9d8bb 100644
--- a/src/include/rundefs.h
+++ b/src/include/rundefs.h
@@ -32,8 +32,8 @@
 #define RUN_FIREJAIL_BANDWIDTH_DIR	RUN_FIREJAIL_DIR "/bandwidth"
 #define RUN_FIREJAIL_PROFILE_DIR	RUN_FIREJAIL_DIR "/profile"
 #define RUN_FIREJAIL_DBUS_DIR RUN_FIREJAIL_DIR "/dbus"
+#define RUN_DIRECTORY_LOCK_FILE		RUN_FIREJAIL_BASEDIR "/firejail-run.lock"
 #define RUN_NETWORK_LOCK_FILE		RUN_FIREJAIL_DIR "/firejail-network.lock"
-#define RUN_DIRECTORY_LOCK_FILE		RUN_FIREJAIL_DIR "/firejail-run.lock"
 #define RUN_RO_DIR			RUN_FIREJAIL_DIR "/firejail.ro.dir"
 #define RUN_RO_FILE			RUN_FIREJAIL_DIR "/firejail.ro.file"
 #define RUN_MNT_DIR			RUN_FIREJAIL_DIR "/mnt"	// a tmpfs is mounted on this directory before any of the files below are created
```

Then `preproc_build_firejail_dir()` can be left as is.

spiiroin:
> Systemd is using temporary directories with random namse (not predictable and unlike to collide) ...

AFAIK the difference is that here the directories are meant to be shared and thus must have predictable known names i.e. the problem was not accidental collisions but a technical glitches on the way to the desired end result.

src/firejail/preproc.c
spiiroin: Well, for the most part this just moves original behavior to helper functions. I did add one errExit, but the original flock() code ignored errors altogether -> suggests that limping forwards is preferred over bailout.

I'll modify the pr to address these points

src/firejail/chroot.c
spiiroin: Littering /run with firejail stuff when /run/firejail is going to be created does not sound like a nice thing to do.

Also note that original preproc_build_firejail_dir() was prepared to create also "/run" directory (that "CentOS 6 doesn't have /run directory" bit). So, having a lock file in /run would not work universally either - well, unless one is willing to create that directory before calling preproc_build_firejail_dir(). Which could be done by adding copy paste code here and there. Or by having a helper function to establish place for lock files. At which point the lock file might as well reside where it already is - under /run/firejail. Which would be what this PR does...

src/firejail/preproc.c
spiiroin: ... or rather, picked suggested changes as-is. My first time bumping into this feature. Thanks,

kmk3: > Well, for the most part this just moves original behavior to helper
> functions.

> I did add one errExit

> ```diff
>  	int lockfd = open(path, O_WRONLY | O_CREAT | O_CLOEXEC, S_IRUSR | S_IWUSR);
> -		if( lockfd == -1 )
> -			errExit(path);
> +	if (lockfd == -1) {
> +		fprintf(stderr, "Error: cannot create a lockfile at %s\n", path);
> +		errExit("open");
> +	}
> ```

This was already changed, but to clarify, when calling errExit please treat the
first argument as if it were the format string in printf (that is, prefer using
a string literal), as this is how it is usually used and eventually it should
support variable arguments like printf.

> but the original flock() code ignored errors altogether -> suggests that
> limping forwards is preferred over bailout.

True, but from what I understand, valid errors in those calls in this code are
so unlikely (like ENOLCK in flock) that if any error happens it's likely that
something went very wrong anyway.

Though admittedly I'm not sure about signal interruption.

src/firejail/chroot.c
kmk3: > Littering /run with firejail stuff when /run/firejail is going to be created
> does not sound like a nice thing to do.

It is less nice than having just one directory to worry about, but it is a
single extra file and it would arguably make the code simpler / easier to
reason about (that is, getting a lock file before creating any path rather than
in between creating paths).

> Also note that original preproc_build_firejail_dir() was prepared to create
> also "/run" directory (that "CentOS 6 doesn't have /run directory" bit). So,
> having a lock file in /run would not work universally either - well, unless
> one is willing to create that directory before calling
> preproc_build_firejail_dir(). Which could be done by adding copy paste code
> here and there. Or by having a helper function to establish place for lock
> files. At which point the lock file might as well reside where it already is
> - under /run/firejail. Which would be what this PR does...

That's true, nevermind.

I was thinking about another change at the same time, that runstatedir would be
set at configure time and so /var/run would be used in distributions that do
not have /run (see also #4535), in which case it would hopefully not be
necessary for firejail to create a common system directory.

I plan to do that eventually, though it is outside the scope of this PR.

src/firejail/preproc.c
```diff
@@ -21,12 +21,86 @@
...
+void preproc_unlock_firejail_dir(int *lockfd_ptr) {
+	preproc_unlock_file(RUN_DIRECTORY_LOCK_FILE, lockfd_ptr);
+}
+
 // build /run/firejail directory
-void preproc_build_firejail_dir(void) {
+void preproc_build_firejail_dir_unlocked(void) {
+	// Use atomic / otherwise timing hazard free operations
+	// to create enough of /run/firejail directory hierarchy
+	// to have a place for lock files
```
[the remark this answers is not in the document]
spiiroin: Mmm, yeah, atomic is maybe a bit loaded / too strong word to use here. But effectively there is a sequence of mkdir() syscalls a firejail process executes / can execute. In case of multiple processes running in parallel, it does not matter which process / processes get to execute each one of these -> timing hazard free -> can be done without locking.

[One could argue that create_empty_dir_as_root() sort of does a bit too much and a bit too little: precheck with stat() to skip mkdir() is a "toctou" hazard and not checking type can lead to false positive, and mkdir() EEXISTS too could be due to something else than a directory. But in this context it does not really matter.] 

Whereas fs_remount_simple() & co perform sets of operations that have interdependence on each other and can easily be interfered by somebody else operating on the same paths / mount points -> locking is needed.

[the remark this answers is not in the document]
kmk3: > Mmm, yeah, atomic is maybe a bit loaded / too strong word to use here. But
> effectively there is a sequence of mkdir() syscalls a firejail process
> executes / can execute. In case of multiple processes running in parallel, it
> does not matter which process / processes get to execute each one of these ->
> timing hazard free -> can be done without locking.
> 
> [One could argue that create_empty_dir_as_root() sort of does a bit too much
> and a bit too little: precheck with stat() to skip mkdir() is a "toctou"
> hazard and not checking type can lead to false positive, and mkdir() EEXISTS
> too could be due to something else than a directory. But in this context it
> does not really matter.]

Agreed.

Misc: I started to refactor util.c some time ago to fix most toctou issues, but
it ended up depending on another WIP thing.

> Whereas fs_remount_simple() & co perform sets of operations that have
> interdependence on each other and can easily be interfered by somebody else
> operating on the same paths / mount points -> locking is needed.

Indeed; how about the following then?

```C
// build /run/firejail directory
//
// Note: This creates the base directory of the rundir lockfile;
// it should be called before preproc_lock_firejail_dir().
void preproc_build_firejail_dir_unlocked(void) {
```

Also, please squash/fixup the commits.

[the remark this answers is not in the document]
spiiroin: Adjusted explanatory comments at preproc_build_firejail_dir_unlocked() as suggested.

Review commits squashed.

kmk3 (commented):
I made a few more changes and pushed it to a `serialize_mounts_v2` branch.

Main changes:

* Fix misc formatting in the code and commit message
* Make lock variables global and use them directly only in preproc.c
* Use the lock/unlock functions also for the network lock (for consistency)
* Log if the path is already locked/unlocked
* Split the original commit into one for the flock improvements and another for
  the main fix (to make it easier to understand the scope of the latter)

If it seems good, you can hard-reset to that branch, add the SOB trailer to the
commits if you want and force-push.  Let me know if you have any questions.

(By the way, note that git trailers cannot have blank lines between them; a
blank line is what separates the trailers from the rest of the message).

spiiroin:
> If it seems good, you can hard-reset to that branch

Was fine by me, took the branch as-is (plus rebase as there were already new commits in the master branch).

kmk3 (approved):
LGTM.

Thanks for the improvements.

kmk3:
> > If it seems good, you can hard-reset to that branch
> 
> Was fine by me, took the branch as-is (plus rebase as there were already new
> commits in the master branch).

Misc: On future PRs, please do a separate force-push before or after rebasing
when possible so that a separate timeline event is created on GitHub.  This
makes the changes not related to the rebase easier to spot in the main
force-push diff.

PR #4656, opened 2021-10-31
WIP: Build firejail with meson

rusty-snake:
Resolves #4642

- [x] test
- [x] go through the TODOs
- [ ] add tests to meson
- [ ] testing
- [ ] gcov
- [ ] decide what to do with autotools
- [ ] did I said to test?

PR #6311, opened 2024-04-13
WIP: Meson build

topimiettinen:
Rebased and updated from #4656 by @rusty-snake.


kmk3 (commented):
It's interesting to see how another build system would do things, but I
personally don't really see this as an improvement.

Even if rough around the edges at times, both firejail and autoconf/make are
relatively self-contained and have minimal dependencies, which is what I
personally find to be one of their biggest appeals as projects.

If the main goal of this is to mitigate what led to the xz attack, I have
posted a few (less drastic) suggestions in #6316.

Other than that, autoconf and make seem to work fine for their intended purpose
and quite a few build system improvements were made since #4642 was opened (and
I have some more in mind).  If there is something else that could be improved
or clarified, feel free to open an issue about it.

topimiettinen:
I personally find this a nice improvement, especially in view of xz attack. Meson's syntax is also cleaner and the builds are faster.

I'd argue that there isn't that great difference between dependencies of auto**** and make vs. meson & ninja. The difference isn't also very meaningful these days, probably it would be a very niche case to build Firejail in such resource constrained embedded system where it would matter.

My feeling for auto**** is that unfortunately the architecture seems to be such a mess that it's not possible to improve it much, especially from our side. But if you find it important to keep using the old way, it's certainly possible to support  both build systems.

PR #6315, opened 2024-04-18
New profile: axel

glitsj16:
Axel
Lightweight CLI download accelerator

https://github.com/axel-download-accelerator/axel

### DIFF
--- a/etc/profile-a-l/axel.profile
+++ b/etc/profile-a-l/axel.profile
@@ -1,0 +1,61 @@
+# Firejail profile for axel
+# Description: Lightweight CLI download accelerator
+quiet
+# This file is overwritten after every install/update
+# Persistent local customizations
+include axel.local
+# Persistent global definitions
+include globals.local
+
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/axel.profile
```diff
@@ -0,0 +1,61 @@
...
+seccomp.block-secondary
+tracelog
+x11 none
+
+#disable-mnt
+private-bin axel
+private-cache
+private-dev
+private-etc @network,@tls-ca,axelrc
+private-lib
```
kmk3, line 54: This can use openssl/wolfssl, so doesn't it need lib access?

glitsj16: I do still build with `--enable-private-lib` and have `private-lib yes` in /etc/firejail/firejail.config. During my testing axel works just fine with `private-lib` for downloading FTPS/HTTPS URI's, if that's what you mean.

kmk3 (approved):

PR #6321, opened 2024-04-22
profiles: audacity: allow networking by default

glitsj16:
Newly release audacity 3.5 supports cloud-saving and remote backup features:

- https://www.audacityteam.org/blog/audacity-3-5/
- https://support.audacityteam.org/additional-resources/changelog/audacity-3.5#cloud-project-saving

### DIFF
--- a/etc/profile-a-l/audacity.profile
+++ b/etc/profile-a-l/audacity.profile
@@ -6,10 +6,9 @@ include audacity.local
 # Persistent global definitions
 include globals.local
 
-# Add the below lines to your audacity.local if you need online plugins.
-#ignore net none
-#netfilter
-#protocol inet6
+# To disable networking, add the following lines to audacity.local:
+#ignore netfilter
+#net none
 
 noblacklist ${HOME}/.audacity-data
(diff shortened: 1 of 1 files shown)


kmk3 (commented):
I think it makes sense to allow it by default to avoid breaking sync features.

Though I think many users would want to block this, considering all the
unfortunate changes that were made after audacity was sold out (data mining,
CLA and licensing issues).

Maybe add a comment where the "Add the below lines" comment is removed for how
to disable it:

```
# To disable networking, add the following lines to audacity.local:
#ignore netfilter
#net none
```

By the way, have you used tenacity?  It appears to be the most active fork and
is packaged on Arch:

* <https://codeberg.org/tenacityteam/tenacity>

glitsj16:
@kmk3
I've added the networking comment, makes sense to do so indeed.

> By the way, have you used tenacity?

Haven't used tenacity yet, hardly do any audio editing these days. But I'll give it a try. Thanks for the info and the review.

kmk3 (approved):

PR #6327, opened 2024-04-30
profiles: add loupe

SkewedZeppelin:
works but can be improved

loupe opens each image in its own sandbox using bwrap if available

largely copy/pasted from eo-common

### DIFF
--- a/etc/profile-a-l/loupe.profile
+++ b/etc/profile-a-l/loupe.profile
@@ -1,0 +1,50 @@
+# Firejail profile for loupe
+# Description: GNOME's modern Image Viewer program
+# This file is overwritten after every install/update
+# Persistent local customizations
+include loupe.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.local/share/Trash
+noblacklist ${HOME}/.Steam
+noblacklist ${HOME}/.steam
+
+#include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-programs.inc
+include disable-write-mnt.inc
+
+#whitelist /usr/share/glycin-loaders
+include whitelist-runuser-common.inc
+#include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
+net none
+nodvd
+nogroups
+noinput
+nonewprivs
+noprinters
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,netlink
+#loupe decodes all images in their own sandbox via glycin
+#https://gitlab.gnome.org/sophie-h/glycin#sandboxing-and-inner-workings
+#seccomp
+seccomp.block-secondary
+tracelog
+
+private-cache
+private-dev
+private-etc @x11
+private-tmp
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -525,6 +525,7 @@ lofromtemplate
 loimpress
 lollypop
 lomath
+loupe
 loweb
 lowriter
 #lrunzip # disable until we fix CLI archivers for makepkg on Arch (see discussion in #3095)


etc/profile-a-l/loupe.profile
```diff
@@ -0,0 +1,48 @@
...
+nodvd
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix,netlink
```
glitsj16, line 41: No `seccomp`?

etc/profile-a-l/loupe.profile
```diff
@@ -0,0 +1,48 @@
...
+seccomp.block-secondary
+tracelog
+
+private-cache
+private-dev
+private-etc alternatives,dconf,fonts,gtk-3.0,ld.so.cache,ld.so.preload
```
glitsj16, line 47: Use the new private-etc @group syntax:

```suggestion
private-etc @x11
```

etc/profile-a-l/loupe.profile
```diff
@@ -0,0 +1,48 @@
...
+# Persistent local customizations
+include loupe.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.local/share/Trash
+noblacklist ${HOME}/.Steam
+noblacklist ${HOME}/.steam
+
+noblacklist /usr/libexec/glycin-loaders
```
glitsj16, line 13: Never blacklisted...
Add the blacklist counterpart to `disable-programs.inc`.

etc/profile-a-l/loupe.profile
```diff
@@ -0,0 +1,48 @@
...
+noblacklist /usr/share/glycin-loaders
```
glitsj16, line 14: Never blacklisted...
Add the blacklist counterpart to disable-programs.inc.

glitsj16 (commented):
Nitpicks

etc/profile-a-l/loupe.profile
SkewedZeppelin: glycin already applies a seccomp filter https://gitlab.gnome.org/sophie-h/glycin#sandboxing-and-inner-workings

glitsj16: Cool. Perhaps a note to that effect might help avoiding someone adding seccomp in the future. But I leave that up to you. Nice seeing you here BTW.

glitsj16 (approved):
LGTM

SkewedZeppelin:
Merged, thanks @glitsj16!

etc/profile-a-l/loupe.profile
```diff
@@ -0,0 +1,50 @@
...
+# Persistent local customizations
+include loupe.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.local/share/Trash
+noblacklist ${HOME}/.Steam
+noblacklist ${HOME}/.steam
+
+#include disable-common.inc
```
kmk3, line 13: @glitsj16 @SkewedZeppelin

This profile has no disable-common and no `whitelist` or `read-only` in
`${HOME}`.

I can modify ~/.bashrc with it.

So what is the point?

If editing images is not a common use case, I'd add `read-only ${HOME}`.

If not, I'd uncomment disable-common.

SkewedZeppelin: I couldn't get disable-common to work

It is still on my TODO list

kmk3: > I couldn't get disable-common to work nothing stuck out when I strace'd it I
> think it is a conflict with how it invokes bwrap
> 
> also simple edits like rotation is common task to not want ro
> 
> It is still on my TODO list

I see.

Then I'd at least add a warning to the profile:

```diff
 # Persistent global definitions
 include globals.local
+
+# Warning: This profile is currently WIP and is not very secure.
```

And maybe remove it from firecfg.config since it's still WIP, to avoid passing
a false sense of security to users.

Note that you can add programs to firecfg in /etc/firejail/firecfg.d/ now (see
#5876).

> edit: also the primary reason I made a profile for it is because when
> hardened_malloc is used images never load due to the rlimit set on its
> sandbox which isn't enough virtual memory for the hmalloc to function

> my typical
> [workaround](https://github.com/divestedcg/rpm-hardened_malloc?tab=readme-ov-file#known-issues)
> is blacklisting /etc/ld.so.preload in my builds of firejail in my rpm repo
> and I saw loupe didn't have a profile yet

That clarifies things a bit; thanks for the details.

PR #6328, opened 2024-05-01
New profile: d-spy

glitsj16:
d-spy
D-Bus debugger for GNOME
https://gitlab.gnome.org/GNOME/d-spy

https://wiki.gnome.org/Apps/DFeet
D-Feet is no longer maintained. Please use d-spy

### DIFF
--- a/etc/profile-a-l/d-spy.profile
+++ b/etc/profile-a-l/d-spy.profile
@@ -1,0 +1,48 @@
+# Firejail profile for d-spy
+# Description: D-Bus debugger for GNOME
+# This file is overwritten after every install/update
+# Persistent local customizations
+include d-spy.local
+# Persistent global definitions
+include globals.local
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-xdg.inc
+
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+net none
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noroot
+nosound
+notv
+nou2f
+novideo
+protocol unix
+seccomp
+
+disable-mnt
+private-bin d-spy
+private-cache
+private-dev
+private-etc dbus-1
+private-tmp
+
+read-only ${HOME}
+restrict-namespaces
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -191,6 +191,7 @@ cryptocat
 cvlc
 cyberfox
 d-feet
+d-spy
 daisy
 darktable
 dconf-editor


etc/profile-a-l/d-spy.profile
```diff
@@ -0,0 +1,48 @@
...
+include whitelist-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+net none
+#no3d
```
kmk3, line 27: Why not enable it?

glitsj16: Fair question. I was planning on opening a dedicated discussion thread on the `no3d` option. GTK4 introduces quite some changes compared to GTK3. During testing my main focus was on the [new renderers](https://blog.gtk.org/2024/01/28/new-renderers-for-gtk/) and [graphics offloading](https://blog.gtk.org/2023/11/15/introducing-graphics-offload/).
Bottomline:
- improved power consumption (especially on laptops this is quite nice to have)
- actually working fractional scaling / improved antialiasing
- unlimited color stops
- ...

I do realize that the above aren't all that important from a security point of view. Not for an app like d-spy anyway (which doesn't neither imaging nor video processing). Still, if we keep cripling all this new stuff in GTK4 app profiles, users that have the hardware will never get the benefits. I'm not sure Firejail should take such decisions without considering this more thoroughly.
That being said, I'm very much willing to enable `no3d` in this profile and wait until we've discussed things like the above at a later date. What do you think?

kmk3: > Fair question. I was planning on opening a dedicated discussion thread on the
> `no3d` option. GTK4 introduces quite some changes compared to GTK3. During
> testing my main focus was on the [new
> renderers](https://blog.gtk.org/2024/01/28/new-renderers-for-gtk/) and
> [graphics
> offloading](https://blog.gtk.org/2023/11/15/introducing-graphics-offload/).
> 
> Bottomline:
> 
> * improved power consumption (especially on laptops this is quite nice to
>   have)

This does seem nice for laptops, though I don't know how much the improvement
would be for programs that do not process media.

> * actually working fractional scaling / improved antialiasing

Wouldn't these be handled by the compositor?

> * unlimited color stops

Not sure what this is.

> I do realize that the above aren't all that important from a security point
> of view. Not for an app like d-spy anyway (which doesn't neither imaging nor
> video processing). Still, if we keep cripling all this new stuff in GTK4 app
> profiles, users that have the hardware will never get the benefits.

The main concern is that graphics drivers are not exactly known for being
bastions of stability and security.

I think performance is important for programs that do heavy graphics
processing, but otherwise personally I'm glad to know that (IME) `no3d` makes a
program much less likely to have annoying artifacts or to crash for reasons
unrelated to the program itself.

> I'm not sure Firejail should take such decisions without considering this
> more thoroughly.

I wish there was a simple global toggle for this (like gtk4-common.profile),
though some gtk4 programs might be graphics-heavy and need 3d, so I don't know
how it could be done properly profile-wise.

> That being said, I'm very much willing to enable `no3d` in this profile and
> wait until we've discussed things like the above at a later date.  What do
> you think?

Sure.

glitsj16: > unlimited color stops
Not sure what this is.

The blog link explains it:
The old GL renderer handles linear, radial and conic gradients with up to 6 color stops. The unified renders allow an unlimited number of color stops. The new renderers also apply antialiasing to gradients, `so sharp edges will have smooth lines`. Again, not relevant for d-spy.

> I wish there was a simple global toggle for this (like gtk4-common.profile),
though some gtk4 programs might be graphics-heavy and need 3d, so I don't know
how it could be done properly profile-wise.

Very true. Haven't found a way to do that properly neither.

TL;DR: I've enabled `no3d` in the profile. Thanks for the review and raising interesting points as always!

kmk3 (approved):

PR #6339, opened 2024-05-13
build: sort.py: use -i by default and add -n

kmk3:
Overwrite in-place by default (`-i`) and add `-n` to override it.

This restores the previous default behavior (from 0.9.72), for the sake
of being consistent with all previous versions and because it's more
likely to be the desired behavior in most cases.

This amends commit aa08aa132 ("build: sort.py: add and require -i to
edit in-place (#6290)", 2024-03-25).

* build: sort.py: support "--" and fail on unknown option

Support "--" to end options and fail if an unknown option is given.

### DIFF
--- a/contrib/sort.py
+++ b/contrib/sort.py
@@ -11,7 +11,7 @@ from sys import argv, exit as sys_exit, stderr
 __doc__ = f"""\
 Sort the arguments of commands in profiles.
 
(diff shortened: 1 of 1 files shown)


glitsj16 (approved):
LGTM. Good idea to change the default behavior.

rusty-snake (approved):

PR #6342, opened 2024-05-17
build: add strip target and simplify install targets

kmk3:
* build: add a standalone strip target

Move the strip invocation into its own target to allow stripping
binaries without having to run the "realinstall" target.

* build: remove redundant realinstall target

Leave just the "install" and "install-strip" targets.

See commit 099925e18 ("added install-strip, make install now without
strip.", 2015-09-10) / PR #60 and commit 0215cbc02 ("make install, make
install-strip", 2015-09-11).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,8 +193,8 @@ clean:
 distclean: clean
 	rm -fr autom4te.cache config.log config.mk config.sh config.status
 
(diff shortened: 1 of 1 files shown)


glitsj16 (approved):
Looks fine.

PR #6343, opened 2024-05-17
build: remove clean dependency from cppcheck targets

kmk3:
Cleaning does not appear to make a difference; the same amount of files
is checked with/without cleaning.

Environment: cppcheck 2.12.0-3 on Artix Linux.

Added on commit 4e22add64 ("llvm scan", 2015-11-29).

This is a follow-up to #6222.


glitsj16 (approved):
LGTM

PR #6354, opened 2024-05-24
build: allow overriding common tools

kmk3:
Tools:

* gzip
* install
* rm
* strip
* tar

For the programs not checked in configure.ac:

From the manual of GNU Autoconf (version 2.71):

> If you use `AC_PROG_INSTALL`, you must include `install-sh` in your
> distribution

So set `install` just in the Makefile.  Use `$(RM)` to ensure that `-f`
is always used and to make it easier to spot when `-r` is used.

See commit 93d623fdf ("build: allow overriding certain tools",
2024-02-23) / PR #6222.

This is a follow-up to #6342.


glitsj16 (approved):
LGTM

glitsj16 (approved):
LGTM

PR #6361, opened 2024-05-26
profiles: blacklist i3 IPC socket & dir except for i3 itself

smheidrich:
This closes the escape route discussed in https://github.com/netblue30/firejail/discussions/6357.

It's left open for i3's own profile, so that people who run i3 itself sandboxed still have the option to use IPC with it at all.

Reference for file paths: https://i3wm.org/docs/userguide.html#_interprocess_communication

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -167,6 +167,10 @@ blacklist ${RUNUSER}/gnome-session-leader-fifo
 blacklist ${RUNUSER}/gnome-shell
 blacklist ${RUNUSER}/gsconnect
 
+# i3 IPC socket (allows arbitrary shell script execution)
+blacklist ${RUNUSER}/i3/ipc-socket.*
+blacklist /tmp/i3-*/ipc-socket.*
+
 # systemd
 blacklist ${HOME}/.config/systemd
 blacklist ${HOME}/.local/share/systemd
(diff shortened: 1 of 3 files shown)


glitsj16 (approved):
LGTM. Thanks for the PR.

etc/inc/disable-common.inc
```diff
@@ -167,6 +167,10 @@ blacklist ${RUNUSER}/gnome-session-leader-fifo
 blacklist ${RUNUSER}/gnome-shell
 blacklist ${RUNUSER}/gsconnect
 
+# i3 IPC socket (allows arbitrary shell script execution)
+blacklist ${RUNUSER}/i3/ipc-socket.*
+blacklist /tmp/i3-*/ipc-socket.*
```
rusty-snake, line 172: Is there anything else in the i3 directory?

XY: Would it make sense to blacklist the directory?

smheidrich: On my system, the only other files in there are an error log and another UNIX domain socket called `log-stream-socket.*`. Seems neither very dangerous nor very useful, so it probably doesn't matter either way.

rusty-snake: Logs could contain sensitive information.

If there is thing important in it we should blacklist the directory IMHO.

glitsj16: FWIW, I agree. Targetting the directory is the better option here.

smheidrich: Fair enough, I changed it to block the entire directories.

glitsj16 (approved):
LGTM

kmk3 (approved):

PR #6365, opened 2024-06-02
New profile: armcord

glitsj16:
ArmCord
Standalone Discord client

https://armcord.app/
https://github.com/NextWork123/ArmCord

Requested in https://github.com/netblue30/firejail/issues/1139#issuecomment-2140174880.

### DIFF
--- a/etc/profile-a-l/armcord.profile
+++ b/etc/profile-a-l/armcord.profile
@@ -1,0 +1,40 @@
+# Firejail profile for armcord
+# Description: Standalone Discord client
+# This file is overwritten after every install/update
+# Persistent local customizations
+include armcord.local
+# Persistent global definitions
+include globals.local
+
+# Modules might depend on nodejs.
+# Add the below lines to your armcord.local if you need this.
+# Allow node (disabled by disable-interpreters.inc)
+#include allow-nodejs.inc
+#private-bin node
+
+# The lines below are needed to find the default Firefox profile name, to allow
+# opening links in an existing instance of Firefox (note that it still fails if
+# there isn't a Firefox instance running with the default profile; see #5352)
+noblacklist ${HOME}/.mozilla
+whitelist ${HOME}/.mozilla/firefox/profiles.ini
+
+noblacklist ${HOME}/.config/ArmCord
+
+mkdir ${HOME}/.config/ArmCord
+whitelist ${HOME}/.config/ArmCord
+whitelist /opt/armcord
+whitelist /usr/share/armcord
+
+ignore novideo
+private-bin armcord
+
+dbus-user filter
+dbus-user.talk org.freedesktop.Notifications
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/armcord.profile
```diff
@@ -0,0 +1,43 @@
...
+# This file is overwritten after every install/update
+# Persistent local customizations
+include armcord.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/ArmCord
+
+# sh is needed to allow Firefox to open links
+#include allow-bin-sh.inc
```
kmk3, line 12: Is the entry supposed to be commented?

It is uncommented in other profiles that also have the firefox part.

etc/profile-a-l/armcord.profile
```diff
@@ -0,0 +1,43 @@
...
+whitelist /usr/share/armcord
+
+# The lines below are needed to find the default Firefox profile name, to allow
+# opening links in an existing instance of Firefox (note that it still fails if
+# there isn't a Firefox instance running with the default profile; see #5352)
+noblacklist ${HOME}/.mozilla
+whitelist ${HOME}/.mozilla/firefox/profiles.ini
+
+ignore novideo
+private-bin armcord
```
kmk3, line 32:
```suggestion
include globals.local

# sh is needed to allow Firefox to open links
#include allow-bin-sh.inc

# Modules might depend on nodejs.
# Add the below lines to your armcord.local if you need this.
# Allow node (disabled by disable-interpreters.inc)
#include allow-nodejs.inc
#private-bin node

# The lines below are needed to find the default Firefox profile name, to allow
# opening links in an existing instance of Firefox (note that it still fails if
# there isn't a Firefox instance running with the default profile; see #5352)
noblacklist ${HOME}/.mozilla
whitelist ${HOME}/.mozilla/firefox/profiles.ini

noblacklist ${HOME}/.config/ArmCord

mkdir ${HOME}/.config/ArmCord
whitelist ${HOME}/.config/ArmCord
whitelist /opt/armcord
whitelist /usr/share/armcord

ignore novideo
private-bin armcord
```

To make it easier to isolate and read the normal profile entries, it might be
useful to put all the optional sections together.

Such as right after globals.local (and maybe also the `ignore` commands) or
right before the redirect.

etc/profile-a-l/armcord.profile
glitsj16: Strictly speaking we can drop both lines: `nothing` down the include chain has `include disable-shell.inc`. That is different in all other profiles that have this 'open links with Firefox' support. For the latter it's a hard-dependency so to speak. I left it in (commented) on purpose. To have a reminder this is needed _if_ something changes this situation.

kmk3 (approved):

PR #6369, opened 2024-06-04
bugfix: profstats: fix restrict-namespaces max count

i-pankrat:


glitsj16 (approved):
LGTM

kmk3 (commented):
Note: This is more than just a typo as it actually fixes the code.  But I'm not
familiar with this code, so if the title is not accurate, feel free to change
it.

Please amend the commit message title to match the PR title so that it's
consistent with the RELNOTES.

kmk3 (approved):

PR #6378, opened 2024-06-11
profiles: claws-mail: add ~/.cache/claws-mail 

glitsj16:
It is apparently used by the (widely used) "Fancy" plugin, which
"Renders HTML e-mail using the WebKit library".

https://www.claws-mail.org/plugins.php

Relates to #6377.

Note: etc/profile-a-l/email-common.profile contains `private-cache`.

### DIFF
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -100,6 +100,7 @@ blacklist ${HOME}/.cache/cantata
 blacklist ${HOME}/.cache/champlain
 blacklist ${HOME}/.cache/chromium
 blacklist ${HOME}/.cache/chromium-dev
+blacklist ${HOME}/.cache/claws-mail
 blacklist ${HOME}/.cache/cliqz
 blacklist ${HOME}/.cache/com.github.johnfactotum.Foliate
 blacklist ${HOME}/.cache/darktable
--- a/etc/profile-a-l/claws-mail.profile
+++ b/etc/profile-a-l/claws-mail.profile
@@ -11,9 +11,12 @@ include globals.local
 # adding the following line to claws-mail.local:
 #ignore no3d
 
+noblacklist ${HOME}/.cache/claws-mail
 noblacklist ${HOME}/.claws-mail
 
+mkdir ${HOME}/.cache/claws-mail
 mkdir ${HOME}/.claws-mail
+whitelist ${HOME}/.cache/claws-mail
 whitelist ${HOME}/.claws-mail
 
 # Add the below lines to your claws-mail.local if you use python-based plugins.


kmk3 (commented):
email-common contains `private-cache`, so wouldn't these changes be a no-op?

Does the plugin not work with the current profile?

glitsj16:
> email-common contains private-cache, so wouldn't these changes be a no-op?

You're absolutely right. This PR makes no sense, my bad.

kmk3:
> > email-common contains private-cache, so wouldn't these changes be a no-op?
> 
> You're absolutely right. This PR makes no sense, my bad.

I was confused about the plugin not working without the changes, but the
changes themselves seem fine (if they are indeed a no-op with `private-cache`).

It seems good for documentation purposes, to have it blocked in
disable-programs and to make it easier to use the real cache directory (by just
using `ignore private-cache`).

kmk3:
@glitsj16 

Hello, do you want to continue with this PR?

It's likely possible to click to restore the branch and then to reopen the PR.

glitsj16:
@kmk3 

Hi, I'll take care of it today. Last days of school exams here, less computer-time for me usually. But it shouldn't take that long to restore and reopen this. Thanks for the reminder!

kmk3:
> Hi, I'll take care of it today. Last days of school exams here, less
> computer-time for me usually. But it shouldn't take that long to restore and
> reopen this. Thanks for the reminder!

Ah I see, I wasn't sure if you had seen previous comment, so I commented again
just in case.

Anyway, no hurry on the PR.

Good luck on the exams!

kmk3 (commented):
Why the new changes?

It seems like the version from yesterday would work and make it easier to use
the real cache directory if desired.

glitsj16:
@kmk3

> Why the new changes?

Oh, I misunderstood. Thought you'd `only` wanted to keep the `disable-common.inc` part and leave `private-cache` as-is. To avoid more mishaps from my side, do you think it would make more sense to `ignore private-cache` for CM and use the dedicated whitelisted `~/.cache/claws-mail` instead?

kmk3:
> To avoid more mishaps from my side, do you think it would make more sense to
> `ignore private-cache` for CM and use the dedicated whitelisted
> `~/.cache/claws-mail` instead?

I think in general it's good to leave `private-cache` on by default to help
keep sandboxes isolated and to avoid unecessarily reducing the lifespan of
storage devices, as IME programs rarely make good use of ~/.cache.

The exceptions would be if the program creates a giant cache (which like
`private-opt` could take too much RAM) or when caching actually makes a
positive noticeable difference (like shader caches in certain programs).

kmk3 (approved):

Issue #6364, opened 2024-06-02
private-dev & private-bin break hashcat


schrotthaufen:
### Description

The default profile for hashcat uses, private-bin, and private-dev, which break hashcat.
I have a AMD RX 7900 XT GPU.

### Steps to Reproduce

1. Run in bash `LC_ALL=C firejail hashcat -b -m 1000` 

### Expected behavior

hashcat starts working.

### Actual behavior

* With `private-bin hashcat`: hashcat throws an error, and quits: `/usr/local/bin/OpenCL/: No such file or directory`
* With `private-dev`: hashcat throws an error, and quits: `No devices found/left.`
### Behavior without a profile

hashcat works as expected.

### Additional context

I think `/dev/kfd` is required to make `private-dev` work, but if I pass `--whitelist=/dev/kfd`, the `/dev/` directory is empty.

### Environment

- Arch Linux, kernel 6.9.3-arch1-1
- ```
firejail version 0.9.72 
Compile time support: 
  - always force nonewprivs support is disabled
  - AppArmor support is enabled
  - AppImage support is enabled
  - chroot support is enabled
  - D-BUS proxy support is enabled
  - file transfer support is enabled
  - firetunnel support is disabled
  - IDS support is disabled
  - networking support is enabled- output logging is enabled                                                                                                                                                           │         │
│ 38 disabl│        - overlayfs support is disabled                                                                                                                                                       │         │
│ 39 #priva│        - private-home support is enabled                                                                                                                                                     │         │
│ 40 privat│        - private-cache and tmpfs as user enabled                                                                                                                                             │         │
│ 41 privat│        - SELinux support is disabled                                                                                                                                                         │         │
│ 42 privat│        - user namespace support is enabled                                                                                                                                                   │         │
│ 43 ⏎     │        - X11 sandboxing support is enabled 
- If you use a development version of firejail, also the commit from which it
  was compiled (`git rev-parse HEAD`).

### Checklist

- [ ] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [ ] I can reproduce the issue without custom modifications (e.g. globals.local).
- [ ] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [ ] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [ ] I have performed a short search for similar issues (to avoid opening a duplicate).
[...]

PR #6380, opened 2024-06-13
modif: private-dev: keep /dev/kfd unless no3d is used

kmk3:
Apparently hashcat needs access to this device for hardware acceleration
on AMD gpus and `private-dev` currently breaks that (see #6364).

It seems to be used by the "amdgpu" and "amdkfd" Linux kernel drivers
and as with /dev/dri/renderD128, this file is owned by the "render"
group.

Relevant udev rules from /usr/lib/udev/rules.d/50-udev-default.rules:

    SUBSYSTEM=="drm", KERNEL=="renderD*", GROUP="render", MODE="0666"
    SUBSYSTEM=="kfd", GROUP="render", MODE="0666"
    SUBSYSTEM=="accel", GROUP="render", MODE="0666"

Environment: udev 255.6-1 on Artix Linux.

Reported-by: @schrotthaufen

### DIFF
--- a/src/firejail/fs_dev.c
+++ b/src/firejail/fs_dev.c
@@ -52,7 +52,8 @@ typedef struct {
 
 static DevEntry dev[] = {
 	{"/dev/snd", RUN_DEV_DIR "/snd", DEV_SOUND},	// sound device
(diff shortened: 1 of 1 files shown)


glitsj16 (approved):
Nice find!

Issue #6119, opened 2023-12-10
Fractal 5 fails to start


mhmdanas:
### Description

Fractal 5 fails to start with Firejail.

### Steps to Reproduce

_Steps to reproduce the behavior_

1. `LC_ALL=C firejail /bin/fractal`
2. Fractal outputs `Could not load gresource file: Error { domain: g-file-error-quark, code: 4, message: "Failed to open file “/usr/share/fractal/resources.gresource”: open() failed: No such file or directory" }`

### Expected behavior

Fractal starts up successfully.

### Actual behavior

Fractal crashes before startup.
### Behavior without a profile

Seems to work fine without a profile.

### Environment

- Linux distribution and version (e.g. "Ubuntu 20.04" or "Arch Linux") Void Linux
- Firejail version (`firejail --version`). 0.9.72

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ ] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

<details>
<summary>Output of <code>LC_ALL=C firejail /path/to/program</code></summary>
<p>

```
Reading profile /etc/firejail/fractal.profile                   
Reading profile /etc/firejail/allow-python2.inc                                                                                 
Reading profile /etc/firejail/allow-python3.inc                 
Reading profile /etc/firejail/disable-common.inc                
Reading profile /etc/firejail/disable-devel.inc                 
Reading profile /etc/firejail/disable-exec.inc                                                                                  
Reading profile /etc/firejail/disable-interpreters.inc          
Reading profile /etc/firejail/disable-programs.inc                                                                                       
Reading profile /etc/firejail/disable-shell.inc                 
Reading profile /etc/firejail/disable-xdg.inc                                                                                            
Reading profile /etc/firejail/whitelist-common.inc              
[...]

PR #6392, opened 2024-06-23
profiles: fractal: add ~/.local/share/fractal

FelixPehla:
I'm on Arch using Fractal 7 and firejail 0.9.72. Even after fixing #6119, using Fractal with firejail would either not work (without Internet access) or show me "Crypto identity incomplete" when using it with, as it stores (encrypted) messages and key material in ~/.local/share/fractal.

This patch fixes that issue for me. Let me know if you'd like me to open an issue first to discuss the problem there before accepting a pull request, or if you have any advice or improvements.

### DIFF
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -950,6 +950,7 @@ blacklist ${HOME}/.local/share/feedreader
 blacklist ${HOME}/.local/share/feral-interactive
 blacklist ${HOME}/.local/share/five-or-more
 blacklist ${HOME}/.local/share/fluffychat
+blacklist ${HOME}/.local/share/fractal
 blacklist ${HOME}/.local/share/freecol
 blacklist ${HOME}/.local/share/gajim
 blacklist ${HOME}/.local/share/gdfuse
--- a/etc/profile-a-l/fractal.profile
+++ b/etc/profile-a-l/fractal.profile
@@ -7,6 +7,7 @@ include fractal.local
 include globals.local
 
 noblacklist ${HOME}/.cache/fractal
+noblacklist ${HOME}/.local/share/fractal
 
 # Allow python (blacklisted by disable-interpreters.inc)
 include allow-python2.inc
@@ -21,7 +22,9 @@ include disable-shell.inc
 include disable-xdg.inc
 
 mkdir ${HOME}/.cache/fractal
+mkdir ${HOME}/.local/share/fractal
 whitelist ${HOME}/.cache/fractal
+whitelist ${HOME}/.local/share/fractal
 whitelist ${DOWNLOADS}
 whitelist /usr/share/fractal
 include whitelist-common.inc


glitsj16 (changes requested):
Needs additional work:

- add `blacklist ${HOME}/.local/share/fractal` to `disable-programs.inc`
- add `noblacklist ${HOME}/.local/share/fractal` to `fractal.profile`

glitsj16 (approved):
Thanks for the PR, LGTM now.

FelixPehla:
Thanks for the looking over it, I amended my previous commit.

I should have probably taken a bit more time to look over the differences between the structure of the last release and master.

By the way, while debugging this I noticed some accessibility-related errors because of missing access to [`org.a11y.bus`](https://www.a11yproject.com/) and the assistive technology service provider interface (`${RUNUSER}/at-spi/bus`).

I didn't see a `.inc` file for accessibility related features, so it might be a good idea to add the option to the profile as well (commented out because of the increased attack surface, with the reference to fractal.local)?

Though I assume people that need this might want to just put it in their global config anyway.

glitsj16:
> By the way, while debugging this I noticed some accessibility-related errors because of missing access to org.a11y.bus and the assistive technology service provider interface (${RUNUSER}/at-spi/bus).

If that is exposed via D-Bus we might consider adding it to fractal.profile:

dbus-user.talk org.a11y.bus

FelixPehla:
only adding that still leaves it with `(fractal:26): Gtk-CRITICAL **: 22:18:28.795: Unable to connect to the accessibility bus at 'unix:path=/run/user/1000/at-spi/bus': Could not connect: No such file or directory`, so I suspect you need `whitelist ${RUNUSER}/at-spi/bus` (and maybe a `noblacklist`) too. But I don't know much about those features so I don't know for sure. How would you feel about adding something like this:

    # Add the next lines to your fractal.local to allow use of accessibility features
    #whitelist ${RUNUSER}/at-spi/bus
    #dbus-user.talk org.a11y.Bus

Is there a preferred way of documenting settings with different keyword but related to the same feature? The most elegant way to do this would probably be a conditional, but creating one (and looking where else it should be used) would definitely be out-of-scope of this PR.

glitsj16:
Thanks for testing a11y. A conditional would be nice, but like you stated, that's out of scope here. Personally I'd leave out any comments regarding a11y and open a separate issue for it so we can add that in later.

kmk3 (changes requested):
@FelixPehla [on Jun 23](https://github.com/netblue30/firejail/pull/6392#issue-2368829116):

> I'm on Arch using Fractal 7 and firejail 0.9.72. Even after fixing #6119,
> using Fractal with firejail would either not work (without Internet access)
> or show me "Crypto identity incomplete" when using it with, as it stores
> (encrypted) messages and key material in ~/.local/share/fractal.
> 
> This patch fixes that issue for me. Let me know if you'd like me to open an
> issue first to discuss the problem there before accepting a pull request

This PR is very simple and self-contained; opening an issue first is more so for
changes that are more complex and/or that affect many files.

> or if you have any advice or improvements.

The current commit message title is very long (>72 characters):

```console
$ git show --pretty='%s' -s ba0425fe09
profiles: fractal: mkdir and whitelist ~/.local/share/fractal as it is used for message and key storage
$ git show --pretty='%s' -s ba0425fe09 | tr -d '\n' | wc -m
103
```

Also, the changes are fairly obvious from the PR title, but for completeness you
could include in the commit message the context from the PR description, such as
the error message.

Suggestion:

```
profiles: fractal: add ~/.local/share/fractal

Fractal 7 on Arch either does not work (without Internet access) or
shows "Crypto identity incomplete", as it stores (encrypted) messages
and key material in ~/.local/share/fractal.
```

kmk3:
@glitsj16 [on Jun 23](https://github.com/netblue30/firejail/pull/6392#issuecomment-2185346804):

> Thanks for testing a11y. A conditional would be nice, but like you stated,
> that's out of scope here. Personally I'd leave out any comments regarding
> a11y and open a separate issue for it so we can add that in later. @kmk3 What
> do you think?

Agreed.

The more self-contained a PR is, the easier to review.

The current changes look good as is and the a11y changes seem a bit trickier.

After this PR you could open another one with the a11y changes for fractal and
if it works we could later discuss in a dedicated issue how to implement it for
more profiles.

FelixPehla:
Thanks for the feedback and the link @kmk3, I'm pretty new to this. I have adjusted my commit message so everything should be good now.

kmk3 (approved):

kmk3:
> Thanks for the feedback and the link @kmk3, I'm pretty new to this. I have
> adjusted my commit message so everything should be good now.

No problem; the new message looks good.

You made a good attempt to follow the existing conventions and feedback, so
this PR went great.

Looking forward to more contributions :)

PR #6401, opened 2024-07-09
modif: remove --noautopulse from --help and zsh comp

kmk3:
This command is deprecated and may be confused for a hardening option.

This amends commit 5a612029b ("rename noautopulse to keep-config-pulse",
2021-05-13) / PR #4278.

This is a follow-up to #6390.


glitsj16 (approved):
LGTM

PR #6408, opened 2024-07-16
profiles: okular: fix "Print to PDF"

gcb:
Fixes #6402.

### DIFF
--- a/etc/profile-m-z/okular.profile
+++ b/etc/profile-m-z/okular.profile
@@ -58,7 +58,7 @@ protocol unix
 seccomp
 tracelog
 
-private-bin kbuildsycoca4,kdeinit4,lpr,okular,unar,unrar
+private-bin kbuildsycoca4,kdeinit4,lpr,okular,ps2pdf,unar,unrar
 private-dev
 private-etc @x11,cups
 # on KDE we need access to the real /tmp for data exchange with email clients


etc/profile-m-z/okular.profile
```diff
@@ -58,7 +58,9 @@ protocol unix
 seccomp
 tracelog
 
-private-bin kbuildsycoca4,kdeinit4,lpr,okular,unar,unrar
+private-bin kbuildsycoca4,kdeinit4,lpr,okular,unar,unrar,ps2pdf
```
glitsj16, line 61: Sort:

```suggestion
private-bin kbuildsycoca4,kdeinit4,lpr,okular,ps2pdf,unar,unrar
```

etc/profile-m-z/okular.profile
```diff
@@ -58,7 +58,9 @@ protocol unix
...
+# ps2pdf - enables print to pdf functionality
```
glitsj16, line 62: Remove both this and the following line. We have the outdated info in https://github.com/netblue30/firejail/discussions/6402 if we want to change that in the future.

glitsj16 (changes requested):
Needs minor changes.

etc/profile-m-z/okular.profile
gcb: k. i usually comment all the things I allow on my app armor profiles. was just going along here too.

glitsj16 (approved):
LGTM

gcb:
it failed on the sort step which you just suggested to fix

glitsj16 (approved):
LGTM. Let's run the checks again...

kmk3 (approved):

Issue #5127, opened 2022-05-02
Spectacle not working under KDE Wayland


wushangwei:
### Description

Spectacle not working under KDE Wayland.  It opens, but complains "Could not take a screenshot". However it works under x11 session.

### Steps to Reproduce
1. Click the spectacle desktop shortcut and doesn't work.
2. kioclient exec /home/nikki/.local/share/applications/org.kde.spectacle.desktop, doesn't work.
Logs are shown in Log section.

3. To reduce the dbus errors above, I created ~/.config/firejail/spectacle.local with the following content:

```shell
dbus-user.talk org.kde.JobViewServer
dbus-user.talk org.kde.kglobalaccel
dbus-user.talk org.kde.KWin
dbus-user.talk org.kde.plasmashell
dbus-user.talk org.kde.KWin.ScreenShot2
```

Run the command above again, DBus errors are gone, but left with Screenshot request failed: "The process is not authorized to take a screenshot". Still doesn't work.

### Expected behavior

Spectacle should take screenshots normally under KDE Wayland.

### Actual behavior

Cannot take screenshots under KDE Wayland.  Does not affect X11 session. Console outputs are provided above. 
If I modify the desktop file, replace "spectacle" with "/usr/bin/spectacle", it will take screenshot normally.

### Behavior without a profile

LC_ALL=C firejail --noprofile  kioclient exec /home/nikki/.local/share/applications/org.kde.spectacle.desktop

Logs are shown in Log Section. Console output is similar with the one after modifying spectacle.local. Doesn't work either.

### Additional context

If simply edit the spectacle desktop file and change the Exec from "spectacle" to "/usr/bin/spectacle", it will work normally.

### Environment

- Arch Linux
- firejail version 0.9.68
- KDE Wayland

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
[...]

PR #6412, opened 2024-07-20
profiles: firecfg.config: disable spectacle

kmk3:
There are various reports in #5127 that the current profile is broken on
wayland (and at least one report that it is broken on xorg as well).

Relates to #6268.


glitsj16 (approved):
LGTM

Issue #2995, opened 2019-10-08
Epiphany needs bwrap


mkdy:
A fresh version of Epiphany browser (3.34.1-1) uses bwrap for some purposes (have no idea why).
I've created `epiphany.local` and filled it with:
`noblacklist ${PATH}/bwrap`
However, it seems that `bwrap` itself needs some permissions such as internet access.

Vincent43:
bwrap is very similar to firejail itself sandboxing tool (used by flatpak) and I guess Epiphany uses it for that. Perhaps we have to drop epiphany support as overlapping sandboxes can't work.

rusty-snake:
Confirming. Fedora 31 BETA VM with firejail-git.
```
$ LC_ALL=C epiphany 
Reading profile /etc/firejail/epiphany.profile
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-devel.inc
Reading profile /etc/firejail/disable-interpreters.inc
Reading profile /etc/firejail/disable-programs.inc
Reading profile /etc/firejail/whitelist-common.inc
Parent pid 41676, child pid 41677
Child process initialized in 142.89 ms

** (epiphany:5): ERROR **: 17:02:33.800: Unable to fork a new child process: Failed to execute child process ?/usr/bin/bwrap? (Permission denied)

Parent is shutting down, bye...
$ LC_ALL=C firejail --noprofile epiphany 
Parent pid 41762, child pid 41763
Child process initialized in 11.65 ms
Warning: an existing sandbox was detected. /usr/bin/epiphany will run without any additional sandboxing features
bwrap: Can't mount proc on /newroot/proc: Operation not permitted

(epiphany:2): GLib-GObject-WARNING **: 17:02:50.308: ../gobject/gsignal.c:2647: instance '0x55c894894390' has no handler with id '2782'

Parent is shutting down, bye...
```

FOSSONLY:
@Vincent43
In fact, Firejail had no support for the Epiphany browser so far. The existing profile called Epiphany refers to a game of the same name. You should better use the Firefox profile for Epiphany.

However, I'm not sure how good it is for programs to bring their own sandbox. So the security is in the hands of the developers, and the user loses any flexibility to define it, if you don't want to change the source code. Personally, I would always use Firejail, because I don't think any program should be able to control its own security. For me, this is something that has to be centrally enforced, to which every program has to subordinate itself. What do the others think about it?

rusty-snake:
[...]

PR #6418, opened 2024-07-26
profiles: bijiben: update webkit var and disable in firecfg

glitsj16:
The current `bijiben.profile` sets an environment variable to disable
its internal webkit/bubblewrap sandbox but now a different variable
needs to be set[1]:

    WEBKIT_FORCE_SANDBOX no longer allows disabling the sandbox. Use WEBKIT_DISABLE_SANDBOX_THIS_IS_DANGEROUS=1 instead.

This may be needed to make the profile work, but disabling the sandbox
affects the security in webkit[2], so update the variable and disable
bijiben by default in firecfg.config.

Note: Upstream replaced bijiben by gnome-notes[3] [4].

Relates to #2995.

[1] https://github.com/WebKit/WebKit/blob/0678a98c864ee36f0114ea4e7d303fd07788a822/Source/WebKit/UIProcess/Launcher/glib/ProcessLauncherGLib.cpp#L117
[2] https://github.com/netblue30/firejail/issues/2995
[3] https://archlinux.org/packages/extra/x86_64/gnome-notes/
[4] https://wiki.gnome.org/Apps/Notes

### DIFF
--- a/etc/profile-a-l/bijiben.profile
+++ b/etc/profile-a-l/bijiben.profile
@@ -59,5 +59,8 @@ dbus-user.talk ca.desrt.dconf
 dbus-user.talk org.freedesktop.Tracker1
 dbus-system none
 
-env WEBKIT_FORCE_SANDBOX=0
+# Warning: Disabling the webkit sandbox may be needed to make firejail work
+# with webkit2gtk, but this is not recommended (see #2995).
+# Add the following line to bijiben.local at your own risk:
+#env WEBKIT_DISABLE_SANDBOX_THIS_IS_DANGEROUS=1
 restrict-namespaces
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -97,7 +97,7 @@ basilisk
 bcompare
 beaker
 bibletime
-bijiben
+#bijiben # webkit2gtk-4.x requires bwrap (see #3647)
 bitcoin-qt
 bitlbee
 bitwarden


rusty-snake:
FTR https://github.com/netblue30/firejail/pull/3926

> In general we should
exclude a program from firecfg until a  solution is found. But
bijiben is special, while epiphany or evolution display random stuff
from the internet is webkit2gtk in bijiben used to display local files
create by the user. Bijiben has a thight profile (net none, whitelist,
private-bin, ...) therefore my decision here was to disable the
webkit2gtk sandbox rather then firejail.

I still consider it less insecure for bijiben because of the trusted input. However every usage of an general insecure practice "teaches" users. And we  already saw all this FUD about internal sandboxing of webkit4gtk/chromium.

glitsj16:
@rusty-snake

Thanks for your response. I wouldn't mind keeping bijiben in firecfg. But I'm not sure how we'd fix the now deprecated env var. Replacing that with the new one (could break users older bijiben)? Just forget about this and wait for people to report problems? Please advise if you find the time. Doesn't look to be anything urgent anyway.

rusty-snake:
No, removing it is fine, just wanted to link back some older discussion.

etc/profile-a-l/bijiben.profile
```diff
@@ -59,5 +59,4 @@ dbus-user.talk ca.desrt.dconf
 dbus-user.talk org.freedesktop.Tracker1
 dbus-system none
 
-env WEBKIT_FORCE_SANDBOX=0
```
kmk3, line 62: Does the profile work at all without this?

If not, I'd suggest the following:

```diff
-env WEBKIT_FORCE_SANDBOX=0
+# Warning: Disabling the webkit sandbox may be needed to make firejail work
+# with webkit2gtk, but this is not recommended (see #2995).
+# Add the following line to bijiben.local at your own risk:
+#env WEBKIT_DISABLE_SANDBOX_THIS_IS_DANGEROUS=1
```

glitsj16: Added. Thanks for the suggestion!

kmk3 (approved):

PR #6419, opened 2024-07-26
profiles: ssh: allow gpgagent socket for custom homedir

ayham-1:
If a custom GPG homedir is used, a hash of its path is used in the path
of the gpg agent socket[1].

For example, when running:

    gpgconf --list-dirs agent-ssh-socket

With a custom homedir it returns:

    /run/user/1000/gnupg/<hashed homedir>/S.gpg-agent.ssh

Environment: gnupg 2.4.5-4 on Arch Linux.

[1] https://github.com/gpg/gnupg/blob/91532dc3f40599b74fff489cb021c0cd3344ae60/common/homedir.c#L1342

### DIFF
--- a/etc/profile-m-z/ssh.profile
+++ b/etc/profile-m-z/ssh.profile
@@ -19,7 +19,8 @@ include disable-exec.inc
 include disable-programs.inc
 
 whitelist ${RUNUSER}/gcr/ssh
-whitelist ${RUNUSER}/gnupg/S.gpg-agent.ssh
+whitelist ${RUNUSER}/gnupg/S.gpg-agent.ssh # default gpg homedir setup
+whitelist ${RUNUSER}/gnupg/*/S.gpg-agent.ssh # custom gpg homedir setup
 whitelist ${RUNUSER}/keyring/ssh
 include whitelist-usr-share-common.inc
 include whitelist-runuser-common.inc


glitsj16:
Hi @ayham-1,

> fixes issues relating to usage of ssh with gpg-agent.

On my Arch Linux box I'm seeing the below when running the `gpgconf` command:

```sh
$ pacman -Q gnupg
gnupg 2.4.5-4
$ gpgconf --list-dirs agent-ssh-socket
/run/user/1001/gnupg/S.gpg-agent.ssh
```

Not saying your different output is incorrect, just wondering `what OS` you're seeing this on. Also, things might be more informative if you could open a `issue` on your specific problems using ssh with gpg-agent.

Regards

etc/profile-m-z/ssh.profile
```diff
@@ -19,7 +19,7 @@ include disable-exec.inc
 include disable-programs.inc
 
 whitelist ${RUNUSER}/gcr/ssh
-whitelist ${RUNUSER}/gnupg/S.gpg-agent.ssh
```
glitsj16, line 22: Removing this whitelist will cause breakage on systems where that path _is_ indeed the one that is being used (as shown by gpgconf). For now it would be better to just `add` a whitelist for other paths instead of `replacing` known good whitelists.

glitsj16 (changes requested):
Needs changes. Also, let's wait for more informative issue report on what this PR actually fixes.

ayham-1:
> Not saying your different output is incorrect, just wondering `what OS` you're seeing this on. Also, things might be more informative if you could open a `issue` on your specific problems using ssh with gpg-agent.
> 

Well that is interesting. I am using ArchLinux with the latest GnuPG version `2.4.5-4`.

Noticing that the directory `${RUNUSER}/gnupg/` contains `S.gpg-agent.ssh`, I emptied the directory and re-logged into the user. Only the jumbled string directory socket thing was back.

The way I start the `gpg-agent` is as follows:
```
gpgconf --launch gpg-agent
gpgconf --reload
```

In the process table, the gpg-agent is launched with the follow arguments:
```
gpg-agent --homedir <dir> --use-standard-socket --daemon
```

My config sets the `$SSH_AUTH_SOCK` variable before running the start commands, however `--use-standard-socket` is passed to the agent.

There could be a chance that we should allow both directory structures.

ayham-1:
Okay so after looking at the code of GnuPG, I think I found what is happening.

[gpg/gnupg/common/homedir.c#L1342](https://github.com/gpg/gnupg/blob/91532dc3f40599b74fff489cb021c0cd3344ae60/common/homedir.c#L1342)

Apparently when using non default home directory for GnuPG, which is what is happening on my setup, GnuPG hashes the homedir path and appends it to `/run/user/1000/gnupg/`. 

@glitsj16 could you confirm that you are using default GnuPG home directory?

If I am reading GnuPG's code correctly, this means that Firejail should have the requested directory from this PR and the original.

etc/profile-m-z/ssh.profile
ayham-1: Understood, I allowed for both conditions in the latest commit. And wrote an explanation as a comment.

glitsj16:
> @glitsj16 could you confirm that you are using default GnuPG home directory?

Yes, I am indeed using the default ${HOME}/.gnupg on my machines.

I assume you have good reasons to `divert` from this default location. Just `be aware of the implications` of doing so.

By default, Firejail protects the default GnuPG path by blacklisting/making it read-only in `disable-common.inc`. In profiles that need access to it, these options are undone and additional whitelisting and/or mkdir options (can) get added. You'd need to `replicate` all this using your non-default GnuPG path in a `disable-common.local` for the sandbox logic to work as expected. Obviously additional `foo.local` overrides need to be created using that path for _all_ programs you run via firejail that reference GnuPG. Quite a bit of work, not easily maintainable and error-prone IMO.

> The way I start the gpg-agent is as follows:
gpgconf --launch gpg-agent
gpgconf --reload

As you can see in the file list on the Arch Linux [gnupg](https://archlinux.org/packages/core/x86_64/gnupg/) page, `systemd socket activation` is enabled by default (see the usr/lib/systemd/user/sockets.target.wants/foo entrees). There's more info on this [here](https://wiki.archlinux.org/title/GnuPG#gpg-agent). I'm assuming you have overrides for these to sandbox them via firejail? If not, please provide more context on your setup :-)

> If I am reading GnuPG's code correctly, this means that Firejail should have the requested directory from this PR and the original.

Correct.

glitsj16 (approved):
LGTM

kmk3 (approved):
Hello, sorry for the delay.

> Change allowed socket file to the correct file returned from running:
> ```
> gpgconf --list-dirs agent-ssh-socket
> ```
> which returns:
> ```
> /run/user/1000/gnupg/<jumbled string>/S.gpg-agent.ssh
> ```
> 
> Ofcourse this does not run the command, but uses the format
> `${RUNUSER}/gnupg/*/S.gpg-agent.ssh`.
> 
> fixes issues relating to usage of ssh with gpg-agent.

Note that the conventional way to mark a line as code in the commit message is
to it indent 4 spaces; triple backticks are a markdown thing.  Example:

```
Change allowed socket file to the correct file returned from running:

    gpgconf --list-dirs agent-ssh-socket

which returns:

    /run/user/1000/gnupg/<jumbled string>/S.gpg-agent.ssh

Ofcourse this does not run the command, but uses the format
`${RUNUSER}/gnupg/*/S.gpg-agent.ssh`.

fixes issues relating to usage of ssh with gpg-agent.
```

I edited the comment and added more details to the commit message.

Feel free to edit it further.

Otherwise, LGTM.

glitsj16 (approved):

PR #6420, opened 2024-07-26
profiles: refactor dbus debugger profiles

glitsj16:
There are a lot of common options in the `d-feet` and `d-spy` profiles.

Create a new common include file and refactor the existing profiles as
redirects.

Relates to #2492 #6328.

### DIFF
--- a/etc/profile-a-l/d-feet.profile
+++ b/etc/profile-a-l/d-feet.profile
@@ -12,45 +12,16 @@ noblacklist ${HOME}/.config/d-feet
 include allow-python2.inc
 include allow-python3.inc
 
-include disable-common.inc
-include disable-devel.inc
-include disable-exec.inc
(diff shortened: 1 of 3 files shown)


kmk3 (commented):
Note that not all options in the new profile (like `tracelog`) are present in
the current ones.

Are you sure that both profiles work with these changes?

Other than that LGTM.

glitsj16:
@kmk3

Yes, I've tested both with the additional `tracelog`. Thanks for the review.

Side note: This refactoring should make it easier to add similar programs. I'm testing a TUI-based dbus debugger at the moment and will add a new profile for that shortly. Just wanted to keep the PR's separate, do this first and open a follow-up later. Trying to improve my git skills ;)

kmk3 (approved):

PR #6425, opened 2024-07-30
man: firejail.1.in: fix whitelist example

glitsj16:


kmk3 (approved):
LGTM.

Note that the quoting works in both cases, but the changes do make them more
consistent with the other examples.

glitsj16:
@kmk3

Thanks for also treating the blacklist example the same way.

> Note that the quoting works in both cases, but the changes do make them more
consistent with the other examples.

Indeed. There was some confusion about the quoting reported on IRC, which was the motivation behind my PR.

Regards

PR #6429, opened 2024-08-01
profiles: blacklist sway IPC socket

FelixPehla:
Much like the i3 IPC socket (#6361), the sway IPC socket also allows arbitrary code execution via the `exec` subcommand (i.e. `swaymsg exec 'echo test > ~/test'`) . Access should only be permitted to sway itself by default.

The location of the IPC socket is set in [sway/ipc-server.c](https://github.com/swaywm/sway/blob/7e74a4914261cf32c45017521960adf7ff6dac8f/sway/ipc-server.c#L126):

```
	const char *dir = getenv("XDG_RUNTIME_DIR");
	if (!dir) {
		dir = "/tmp";
	}
	if (path_size <= snprintf(ipc_sockaddr->sun_path, path_size,
			"%s/sway-ipc.%u.%i.sock", dir, getuid(), getpid())) {
		sway_abort("Socket path won't fit into ipc_sockaddr->sun_path");
	}
```

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -171,6 +171,10 @@ blacklist ${RUNUSER}/gsconnect
 blacklist ${RUNUSER}/i3/ipc-socket.*
 blacklist /tmp/i3-*/ipc-socket.*
 
+# sway IPC socket (allows arbitrary shell script execution)
+blacklist ${RUNUSER}/sway-ipc.*
+blacklist /tmp/sway-ipc.*
+
 # systemd
 blacklist ${HOME}/.config/systemd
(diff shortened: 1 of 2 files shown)


rusty-snake (approved):

FelixPehla:
Realized that I used slightly different paths (`${RUNUSER}/sway-ipc.*` vs. `${RUNUSER}/sway-ipc.*.sock`), should be consistent now.

rusty-snake (approved):

glitsj16 (approved):
Nice to have this in.

kmk3 (approved):
Nicely done.

PR #6431, opened 2024-08-04
feature: fshaper.sh: support tc on NixOS

### LINKED ISSUE #6426: Traffic shaping does not work with NixOS
### Description

It seems that `fshaper.sh` hardcodes the path to `tc`, which causes setting bandwidth to fail on NixOS given that it has neither `/sbin` nor `/usr/sbin`.  I think this could resolved fairly easily by just letting users set a variable with the path to `tc` or something like that.  Alternatively, you could just allow configuring the path when building the project.

### Steps to Reproduce

1. Use NixOS.
2. `firejail --noprofile --name=blah --net=eth0`
3. `firejail --bandwidth=blah set eth0 1 1`

### Expected behavior

The bandwidth to be set properly.

### Actual behavior

An error saying that `tc` could not be found.

No difference since this is an issue with the script's paths.

### Additional context

This is basically the same issue as #3620.

### Environment

- NixOS 24.05
- Firejail version 0.9.72

### Checklist

- [x] The issue is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).

### Log

<details>
<summary>Output of <code>LC_ALL=C firejail --bandwidth=blah set enp5s0 1 1</code></summary>
<p>

```
Switching to pid 5297, the first child process inside the sandbox
Error: traffic control utility (tc) not found
```

</p>
</details>

<details>
<summary>Output of <code>LC_ALL=C firejail --debug --bandwidth=blah set enp5s0 1 1</code></summary>
<p>

```
Switching to pid 5297, the first child process inside the sandbox
sbox exec: /bin/sh -c /nix/store/w2wdpq3m7qlhg13pwpwq0g80jlqcvwn7-firejail-0.9.72/lib/firejail/fshaper.sh --set eth0-5296 1 1 
Set caps filter 3000
Error: traffic control utility (tc) not found
```

</p>
</details>

glitsj16:
Fixes #6426.

### DIFF
--- a/src/fshaper/fshaper.sh
+++ b/src/fshaper/fshaper.sh
@@ -3,13 +3,9 @@
 # Copyright (C) 2014-2024 Firejail Authors
 # License GPL v2
 
-TCFILE=""
-if [ -x "/usr/sbin/tc" ]; then
-	TCFILE="/usr/sbin/tc"
-elif [ -x "/sbin/tc" ]; then
-	TCFILE="/sbin/tc";
-else
-	echo "Error: traffic control utility (tc) not found";
+TCFILE="$(PATH=/usr/sbin:/sbin:/run/current-system/sw/bin command -v tc)"
+if [ -z "$TCFILE" ]; then
+	echo "Error: traffic control utility (tc) not found"
 	exit 1
 fi
 


src/fshaper/fshaper.sh
```diff
@@ -4,12 +4,10 @@
 # License GPL v2
 
 TCFILE=""
-if [ -x "/usr/sbin/tc" ]; then
-	TCFILE="/usr/sbin/tc"
-elif [ -x "/sbin/tc" ]; then
-	TCFILE="/sbin/tc";
+if [ -x "$(PATH=/usr/sbin:/sbin:/run/current-system/sw/bin command -v tc)" ]; then
```
rusty-snake, line 7: Use `if command >/dev/null` if you still can set PATH.

src/fshaper/fshaper.sh
```diff
@@ -4,12 +4,10 @@
...
+	TCFILE="$(PATH=/usr/sbin:/sbin:/run/current-system/sw/bin command -v tc)"
```
rusty-snake, line 8: Double call to command.

Assigne it to TCFILE, then check TCFILE

glitsj16: I've reworked the test. Thanks for reviewing!

src/fshaper/fshaper.sh
```diff
@@ -3,10 +3,8 @@
 # Copyright (C) 2014-2024 Firejail Authors
 # License GPL v2
 
-TCFILE=""
-if [ -x "$(PATH=/usr/sbin:/sbin:/run/current-system/sw/bin command -v tc)" ]; then
-	TCFILE="$(PATH=/usr/sbin:/sbin:/run/current-system/sw/bin command -v tc)"
-else
+TCFILE="$(PATH=/usr/sbin:/sbin:/run/current-system/sw/bin command -v tc >/dev/null)"
```
rusty-snake, line 6:
```suggestion
TCFILE="$(PATH=/usr/sbin:/sbin:/run/current-system/sw/bin command -v tc)"
```

src/fshaper/fshaper.sh
```diff
@@ -3,10 +3,8 @@
...
+if [ ! -x "$TCFILE" ]; then
```
rusty-snake, line 7:
```suggestion
if [ -z "$TCFILE" ]; then
```

glitsj16: Not my best argh. Should be fixed now, thanks!

kmk3 (commented):
Using `command` with a specific `PATH` is nice.

It looks like this script is executed as root (`SBOX_ROOT`):

* <https://github.com/netblue30/firejail/blob/03ea436e1cb35fdcba8fb6588de209d0a10174e9/src/firejail/bandwidth.c#L365-L394>

I'm not familiar with NixOS; is `/run/current-system/sw/bin` a path that can
only be created/modified by root on NixOS?

If so, LGTM.

glitsj16:
> I'm not familiar with NixOS; is /run/current-system/sw/bin a path that can
only be created/modified by root on NixOS?

Had to dig deep to find the `/run/current-system/sw/bin` path to properly fix this script on NixOS. I'm not at all familiar with it either. That being said, [here](https://nixos.wiki/index.php?title=Ubuntu_vs._NixOS&useskin=vector)'s some details from their wiki. I additionally asked several users on #nixos IRC and they all confirmed this works.

HTH

kmk3:
> > I'm not familiar with NixOS; is /run/current-system/sw/bin a path that can
> > only be created/modified by root on NixOS?
> 
> Had to dig deep to find the `/run/current-system/sw/bin` path to properly fix
> this script on NixOS.

That is a pretty good finding indeed.

> I'm not at all familiar with it either. That being said,
> [here](https://nixos.wiki/index.php?title=Ubuntu_vs._NixOS&useskin=vector)'s
> some details from their wiki. I additionally asked several users on #nixos
> IRC and they all confirmed this works.

To be clear, I don't doubt that this works; the concern is that if the path can
be modified by normal users, then they could potentially gain root simply by
putting a malicious `tc` executable in there.

But if it's just a normal system-managed path (and there are no user overrides
involving containers, etc) then it should be fine.

Issue #6400, opened 2024-07-07
private-etc groups meta-issue


kmk3:
Currently #5610 is used for the discussion of the new private-etc groups and
associated refactoring.

Though unlike issues, when referencing a discussion a timeline event for the
reference is not created, so links between discussions and issues/PRs/commits
have to be created manually in comments (see
https://github.com/orgs/community/discussions/2971).

So create this new issue to use as the main issue to be referenced in this case
for the sake of automatic links.

Relates to:

* https://github.com/netblue30/firejail/discussions/5597#discussioncomment-4739664
* #5607
* #5608
* #5609
* #5610
* #5629
* #5638
* #5641
* #5642
* #5643
* #5645

(The list above was copied from
https://github.com/netblue30/firejail/discussions/5610#discussion-4780332)

PR #6435, opened 2024-08-17
profiles: firefox-common: fix private-etc in includers

kmk3:
It was enabled in firefox-common.inc on commit 34d004892 ("private-etc:
corss-distro test for curl, gimp, inkscape, firefox, warzone2100",
2023-01-28), but not in the profiles that include it.

Enable it in the including profiles as well.

Note: This was already done for firefox.profile on commit 76249284f
("firefox: fix private-etc firefox", 2023-06-02) / PR #5844.

Relates to #6400.


glitsj16 (approved):
LGTM

PR #6463, opened 2024-09-04
New profile: singularity (Endgame: Singularity)

NetSysFire:
Tested it myself of course. This is a small, python-based game which needs no network access and is generally quite frugal in terms of resources.

I would deem it fine if this is added to firecfg, as there are little to no configurable options and will work out of the box.

### DIFF
--- a/etc/profile-a-l/singularity.profile
+++ b/etc/profile-a-l/singularity.profile
@@ -1,0 +1,47 @@
+# Firejail profile for singularity
+# Description: Simulation game about playing as an artificial intelligence
+# This file is overwritten after every install/update
+# Persistent local customizations
+include singularity.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/singularity
+noblacklist ${HOME}/.local/share/singularity
+
+# Allow python (blacklisted by disable-interpreters.inc)
+include allow-python3.inc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+
+mkdir ${HOME}/.config/singularity
+mkdir ${HOME}/.local/share/singularity
+whitelist ${HOME}/.config/singularity
+whitelist ${HOME}/.local/share/singularity
+include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+net none
+#no3d
+nodvd
+noroot
+#nosound
+nou2f
+
+disable-mnt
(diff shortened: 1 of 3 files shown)


etc/profile-a-l/singularity.profile
```diff
@@ -0,0 +1,38 @@
+include singularity.local
+include globals.local
```
rusty-snake, line 2: Add comments as everywhere.

etc/profile-a-l/singularity.profile
```diff
@@ -0,0 +1,38 @@
...
+
+noblacklist ${HOME}/.config/singularity
+noblacklist ${HOME}/.local/share/singularity
```
rusty-snake, line 5: Add blacklist to disable-programs.inc

etc/profile-a-l/singularity.profile
```diff
@@ -0,0 +1,38 @@
...
+include whitelist-common.inc
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+net none
+caps.drop all
+disable-mnt
```
rusty-snake, line 29: Sort disable-mnt down as per template

etc/profile-a-l/singularity.profile
```diff
@@ -0,0 +1,38 @@
...
+nodvd
+noroot
+nou2f
```
rusty-snake, line 32:
```suggestion
nou2f

```

etc/profile-a-l/singularity.profile
NetSysFire: sort.py did not point this out.

rusty-snake: sort.py does inline sorting of arguments too `private-*`, `seccomp`, ... but no multiline sorting across the file.

kmk3:
@NetSysFire

Disabled more things in the profile.

Does it still work with the changes?

NetSysFire:
This is a small python game that does not need any networking or even external access. I already thought what I did was already a bit overkill as my initial profile for this was even smaller. There is simply no attack surface to justify additional effort in my opinion.

kmk3:
> This is a small python game that does not need any networking or even
> external access. I already thought what I did was already a bit overkill as
> my initial profile for this was even smaller.

This idea seems backwards.

Why give a program more access than it needs, especially if it is unlikely to
cause problems?

> There is simply no attack surface to justify additional effort in my opinion.

What would be the additional effort?  I mostly just added the rest of the
standard entries from profile.template.

Again, do the changes cause any breakage?

NetSysFire:
I do not have any more motivation to test the new directives. I'd prefer to stick to the things I already tested.

netblue30:
All in, thanks!

PR #924, opened 2016-11-19
Many new profiles

valoq:
New profiles:

amarok,ark,atool,bleachbit,brasero,dolphin,dragon,elinks,enchant,exiftool,file-roller,gedit
gjs,gnome-books,gnome-documents,gnome-maps,gnome-music,gnome-photos,gnome-weather,goobox
gpa,gpg,gpg-agent,highlight,img2txt,k3b,kate,lynx,mediainfo,nautilus,odt2txt,pdftotext
simple-scan,skanlite,ssh-agent,transmission-cli,tracker,transmission-show,w3m,xfburn,xpra

added nogroups and shell none to default.profile

PR #6477, opened 2024-09-14
profiles: firecfg: disable text editors

### LINKED ISSUE #6002: text-editors: cannot access all text files in the user home
Sometimes, packages require edits, sometimes those are larger than I'd prefer to use with vim.

I question the use in even firejailing geany at this point, though...

kmk3:
Disable common general-purpose text editors.

They are likely to be the default OS text editor and users may want to
use them for editing most/all files, which could include common
sensitive files such as ~/.bashrc and profiles in ~/.config/firejail.

Fixes #6002.

Relates to #924 #941 #1154.

Reported-by: @ilikenwf

### DIFF
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -320,9 +320,9 @@ gcalccmd
 gcloud
 gconf-editor
 gdu
-geany
+#geany # text editor (see #6002)
 #geary # webkit2gtk-4.x requires bwrap (see #3647)
-gedit
+#gedit # text editor (see #6002)
 geekbench
 geeqie
 geki2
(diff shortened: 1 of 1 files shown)


kmk3:
Let me know if I missed any text editor.

src/firecfg/firecfg.config
```diff
@@ -320,9 +320,9 @@ gcalccmd
 gcloud
 gconf-editor
 gdu
-geany
+#geany # text editor
```
rusty-snake, line 323: FTR: More IDE than just editor.

rusty-snake (approved):

src/firecfg/firecfg.config
kmk3: > FTR: More IDE than just editor.

Sure, but it's probably lightweight enough to be used as the default text
editor.

Similarly to kate.

PR #6498, opened 2024-10-01
profiles: signal-desktop - Add access to D-Bus freedesktop.org secret API

corsac-s:
Signal recently started storing a local key in the freedesktop.org secret API so allow access in the profile

Also see https://github.com/signalapp/Signal-Desktop/issues/6970


rusty-snake (approved):

netblue30:
Merged, thanks for the patch!

PR #6525, opened 2024-10-29
docs: use GitHub issues as the bug reporting address

tredondo:


rusty-snake (changes requested):
`configure` should not be edited directly, instead `configure.ac` should be edited and `configure` regenerated.

kmk3 (commented):
```diff
-# Copyright (C) 1992-1996, 1998-2017, 2020-2021 Free Software Foundation,
+# Copyright (C) various years Free Software Foundation,
```

It would not be helpful (to say the least) to remove information from a
copyright statement (especially for code you didn't write), so this change was
undone.

Additionally, note that this line is from another project, so it would
potentially also be a license violation.

kmk3 (commented):
It looks like this parameter is intended to be an email address, probably to
allow people to contribute patches without needing an account, though I'm not
sure how often that happens.

We do not have a mailing list and the same email address is already listed on
README for reporting security vulnerabilities, so unless this email is used for
bug reports/patches, I suppose that it would make sense to reference just the
issue tracker.

@netblue30 Do you get bug reports/patches by email?

netblue30:
I'm merging it as is, thanks @tredondo. I'll generate a new configure script.

Issue #6121, opened 2023-12-10
Virtual network fails with firejail enabled for dnsmasq


marek22k:
### Description

I cannot activate the virtual network when firejail is activated.

_Steps to reproduce the behavior_

1. Install Arch Linux
2. Install qemu/kvm
3. Run `sudo virsh net-start default`

### Expected behavior

The network starts.

### Actual behavior

The network does not start.

### Behavior without a profile

Since dnsmasq is called by libvirt, it is difficult to do this manually. However, running it after `firecfg --clean` works.

### Additional context

```
$ sudo virsh net-start default
error: Failed to start network default
error: internal error: Child process (VIR_BRIDGE_NAME=virbr0 /usr/local/bin/dnsmasq --conf-file=/var/lib/libvirt/dnsmasq/default.conf --leasefile-ro --dhcp-script=/usr/lib/libvirt/libvirt_leaseshelper) unexpected exit status 1: Error: PATH environment variable not set


$ sudo firecfg --clean
Removing all firejail symlinks:
   alacarte removed
   atril removed
   atril-previewer removed
   atril-thumbnailer removed
   calibre removed
   chromium removed
   claws-mail removed
   codium removed
   com.github.tchx84.Flatseal removed
   conplay removed
   cvlc removed
   dig removed
   display removed
   dnsmasq removed
   drill removed
   ebook-convert removed
   ebook-edit removed
   ebook-meta removed
   ebook-polish removed
   ebook-viewer removed
   enchant-2 removed
   enchant-lsmod-2 removed
   ffplay removed
   ffprobe removed
   filezilla removed
   ftp removed
   gajim removed
   gapplication removed
   geany removed
   gimp removed
   gimp-2.10 removed
   gpa removed
   hexchat removed
   host removed
   img2txt removed
   inkscape removed
   inkview removed
   keepassxc removed
   keepassxc-cli removed
   keepassxc-proxy removed
   libreoffice removed
   librewolf removed
   lobase removed
   localc removed
   lodraw removed
   loffice removed
   lofromtemplate removed
   loimpress removed
   lomath removed
   loweb removed
   lowriter removed
   man removed
   mate-color-select removed
   meld removed
   mousepad removed
   mpg123 removed
   mpg123-id3dump removed
   mpg123-strip removed
   mpv removed
   nslookup removed
   out123 removed
   parole removed
   patch removed
   pdftotext removed
   ping removed
   pluma removed
   qt-faststart removed
   qtox removed
[...]

PR #6533, opened 2024-11-07
profiles: firecfg.config: disable dnsmasq

kmk3:
There are multiple reports in #6121 that dnsmasq does not work when
called by libvirt:

    $ sudo virsh net-start default
    error: Failed to start network default
    error: internal error: Child process (VIR_BRIDGE_NAME=virbr0 /usr/local/bin/dnsmasq [...]) unexpected exit status 1: Error: PATH environment variable not set

Also, note that this is a server program, so it might be better to
disable it by default anyway.

Reported-by: @marek22k

### DIFF
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -219,7 +219,7 @@ display
 display-im6.q16
 dnox
 dnscrypt-proxy
-dnsmasq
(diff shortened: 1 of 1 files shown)


rusty-snake (approved):
LGTM

I disabled it on my system long time ago (Fedora).

Issue #295, opened 2016-02-14
Syncthing


Tanath:
`$ firejail syncthing-gtk`
`Reading profile /etc/firejail/syncthing-gtk.profile`
`Reading profile /etc/firejail/disable-mgmt.inc`
`Reading profile /etc/firejail/disable-secret.inc`
`Error: cannot open profile file /etc/firejail/disable-common.inc syncthing`

`$ file /etc/firejail/disable-common.inc`
`/etc/firejail/disable-common.inc: ASCII text`

What's going on here, and how do I troubleshoot it?

genodeftest:
Do you have permissions to open this file? What's the output of `ls -Al /etc/firejail/disable-common.inc`?

Tanath:
Yes, I've looked at it in vim.
`-rw-r--r-- 1 root root 3659 Feb 13 21:50 /etc/firejail/disable-common.inc`

netblue30:
I think you run into some feature I deprecated a few releases back. Pint here your /etc/firejail/syncthing-gtk.profile, I'll need to modify it.

Tanath:
`# syncthing-gtk profile  

include /etc/firejail/disable-mgmt.inc  

include /etc/firejail/disable-secret.inc  

include /etc/firejail/disable-common.inc syncthing  

caps  

seccomp  

noroot`

netblue30:
Try this:
`````
# syncthing-gtk profile
include /etc/firejail/disable-mgmt.inc
include /etc/firejail/disable-secret.inc
include /etc/firejail/disable-common.inc
caps
seccomp
noroot
`````

Tanath:
Runs now, but isn't that basically just using generic profile then?

netblue30:
Yes, you don't really need a special profile, the default one will do it.

PR #6536, opened 2024-11-08
New profile: syncthing

gcb:
https://docs.syncthing.net/

tested with v1.28.0

### DIFF
--- a/etc/profile-m-z/syncthing.profile
+++ b/etc/profile-m-z/syncthing.profile
@@ -1,47 +1,116 @@
 # Firejail profile for syncthing
+# Description: File synchronization using public networks
 # This file is overwritten after every install/update
+
+##quiet
 # Persistent local customizations
 include syncthing.local
 # Persistent global definitions
 include globals.local
 
-# NOTE: this assumes a ~/Sync directory to be shared.
+noblacklist ${HOME}/.local/state/syncthing
+# NOTE: this assumes a ~/Sync directory to be shared by default. Maybe leave a requirement to edit local file to set dirs?
+noblacklist ${HOME}/Sync
 
+# NOTE: will cause WARNING: Failed to lower process
+#       priority: set I/O priority: operation not permitted
+#       So, we try to preemptively set it here:
+nice 2
 
-# Allow python3 (blacklisted by disable-interpreters.inc)
-include allow-python3.inc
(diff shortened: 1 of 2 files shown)


gcb:
previously https://github.com/netblue30/firejail/issues/295

src/firecfg/firecfg.config
```diff
@@ -851,6 +851,7 @@ surf
 sushi
 swell-foop
 sylpheed
+syncthing
```
kmk3, line 854:
```suggestion
#syncthing # server
```

See:

* #6533

etc/profile-m-z/syncthing.profile
```diff
@@ -0,0 +1,116 @@
+# Firejail profile for syncthing
+# Description: File synchronization using public networks
+# This file is overwritten after every install/update
+
+##quiet
```
kmk3, line 5:
```suggestion
```

Remove `##` lines and other noise like in #6549.

Same for the rest of the file.

etc/profile-m-z/syncthing.profile
```diff
@@ -0,0 +1,116 @@
...
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+#include disable-write-mnt.inc # we set disable-mnt
+#include disable-x11.inc # this causes an error...
+include disable-xdg.inc
+
+
+mkdir ${HOME}/.local/state/syncthing
+whitelist ${HOME}/.local/state/syncthing
```
kmk3, line 45: Is this program intended to be executed at the user or system level?

Does the `syncthing` executable act as the client, server or both?

kmk3 (changes requested):

etc/profile-m-z/syncthing.profile
gcb: Answer is both for both questions.

Ideally it is a user service. Most commonly it is system service. But upstream likes everyone to just run the program in a terminal and then minimize it.

Then it is a peer in a network which will connect to all sorts of things (announce servers, relay servers, other peers), while also exposing servers for LAN auto-discovery and to receive other peer's connections.

netblue30:
Let's go with it, thanks @gcb. I will remove the program from src/firecfg/firecfg.config for now (see kmk3 discussion above).

PR #6542, opened 2024-11-11
profiles: wget: allow ~/.local/share/wget

celenityy:
wget appears to require access to this directory for HSTS & HPKP.

Without access to this directory, I get the following error when running wget:

```sh
Failed to read HSTS data
Failed to read HPKP data
Failed to write HSTS file
```

This fixes it.

### DIFF
--- a/etc/profile-m-z/wget.profile
+++ b/etc/profile-m-z/wget.profile
@@ -7,6 +7,7 @@ include wget.local
 # Persistent global definitions
 include globals.local
 
+noblacklist ${HOME}/.local/share/wget
 noblacklist ${HOME}/.netrc
 noblacklist ${HOME}/.wget-hsts
 noblacklist ${HOME}/.wgetrc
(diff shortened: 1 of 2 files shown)


rusty-snake (changes requested):
Distributions started to replace wget with wget2 (I.e. `wget` and `wget2` are the same binary where one of them is a symlink to the other).

We should reorganize our profile for this fact.

Can you also add `.config/wget` and remove the then doubled noblacklists from wget2.profile.

kmk3:
@rusty-snake

> Distributions started to replace wget with wget2 (I.e. `wget` and `wget2` are
> the same binary where one of them is a symlink to the other).

Good catch.

> We should reorganize our profile for this fact.
> 
> Can you also add `.config/wget` and remove the then doubled noblacklists from
> wget2.profile.

Since this is a self-contained fix, how about merging this as is and then I
open a PR to refactor them after this PR?

rusty-snake (approved):
> Since this is a self-contained fix, how about merging this as is and then I
open a PR to refactor them after this PR?

If you do, I approve.

PR #6549, opened 2024-11-18
New profile: irssi

gcb:
cross irssi off the wanted list https://github.com/netblue30/firejail/issues/1139

I think it will work, but I don't have how to test DCC, because well, like the entire universe i'm behind layers of NAT, CGNAT, etc... and nobody should use DCC anyway.


etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,67 @@
...
+protocol unix,inet,inet6
+seccomp
+# NOTE: comments here are things that can be improved, if you can spare the time.
+##seccomp.drop SYSCALLS (see syscalls.txt)
+##seccomp-error-action log (only for debugging seccomp issues)
+#shell none
+#tracelog
+disable-mnt
+##private-opt NAME
+#private-tmp
```
rusty-snake, line 50: `private-dev`? `private-cache`?

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,67 @@
...
+##writable-run-user
+##writable-var
+##writable-var-log
```
rusty-snake, line 53:
```suggestion
```

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,67 @@
+# Firejail profile for irssi
```
rusty-snake, line 1: firecfg.config?

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,67 @@
...
+noblacklist ${HOME}/.irssi/
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
```
rusty-snake, line 19:
```suggestion
```

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,67 @@
...
+# Description: IRC client
+# This file is overwritten after every install/update
+# Persistent local customizations
+include irssi.local
+
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.irssi/
+
```
rusty-snake, line 11:
```suggestion

blacklist ${RUNUSER}/wayland-*

```

rusty-snake (dismissed):

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,67 @@
+# Firejail profile for irssi
+# Description: IRC client
+# This file is overwritten after every install/update
+# Persistent local customizations
+include irssi.local
+
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.irssi/
```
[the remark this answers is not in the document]
gcb: habit of always adding this when allowing access later :) 

will drop

etc/profile-a-l/irssi.profile
gcb: lost those when moving my profile to the template. thanks.

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,67 @@
...
+include disable-shell.inc
+include disable-write-mnt.inc
+include disable-X11.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.irssi/
+whitelist ${HOME}/.irssi/
+
+#include whitelist-usr-share-common.inc
+#include whitelist-var-common.inc
```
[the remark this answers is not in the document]
gcb: yep. for some reason i was assuming they were subsets of the whitelist-common if you needed only those root dirs. :facepalm:

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,54 @@
...
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-X11.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.irssi/
+whitelist ${HOME}/.irssi/
```
kmk3, line 22: Remove trailing `/`, add `noblacklist`, add to disable-programs.inc.

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,54 @@
...
+
```
kmk3, line 23:
```suggestion
```

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,54 @@
...
+no3d
+nodvd
+nogroups
+nonewprivs
+noprinters
+noroot
+nosound
+notv
+nou2f
+novideo
```
kmk3, line 39: Add and sort:

```
noinput
notpm
```

Note: `notpm` is in firejail-git.  When creating a new profile it's recommended
to use the latest development version to ensure that an up-to-date
profile.template is used.

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,54 @@
...
+disable-mnt
+private-cache
+private-dev
+private-tmp
+
+dbus-user none
+dbus-system none
+
+# NOTE: almost sure irssi uses perl, but all seems to work without allowing it.
+
```
kmk3, line 53:
```suggestion
```

kmk3 (changes requested):

etc/profile-a-l/irssi.profile
gcb: I did use the template 4mo ago, when notpm was added to the template :)

gcb: hum... i thought i had issues with noinput but now i tested and can't find why.  adding it back then

etc/profile-a-l/irssi.profile
```diff
@@ -0,0 +1,58 @@
...
+notv
+nou2f
+novideo
+protocol unix,inet,inet6
+seccomp
+seccomp.block-secondary
+
+disable-mnt
+private-cache
+private-dev
```
kmk3, line 52:
```suggestion
private-dev
private-etc @network,@tls-ca
```

Does it work with `private-etc`?

gcb: I had installed -git to test with the `notpm` line, but now i'm back to my distro version and did not test `private-etc`. I will test with -git later on and report. But I don't see why not.

gcb: instead I ran it now in debugger, also watching `lsof` for the pid, and nothing on etc is ever touched. I think it is OK.

kmk3: > instead I ran it now in debugger, also watching `lsof` for the pid, and
> nothing on etc is ever touched. I think it is OK.

Can you test connecting to a server in firejail-git to make sure?

It seems strange that nothing on /etc would be accessed (such as TLS-related
paths).

What is the output of trace.txt in the following?

```sh
firejail --trace=trace.txt --profile=irssi /usr/bin/irssi
```

netblue30:
Thanks for the patch, merging in!

netblue30:
Added and commented out (see kmk3 above):
`````
# private-etc @network,@tls-ca
`````
@gcb give it a try, maybe is working. Thanks.

etc/profile-a-l/irssi.profile
kmk3: > My bad. My irssi profile is very uncommon. Retested this with a clean irssi
> profile, and indeed, it does break dns as expected.
> 
> with private-etc: Irssi: Unable to connect server irc.freenode.net port 6667
> [Temporary failure in name resolution]
> 
> without: resolves names fine.

Strange, what program do you use to manage dns?

On what distribution name/version?

Also, could you retest with the current irssi.profile in firejail-git (there is
one new change) and post the output of trace.txt?

PR #6571, opened 2024-12-17
Add bubblejail files to `disable-common.inc`

lucasmz-dev:
I noticed I was able to access bubblejail files when running `firejail bash`, block this in this permissive sceneario to be sure


rusty-snake (approved):

netblue30:
Merged, thanks!

PR #6574, opened 2024-12-19
feature: add aarch64 syscalls

### LINKED ISSUE #5821: No syscall table provided in firejail on ARM64 platform
### Description

No syscall listed on ARM64 platform

### Steps to Reproduce

When run command 'firejail --debug-syscalls' in order to get syscall list on our bcm-2xxx-rpi4 platform which is arm64 arch, there is nothing printed, but if run the same command on x86 or arm32 platforms the syscall can be listed properly.
And there is same problem in firejail command about seccomp functions(for example command ' firejail --seccomp.drop=execve') on our side. 
I check below code in "src/lib/syscall.c", I find there is only variable 'syslist' defined for arch x86_64, i386 and arm, there is no such variable defined for arm64 arch.

_**_static const SyscallEntry syslist[] = {
#if defined(__x86_64__)
// code generated using
// awk '/__NR_/ { print "{ \"" gensub("__NR_", "", "g", $2) "\", " $3 " },"; }' < /usr/include/x86_64-linux-gnu/asm/unistd_64.h
#include "../include/syscall_x86_64.h"
#elif defined(__i386__)
// awk '/__NR_/ { print "{ \"" gensub("__NR_", "", "g", $2) "\", " $3 " },"; }' < /usr/include/x86_64-linux-gnu/asm/unistd_32.h
#include "../include/syscall_i386.h"
#elif defined(__arm__)
#include "../include/syscall_armeabi.h"
#else
#warning "Please submit a syscall table for your architecture"
#endif
};_**_

And there are only syscall table files about x86-64, i386 and arm32 provided in src/include/ as src/include/syscall_armeabi.h  src/include/syscall_i386.h  src/include/syscall_x86_64.h.

So my question is why firejail didn't provide syscall table for arch ARM64? Is there any risk when provide syscall table for arch arm64 in firejail?
When will firejail provide syscall table for arch ARM64?
If I provide the syscall table for arm64 arch privately, will it cause any potential risk for firejail? 

My current firejail version is 0.9.72, but it looks there should be same issue in lasted version.

AndreySV:
syscall_aarch64.h was generated from 6.12 kernel headers using the
following commands:

    awk '/__NR_/ { print "{ \"" gensub("__NR_", "", "g", $2) "\", " $3 " },"; }' \
      < unistd_64.h | grep -v 'syscalls'

Closes #5821.


kmk3 (approved):
Nicely done, thanks.

Misc: For future PRs, please use 4 leading spaces (instead of backticks) to
indent code blocks in the commit message to make them easier to spot in the
terminal.

Issue #2820, opened 2019-07-02
fcopy size limit is not adjustable


maxhaslbeck:
hi,

I add an folder to private-home that happens to be bigger than 500MB in

```
firejail --noprofile --private-home=MYFOLDER MYCOMMAND
```

i get the error message

```
Error fcopy: size limit of 500MB reached
```

is there a reason for the size limit in fcopy? can I increase it easily?
are large files in the private-home discouraged?
Is there a simple workaround to get MYFOLDER available within the jail?

cheers,
max

rusty-snake:
limit is set [here](https://github.com/netblue30/firejail/blob/master/src/fcopy/main.c#L30)

> are large files in the private-home discouraged?

Becauce the fiels are copied, there are large performace losing while starting the programm.

> Is there a simple workaround to get MYFOLDER available within the jail?

I recommend to use `whitelist` and `overlayfs-tmpfs` instead of `private-home`, because it is faster.

Vincent43:
Unlike `private-home`, `overlayfs-tmpfs` isn't persistent. More equivalent option is `overlay-named` but we have reports that [it may be broken in recent kernels](https://github.com/netblue30/firejail/issues/2799).

rusty-snake:
> Unlike `private-home`, `overlayfs-tmpfs` isn't persistent.

> --private-home=file,directory
              Build  a  new user home in a temporary filesystem, and copy the files
              and directories in the list in the new home.  **All  modifications  are
              discarded when the sandbox is closed.**

Vincent43:
@rusty-snake you are right, I confused `--private-home=` with `--private=`, the latter one is persistent, sorry.

netblue30:
@maxhaslbeck

Load the latest from mainline and in /etc/firejail/firejail.config set file-copy-limit to whatever you need.
The files get copied in a temporary filesystem in RAM, and if you run out of memory the kernel will start swapping - this is why we put a limit on it.

PR #6579, opened 2024-12-25
bugfix: parse --debug before using it

powerjungle:
The `arg_debug` flag is set after some functions check for it.
https://github.com/netblue30/firejail/blob/f27775c5a810729b7dc8f144f522c5e79ec4ee55/src/firejail/main.c#L1347
An example of a function in this situation is `check_kernel_procs()`.
https://github.com/netblue30/firejail/blob/f27775c5a810729b7dc8f144f522c5e79ec4ee55/src/firejail/main.c#L1129
https://github.com/netblue30/firejail/blob/f27775c5a810729b7dc8f144f522c5e79ec4ee55/src/firejail/no_sandbox.c#L102

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -1110,6 +1110,12 @@ int main(int argc, char **argv, char **envp) {
 	if (check_arg(argc, argv, "--quiet", 1) || (env_quiet && strcmp(env_quiet, "yes") == 0))
 		arg_quiet = 1;
 
+	// process --debug
+	if (check_arg(argc, argv, "--debug", 1)) {
+		arg_debug = 1;
+		arg_quiet = 0;
+	}
+
 	// check if the user is allowed to use firejail
 	init_cfg(argc, argv);
 


src/firejail/main.c
```diff
@@ -1110,6 +1110,13 @@ int main(int argc, char **argv, char **envp) {
 	if (check_arg(argc, argv, "--quiet", 1) || (env_quiet && strcmp(env_quiet, "yes") == 0))
 		arg_quiet = 1;
 
+	// process --debug
+	const char *env_debug = env_get("FIREJAIL_DEBUG");
```
kmk3, line 1114:
```suggestion
```

It looks like the env var would end up overriding a CLI argument (`--quiet`),
so remove this for now.

Also, I'm not sure why, but this variable is currently only set and never read.
It was added on commit ccd01529a ("fix #2820 - adjustable file copy limit;
export FIREJAIL_DEBUG into sbox", 2019-07-04).

Anyway, I have an idea of how to refactor this later to include the env var and
improve the handling of both quiet/debug (it would involve more code, while the
fix in this PR is rather straightforward).

kmk3:
I see that you reworded the commit message (and undid it).

I was trying to improve the grammar with my force-push, but now I'm not sure if
you meant just the `check_kernel_procs()` function or that multiple functions
are affected by this issue.

I noticed that the `preproc_build_firejail_dir_unlocked()` function (and maybe
the other ones below) seem to also be affected:

* <https://github.com/netblue30/firejail/blob/afce4679bf66c9d204ea48d33041fdb3e6801d91/src/firejail/main.c#L1170>
* <https://github.com/netblue30/firejail/blob/afce4679bf66c9d204ea48d33041fdb3e6801d91/src/firejail/main.c#L1223>
* <https://github.com/netblue30/firejail/blob/afce4679bf66c9d204ea48d33041fdb3e6801d91/src/firejail/main.c#L1335>

In which case the current version of the message should be accurate, but if
that's not the case, the message could be simplified to:

```
The `arg_debug` flag is being set after the `check_kernel_procs()` function
checks for it.
```

Anyway, sorry for the confusion.

powerjungle:
> I see that you reworded the commit message (and undid it).

It was by accident, I didn't notice that you force pushed and I tried to fix the message later.

> I was trying to improve the grammar with my force-push, but now I'm not sure if
> you meant just the `check_kernel_procs()` function or that multiple functions
> are affected by this issue.

I just gave it as an example, but yes there are more functions.

> In which case the current version of the message should be accurate

I think the current version is fine.

> Anyway, sorry for the confusion.

Sorry for the confusion too, I rushed through writing the message and I didn't notice how unclear and full with mistakes it was. 😅

src/firejail/main.c
powerjungle: > It looks like the env var would end up overriding a CLI argument (--quiet),
> so remove this for now.

How does it override it in this case?

Also if we remove `env_debug` we also need to remove the `||` statement on the next like that uses `env_debug`.

kmk3: > > It looks like the env var would end up overriding a CLI argument (--quiet),
> > so remove this for now.
> 
> Ah you mean because it would set `arg_quiet` to 0

Yes.

> but if somebody wants to enable debugging, isn't that expected?

The usual convention on Linux etc is that CLI options always override their
corresponding environment variables, probably because the options are usually
easier to type and may have shell completion.  Also, the variables may come
from ~/.bashrc or somewhere not immediately obvious during usage and are
probably less likely to be well-documented compared to CLI options.

Personally I'd find it surprising if the override did not work this way in a
given tool.

> Also if we remove `env_debug` we also need to remove the right side of the
> `||` statement on the next line that uses `env_debug`.

Yes, good catch.

powerjungle: > The usual convention on Linux etc is that CLI options always override their
corresponding environment variables, probably because the options are usually
easier to type and may have shell completion. Also, the variables may come
from ~/.bashrc or somewhere not immediately obvious during usage and are
probably less likely to be well-documented compared to CLI options.

Ah I see, makes sense.

> Yes, good catch.

I'll force push with the new changes and resolve the conversation.

PR #6580, opened 2024-12-26
Improve profiles for anki and foliate. Add new profiles.

amano-kenji:
On gentoo linux, there is /usr/bin/foliate instead of /usr/bin/com.github.johnfactotum.Foliate

### DIFF
--- a/etc/profile-a-l/anki.profile
+++ b/etc/profile-a-l/anki.profile
@@ -17,6 +17,8 @@ noblacklist ${HOME}/.config/mpv
 noblacklist ${HOME}/.local/share/Anki2
 noblacklist ${HOME}/.mplayer
 
+# sh and dbus-send are used by aqt/theme.py to query dark mode through org.freedesktop.portal.Desktop
+include allow-bin-sh.inc
 # Allow lua (blacklisted by disable-interpreters.inc)
 include allow-lua.inc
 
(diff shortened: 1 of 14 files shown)


kmk3 (changes requested):
There are too many unrelated changes in the same PR, which makes review harder.

Split the changes into more PRs (this PR can be the anki one):

* profiles: anki: fix dark mode detection & misc changes
* profiles: foliate: add foliate to private-bin
* New profile: aria2c/aria2rpc
* New profile: hledger/hledger-ui
* New profile: cryptonote-wallet-cli / monero-wallet-cli

Unless there is something in common about some of these, they should probably
also get 1 PR each:

* New profile: buku
* New profile: ncmpcpp
* New profile: nsxiv
* New profile: pyradio
* New profile: tremc

As for a basic overall review:

* The new profiles are missing the description and the rest of the comments
  that go before `quiet` (see profile.template /
  `redirect_alias-profile.template`)
* The new profiles are missing the `# Allow ...` comment that goes before each
  `include allow-` entry (see profile.template)
* Add `notpm` to the new profiles if they do not use TPM devices (which
  probably applies to all of them except for maybe the crypto ones) (see
  profile.template)

Also, add links to the homepage (or source code) of each program in the commit
message and PR.  For example, I have no idea what `buku` and `tremc` are
supposed to be.

PR #6583, opened 2024-12-27
New profile: aria2p/aria2rpc

amano-kenji:
https://github.com/pawamoy/aria2p is Command-line tool and library to interact with an aria2c daemon process with JSON-RPC.

aria2rpc from https://github.com/aria2/aria2 is command line tool for connecting to a remote instance of aria2c

### DIFF
--- a/etc/profile-a-l/aria2p.profile
+++ b/etc/profile-a-l/aria2p.profile
@@ -1,0 +1,65 @@
+# Firejail profile for aria2p
+# Description: Command-line tool and library to interact with an aria2c daemon process with JSON-RPC.
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include aria2p.local
+# Persistent global definitions
+include globals.local
+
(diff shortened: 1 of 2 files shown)


etc/profile-a-l/aria2p.profile
```diff
@@ -0,0 +1,65 @@
...
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+#include disable-write-mnt.inc
+# xclip and xsel require X11.
+#include disable-X11.inc
```
kmk3, line 24: @amano-kenji

Does the program itself use xclip/xsel directly?

Or do you just do something like `aria2p | xclip` on the CLI?

kmk3: > I don't really remember now, but aria2p itself doesn't use xclip or xsel
> directly.

(Continued at #6609)

PR #6585, opened 2024-12-27
New profile: hledger/hledger-ui

amano-kenji:
https://hledger.org/

Relates to #6580.

### DIFF
--- a/etc/profile-a-l/hledger-ui.profile
+++ b/etc/profile-a-l/hledger-ui.profile
@@ -1,0 +1,8 @@
+# Firejail profile for hledger-ui
+# Description: Terminal interface (TUI) for hledger
+# This file is overwritten after every install/update
+# Persistent local customizations
+include hledger-ui.local
+
+# Redirect
+include hledger.profile
--- a/etc/profile-a-l/hledger.profile
+++ b/etc/profile-a-l/hledger.profile
@@ -1,0 +1,69 @@
+# Firejail profile for hledger
+# Description: Plain text accounting software (CLI)
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include hledger.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.hledger.journal
+
+# Allow /bin/sh (blacklisted by disable-shell.inc)
+include allow-bin-sh.inc
+
+blacklist ${RUNUSER}
+blacklist /usr/libexec
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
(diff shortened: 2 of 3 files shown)


kmk3 (changes requested):
Both profiles are nearly identical, so I don't see a need for a -common
profile.

To simplify it:

* Move hledger-common into hledger
* Make hledger-ui redirect to hledger
* Remove hledger-common

etc/profile-a-l/hledger-ui.profile
```diff
@@ -0,0 +1,11 @@
+# Firejail profile for hledger-ui
+# Description: terminal interface (TUI) for hledger
+# This file is overwritten after every install/update
+# Persistent local customizations
+include hledger-ui.local
+
+# Allow /bin/sh (blacklisted by disable-shell.inc)
+# This seems required for launching editor in hledger-ui
+include allow-bin-sh.inc
```
kmk3, line 9:
```suggestion
# This seems to be required for launching the editor in hledger-ui.
# Allow /bin/sh (blacklisted by disable-shell.inc)
include allow-bin-sh.inc
```

The `Allow ...` comment should always immediately precede the `include allow`
entry (see profile.template).

This also applies to the other PRs.

kmk3 (changes requested):

kmk3 (commented):
Can you try to add `private-etc` entries to the new profiles?

You can check what paths a program tries to access with `--trace=`.

Example:

```sh
firejail --trace=trace.txt /usr/bin/foo
```

The output in trace.txt should contain the relevant /etc paths.

Other than program-specific files, generally a profile will have `private
@network,@tls-ca` if it needs internet access and `private-etc @x11` if it has
a GUI.

See `src/include/etc_groups.h` for the other groups.

---

Also, #6593 was merged and it catches (among other things) a missing newline at
EOF (which affects some of the open PRs), so please rebase all PRs to master.

etc/profile-a-l/hledger-ui.profile
```diff
@@ -0,0 +1,8 @@
+# Firejail profile for hledger-ui
+# Description: terminal interface (TUI) for hledger
```
kmk3, line 2:
```suggestion
# Description: Terminal interface (TUI) for hledger
```

Formatting.

etc/profile-a-l/hledger.profile
```diff
@@ -0,0 +1,69 @@
+# Firejail profile for hledger
+# Description: hledger is plain text accounting software
```
kmk3, line 2:
```suggestion
# Description: Plain text accounting software (CLI)
```

Formatting.

kmk3 (approved):

PR #6587, opened 2024-12-27
New profile: ncmpcpp

amano-kenji:
https://github.com/ncmpcpp/ncmpcpp

Relates to #6580.

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -345,6 +345,7 @@ read-only ${HOME}/.cargo/env
 read-only ${HOME}/.config/mpv
 read-only ${HOME}/.config/msmtp
 read-only ${HOME}/.config/nano
+read-only ${HOME}/.config/ncmpcpp/config
 read-only ${HOME}/.config/nvim
 read-only ${HOME}/.config/pkcs11
 read-only ${HOME}/.dotfiles
--- a/etc/profile-m-z/ncmpcpp.profile
+++ b/etc/profile-m-z/ncmpcpp.profile
@@ -1,0 +1,75 @@
+# Firejail profile for ncmpcpp
+# Description: Featureful ncurses-based MPD client inspired by ncmpc
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include ncmpcpp.local
+# Persistent global definitions
+include globals.local
+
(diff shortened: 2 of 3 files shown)


etc/profile-m-z/ncmpcpp.profile
```diff
@@ -0,0 +1,69 @@
...
+include disable-programs.inc
+include disable-shell.inc
+#include disable-write-mnt.inc
+include disable-X11.inc
+include disable-xdg.inc
+
+mkdir ${HOME}/.config/ncmpcpp
+mkdir ${HOME}/.lyrics
+whitelist ${HOME}/.config/ncmpcpp
+whitelist ${HOME}/.lyrics
```
kmk3, line 30: Add these paths to disable-programs.inc and add corresponding `noblacklist`
entries in this profile.

This should be done whenever adding program-specific paths.

Also, if any of the configuration files allow arbitrary code execution, add
`read-only $path` to disable-common.inc (and `read-write $path` to this profile
if the configuration file can be edited by the program itself).

This also applies to the other PRs.

amano-kenji: There are a lot of suggested changes. I will try to apply all changes at once tomorrow.

amano-kenji: I applied this change, but I also added /var/lib/mpd to disable-programs.inc

etc/inc/disable-common.inc
```diff
@@ -739,3 +739,6 @@ blacklist ${HOME}/.local/opt/tor-browser
 
 # pass utility (pass package in Debian etc.)
 blacklist ${HOME}/.password-store
+
+# ncmpcpp
+read-only ${HOME}/.config/ncmpcpp/config
```
kmk3, line 744: Move it below `read-only ${HOME}/.config/nano`.

etc/profile-m-z/ncmpcpp.profile
```diff
@@ -0,0 +1,75 @@
+# Firejail profile for ncmpcpp
+# Description: Featureful ncurses based MPD client inspired by ncmpc
```
kmk3, line 2:
```suggestion
# Description: Featureful ncurses-based MPD client inspired by ncmpc
```

kmk3 (approved):

PR #6588, opened 2024-12-27
New profile: nsxiv

amano-kenji:
https://nsxiv.codeberg.page/

Relates to #6580.

### DIFF
--- a/etc/inc/disable-common.inc
+++ b/etc/inc/disable-common.inc
@@ -346,6 +346,7 @@ read-only ${HOME}/.config/mpv
 read-only ${HOME}/.config/msmtp
 read-only ${HOME}/.config/nano
 read-only ${HOME}/.config/ncmpcpp/config
+read-only ${HOME}/.config/nsxiv/exec
 read-only ${HOME}/.config/nvim
 read-only ${HOME}/.config/pkcs11
 read-only ${HOME}/.dotfiles
--- a/etc/profile-m-z/nsxiv.profile
+++ b/etc/profile-m-z/nsxiv.profile
@@ -1,0 +1,57 @@
+# Firejail profile for nsxiv
+# Description: Neo Simple X Image Viewer
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include nsxiv.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.cache/nsxiv
+noblacklist ${HOME}/.config/nsxiv
+
+# Allow /bin/sh (blacklisted by disable-shell.inc)
+include allow-bin-sh.inc
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-write-mnt.inc
+
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+
+apparmor
+caps.drop all
+machine-id
+net none
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noprinters
+noroot
+nosound
+notpm
+notv
+nou2f
+novideo
+seccomp
+seccomp.block-secondary
+
+private-cache
+private-dev
+private-etc
+private-tmp
+
+dbus-user none
+dbus-system none
+
+deterministic-shutdown
+memory-deny-write-execute
+read-only ${HOME}
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -193,6 +193,7 @@ blacklist ${HOME}/.cache/mypaint
 blacklist ${HOME}/.cache/netsurf
 blacklist ${HOME}/.cache/nheko
 blacklist ${HOME}/.cache/nhex
+blacklist ${HOME}/.cache/nsxiv
 blacklist ${HOME}/.cache/nvim
 blacklist ${HOME}/.cache/ocenaudio
 blacklist ${HOME}/.cache/okular
@@ -598,6 +599,7 @@ blacklist ${HOME}/.config/newsboat
 blacklist ${HOME}/.config/newsflash
 blacklist ${HOME}/.config/nheko
 blacklist ${HOME}/.config/nomacs
+blacklist ${HOME}/.config/nsxiv
 blacklist ${HOME}/.config/nuclear
 blacklist ${HOME}/.config/nvim
 blacklist ${HOME}/.config/obs-studio
--- a/src/firecfg/firecfg.config
+++ b/src/firecfg/firecfg.config
@@ -667,6 +667,7 @@ nitroshare-ui
 nomacs
 notable
 nslookup
+nsxiv
 nuclear
 nylas
 nyx


kmk3 (changes requested):

etc/profile-m-z/nsxiv.profile
```diff
@@ -0,0 +1,49 @@
...
+# This file is overwritten after every install/update
+# Persistent local customizations
+include nsxiv.local
+# Persistent global definitions
+include globals.local
+
+# Allow /bin/sh (blacklisted by disable-shell.inc)
+include allow-bin-sh.inc
+
+include disable-devel.inc
```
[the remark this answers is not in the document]
kmk3: > Also, add the paths to disable-programs.inc.
> 
> And the following to "Initialization files" in disable-common.inc:
> 
> ```
> read-only ${HOME}/.config/nsxiv/exec
> ```

etc/inc/disable-common.inc
```diff
@@ -739,3 +739,6 @@ blacklist ${HOME}/.local/opt/tor-browser
 
 # pass utility (pass package in Debian etc.)
 blacklist ${HOME}/.password-store
+
+# nsxiv
+read-only ${HOME}/.config/nsxiv/exec
```
kmk3, line 744: Move it below `read-only ${HOME}/.config/nano`.

etc/profile-m-z/nsxiv.profile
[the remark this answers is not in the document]
kmk3: > Also, note that without including disable-common.inc, the profile arguably
> does not do all that much for security (as sensitive paths like ~/.bashrc and
> ~/.ssh would remain available).

etc/profile-m-z/nsxiv.profile
```diff
@@ -0,0 +1,52 @@
...
+noroot
+nosound
+notpm
+notv
+nou2f
+novideo
+seccomp
+seccomp.block-secondary
+
+private-dev
```
kmk3, line 45:
```suggestion
private-cache
private-dev
private-etc
```

Does it work with `private-cache` and `private-etc`?

etc/profile-m-z/nsxiv.profile
```diff
@@ -0,0 +1,52 @@
+# Firejail profile for nsxiv
+# Description: Neo Simple X Image Viewer
+# This file is overwritten after every install/update
+# Persistent local customizations
```
kmk3, line 4:
```suggestion
# This file is overwritten after every install/update
quiet
# Persistent local customizations
```

Use `quiet` for CLI programs.

etc/profile-m-z/nsxiv.profile
kmk3, line 1: Add to firecfg.config.

kmk3:
(Offtopic)

If you are going to rebase to master, please do so in a separate force-push.
This makes it easier to review PR-specific changes if they are also
force-pushed, as each force-push gets its own timeline event.

For example:

```sh
git rebase master
git push --force-with-lease
# (modify PR-related files)
git push --force-with-lease
```

etc/profile-m-z/nsxiv.profile
amano-kenji: nsxiv is an X11 image viewer.

etc/profile-m-z/nsxiv.profile
```diff
@@ -0,0 +1,56 @@
...
+notpm
+notv
+nou2f
+novideo
+seccomp
+seccomp.block-secondary
+
+private-cache
+private-dev
+private-etc fonts
```
kmk3, line 48:
```suggestion
private-etc
```

Already in `@default`; see `src/include/etc_groups.h`.

etc/profile-m-z/nsxiv.profile
kmk3: > nsxiv is an X11 image viewer.

Yes, what I meant is that such minimalist/suckless tools are commonly executed
from a terminal (at least in the case of sxiv).

I'd guess that this would be the case for most users of nsxiv.

amano-kenji: nsxiv is almost always launched from a TUI/GUI file manager on my machine. I may launch it on a terminal emulator for debugging purposes from time to time.

kmk3: > nsxiv is almost always launched from a TUI/GUI file manager on my machine. I
> may launch it on a terminal emulator for debugging purposes from time to
> time.
> 
> Using it on a command line is uncomfortable because I have to pass every
> image file to nsxiv command arguments. TUI/GUI file managers automatically
> pass all image files in a directory to nsxiv.

What do you mean?

Both sxiv and nsxiv can open directories (see also `-r` for recursive).

Using `nsxiv foo` or `cd foo && nsxiv .`, it correctly opens all images in a
`foo` directory, skipping all non-image files.

amano-kenji: If I want to open a specific image in a directory among all other images in the directory, I wouldn't want to determine the value of `--start-at` manually everytime unless someone forces me to do so against my will.

kmk3: > If I want to open a specific image in a directory among all other images in
> the directory, I wouldn't want to determine the value of `--start-at`
> manually everytime unless someone forces me to do so against my will.
> 
> nsxiv wasn't designed as an image browser.

What?

It quite literally lets you browse all images by thumbnail if you press Enter
("thumbnail mode").

And then mark/select the thumbnails with `m` and perform actions on all marked
images.

What else would an image browser need?

amano-kenji: If you are trying to use nsxiv as a command line image viewer, you pass an image file to it and want to browse a directory afterwards. This usage is not supported by plain nsxiv.

I don't just open up a directory in nsxiv. I want to start with a specific image file in a directory.

kmk3: > If you are trying to use nsxiv as a command line image viewer, you pass an
> image file to it and want to browse a directory afterwards. This usage is not
> supported by plain nsxiv.

Indeed, it is not inteded to be used as a GUI file manager.

In other words, it does not support opening anything in its GUI, but does
support opening them in its CLI.

So unlike most GUI image viewers (such as gwenview), nsxiv cannot be started
and used through its GUI alone.  It does not even open a GUI and returns an
error if it is called without CLI arguments.  That is, it requires using its
CLI in order to be useful, therefore it is primarily a CLI tool.

> I don't just open up a directory in nsxiv. I want to start with a specific
> image file in a directory.
> 
> I use command line extensively everyday, but I don't use it for image
> browsing. I use yazi TUI file manager for image browsing nowadays. Inside
> yazi, I open an image file with nsxiv or mpv. If you are not using yazi or
> joshuto, you are missing out.
> 
> After you get used to directory shortcuts in file managers, you don't use
> shell as a file browser.

I'm glad that this setup works for you, but the same argument can be made for
any external program that you decide to execute through a file manager (or
something like dmenu).

Does scrot cease to be a CLI tool if it's opened through a keyboard shortcut
rather than a terminal?

Is gwenview primarily a CLI tool because you can open images through its CLI in
a terminal?

If the file manager calls fzf and displays the result in its TUI, does that
make fzf a TUI tool?

amano-kenji: > In other words, it does not support opening additional files and directories in its GUI, but does support opening them in its CLI.

Do you actually use nsxiv or sxiv everyday? I do.

> So unlike most GUI image viewers (such as gwenview), nsxiv cannot be started and used through its GUI alone. It does not even open a GUI and returns an error if it is called without CLI arguments. That is, it requires using its
CLI in order to be useful, therefore it is primarily a CLI tool.

ranger file manager just passes the right arguments to nsxiv when I open an image. yazi file manager doesn't yet know how to handle nsxiv, so I had to write a little program to handle nsxiv for yazi.

nsxiv only opens either a directory or a file. If you don't write a program that calls nsxiv with the right arguments, it can't open a directory but start with a specific image.

If I didn't know how to write scripts, I wouldn't use nsxiv. I would use gwenview or something else that can open an image file but also browser the directory containing the image without forcing me to write a script.

It is not a CLI program that you use on shells. It is a scriptable program that you use indirectly via a file manager.

kmk3: > > In other words, it does not support opening additional files and
> > directories in its GUI, but does support opening them in its CLI.
> 
> Do you actually use nsxiv everyday? I do.

I use it occasionally to quickly open a few images when I'm already on the
terminal, as it's faster and has a keyboard-centric UI compared to common GUI
image viewers.

If I want to actually interact with the images, I'd probably use something with
good mouse support.  For example, AFAIK there is no way to drag images around
in nsxiv.

> nsxiv only opens either a directory or a file. If you don't write a program
> that calls nsxiv with the right arguments, it can't open a directory but
> start with a specific image.

What is the functional difference between the following?

* Opening a file manager, visually browsing to a specific image and running
  `nsxiv --start-at "$image" .`
* Running `nsxiv .` and visually browsing to a specific image

> If I didn't know how to write scripts, I wouldn't use nsxiv. I would use
> gwenview or something else that can open an image file but also browser the
> directory containing the image without forcing me to write a script.
> 
> It is not a CLI program that you use on shells. It is a scriptable program
> that you use indirectly via a file manager, window manager, or some other
> programs. I even have a sway keyboard shortcut that opens a directory with
> nsxiv. You don't execute nsxiv directly on shells.

Do you think that the people that created sxiv/nsxiv would agree with that?
I'd be surprised.

It works just fine when using it through a shell, so I don't see what the
problem is.  File managers are nice, but they are still just an extra, optional
middleman.

> If I wanted an image viewer that I wanted to execute directly on shells, I
> would use gwenview, ristretto, or something else.

I don't see why the program opened would be different based on whether it's
opened through the terminal or a file manager (as opposed to being based on
something like the use case).

If gwenview/ristretto had `--start-at`, would you use them through the file
manager instead of nsxiv?

> I never execute nsxiv directly on shells. nsxiv is not suitable for that
> usage. I wouldn't execute nsxiv directly on shell when I'm not debugging
> something. When I debug something, I want to see some logs from firejail,
> too. When I can't open an image in nsxiv, I have to debug it on shells.
> Perhaps, I didn't whitelist a directory in firejail.
> 
> My argument is that not having `quiet` is the right default.

So this is the main point: The default firejail output is very verbose.  For
programs not commonly used in the terminal, it's usually fine to omit `quiet`.
Otherwise, it just adds a lot of noise when running/debugging the program
itself, especially when in a generally interactive shell session (as opposed to
a program that you just open through a shell and leave it running).

Only when debugging firejail itself or the profile is the output actually
useful.

In your specific case, unless you are frequently debugging the firejail
profile, you are not going to see the firejail output anyway.

And when you actually want to see it, you can use `firejail --ignore=quiet
/usr/bin/nsxiv` or temporarily (or permanently) add `ignore quiet` to
~/.config/firejail/nsxiv.local.

amano-kenji: It is not `nsxiv --start-at "$image"`. It is `nsxiv --start-at "$num"`.

It is very time-consuming to manually calculate the numerical position of an image file in a directory. I wrote a script to handle that.

kmk3: > It is not `nsxiv --start-at "$image" .`. It is `nsxiv --start-at "$num" .`.

Good catch; I originally wrote `nsxiv --start-at 10 .`, but then I tried to
make it more generic.

> It is very time-consuming to manually calculate the numerical position of an
> image file in a directory. I wrote a script to handle that.
> 
> > --start-at
> > Start at picture number NUM.
> 
> This is why I don't use nsxiv on shell directly.

I understand that, what I mean is that I don't think that this use case (where
using a file manager + `--start-at` is basically a requirement) is
(overwhelmingly) more common than just using nsxiv directly on the terminal.

amano-kenji: Let's make it clear here. Do you think `quiet` might be beneficial for other people who use nsxiv on shells frequently? Or, do you personally use nsxiv on shells frequently?

kmk3: > Let's make it clear here. Do you think `quiet` might be beneficial for other
> people who use nsxiv on shells frequently?

Yes, and that also goes for any tool in this scenario:

* The program is commonly used on the terminal ->
* The default firejail output is too verbose (and usually unnecessary) to see
  frequently on the terminal ->
* Use `quiet` in the profile

I think that this use case is either the default or at least a common enough
case, especially considering the ethos of suckless tools (and that some of
these tools are explicitly mentioned in nsxiv's README.md).

> Or, do you personally use nsxiv on shells frequently?

I use it occasionally as I mentioned and I'd rather not see unnecessary output.

amano-kenji:
I looked into swayimg, and it's worse than nsxiv. I wish I could find a wayland-native alternative to nsxiv before Xorg dies.

kmk3 (approved):

kmk3:
> I looked into swayimg, and it's worse than nsxiv. I wish I could find a
> wayland-native alternative to nsxiv before Xorg dies.

Even if all DEs/WMs somehow stop working, xwayland will probably be around for
a long while, as something is needed to keep existing programs working.

> swayimg starts as a floating window, and I can't make it start as a
> non-floating window. Its default background color is transparent, so it looks
> ugly without further configuration.
> 
> I something like nsxiv that doesn't force me to write C for config, supports
> wayland, and doesn't automatically write text config files. Qutebrowser has
> been exactly that. It supports python for config, supports wayland, and
> doesn't automatically write text config files.
> https://hackage.haskell.org/package/glirc configuration system is also okay.
> 
> I haven't seen any program that seamlessly mixes GUI config with manual text
> config. Qutebrowser is close to that ideal.

I think the main issue is that there is nothing as standard and as unified as
libx11 (and related libs) for wayland AFAIK, so you'd probably have to link to
something much bigger and framework-specific like libadwaita or qt and deal
with the potential design and maintenance churn.

PR #6591, opened 2024-12-27
bugfix: do not interact with dbus directory if dbus proxy is disabled

powerjungle:


kmk3 (commented):
For clarity, does the current code cause any issues or is this intended to just
keep /run/firejail cleaner?

powerjungle:
It did cause an issue when I was running firejail inside a chroot environment. That's not important, but I also realized there's no point in that code running if we disable the dbus proxy code anyway, just to keep thing cleaner.

kmk3:
> It did cause an issue when I was running firejail inside a chroot
> environment. That's not important, but I also realized there's no point in
> that code running if we disable the dbus proxy code anyway, just to keep
> things cleaner.

I see, but it is currently marked as a bugfix and it's not immediately obvious
what the bug is/what are the practical effects of creating/bind-mounting the
directory.

Could you add those details to the commit message?

By the way, firejail in a chroot is likely an uncommon environment, but if you
have any tips, use cases, etc about it, feel free to add them to the wiki.

powerjungle:
> I see, but it is currently marked as a bugfix and it's not immediately obvious what the bug is

Sorry about that. Well the bug was, trying to run `firejail` inside of a `firejail --chroot=/somedir --noprofile` causes an `assert` to fail which checks whether the mode for `/run/firejail/dbus` is correct, which in this case means `0755`, but the issue is that when inside of a "firejail chroot", that directory is mounted with mode `0400` probably to isolate the `chroot` better, which makes sense.

I said that it's not important, because this use case shouldn't change the fact that it shouldn't create and care about the directory if `firejail` is built with `dbus` disabled. It's just a small cleanup.

> By the way, firejail in a chroot is likely an uncommon environment, but if you have any tips, use cases, etc about it, feel free to add them to the wiki.

Oh I forgot about the wiki, that's not a bad idea. Yeah I realize it's uncommon, actually I realized it even had a `chroot` argument very recently. I think it's a nice and relatively user friendly way to setup a "development environment" where you can install all dev packages in there and just isolate it from the "host system". Also testing out unstable software in that environment is really nice. Yes there are `LXC` and `chroot/schroot`, but all of those are a lot more tricky in my opinion and since I'm already using `firejail` for half of my programs anyway, why not use it for a simple "isolated environment" too. Yes, there are virtual machines too, but they're waaay too heavy for most daily use cases in my opinion, especially if you work on a laptop.

kmk3:
@powerjungle

(Offtopic)

Hello, please avoid rebasing to master without need (especially for branches
with only 1 commit, which usually end up being squashed/rebased anyway), as it
generates noise with notifications/timeline events, which likely results in
wasted time re-reviewing.

I think it's helpful to rebase to master when:

* There have been a lot of new commits to master
* There are new commits related to the PR or that touch related code
* There are new commits that improve development/debugging of the PR (#6579
  might be a good example)

kmk3 (commented):
> Sorry about that. Well the bug was, trying to run `firejail` inside of a
> `firejail --chroot=/somedir --noprofile` causes an `assert` to fail which
> checks whether the mode for `/run/firejail/dbus` is correct, which in this
> case means `0755`, but the issue is that when inside of a "firejail chroot",
> that directory is mounted with mode `0400` probably to isolate the `chroot`
> better, which makes sense. Even when running `sudo firejail --chroot=/somedir
> --noprofile`, the `/run/firejail/dbus` directory still has the mode `0400`.

Could you mention this issue in the commit message and how this code fixes it?
It would be nice to have it there for future reference.

It can be a short summary if you don't want to get into the details.

Also, it would be good to include the assert error message.

> I said that it's not important, because this use case shouldn't change the
> fact that it shouldn't create and care about the directory if `firejail` is
> built with `dbus` disabled. It's just a small cleanup. I probably shouldn't
> have called it a `bugfix`, but I just felt like it might cause issues in some
> other weird situations as well.

I think it's fine to call this a bugfix, as it makes `--chroot=` work with a
`--disable-dbusproxy` build.

powerjungle:
> Could you mention this issue in the commit message and how this code fixes it?

Done, I hope the commit description is clear.

netblue30:
Merged, thanks!

PR #6592, opened 2024-12-27
feature: build: add --disable-sandbox-check configure flag

powerjungle:
This flag disables the code which checks whether the current instance of
firejail is running within a sandbox like LXC, chroot or firejail itself.
If we want to develop firejail inside of a sandbox, to keep the "host system"
clean of unnecessary installed dependencies and changes to the system,
we might want to force firejail to run normally, so that we can test different
profiles inside of the sandbox. This is only meant for people who are working
on the firejail code, not someone attempting to run firejail inside of a
sandbox as a user, because it needs to run as root and it can escape the
sandbox easily.


netblue30:
Merging in, thanks!

PR #6556, opened 2024-11-29
build: sort.py: strip whitespace in profiles

kmk3:
build: sort.py: strip trailing whitespace in all lines

Currently the output is mangled if the last item on the line contains
trailing whitespace and is moved when sorting.

So remove trailing whitespace in all lines (that is, not just in lines
containing supported commands).

Leave leading whitespace as is for now since it could potentially be
used for indentation.

Before:

    $ printf '# hello world  \nprivate-bin a,b  \nprivate-bin b,a  \nprivate-bin  a,b\n' \
      >foo.profile
    $ ./contrib/sort.py -n foo.profile | tr ' ' .
    sort.py:.checking.1.profile(s)...
    foo.profile:3:-private-bin.b,a..
    foo.profile:3:+private-bin.a..,b

After:

    $ printf '# hello world  \nprivate-bin a,b  \nprivate-bin b,a  \n' \
      >foo.profile
    $ ./contrib/sort.py -n foo.profile | tr ' ' .
    sort.py:.checking.1.profile(s)...
    foo.profile:1:-#.hello.world..
    foo.profile:1:+#.hello.world
    foo.profile:2:-private-bin.a,b..
    foo.profile:2:+private-bin.a,b
    foo.profile:3:-private-bin.b,a..
    foo.profile:3:+private-bin.a,b

---

build: sort.py: strip whitespace in commands

Currently whitespace is left as is within an entry.

In a `protocol` entry, if there is whitespace between the command and
its argument or around an item, the item in question is dropped from the
output.

Changes:

* `protocol`: Strip all whitespace in the argument
* Other commands: Strip leading/trailing whitespace around each item,
  including any extra whitespace between a command and its argument

Note: Whitespace characters inside paths are left as is, as some paths
(such as `Foo Bar` may contain spaces.

Before:

    $ printf 'private-bin a,b\nprivate-bin  a,b\nprivate-bin  b,a\nprivate-bin  C,A  B\nprotocol  unix,net\nprotocol  inet,unix\n' \
      >foo.profile
    $ ./contrib/sort.py -n foo.profile
    sort.py: checking 1 profile(s)...
    foo.profile:5:-protocol  unix,net
    foo.profile:5:+protocol
    foo.profile:6:-protocol  inet,unix
    foo.profile:6:+protocol unix

After:

    $ printf 'private-bin a,b\nprivate-bin  a,b\nprivate-bin  b,a\nprivate-bin  C,A  B\nprotocol  unix,net\nprotocol  inet,unix\n' \
[...]

PR #6593, opened 2024-12-28
build: sort.py: fix whitespace in entire profile

kmk3:
Changes:

* Strip whitespace at the beginning
* Strip whitespace at the end
* Ensure exactly one newline at the end
* Strip extraneous newlines

Also, for clarity print the git diff in the sort.py ci job, since the
specific lines changed are not printed by the sort.py script in this
case (as whitespace is fixed in the entire profile at once).

Command used to search and replace:

    ./contrib/sort.py etc/inc/*.inc etc/profile*/*.profile

This is a follow-up to #6556.

### DIFF
--- a/contrib/sort.py
+++ b/contrib/sort.py
@@ -99,11 +99,20 @@ def check_profile(filename, overwrite):
                 )
             fixed_profile.append(line)
 
+        fixed_profile_str = "\n".join(fixed_profile)
+        stripped_profile_str = fixed_profile_str.strip() + "\n"
+        while "\n\n\n" in stripped_profile_str:
+            stripped_profile_str = stripped_profile_str.replace("\n\n\n", "\n\n")
+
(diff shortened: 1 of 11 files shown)


contrib/sort.py
```diff
@@ -99,11 +99,19 @@ def check_profile(filename, overwrite):
                 )
             fixed_profile.append(line)
 
+        fixed_profile_str = "\n".join(fixed_profile)
+        stripped_profile_str = fixed_profile_str.strip() + "\n"
+        while stripped_profile_str.find("\n\n\n") != -1:
```
rusty-snake, line 104:
```suggestion
        while " \n\n\n" in stripped_profile_str:
```

kmk3: Good catch.

rusty-snake (approved):

PR #6598, opened 2025-01-03
bugfix: fix possible memory leak in fs_home.c

peter15914:
The pointer returned by ```asprintf()``` should be passed to free() to release the allocated storage when it is no longer needed.
There was one exit point without previous freeing ```src```.


kmk3 (approved):

kmk3:
@peter15914

By the way, I see that you fixed similar issues on multiple other projects.

Are you using any specific tool to check for this?

peter15914:
> @peter15914
> 
> By the way, I see that you fixed similar issues on multiple other projects.
> 
> Are you using any specific tool to check for this?

Yes, I use a specific tool, a kind of static code analyzer, but this tool is not public.

PR #6609, opened 2025-01-14
profiles: aria2p: disable x11 and clipboard managers

kmk3:
aria2p is a command-line tool, so these should not be needed (and it's
unclear how/why they would be used by the program).

See also:
https://github.com/netblue30/firejail/pull/6583#discussion_r1912891807

Added on commit c869f11d5 ("New profile: aria2p/aria2rpc", 2024-12-27) /
PR #6583.

Cc: @amano-kenji


amano-kenji:
After thinking for a while, I concluded that aria2p doesn't need x11 or clipboard managers.

PR #6616, opened 2025-01-15
New profile: device-flasher.linux (CalyxOS)

lucasmz-dev:
The CalyxOS CLI device flasher.

### DIFF
--- a/etc/profile-a-l/device-flasher.linux.profile
+++ b/etc/profile-a-l/device-flasher.linux.profile
@@ -1,0 +1,72 @@
+# Firejail profile for device-flasher.linux
+# Description: CalyxOS' device flasher
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include device-flasher.linux.local
+# Persistent global definitions
+include globals.local
+
+# Usage: Run firejail ./device-flasher.linux in the folder with it and your
+# factory image.
+
+ignore noexec ${HOME}
+
+# Allow /bin/sh (blacklisted by disable-shell.inc)
+include allow-bin-sh.inc
+
+blacklist ${RUNUSER}
+blacklist /opt
+blacklist /srv
+blacklist /usr/libexec
+
+include disable-common.inc
+include disable-devel.inc
+include disable-exec.inc
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-x11.inc
+include disable-xdg.inc
+
+whitelist ${DOWNLOADS}
+
+include whitelist-run-common.inc
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+apparmor
+caps.drop all
+ipc-namespace
+machine-id
+netfilter
+no3d
+nodvd
+nogroups
+noinput
+nonewprivs
+noprinters
+noroot
+nosound
+notpm
+notv
+nou2f
+novideo
+protocol inet,inet6,netlink
+seccomp
+seccomp.block-secondary
+
+disable-mnt
+private-bin cut,grep,sed,sh,sleep,which
+private-cache
+private-etc @network,@tls-ca
+private-tmp
+
+dbus-user none
+dbus-system none
+
+deterministic-shutdown
+memory-deny-write-execute
+restrict-namespaces


etc/profile-a-l/device-flasher.linux.profile
```diff
@@ -0,0 +1,82 @@
...
+# Persistent local customizations
+include globals.local
+# Persistent global definitions
+
+ignore noexec ${HOME}
+
+include allow-bin-sh.inc
+
+blacklist /opt
+blacklist /srv
```
lucasmz-dev, line 25: Added this only because of rusty-snake's kyst, works fine so might as well keep it?
I'm not entirely sure what /opt and /srv are for.

etc/profile-a-l/device-flasher.linux.profile
```diff
@@ -0,0 +1,82 @@
...
+# reboot the device into bootloader mode first. To do this, hold Volume Down + Power until you get
+# into a screen with an Android robot being repaired. You'll need OEM Unlocking beforehand.
+
+# Fedora: you may need to manually install a third-party resource for udev rules, as Fedora ships broken ones on android-tools.
+
+quiet
+include device-flasher.linux.local
+# Persistent local customizations
+include globals.local
+# Persistent global definitions
```
kmk3, line 18:
```suggestion
# Firejail profile for device-flasher.linux
# Description: CalyxOS' device flasher
# This file is overwritten after every install/update
quiet
# Persistent local customizations
include device-flasher.linux.local
# Persistent global definitions
include globals.local

# Usage: Run firejail ./device-flasher.linux in the folder with it and your
# factory image.

# Warning: ADB does not work under this sandbox, this is not a problem however,
# it just means you need to reboot the device into bootloader mode first. To do
# this, hold Volume Down + Power until you get into a screen with an Android
# robot being repaired. You'll need OEM Unlocking beforehand.

# Fedora: You may need to manually install a third-party resource for udev
# rules, as Fedora ships broken ones on android-tools.
```

etc/profile-a-l/device-flasher.linux.profile
```diff
@@ -0,0 +1,82 @@
...
+
+ignore noexec ${HOME}
+
+include allow-bin-sh.inc
```
kmk3, line 22:
```suggestion
# Allow /bin/sh (blacklisted by disable-shell.inc)
include allow-bin-sh.inc
```

etc/profile-a-l/device-flasher.linux.profile
```diff
@@ -0,0 +1,82 @@
...
+
+blacklist /opt
+blacklist /srv
+blacklist ${RUNUSER}
+blacklist /usr/libexec
```
kmk3, line 27:
```suggestion
blacklist ${RUNUSER}
blacklist /opt
blacklist /srv
blacklist /usr/libexec
```

Sort.

etc/profile-a-l/device-flasher.linux.profile
```diff
@@ -0,0 +1,82 @@
...
+include disable-programs.inc
+include disable-shell.inc
+include disable-x11.inc
+include disable-xdg.inc
+
+whitelist ${DOWNLOADS}
+# The Downloads folder is the only folder available to the flasher, use it or manually whitelist another.
+# Preferably use a sub-folder such as Downloads/CalyxOS/ to prevent issues.
+
+include whitelist-run-common.inc
```
kmk3, line 43:
```suggestion
whitelist ${DOWNLOADS}

include whitelist-common.inc
include whitelist-run-common.inc
```

The comment is kind of redundant, as it generally applies to any profile with
`whitelist ${DOWNLOADS}` (or another XDG user directory).

Usually whitelist-common.inc is included when `whitelist ${HOME}/...` is used.

etc/profile-a-l/device-flasher.linux.profile
```diff
@@ -0,0 +1,82 @@
...
+include whitelist-runuser-common.inc
+include whitelist-usr-share-common.inc
+include whitelist-var-common.inc
+
+include landlock-common.inc
```
kmk3, line 48:
```suggestion
```

Not currently used.

etc/profile-a-l/device-flasher.linux.profile
```diff
@@ -0,0 +1,82 @@
...
+nou2f
+novideo
+protocol inet,inet6
+seccomp
+seccomp.block-secondary
+
+disable-mnt
+private-bin bash,cut,grep,ls,sed,sh,sleep,which
+private-cache
+private-etc alternatives,ca-certificates,crypto-policies,host.conf,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,mime.types,nsswitch.conf,pki,protocols,resolv.conf,rpc,services,ssl,xdg
```
kmk3, line 74:
```suggestion
private-etc @network,@tls-ca,host.conf,mime.types,rpc,services,xdg
```

Use new `private-etc` groups (see #6400).

kmk3 (changes requested):

etc/profile-a-l/device-flasher.linux.profile
kmk3: > Added this only because of rusty-snake's kyst, works fine so might as well
> keep it?

Seems fine.

> I'm not entirely sure what /opt and /srv are for.

/opt is sometimes used for putting the entire installation files of third-party
programs into (as with `Program Files` in Windows).

/srv is used by things like HTTP servers.

etc/profile-a-l/device-flasher.linux.profile
lucasmz-dev: I don't see much reason to use whitelist-common.inc, it isn't needed at all, especially being a CLI tool.

(I'll go ahead soon enough and apply the changes requested, I'd like to confirm flashing works, this is the only one I don't see why use whitelist-common in specific, removing the comments is fine)

etc/profile-a-l/device-flasher.linux.profile
```diff
@@ -0,0 +1,82 @@
...
+private-tmp
+
+dbus-system none
+dbus-user none
```
rusty-snake, line 78:
```suggestion
dbus-system none
```

etc/profile-a-l/device-flasher.linux.profile
```diff
@@ -0,0 +1,81 @@
...
+include disable-interpreters.inc
+include disable-proc.inc
+include disable-programs.inc
+include disable-shell.inc
+include disable-x11.inc
+include disable-xdg.inc
+
+whitelist ${DOWNLOADS}
+
+include whitelist-common.inc
```
kmk3, line 43:
```suggestion
```

> I don't see much reason to use whitelist-common.inc, it isn't needed at all,
> especially being a CLI tool.

After looking through whitelist-common.inc again, I think it makes sense to
leave it out, as almost all paths are for GUI programs.

Misc: Maybe it would make sense to eventually rename it to
whitelist-gui-common.inc.

lucasmz-dev:
I'll do a test later today with the changes (even if they're only really reordering AFAICT) and rebase/re-sign the commits with the proper credits

I should probably also look into the stuff in private-etc to see what's truly needed

lucasmz-dev:
It wouldn't be possible to print out a message right? The ADB thing would be useful to print.

lucasmz-dev:
I suspect it may have something to do with the plugdev group

lucasmz-dev:
found the culprit for adb not working
something inside protocol, needs investigating

lucasmz-dev:
adding netlink solves it!

lucasmz-dev:
works perfectly now! i also went and removed some of the entries in private-bin that were debugging only (i thought in keeping it for someone who may wanna mess with other things themselves but... eh)

confirmed working with a flash to a devon device (Moto G32)
I'd assume Pixel to be safe given they're very similar, Fairphones may be different but I still don't think there'd be anything weird

kmk3 (approved):

PR #6623, opened 2025-01-19
build(config.mk.in): add comment explaining the different flags

powerjungle:
To make this less likely in the future: https://github.com/netblue30/firejail/commit/5c6fa6ab584eb6241467b5d5aea76f26dcc922fb


kmk3 (commented):
Documenting this is a good idea, but please ignore this for now.

Issue #2799, opened 2019-06-26
--overlay-named exits with error as of linux 5.1.15


lskrejci:
Distribution: Arch Linux
firejail: 0.9.60-1
linux: 5.1.15.arch1-1

Running `firejail --noprofile --overlay-named=test` fails:

- on a system with a separate /home partition:
> Error mounting overlayfs for mounted home directory: fs.c:1132 fs_overlayfs: Too many levels of symbolic links

- on a system with a single root partition:
> Error mounting overlayfs: fs.c:1077 fs_overlayfs: Too many levels of symbolic links

dmesg on both system configurations contains:
> overlayfs: overlapping upperdir path

The same command works on linux 5.1.14.
Upstream commit that appears to have introduced the issue: [ovl: detect overlapping layers](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-5.1.y&id=99eb836cd9a4df455ae90807bc00ee635be342f0)

Hello71:
I think --rbind might be able to work around this.

Vincent43:
Fix for docker: https://github.com/moby/moby/commit/477bf1e413708076f9ed8cd316102765cc5bdb11

smitsohu:
Maybe I don't fully understand the docker fix, but I wonder a bit what's the point of an overlay just to make the file system read-only.

@netblue30 do you have an idea regarding the way forward?

smitsohu:
> but I wonder a bit what's the point of an overlay just to make the file system read-only.

Or maybe it could serve as a simple switch for apps that really don't need to write anything at all.

Vincent43:
There is some [kernel fix available](https://github.com/torvalds/linux/commit/0be0bfd2de9dfdd2098a9c5b14bdd8f739c9165d) in Linux 5.3.1 and 5.2.17, @lskrejci could you try if you can still reproduce issue no those kernels?

Vincent43:
Still fails on Linux 5.2.17 so some firejail chnages are still needed:
```
firejail --noprofile --overlay-named=test
Parent pid 9167, child pid 9168

**     Warning: dropping all Linux capabilities     **
Error mounting overlayfs: fs.c:977 fs_overlayfs: Too many levels of symbolic links
Error: proc 9167 cannot sync with peer: unexpected EOF
Peer 9168 unexpectedly exited with status 1
```

schendstok:
[...]

PR #6632, opened 2025-01-25
docs: clarify unmaintained status of overlayfs in configure.ac

kmk3:
Since users are still asking about it.

Relates to #2799 #4178.


configure
```diff
@@ -4248,7 +4248,15 @@ then :
 
 fi
 
-# overlayfs features temporarily disabled pending fixes
+# The --overlay and related options are not supported anymore due to security
+# concerns (see CVE-2021-26910 and #4178), issues on newer kernels (see #2799)
+# and overall lack of interest in fixing and maintaining them.
```
rusty-snake, line 4253: (correct)

configure
```diff
@@ -4248,7 +4248,15 @@ then :
...
+#
+# Support is unlikely to be restored, as it would require auditing the
+# overlayfs code to ensure that it is secure and that it works properly.
```
rusty-snake, line 4256: Do we want "unlikely to be restored" or "disable until someone picks up the work".

rusty-snake, line 4256: Technical notes:
- crablock and bubblewrap have integrated support for overlayfs. This was possible because Linux >= 5.11 allows `overlayfs` mounts from unprivileged user-namespace. An "unprivileged" implementation in firejail might be the most likely comeback as it means less auditing.
- The usage in firejail was broken at least two times in the past because of kernel changes (IIRC). And is likely to break again in the future.
  https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html
  > Changes to the underlying filesystems while part of a mounted overlay filesystem are not allowed. If the underlying filesystem is changed, the behavior of the overlay is undefined, though it will not result in a crash or deadlock.

rusty-snake (approved):

configure
kmk3: > Do we want "unlikely to be restored" or "disable until someone picks up the
> work".

Misc: I used stronger language to avoid giving false hope (as with the current
"temporarily disabled") that could lead to more new issues asking to restore
overlayfs, but I agree that it's good to note that it's still possibility with
enough work.

Updated with a bit of both; thoughts?

kmk3: > Technical notes:

Thanks for the all details, they seem to be very relevant.

Though this information will likely get lost in this PR; thoughts on adding it
to #4178, the wiki or a new dedicated issue?

rusty-snake: Sounds good. Gives users no false expectation but underlines the reason (open-source development in free time).

PR #6648, opened 2025-02-15
profiles: qutebrowser: add @sound to private-etc

amano-kenji:


kmk3 (changes requested):
The `@sound` group is automatically included when `nosound` is not used.

kmk3:
But to be fair, some profiles do have `@sound`, so it's unclear when it's
intended to be used.

@rusty-snake Thoughts?

rusty-snake:
> @rusty-snake Thoughts?

No. 
It is document.

kmk3:
> No. It is document.

I have no idea what you mean, can you clarify?

kmk3:
> No thought about `@sound` usage.
> 
> The behavior of `private-etc` in the absence of `nosound` is documented.

Since it's already working as intended and it's documented, let's leave it as
is for now.

Issue #6651, opened 2025-02-19
Cannot use betterbird with firejail


PWungsten:
### Description

I am trying to run betterbird in firejail. I cannot get a profile that works. The automatically-built profile does not work.

### Steps to Reproduce

1. Run `/usr/bin/firejail --build=betterbird.profile  /usr/local/bin/betterbird/betterbird &` . It runs, I can do normal stuff in betterbird.
2. When betterbird exits, the file betterbird.profile is created. All OK so far.
3. Move betterbird.profile to $HOME/.config/firejail
4. Run  `/usr/bin/firejail /usr/local/bin/betterbird/betterbird` . This fails, reporting

> Reading profile /home/ME/.config/firejail/betterbird.profile
> Reading profile /etc/firejail/whitelist-common.inc
> Reading profile /etc/firejail/whitelist-run-common.inc
> Reading profile /etc/firejail/whitelist-runuser-common.inc
> Reading profile /etc/firejail/whitelist-usr-share-common.inc
> Reading profile /etc/firejail/whitelist-var-common.inc
> Warning: networking feature is disabled in Firejail configuration file
> Warning: networking feature is disabled in Firejail configuration file
> Warning: "shell none" command in the profile file is done by default; the command will be deprecated
> Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
> Parent pid 5647, child pid 5648
> 4 programs installed in 6.20 ms
> Warning: skipping drirc for private /etc
> Private /etc installed in 27.52 ms
> Private /usr/etc installed in 0.00 ms
> Warning: cleaning all supplementary groups
> Warning: cleaning all supplementary groups
> Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
> Warning: cleaning all supplementary groups
> Child process initialized in 107.37 ms
> Error: no suitable /usr/local/bin/betterbird/betterbird executable found

Here is the automatically-created profile (comments and commented-out lines deleted)

> include disable-common.inc
> include disable-programs.inc
> whitelist ${HOME}/.mozilla/firefox
> whitelist ${HOME}/.mozilla/firefox/Crash Reports
> whitelist ${HOME}/.cache/thumbnails
> whitelist ${HOME}/.hidden
> whitelist ${HOME}/.mozilla/extensions
> whitelist ${HOME}/.mailcap
> whitelist ${HOME}/.local/bin
> whitelist ${HOME}/bin
> whitelist ${HOME}/.local/share/glib-2.0/schemas
[...]

PR #6653, opened 2025-02-20
docs: note that --build may generate a non-functional profile

kmk3:
Users might expect `--build` to generate a complete and working profile
(see #6651), so clarify that this is not really the case.


rusty-snake (approved):
> may be incomplete

This also means that there is currently no support to generate D-Bus rules or other important restrictions (`restrict-namespaces`, ...) meaning it is not that secure.

kmk3:
> > may be incomplete
> 
> This also means that there is currently no support to generate D-Bus rules or
> other important restrictions (`restrict-namespaces`, ...) meaning it is not
> that secure.

Yes; added a part about security as well.

rusty-snake (approved):

PR #6659, opened 2025-02-22
profiles: thunderbird: fix access to wayland socket

sudoAlphaX:
Denying access to `${RUNUSER}` (`/run/user/$(id -u)`) denies access to
Wayland socket which results in thunderbird running under X11.

whitelist-runuser-common.inc is included in firefox-common.profile, which is
included by firefox.profile, which is in turn included by
thunderbird.profile. But it is ignored by the thunderbird.profile.
This patch removes the ignore whitelist-runuser-common.inc to allow
thunderbird to access wayland sockets.

### DIFF
--- a/etc/profile-m-z/thunderbird.profile
+++ b/etc/profile-m-z/thunderbird.profile
@@ -6,8 +6,6 @@ include thunderbird.local
 # Persistent global definitions
 include globals.local
 
-ignore include whitelist-runuser-common.inc
-
 # TB stopped supporting enigmail in 2020 (v78) - let's harden D-Bus
 # https://support.mozilla.org/en-US/kb/openpgp-thunderbird-howto-and-faq
 ignore dbus-user none


kmk3 (commented):
With this change I see less paths in `${RUNUSER}` than before (which makes
sense when using whitelisting) with the following command:

```sh
firejail --ignore='include globals.local' \
  --ignore='include thunderbird.local' --profile=thunderbird \
  ls -l "/run/user/$(id -u)"
```

Are you sure that this increases the access to the wayland socket?

Access to that socket is available by default AFAIK (see the
`blacklist ${RUNUSER}/wayland-*` entry in many profiles).

Have you tested this with firejail-git and no profile modifications?

rusty-snake (changes requested):
Thunderbird does not use wruc for reason. If you enabled `whitelist ${RUNUSER}/app/org.keepassxc.KeePassXC` or something else in your firefox-common.local, you need to ignore it in your Thunderbird.local.

sudoAlphaX:
> Have you tested this with firejail-git and no profile modifications?

I current use firejail-git without any modifications to the thunderbird.local. I have a few changes to globals.local and they are as follows:
```
# Allow the runtime clipboard
include runtime-clipboard.local

# Do not join the same sandbox (multi-user fix)
ignore join-or-start
```

runtime-clipboard.local

```
mkdir ${RUNUSER}/clipboard
noblacklist ${RUNUSER}/clipboard
whitelist ${RUNUSER}/clipboard
```

~~Which i dont think would affect RUNUSER access.~~

~~Just to be extra sure, I removed all my firejail local overrides launched thunderbird again~~

What the heck? whitelisting ${RUNUSER}/clipboard in globals.local blocks access to `/run/user/$(id -u)` (from runtime-clipboard.local).
**This is a problem with my configuration.** I have removed the file and thunderbird launches with wayland. Any ideas on how to fix this?

Here's are my local overrides:

firefox-common.local

```
# private-etc support
private-etc alternatives,asound.conf,ca-certificates,crypto-policies,dconf,fonts,group,gtk-2.0,gtk-3.0,hostname,hosts,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,localtime,machine-id,mailcap,mime.types,nsswitch.conf,pango,passwd,pki,pulse,resolv.conf,selinux,ssl,X11,xdg
```

firefox.local

```
# Private dirs support
private-bin bash,dbus-launch,dbus-send,env,firefox,sh,which
private-etc firefox

# Inhibit screensaver

# Native notifications

# Wayland Screensharing
#dbus-user.talk org.freedesktop.portal.Desktop
# Add the next line to your firefox.local if screen sharing sharing still does not work
# with the above lines (might depend on the portal implementation).
#ignore noroot
```

globals.local

```
# Allow the runtime clipboard
include runtime-clipboard.local

# Do not join the same sandbox (multi-user fix)
ignore join-or-start
```

runtime-clipboard.local

```
mkdir ${RUNUSER}/clipboard
noblacklist ${RUNUSER}/clipboard
whitelist ${RUNUSER}/clipboard
```

sudoAlphaX:
> Thunderbird does not use wruc for reason. If you enabled `whitelist ${RUNUSER}/app/org.keepassxc.KeePassXC` or something else in your firefox-common.local, you need to ignore it in your Thunderbird.local.

Ah i see. I have whitelisted a particular directory under ${RUNUSER} globally. That is the reason. How do I do something like that without affecting other files? I was ${RUNUSER}/clipboard to be available everywhere but not any parent directory.

Do i use noblacklist?

Thanks for the help.

rusty-snake:
If you want to make sure all sandboxes can access `/run/user/$UID/clipboard`:

`whitelist-runuser-common.local`:

```
whitelist ${RUNUSER}/clipboard
```

`~/.config/user-tmpfiles.d/runuser-clipboard.conf`

```
d %t/clipboard 700 - - - -
```

PR #6660, opened 2025-02-22
feature: block /dev/ntsync & add keep-dev-ntsync command

### LINKED ISSUE #6655: include /dev/ntsync in private-dev
it will be useful when ntsync lands in stable linux / wine since private-dev is in the wine profile

weebnix:
Changes:

* Block access to /dev/ntsync by default
* Add the `keep-dev-ntsync` command to allow access to /dev/ntsync (even
  if `private-dev` is used)
* Add `keep-dev-ntsync` to wine.profile and similar profiles

Closes #6655.

### DIFF
--- a/etc/templates/profile.template
+++ b/etc/templates/profile.template
@@ -195,6 +195,7 @@ include globals.local
 ##x11 none
 
 #disable-mnt
+##keep-dev-ntsync
 ##private
 # It's common practice to refer to the python executable(s) in private-bin with `python*`, which covers both v2 and v3
 #private-bin PROGRAMS
--- a/contrib/syntax/lists/profile_commands_arg0.list
+++ b/contrib/syntax/lists/profile_commands_arg0.list
@@ -9,6 +9,7 @@ deterministic-shutdown
 disable-mnt
 ipc-namespace
 keep-config-pulse
+keep-dev-ntsync
 keep-dev-shm
 keep-shell-rc
 keep-var-tmp
--- a/etc/profile-a-l/lutris.profile
+++ b/etc/profile-a-l/lutris.profile
@@ -79,6 +79,7 @@ protocol unix,inet,inet6,netlink
 seccomp !clone3,!modify_ldt,!process_vm_readv,!ptrace
 seccomp.32 !modify_ldt
 
+keep-dev-ntsync
 # Add the next line to your lutris.local if you do not need controller support.
 #private-dev
 private-tmp
--- a/etc/profile-m-z/noprofile.profile
+++ b/etc/profile-m-z/noprofile.profile
@@ -21,6 +21,7 @@ ignore read-only /sys/module/nvidia*
 allow-debuggers
 allusers
 keep-config-pulse
+keep-dev-ntsync
 keep-dev-shm
 keep-fd all
 keep-var-tmp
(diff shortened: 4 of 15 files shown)


rusty-snake:
I agree that `private-dev a,b,c` is the best way forward. However, it is also the most complicated way to implement. If @weebnix or someone else is willing to implement this, we should. If not it, I think a short/mid term way for `ntsync`  w/o `private-dev ntsync` is better than not supporting it.

Given that `ntsync` seems to be a feature developed for wine and used only by wine (or other WinNT compatibility tools), a `keep-[dev-]ntsync` (like `keep-dev-shm`) would be better and drastically reduce the noise.

kmk3:
> I agree that `private-dev a,b,c` is the best way forward. However, it is also
> the most complicated way to implement. If @weebnix or someone else is willing
> to implement this, we should. If not it, I think a short/mid term way for
> `ntsync` w/o `private-dev ntsync` is better than not supporting it.

Yes, it seems doable but it would not be trivial.

> Given that `ntsync` seems to be a feature developed for wine and used only by
> wine (or other WinNT compatibility tools), a `keep-[dev-]ntsync` (like
> `keep-dev-shm`) would be better and drastically reduce the noise.

+1 for `keep-dev-ntsync`, that seems like a big improvement.

Not only in the noise, but also in that it would be a whitelisting command.

@weebnix Can you try that in the PR?

The PR implementation can remain basically the same, as `keep-dev-shm` does
some things differently with /dev/shm.

That is:

* Keep `DEV_NTSYNC` etc as is
* Rename `nontsync` to `keep-dev-ntsync`
* Try to keep the commands sorted (mostly by keeping `keep-dev-ntsync` right
  above `keep-dev-shm` in the code), especially in the man pages
* Add it to etc/templates/profile.template

Feel free to amend and force-push.

kmk3:
Organized the code and made /dev/ntsync blocked by default, so now
`keep-dev-ntsync` allows access regardless of `private-dev`.

weebnix:
not really sure if its really that necessary to block access to /dev/ntsync by default, because when --noprofile is being used everything except for /dev/ntsync will be available in that sandbox, and that is a pretty confusing behavior for the user.

rusty-snake:
> because when --noprofile is being used everything except for /dev/ntsync will be available in that sandbox

No.

- `/sys/fs` will not be available
- `/sys/module` will not be available
- `/var` will be read-only
- All file descriptors will be closed
- `/home/*` except for `/home/$USER` will not be available
- ...
- https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/noprofile.profile

weebnix:
> No.

oh, alright, i guess i shouldn't have trusted the name

> https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/noprofile.profile

shouldn't we add keep-dev-ntsync here too?

kmk3:
> > https://github.com/netblue30/firejail/blob/master/etc/profile-m-z/noprofile.profile
> 
> shouldn't we add keep-dev-ntsync here too?

It's already there; see the new commits in the PR.

kmk3:
Squashed/reworded the commits and force-pushed.

kmk3 (approved):

etc/templates/profile.template
```diff
@@ -195,6 +195,7 @@ include globals.local
 ##x11 none
 
 #disable-mnt
+##keep-dev-ntsync
```
rusty-snake, line 198: Placement here?

kmk3: > Placement here?

I agree that it does not seem like the best place, though putting it in the
`no3d` section (for example) does not seem much better either so I just left it
as is.  But I can move it somewhere else if you want.

I plan to create a PR after this to create an "allow" section above the `no3d`
etc section with only allow commands (like the second section in
noprofile.profile) and move the relevant commands there.

PR #6669, opened 2025-02-27
New profile: remmina-file-wrapper

northboot:
Remmina may install this wrapper binary on some distributions.

On Void Linux, this is the default binary launched via application
launchers (e.g. rofi):

    $ grep "Exec" /usr/share/applications/org.remmina.Remmina.desktop
    TryExec=remmina-file-wrapper
    Exec=remmina-file-wrapper %U
    [...]


kmk3 (commented):
Remmina is in firecfg.config, so add the new profile there as well.

Also, for completeness use the current PR message as the commit message.

kmk3 (approved):

Issue #6657, opened 2025-02-21
firecfg: gedit is not sandboxed (.desktop file)


ginto37:
### Description

gedit AKA TextEdit is not sandboxed with firejail if it is launched using the Search feature from the Activities/Overview mode in Ubuntu 22.04 LTS. It _is_ sandboxed if launched from the Terminal with `gedit`.

### Steps to Reproduce

_Steps to reproduce the behavior_

1. Open Activities/Overview mode from the top panel or using the keyboard shortcut
2. Search for TextEdit or gedit and launch it
3. In a Terminal, check the output of `firejail --list`

### Expected behavior

Output should be similar to the following:

`3233:USERNAME::/usr/bin/firejail /usr/bin/gedit`

### Actual behavior

There is no output.

### Behavior without a profile

N/A

### Additional context

gedit AKA TextEdit is sandboxed if launched from the Terminal using `gedit`.

### Environment

- Name/version/arch of the Linux kernel (`uname -srm`): Linux 6.8.0-52-generic x86_64
- Name/version of the Linux distribution (e.g. "Ubuntu 20.04" or "Arch Linux"): Ubuntu 22.04.5 LTS
- Name/version of the relevant program(s)/package(s) (e.g. "firefox 134.0-1,
  mesa 1:24.3.3-2"): gedit 41.0
- Version of Firejail (`firejail --version`): firejail version 0.9.72

Compile time support:
	- always force nonewprivs support is disabled
	- AppArmor support is enabled
	- AppImage support is enabled
	- chroot support is enabled
	- D-BUS proxy support is enabled
	- file transfer support is enabled
	- firetunnel support is disabled
	- IDS support is enabled
	- networking support is enabled
	- output logging is enabled
	- overlayfs support is disabled
	- private-home support is enabled
	- private-cache and tmpfs as user enabled
	- SELinux support is enabled
	- user namespace support is enabled
	- X11 sandboxing support is enabled
- If you use a development version of firejail, also the commit from which it
  was compiled (`git rev-parse HEAD`):

### Checklist

- [N/A ] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [N/A ] I can reproduce the issue without custom modifications (e.g. globals.local).
- [X ] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
[...]

PR #6677, opened 2025-03-07
docs: clarify to run "sudo firecfg" as a normal (desktop) user

kmk3:
End users following the steps for desktop integration may end up running
`sudo firecfg` as root (or as a similar account) rather than as a normal
desktop user.

In that case, programs opened through a desktop launcher would still not
be running under firejail, which might surprise users.

So clarify that `sudo firecfg` should be executed as a normal (desktop)
user for desktop integration.

Relates to #6657.

Kind of relates to #5812.

Reported-by: @ginto37

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,9 @@ $ firejail --list
 
 ## Desktop integration
 
+Note: Desktop integration is only applied to the user running firecfg, so make
+sure to run `sudo firecfg` as a normal (desktop) user, not as root.
+
 Integrate your sandbox into your desktop by running the following two commands:
 
 ```sh
@@ -227,6 +230,8 @@ number goes up with every new release.
 We keep the application list in
 [src/firecfg/firecfg.config](src/firecfg/firecfg.config)
 (/etc/firejail/firecfg.config when installed).
+
+See `man firecfg` for details.
 
 ## Security profiles
 
--- a/src/man/firecfg.1.in
+++ b/src/man/firecfg.1.in
@@ -26,6 +26,11 @@ desktop managers are supported in this moment
 .PP
 Note: The examples use \fBsudo\fR, but \fBdoas\fR is also supported.
 .PP
+Note: Changes inside the user home (such as for desktop integration) are only
+made for the user running firecfg, so make sure to run "sudo firecfg" as a
+normal (desktop) user, not as root.
+See \fB\-\-fix\fR for details.
+.PP
 To set it up, run "sudo firecfg" after installing Firejail software.
 The same command should also be run after
 installing new programs. If the program is supported by Firejail, the symbolic link in /usr/local/bin


README.md
```diff
@@ -255,6 +260,9 @@ You can also use this tool to get a list of syscalls needed by a program:
 
 ## Uninstalling
 
+Note: To remove desktop integration for a given user, run `sudo firecfg
+--clean` as that user.
```
rusty-snake, line 264: IIRC we never delete in `~/.local/share/applications`. `--clean` only removes the symlinks. Therefore `sudo`-vs-root does not matter.

README.md
```diff
@@ -206,6 +206,9 @@ $ firejail --list
 
 ## Desktop integration
 
+Note: Desktop integration is only applied to the user running firecfg, so make
+sure to run `sudo firecfg` as a normal (desktop) user, not as root.
```
rusty-snake, line 210: The alternatives is

```console
# firecfg
# firecfg --add-users YOUR_USER
$ firecfg --fix
```

README.md
kmk3: > IIRC we never delete in `~/.local/share/applications`. `--clean` only removes
> the symlinks. Therefore `sudo`-vs-root does not matter.

Good catch; removed.

ginto37:
I'm not sure the phrase "normal (desktop) user" really helps here, and I'd suggest not going with that. My admin account, the one with `sudo` privileges, is a normal, desktop user, and so is my standard user account, the one without the `sudo` privileges and the one I work in 90% of the time. But only the admin account can run `sudo firecfg`. Also, most desktop users won't have direct access to root - Fedora, Ubuntu and Debian don't even enable it by default. So the average user will be working with just the admin account and the standard account, only one of which will work with that command.

I feel like there's a fundamental difference in views here regarding the use of accounts on Linux. Basically I feel like you're looking at this from a different angle and it's leading you to write up the instructions from a slightly unhelpful viewpoint that's going to confuse users rather than help them.

For that reason, Rusty-snake's suggestion should be the default method, not the alternative, IMO, otherwise users like myself are going to wonder why `sudo firecfg` won't work when run in an account which doesn't have `sudo` privileges for `firecfg`.

Rusty-snake's method doesn't involve modifying the `sudoers` file - which would be necessary for a non-admin account, and which you don't cover in the steps, despite it being crucial and capable of breaking the system if performed incorrectly - and encourages separation of admin and standard user accounts, which is good security practice.

In essence,

```
# firecfg
# firecfg --add-users <EACH USER WHO NEEDS TO USE FIREJAIL>
As each of those users, for those who want desktop integration:
$ firecfg --fix
```

OOC, where would `firecfg --fix-sound` come in that list? Does it also have to be run for each user intending to use Firejail?

rusty-snake:
> sudoers file

I still don't get why you want to modify it.

> OOC, where would firecfg --fix-sound come in that list? Does it also have to be run for each user intending to use Firejail?

Each user.

Keep in mind that firejails primary target are single user desktop systems.

ginto37:
OK, I'll start from the official installation method.

- The official README.md calls out two distros for explicit mention: Ubuntu and Debian.
 
- In [Ubuntu](https://linuxize.com/post/how-to-enable-and-disable-root-user-account-in-ubuntu/), the root account is disabled by default. On [Debian](https://wiki.debian.org/Root), it is only enabled if configured during installation. If it is, the official documentation still recommends using `sudo` instead of `su`ing to root for administration.

- It's generally recognized that using the root account is not recommended (see [here](https://www.ceos3c.com/linux/understanding-linux-root-user-a-complete-security-guide/#1-avoid-using-root-account-directly) and [here](https://superuser.com/questions/666942/why-it-is-not-recommend-to-use-root-login-in-linux)). For the same reasons, `sudo` rights should not be granted without reasonable justification. For example, the `sudo` manpage on Debian 12 says this:

> There is no easy way to prevent a user from gaining a root shell if that user is allowed to run arbitrary commands via sudo.

- If root is not enabled, there are only two account types available on Ubuntu and Debian for ordinary desktop users: Standard and Administrator. As security is the primary focus of `firejail`, let's assume that a Debian user didn't enable root during installation, and will be relying on `sudo`.

- On both distros, only Administrator user accounts can use `sudo` by default. This means that, by default, without making changes to `sudoers`, only an Administrator user account can run `sudo firecfg`. A Standard user trying to run it will see a message like this:

> Sorry, user foo is not allowed to execute '/usr/bin/firecfg' as root on ubuntu-desktop.

Again because security is the primary focus of `firejail`, let's assume a user on either distro is using a Standard user for his daily activities, and only switches to an Administrator account when necessary.

- As per the README.md, desktop integration for `firejail` requires running the following commands:

```
firecfg --fix-sound
sudo firecfg
```

- Any security-conscious user running those commands as a Standard user will not be able to run the second one, and will see the error message described above. This will prevent the creation of the files necessary for that particular user to launch applications from Activities view using Search (this applies to both Ubuntu and Debian).

- In attempting to fix this, the user can:

a) run `sudo firecfg` as an admin user. This would not actually create the desktop integration for the intended Standard user (this was my experience), so the user would then have to run `firecfg --fix` from the Standard user account

Or b) Give the Standard user general `sudo` privileges, which is not recommended (see above)

Or c) Give the Standard user `sudo` privileges for the `firecfg` command by modifying the `sudoers` file.

Of those three, the most secure and least complex option is, IMO, the first: a).

Ideally, and IMO, the steps for desktop integration should be described something like as follows:

1. As an administrator, run `firecfg` to create the symlinks:

```
sudo firecfg
```

2. Enable `firejail` for each user who needs to use it:

```
sudo firecfg --add-users foo bar
```

3. As each Standard user, run:

```
firecfg --fix
```

That's really as far as my thinking takes me. I'm not a `firejail` power user. I can't comment on whether this solves every conceivable problem. But I think it's an improvement on the current instructions which are almost guaranteed to make a user wonder why his desktop integration doesn't always work.

Going one step further, I'd like to suggest that `firecfg` should be modified so that it takes into consideration any users added with --add-users and automatically performs desktop integration for them when run as an administrator. The only time this might not work is if the system uses encrypted home directories (which is supported on Ubuntu by default in addition to full disk encryption). In that case, `firecfg --fix` would be the answer.

HTH.

rusty-snake:
> let's assume

To many axioms.

> Standard user for his daily activities, and only switches to an Administrator account when necessary.

I doubt that this is the common setup. The most users of firejail will have one user account. That is in the wheel group.

> c) Give the Standard user sudo privileges for the firecfg command by modifying the sudoers file.

Should be treated equal to "Or b) Give the Standard user general sudo privileges, which is not recommended (see above)".

> Ideally, and IMO, the steps for desktop integration should be described something like as follows:

Then send a PR.

> As an administrator, 

This could also be understood as "as root user" as a lot users thing "administrator = root" and "root = administrator".

> Enable firejail for each user who needs to use it:

Add on this, if you never do `--add-users`, all users are allowed to use firejail.

> firecfg --fix

> Going one step further, I'd like to suggest that firecfg should be modified so that it takes into consideration any users added with --add-users and automatically performs desktop integration for them when run as an administrator. The only time this might not work is if the system uses encrypted home directories (which is supported on Ubuntu by default in addition to full disk encryption). In that case, firecfg --fix would be the answer.

We could also implement a `firecfg --fix-system`.

ginto37:
The first sentence in the README.md:

> Firejail is a lightweight security tool intended to protect a Linux system _by setting up a restricted environment_...

As the devs behind `firejail`, it's counter-productive to promote `firejail` but also to tacitly and implicitly promote a fundamentally less secure initial working environment.

If you genuinely believe most of your users are working day-to-day in root or admin environments, then you have a great opportunity to educate them out of that position before recommending `firejail` as additional hardening.

kmk3:
Closing to clarify that the PR is not intended for merging in this state.

PR #6678, opened 2025-03-07
New profile: ouch

northboot:
CLI utility for easily compressing and decompressing files and dirs.

https://github.com/ouch-org/ouch

### DIFF
--- a/etc/profile-m-z/ouch.profile
+++ b/etc/profile-m-z/ouch.profile
@@ -1,0 +1,11 @@
+# Firejail profile for ouch
+# Description: CLI utility for easily compressing and decompressing files and dirs
+# This file is overwritten after every install/update
+quiet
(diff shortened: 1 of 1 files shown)


etc/profile-m-z/ouch.profile
```diff
@@ -0,0 +1,13 @@
+# Firejail profile for ouch
+# Description: CLI utility for easily compressing and decompressing files and dirs
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include ouch.local
+# Persistent global definitions
+include globals.local
+
+noroot
```
kmk3, line 10:
```suggestion
```

This is commented in archiver-common, probably to avoid issues with group
permissions.

kmk3 (approved):

Issue #6444, opened 2024-08-23
firefox: cannot drag and drop files from Dolphin


Utini2000:
### Description

Drag & Drop of files (e.g. adding attachments in a webmail) does not work with Firefox & KDE Dolphin.

### Steps to Reproduce

Open Firefox
Browser gmail
Create Mail
Drag & Drop a .pdf from /home/Downloads/ via Dolphin

### Expected behavior

Working drag & drop.

### Actual behavior

Drag & Drop not working

### Behavior without a profile

_What changed calling `LC_ALL=C firejail --noprofile firefox` in a
terminal?_
Then drag & drop will work fine

### Additional context

Nope :) 

### Environment

- 6.10.6-arch1-1
- Firejail version: 0.9.72

firefox.local:
```
private-bin bash,dbus-launch,dbus-send,env,firefox,sh,which,keepassxc-proxy
whitelist ${RUNUSER}/org.keepassxc.KeePassXC.BrowserServer
whitelist ${RUNUSER}/kpxc_server
noblacklist ${RUNUSER}/app
mkdir ${RUNUSER}/app/org.keepassxc.KeePassXC
whitelist ${RUNUSER}/app/org.keepassxc.KeePassXC

# Add the next line to your firefox.local to enable native notifications.
dbus-user.talk org.freedesktop.Notifications
# Add the next line to your firefox.local to allow inhibiting screensavers.
#dbus-user.talk org.freedesktop.ScreenSaver
# Add the next lines to your firefox.local for plasma browser integration.
dbus-user.own org.mpris.MediaPlayer2.plasma-browser-integration
dbus-user.talk org.kde.JobViewServer
dbus-user.talk org.kde.kuiserver
# Add the next line to your firefox.local to allow screen sharing under wayland.
dbus-user.talk org.freedesktop.portal.Desktop
# Add the next line to your firefox.local if screen sharing sharing still does not work
# with the above lines (might depend on the portal implementation).
ignore noroot
```

### Checklist

- [ x ] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [ x ] I can reproduce the issue without custom modifications (e.g. globals.local).
- [ x ] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [ x  ] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
[...]

PR #6679, opened 2025-03-07
New profile: xarchiver

northboot:
Lightweight desktop-independent archive manager.

https://github.com/ib/xarchiver

### DIFF
--- a/etc/profile-m-z/xarchiver.profile
+++ b/etc/profile-m-z/xarchiver.profile
@@ -1,0 +1,20 @@
+# Firejail profile for xarchiver
+# Description: Lightweight desktop-independent archive manager
+# This file is overwritten after every install/update
+# Persistent local customizations
+include xarchiver.local
+# Persistent global definitions
+include globals.local
+
+ignore include disable-shell.inc
+ignore x11
+ignore dbus-user none
+
+noblacklist ${HOME}/.config/xarchiver
+noblacklist ${RUNUSER}
+
+dbus-user filter
+dbus-user.talk ca.desrt.dconf
+
+# Redirect
+include archiver-common.profile
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -699,6 +699,7 @@ blacklist ${HOME}/.config/wesnoth
 blacklist ${HOME}/.config/wget
 blacklist ${HOME}/.config/wireshark
 blacklist ${HOME}/.config/wormux
+blacklist ${HOME}/.config/xarchiver
 blacklist ${HOME}/.config/xchat
 blacklist ${HOME}/.config/xed
 blacklist ${HOME}/.config/xfburn


etc/profile-m-z/xarchiver.profile
```diff
@@ -0,0 +1,21 @@
...
+include globals.local
+
+ignore blacklist ${RUNUSER}
+ignore dbus-user none
+ignore x11
+ignore include disable-shell.inc
+
+noblacklist ${HOME}/.config/xarchiver
+
+noroot
```
kmk3, line 16:
```suggestion
```

This is commented in archiver-common, probably to avoid issues with group
permissions.

etc/profile-m-z/xarchiver.profile
```diff
@@ -0,0 +1,21 @@
...
+# This file is overwritten after every install/update
+# Persistent local customizations
+include xarchiver.local
+# Persistent global definitions
+include globals.local
+
+ignore blacklist ${RUNUSER}
+ignore dbus-user none
+ignore x11
+ignore include disable-shell.inc
```
kmk3, line 12:
```suggestion
noblacklist ${RUNUSER}
ignore include disable-shell.inc
ignore x11
ignore dbus-user none
```

Sort/simplify.

etc/profile-m-z/xarchiver.profile
```diff
@@ -0,0 +1,19 @@
...
+include globals.local
+
+noblacklist ${RUNUSER}
+ignore include disable-shell.inc
+ignore x11
+ignore dbus-user none
+
+noblacklist ${HOME}/.config/xarchiver
+
+dbus-user.talk ca.desrt.dconf.service
```
rusty-snake, line 16: Do we ever enable filtering?

kmk3: > Do we ever enable filtering?

@northboot

You can check what dbus paths are used with d-feet or busctl, then add the
relevant filter/own (/talk) commands:

```
dbus-user filter
dbus-user.own foo.bar.baz
```

See for example:

* <https://github.com/netblue30/firejail/issues/6444#issuecomment-2322650453>

kmk3 (changes requested):
dbus:

* <https://github.com/netblue30/firejail/pull/6679#discussion_r1988712971>

etc/profile-m-z/xarchiver.profile
kmk3: > Updated the commands

Still missing a `dbus-user.own` entry.

rusty-snake: Does it need one?

kmk3: > Does it need one? / have one?

I was thinking that common GUI programs that use dbus would generally register
their own dbus name even if it's not currently used.

But now I see that engrampa.profile and some other profiles have .talk without
.own.

> No, it doesn't have one: `$ dbus-send --print-reply --session
> --dest=org.freedesktop.DBus /org/freedesktop/DBus
> org.freedesktop.DBus.ListNames | grep -i archiver` does not output anything
> (while running the application)

I see, thanks for confirming.

kmk3 (approved):

PR #6680, opened 2025-03-09
contrib/vim: add ftplugin file (based on cfg.vim)

dringsim:
Add a simple vim filetype plugin.


kmk3 (approved):
Very nice addition, thanks.

PR #6686, opened 2025-03-21
profiles: godot: ignore noexec in home to fix addons

NetSysFire:
I had weird errors when working on a godot project:

    addons/godot-sqlite/bin/libgdsqlite.linux.template_debug.x86_64.so: failed to map segment from shared object.
    ERROR: Can't open GDExtension dynamic library: 'res://addons/godot-sqlite/gdsqlite.gdextension'.

These addons are executable files. Addons like these are common enough
and `noexec` breaks them. I confirmed that the change fixes this error
and allows loading addons.

### DIFF
--- a/etc/profile-a-l/godot.profile
+++ b/etc/profile-a-l/godot.profile
@@ -12,7 +12,6 @@ noblacklist ${HOME}/.local/share/godot
 
 include disable-common.inc
 include disable-devel.inc
-include disable-exec.inc
 include disable-interpreters.inc
 include disable-programs.inc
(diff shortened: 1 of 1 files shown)


rusty-snake (changes requested):
Keep it and ignore the path that is necessary (likely `$HOME`).

NetSysFire:
That is a fair point to do it this way instead.

rusty-snake (approved):

kmk3 (approved):

PR #6687, opened 2025-03-23
profiles: split commands that increase/reduce access

kmk3:
This should clarify which commands do what (increase/reduce access) and
also make etc/templates/profile.template more consistent with
etc/profile-m-z/noprofile.profile.

Command used to search for the relevant profiles:

    git grep -El 'allow-debuggers|allusers|keep-|writable-' -- etc

This is a follow-up to #6660.

See also the discussion at [1].

[1] https://github.com/netblue30/firejail/pull/6660#discussion_r1975233984


kmk3:
@rusty-snake

Thoughts on merging this?

rusty-snake:
Looks fine to me.

PR #6683, opened 2025-03-16
profiles: floorp: add profile sync daemon paths

Renkoto:
Fix typo.

Denying read-write access to ${HOME}/.local/share/applications prevents Floorp from creating shortcuts to the application launcher, which breaks the (experimental) PWA feature on Linux.

Additionally i whitelisted the directories `${RUNUSER}/*floorp* `and
whitelist `${RUNUSER}/psd/*floorp* `to enable profile-sync-daemon support.

Since this is my first time doing a PR, I may have missed something, feel free to edit if needed.

PR #6689, opened 2025-03-23
profiles: firefox: add comment about creating PWA shortcuts

Renkoto:
This is used by at least floorp and FirefoxPWA.

This is a follow-up to #6683.

### DIFF
--- a/etc/profile-a-l/firefox-common-addons.profile
+++ b/etc/profile-a-l/firefox-common-addons.profile
@@ -2,6 +2,13 @@
 # Persistent customizations should go in a .local file.
 include firefox-common-addons.local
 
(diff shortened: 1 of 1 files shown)


kmk3:
@Renkoto

This would allow easily bypassing the sandbox, so I turned it into a comment
and added a warning.

Thoughts?

Renkoto:
> @Renkoto
> 
> This would allow easily bypassing the sandbox, so I turned it into a comment and added a warning.
> 
> Thoughts?

I see that everything is correct, Firefox is on the way to re-add native PWA support, I don't know if they will keep to use the same methods as Floorp and FirefoxPWA or rather they will go for a safer desktop integration like Chromium-based browsers does

kmk3 (approved):

Issue #6379, opened 2024-06-13
ssh: cannot access private key stored in TPM (private-dev)


qdii:
### Description

Security freaks like me do like to store their SSH private keys in protected places. Most laptops nowadays come with a TPM, which can store them (see https://jade.fyi/blog/tpm-ssh/ for instance).

### Steps to Reproduce

1. Follow the configuration guide for ssh with TPM2 linked above
2. Add the following configuration in `ssh.local`
   
   ```
   private-dev
   whitelist /usr/lib/libtss2-*
   ```

   Note: the SSH connection is also broken without adding these lines, but for a different reasons (no access to `/dev/tpm0`, no access to the libraries). 

3. Run `firejail ssh <some host>`

### Expected behavior

A SSH connection is opened to the host.

### Actual behavior

An error message appears and no connection is established.

### Behavior without a profile

A SSH connection is established.

### Environment

- Arch Linux
- Firejail version 0.9.72

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [x] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

```
$ firejail ssh $host
fexecve: No such file or directory
Error: failed to run /run/firejail/lib/fseccomp, exiting...
Error: proc 18455 cannot sync with peer: unexpected EOF
Peer 18456 unexpectedly exited with status 1
```

rusty-snake:
Instead of removing `private-dev` from profiles, we should enhance `private-dev`.

Proposal 1.

- `private-dev` includes TPM devices
- `notpm` blocks them

Proposal 2.

- `private-dev` supports subgroups.
  - `private-dev` only basic devices
  - `private-dev tpm,video,3d` basic with TPM devices, video devices and GPU.
- Unresolved to keep blacklisting approach.

glitsj16:
@qdii

[...]

PR #6698, opened 2025-04-01
modif: block TPM devices & turn notpm command into keep-dev-tpm

kmk3:
Instead of having a `notpm` command and potentially adding it to almost
all profiles (as few programs should need direct access to TPM devices),
add a `keep-dev-tpm` command and use it only in profiles that need
access to TPM devices.

Changes:

* Turn `notpm` command into `keep-dev-tpm` command
* Warn and ignore if `notpm` is used
* Block `/dev/tpm*` devices by default
* Allow `/dev/tpm*` devices with `keep-dev-tpm` (even if `private-dev`
  is used)

Added on commit 001320226 ("feature: add notpm command & keep tpm
devices in private-dev (#6390)", 2024-07-09).

See also commit ee1c264c5 ("feature: block /dev/ntsync & add
keep-dev-ntsync command (#6660)", 2025-03-06) and the discussion at
PR #6660.

This is a follow-up to #6687.

Relates to #6379 #6536.


src/firejail/main.c
```diff
@@ -2224,8 +2227,9 @@ int main(int argc, char **argv, char **envp) {
 			arg_notv = 1;
 		else if (strcmp(argv[i], "--nodvd") == 0)
 			arg_nodvd = 1;
+		// TODO: Fully remove notpm after 0.9.76.
 		else if (strcmp(argv[i], "--notpm") == 0)
-			arg_notpm = 1;
+			fwarning("ignoring removed command: --notpm (see --keep-dev-tpm)\n");
```
rusty-snake, line 2232: Just warn or hard fail afterwards?

kmk3: > Just warn or hard fail afterwards?

Both options seem valid, but I find it confusing when the program behaves
differently depending only on whether a command is used in the CLI or in a
profile.

For example, I remember one time discussing for a while about errors when
trying to use `private-lib` when building without `--enable-private-lib` until
someone clarified that the error only happens with `--private-lib` in the CLI
but not with `private-lib` in profiles, where it would just be ignored when not
supported.

In this case the `--notpm` command is a no-op (as what it does is made the
default) and when the remaining `notpm` sections are fully removed, trying to
use it will result in a hard fail anyway (unknown command).  Also, it seems
easier to reason about the relevant code/docs when they are consistent in all 4
places.

rusty-snake (approved):

PR #6704, opened 2025-04-10
feature: use globbing in hardcoded numbered /dev paths

kmk3:
Certain devices appear in numbered paths in /dev (such as `/dev/hidraw0`
or `/dev/hidraw1`) when they are plugged in.  When private-dev is used,
it attempts to include a hardcoded list of /dev paths in the sandbox.

Since the hardcoded paths only go up to a specific number (mostly from 0
to 9), devices that appear at higher numbered paths (such as
`/dev/hidraw10` or `/dev/hidraw20`) do not show up in /dev when
private-dev is enabled.

This issue also affects options that attempt to disable numbered paths
(such as --no3d and --nou2f), independently of whether private-dev is
used.

Fix this by using glob patterns (such as `/dev/hidraw[0-9]*`) to ensure
that all relevant numbered paths are matched regardless of number.

The globbing is similar to the one from commit 2993298aa ("firecfg:
parse config files in /etc/firejail/firecfg.d", 2023-06-29) / PR #5876.

Closes #2723.

Reported-by: @tsankuanglee
Reported-by: @WPettersson

Issue #6688, opened 2025-03-23
cannot blacklist keyboard/mouse (multiseat)


Sx1Fl0NdqlkM6N4s2pswXKUI0k9FKTaOckLbNXu:
### Description

Adding keyboard or mouse to blacklist doesn't work.

### Steps to Reproduce

```sh
firejail --noprofile --blacklist="/dev/input/*" \
  --blacklist="/dev/input/by-id/*" \
  --blacklist="/dev/input/by-path/*" \
  --blacklist="/dev/input/event*" application
```

or pointing out devices individually, or using `--noinput` flag also shows no results.

### Additional context

I'd assume it's an intended feature for most users, with no way to turn it off.
My use-case is a multiseat setup where kb&m inputs won't be sent to all selected windows.

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [ ] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [ ] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
- [ ] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ ] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

rusty-snake:
> firejail --noprofile --blacklist="/dev/input/" --blacklist="/dev/input/by-id/" --blacklist="/dev/input/by-path/" --blacklist="/dev/input/event" application

***What did you expected to happen?***

Who did you checked? if blacklisting does not work, this would be a security problem. But I don't believe.

Note that neither GUI nor TUI programs read raw kb/m input from device files.

Sx1Fl0NdqlkM6N4s2pswXKUI0k9FKTaOckLbNXu:
> _**What did you expected to happen?**_
Formatting have eaten stars after every directory.
[...]

PR #6707, opened 2025-04-13
profiles: godot: remove noinput so gamepads work

NetSysFire:
As you might know, Godot is a game engine. For this reason, they support the very commonly used gamepads and other potentially exotic input devices. `noinput` is incompatible with that and interferes with normal operation.

### DIFF
--- a/etc/profile-a-l/godot.profile
+++ b/etc/profile-a-l/godot.profile
@@ -26,7 +26,7 @@ caps.drop all
 netfilter
 nodvd
 nogroups
-noinput
+#noinput # may break gamepads (see #6707)
 nonewprivs
 noroot
 notv
--- a/etc/profile-a-l/godot.profile
+++ b/etc/profile-a-l/godot.profile
@@ -26,7 +26,7 @@ caps.drop all
 netfilter
 nodvd
 nogroups
-#noinput # may break gamepads (see #6707)
+#noinput # breaks gamepads (see #6707)
 nonewprivs
 noroot
 notv


kmk3 (commented):
> As you might know, Godot is a game engine. For this reason, they support the
> very commonly used gamepads and other potentially exotic input devices.
> `noinput` is incompatible with that and interferes with normal operation.

Can you give an example of such an input device and which path godot accesses
in /dev/input?

From what I've seen so far it's usually enough to use `ignore nou2f` to allow
`/dev/hidraw*`.

Now that #6704 has been merged, does it work with just `ignore nou2f` in
[firejail-git](https://github.com/netblue30/firejail?tab=readme-ov-file#building)?

Edit: I see now that steam.profile and lutris.profile do not have `noinput`
either, but wine.profile does.  If disabling it in godot.profile is needed, it
would probably also make sense to do it for wine.profile.  If not, maybe
`noinput` could be added to the other profiles as well.

Relates to:

* #6534
* #6704
* <https://github.com/netblue30/firejail/issues/6688#issuecomment-2747210433>

NetSysFire:
For example, /dev/input/js0.

kmk3:
> For example, /dev/input/js0.

Good point, I forgot about the `/dev/input/js*` paths.

kmk3 (approved):

kmk3:
> Well, it _does_ break gamepads, regarding your comment. It is not a question
> of may or may not, it will always do that.

I used "may break" because Godot could release a new version that supports some
(but not all) gamepads through `/dev/hidraw*` (which is not affected by
`noinput`), in which case the "may break" comment would remain correct (the
same is done for the `nou2f` comment in lutris.profile and steam.profile).

> Pretty much all the traditional gamepads are always under `/dev/input/js*` as
> far as I can see. Dedicated retro joysticks may be somewhere else and there
> are other more exotic gamepads, like those guitars or guns, so yes, `nou2f`
> does not suffice.

So I had a look at the kernel docs to clarify the paths:

From <https://github.com/torvalds/linux/blob/v6.14/Documentation/input/joydev/joystick-api.rst>:

> This document describes legacy `js` interface. Newer clients are encouraged
> to switch to the generic event (`evdev`) interface.

> /dev/input/js0

From <https://github.com/torvalds/linux/blob/v6.14/Documentation/input/input.rst>:

> `evdev` is the generic input event interface. It passes the events generated
> in the kernel straight to the program, with timestamps. The event codes are
> the same on all architectures and are hardware independent.
> 
> This is the preferred interface for userspace to consume user input, and all
> clients are encouraged to use it.

> /dev/input/eventX

My conclusion:

`/dev/input/js*` is the path for the older API that only supports joysticks and
`/dev/input/event*` is the path for the newer API that also supports other
device types.  Where the gamepad appears seemingly depends on how support is
implemented for it in the kernel (if it is supported).

From <https://github.com/torvalds/linux/blob/v6.14/Documentation/hid/hidraw.rst>:

> The hidraw driver provides a raw interface to USB and Bluetooth Human
> Interface Devices (HIDs).  It differs from hiddev in that reports sent and
> received are not parsed by the HID parser, but are sent to and received from
> the device unmodified.
> 
> Hidraw should be used if the userspace application knows exactly how to
> communicate with the hardware device, and is able to construct the HID
> reports manually.  This is often the case when making userspace drivers for
> custom HID devices.

> /dev/hidraw0

Modern gamepads usually support USB/Bluetooth HID and show up in
`/dev/hidraw*`, in which case no kernel driver is needed if the application can
deal with the device directly (userspace driver).

For programs that use SDL, it supports some common gamepads directly through
hidraw (besides the js/evdev interfaces):

From <https://github.com/libsdl-org/SDL/blob/release-3.2.10/src/hidapi/README.md>:

> ### HIDAPI has four back-ends:
> * Windows (using `hid.dll`)
> * Linux/hidraw (using the Kernel's hidraw driver)
> * libusb (using libusb-1.0 - Linux/BSD/other UNIX-like systems)
> * macOS (using IOHidManager)

> #### __Linux/hidraw__ (`linux/hid.c`):
> 
> This back-end uses the hidraw interface in the Linux kernel, and supports
> both USB and Bluetooth HID devices. It requires kernel version at least
> 2.6.39 to build. In addition, it will only communicate with devices which
> have hidraw nodes associated with them.
> 
> Keyboards, mice, and some other devices which are blacklisted from having
> hidraw nodes will not work. Fortunately, for nearly all the uses of hidraw,
> this is not a problem.

See also:

* <https://github.com/libsdl-org/SDL/blob/release-3.2.10/src/joystick/usb_ids.h>
* <https://github.com/libsdl-org/SDL/tree/release-3.2.10/src/joystick/hidapi>

> > Edit: I see now that steam.profile and lutris.profile do not have noinput
> > either
> 
> Yeah this can cause trouble and may need to be adjusted. I use none of those
> so I can not say. I do use wine however and my gamepads work as expected
> there.

It seems that steam and wine (and many games) use SDL (which can use hidraw),
while godot doesn't, which might explain the difference in gamepad support.

Other related links:

* <https://github.com/torvalds/linux/blob/v6.14/Documentation/hid/hiddev.rst>
* <https://github.com/torvalds/linux/blob/v6.14/Documentation/hid/hidintro.rst>
* <https://github.com/torvalds/linux/blob/v6.14/Documentation/input/gamepad.rst>
* <https://github.com/torvalds/linux/blob/v6.14/Documentation/input/joydev/joystick.rst>
* <https://github.com/torvalds/linux/blob/v6.14/drivers/hid/hid-ids.h>
* <https://github.com/torvalds/linux/blob/v6.14/drivers/input/joydev.c>

kmk3:
@NetSysFire

Thoughts on merging this?

NetSysFire:
I updated the comment, as a drivers or a godot version that uses hidraw will be in the far future, this can be squashed and merged. I already use this change locally.

PR #6710, opened 2025-04-15
feature: add warn command

kmk3:
And use it in etc/inc/disable-X11.inc.

This allows printing a warning message from inside a profile.

Everything after the command is printed as is to stderr.

Relates to #6294.

This is a follow-up to #6709.


kmk3:
@rusty-snake

Thoughts on merging this?

PR #6756, opened 2025-05-20
profiles: centralize nvm allow line (archivers/hashers)

kmk3:
For simplicity and to avoid having to update the profiles if nvm changes
which archiver/hasher it uses, move the entry from a specific
archiver/hasher into archiver-common and hasher-common.

Related commits:

* 713249c98 ("sha256sum: add nvm support comment", 2022-03-20) /
  PR #5058
* d96cf4c8c ("tar: add nvm support comment", 2022-03-20) /
  PR #5058


kmk3:
Ignore for now.

Issue #6729, opened 2025-05-01
Deadlock on /run/firejail/firejail-run.lock when a firejailed process is suspended


hlein:
### Description

If a firejail'ed process is `^Z`'d at the wrong time during startup, other firejail processes will be unable to proceed because they'll wait forever for a flock on `/run/firejail/firejail-run.lock`

### Steps to Reproduce

1. Run something firejail'ed that will take a while to start. In my case I was running a `make` target that `ssh`'ed to a remote host via a SOCKS proxy listener that was dead.
2. control-z the parent (the `make` in my case)
3. in another terminal try to run any firejail'ed thing
4. never see startup messages...

Note, this is racy (interrupting between lock and unlock) so I don't expect it to be reproducible every time.

### Expected behavior

Separate/independent firejails should proceed, or if not, perhaps output an error (upon timeout?).

### Actual behavior

Any new firejail process hangs indefinitely at startup.

### Behavior without a profile

I think as long as we grab a lock on `/run/firejail/firejail-run.lock`, and allow `SIGSTP` to interrupt while we hold the lock, we are susceptible.

### Additional context

Maybe we can block `SIGSTP` in the time between `preproc_lock_firejail_dir()` and `preproc_unlock_firejail_dir()`?  And/or use `LOCK_NB` and busy-wait a reasonable amount of time so we can abort.

### Environment

- Name/version/arch of the Linux kernel (`uname -srm`): Linux 6.6.83-gentoo x86_64
- Name/version of the Linux distribution (e.g. "Ubuntu 20.04" or "Arch Linux"): Gentoo Linux
- Name/version of the relevant program(s)/package(s) (e.g. "firefox 134.0-1,
  mesa 1:24.3.3-2"): openssh-10.0_p1 (but I think it's not dependent on that)
- Version of Firejail (`firejail --version`): 0.9.74
- If you use a development version of firejail, also the commit from which it
  was compiled (`git rev-parse HEAD`):

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
[...]

PR #6761, opened 2025-05-28
feature: use non-blocking flock calls

jlimor-kl:
As reported by @hlein in #6729, if a firejailed process is `^Z`'d at the
wrong time during startup, other firejail processes will be unable to
proceed because they'll wait forever for an flock on
`/run/firejail/firejail-run.lock`.

This builds upon commit f4b8c6dbb ("bugfix: fix potential deadlock with
flock + SIGTSTP (#6750)", 2025-05-21) to add a busy wait while
attempting to acquire a lock. This allows the application to bail out if
it fails to acquire a lock after a given timeout.

Behavior after this patch (we use SIGSTOP, as that cannot be
caught/ignored):

    $ firejail --debug id & P=$!;sleep 0.00001;kill -STOP $P
    [1] 14411
    Looking for kernel processes
    Found kthreadd process, we are not running in a sandbox
    pid=14411: locking /run/firejail/firejail-run.lock ...
    pid=14411: locked /run/firejail/firejail-run.lock

    [1]+  Stopped                 firejail --debug id
    $ firejail --debug ls
    Looking for kernel processes
    Found kthreadd process, we are not running in a sandbox
    pid=14413: locking /run/firejail/firejail-run.lock ...
    pid=14413: sleeping 500us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 1000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 2000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 4000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 8000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 16000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 32000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 64000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 128000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 256000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 500000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 500000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 500000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 500000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 500000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 500000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 500000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 500000us while trying to lock /run/firejail/firejail-run.lock
    pid=14413: sleeping 500000us while trying to lock /run/firejail/firejail-run.lock
    Error: timeout occurred while trying to lock /run/firejail/firejail-run.lock
    Error: ../../src/firejail/preproc.c:112: preproc_lock_file: flock: Resource temporarily unavailable

Enhances commit f4b8c6dbb ("bugfix: fix potential deadlock with flock +
SIGTSTP (#6750)", 2025-05-21).

Suggested-by: @hlein
Co-authored-by: @kmk3

### DIFF
--- a/src/firejail/preproc.c
+++ b/src/firejail/preproc.c
@@ -23,8 +23,13 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <dirent.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <signal.h>
+
+static const useconds_t    LOCK_INITIAL_SLEEP_USEC =  500U;
+static const useconds_t    LOCK_MAX_SLEEP_USEC =   500000U;
+static const unsigned long LOCK_TIMEOUT_USEC =    5000000U;
 
 static int tmpfs_mounted = 0;
 static volatile sig_atomic_t caught_tstp = 0;
@@ -97,9 +102,31 @@ static void preproc_lock_file(const char *path, int *lockfd_ptr) {
 		errExit("fchown");
 	}
 
-	if (flock(lockfd, LOCK_EX) == -1) {
-		fprintf(stderr, "Error: cannot lock %s\n", path);
-		errExit("flock");
+	unsigned long sleep_usec = LOCK_INITIAL_SLEEP_USEC;
+	unsigned long wait_total = 0U;
+
+	while (flock(lockfd, LOCK_EX | LOCK_NB) == -1) {
+		if (errno == EWOULDBLOCK) {
+			if (wait_total >= LOCK_TIMEOUT_USEC) {
+				fprintf(stderr, "Error: timeout occurred while trying to lock %s\n", path);
+				errExit("flock");
+			}
+
+			if (arg_debug) {
+				printf("pid=%ld: sleeping %luus while trying to lock %s\n",
+				       pid, sleep_usec, path);
+			}
+
+			usleep(sleep_usec);
+			wait_total += sleep_usec;
+
+			sleep_usec *= 2U;
+			if (sleep_usec > LOCK_MAX_SLEEP_USEC)
+				sleep_usec = LOCK_MAX_SLEEP_USEC;
+		} else {
+			fprintf(stderr, "Error: cannot lock %s\n", path);
+			errExit("flock");
+		}
 	}
 
 	*lockfd_ptr = lockfd;


kmk3:
@spiiroin

Any thoughts on the PR and on a timeout of 5 seconds?

From #6307:

> There are reports of firejail sandboxed applications occasionally taking long
> time (12 seconds) to start up. When this happens, it affects all sandboxed
> applications until the device is rebooted.

So before using locks it could take up to 12 seconds.

Any idea on what is the longest time that an instance spends waiting for a
lock?

> The reason for the slowdown seems to be a timing hazard in the way remounts
> under /run/firejail are handled. This gets triggered when multiple firejail
> processes are launched in parallel as part of user session bring up and
> results in some, dozens, hundreds, or even thousands of stray
> /run/firejail/xxx mounts. The amount of mount points then affects every mount
> operation that is done during sandbox filesystem construction.

Are there hundreds of sandboxes being started in parallel or was this just
about accumulated stray mounts over time?

Any idea on how many sandboxes are started in parallel on a normal system and
on a system with a lot of things installed (/worst case scenario)?

kmk3:
@jlimor-kl

Hello, are you still interested in this PR?

I can fix the issues and merge it if you want.

jlimor-kl:
Ah yes, I am still interested in working on this, had to put it on the back burner for a bit as I had other things I was working on.

jlimor-kl:
did some research to see if the error output could be improved by showing what process holds the lock, but it seems like `flock` is not implemented in a way that is queryable. Assuming there are no other suggestions, this should be ready to merge.

kmk3 (approved):
I squashed the commits, re-ran the code to use the updated values in the commit
message and formatted it.

Looks good now.

Some suggestions for future PRs:

> feature: use non-blocking flock calls (#6750)

Avoid manually adding a reference to the end of the commit message title, as it
can be confused with the commit's actual PR reference (which is automatically
added by GitHub when using "squash and merge" to merge the commit on the web
UI).

With "squash and merge", it would have ended up as the following:

> feature: use non-blocking flock calls (#6750) (#6761)

Without "squash and merge", it would have ended up as the following:

> feature: use non-blocking flock calls (#6750)

Instead of the intended:

> feature: use non-blocking flock calls (#6761)

---

>     ```sh
>     $ firejail --debug id & P=$!;sleep 0.00001;kill -STOP $P
>     [1] 14411
>     Looking for kernel processes
>     [...]
>     ```

Instead of backticks, use 4 spaces to indent code blocks to make it more
readable as plaintext in the terminal:

```
    $ firejail --debug id & P=$!;sleep 0.00001;kill -STOP $P
    [1] 14411
    Looking for kernel processes
    [...]
```

kmk3:
> Ah yes, I am still interested in working on this, had to put it on the back
> burner for a bit as I had other things I was working on.

All good.

> did some research to see if the error output could be improved by showing
> what process holds the lock, but it seems like `flock` is not implemented in
> a way that is queryable. Assuming there are no other suggestions, this should
> be ready to merge.

That would be nice, though indeed it can be left for later.

I see that `lslocks(1)` in the CLI shows the PIDs/paths for all locks.

Maybe `fcntl(3p)` can do the same in the code, though in that case I think that
`flock(3)` calls would first have to be converted into `lockf(3p)` calls.

kmk3:
The current code seems good enough, but for completeness, there may be another
potential improvement if you are interested in looking into it (though I'm not
sure if the complexity is worth it):

It seems that `usleep(3)` was deprecated / removed from POSIX:

<details>
<summary>usleep(3)</summary>
<p>

```
STANDARDS
       None.

HISTORY
       4.3BSD, POSIX.1-2001.  POSIX.1-2001 declares it obsolete, suggesting
       nanosleep(2) instead.  Removed in POSIX.1-2008.

       On the original BSD implementation, and before glibc 2.2.2, the return
       type of this function is void.  The POSIX version returns int, and this
       is also the prototype used since glibc 2.2.2.

       Only the EINVAL error return is documented by SUSv2 and POSIX.1-2001.

CAVEATS
       The interaction of this function with the SIGALRM signal, and with
       other timer functions such as alarm(2), sleep(3), nanosleep(2),
       setitimer(2), timer_create(2), timer_delete(2), timer_getoverrun(2),
       timer_gettime(2), timer_settime(2), ualarm(3) is unspecified.
```

</p>
</details>

It suggests using `nanosleep(3p)` instead, in which case it may be necessary to
manually do arithmetic converting between seconds and nanoseconds.  Also, it
might need to be done using 64-bit integers instead of 32/64-bit.

And even then, `nanosleep` may be impacted by signals:

<details>
<summary>nanosleep(2)</summary>
<p>

```
VERSIONS
       POSIX.1 specifies that nanosleep() should measure time against the
       CLOCK_REALTIME clock.  However, Linux measures the time using the
       CLOCK_MONOTONIC clock.  This probably does not matter, since the
       POSIX.1 specification for clock_settime(2) says that discontinuous
       changes in CLOCK_REALTIME should not affect nanosleep():

              Setting the value of the CLOCK_REALTIME clock via
              clock_settime(2) shall have no effect on threads that are
              blocked waiting for a relative time service based upon this
              clock, including the nanosleep() function; ...  Consequently,
              these time services shall expire when the requested duration
              elapses, independently of the new or old value of the clock.

STANDARDS
       POSIX.1-2008.

HISTORY
       POSIX.1-2001.
[...]
BUGS
       If a program that catches signals and uses nanosleep() receives signals
       at a very high rate, then scheduling delays and rounding errors in the
       kernel's calculation of the sleep interval and the returned remain
       value mean that the remain value may steadily increase on successive
       restarts of the nanosleep() call.  To avoid such problems, use
       clock_nanosleep(2) with the TIMER_ABSTIME flag to sleep to an absolute
       deadline.

       In Linux 2.4, if nanosleep() is stopped by a signal (e.g., SIGTSTP),
       then the call fails with the error EINTR after the thread is resumed by
       a SIGCONT signal.  If the system call is subsequently restarted, then
       the time that the thread spent in the stopped state is not counted
       against the sleep interval.  This problem is fixed in Linux 2.6.0 and
       later kernels.
```

</p>
</details>

In which case, to avoid that one would have to use `clock_nanosleep(3p)`:

<details>
<summary>clock_nanosleep(2)</summary>
<p>

```
DESCRIPTION
       Like nanosleep(2), clock_nanosleep() allows the calling thread to sleep
       for an interval specified with nanosecond precision.  It differs in
       allowing the caller to select the clock against which the sleep
       interval is to be measured, and in allowing the sleep interval to be
       specified as either an absolute or a relative value.
[...]
STANDARDS
       POSIX.1-2008.

HISTORY
       POSIX.1-2001.  Linux 2.6, glibc 2.1.
[...]
NOTES
       If the interval specified in t is not an exact multiple of the
       granularity underlying clock (see time(7)), then the interval will be
       rounded up to the next multiple.  Furthermore, after the sleep
       completes, there may still be a delay before the CPU becomes free to
       once again execute the calling thread.

       Using an absolute timer is useful for preventing timer drift problems
       of the type described in nanosleep(2).  (Such problems are exacerbated
       in programs that try to restart a relative sleep that is repeatedly
       interrupted by signals.)  To perform a relative sleep that avoids these
       problems, call clock_gettime(2) for the desired clock, add the desired
       interval to the returned time value, and then call clock_nanosleep()
       with the TIMER_ABSTIME flag.

       clock_nanosleep() is never restarted after being interrupted by a
       signal handler, regardless of the use of the sigaction(2) SA_RESTART
       flag.

       POSIX.1 specifies that clock_nanosleep() has no effect on signals
       dispositions or the signal mask.
```

</p>
</details>

PR #6779, opened 2025-06-15
profiles: ani-cli: add mpv to private-etc for plugins access

sudoAlphaX:
Allow ani-cli to access /etc/mpv for mpv plugins/themes/scripts access.

mpv playing under ani-cli cannot use plugins.
Example, unable to use mpv-mpris plugin when playing anime using ani-cli.

This patch just adds mpv to private-etc for plugin access.
The profile imports mpv.profile later so all that magic is taken care of.

To note that in mpv.profile, there is no private-etc definition, but it is defined in ani-cli.profile.
Thats why it was broken.


sudoAlphaX:
On that note, I would like to request addition of private-etc to mpv.profile. I don't know what exactly to allow in /etc for mpv, so it will be great if someone can figure it out.

sudoAlphaX:
On a second thought, it is better to just add a proper private-etc to mpv.profile. This patch is just a hack until that is fixed. If a private-etc definition exists in mpv.profile, this can be reverted.

kmk3:
> On that note, I would like to request addition of private-etc to mpv.profile.
> I don't know what exactly to allow in /etc for mpv, so it will be great if
> someone can figure it out.

Usually you can run firejail with `--trace=` and it will say which paths the
program tries to access.

If you could test mpv with the plugins/themes/scripts that you use, check the
trace output and post the relevant paths (or open a PR), that would be very
helpful.

Example:

```sh
firejail --trace=trace.txt /usr/bin/mpv
```

For example, IIRC yt-dlp has a path in /etc.

By the way, I was working on adding the missing mpv paths (and other related
paths) to profiles a while back, though there are many profiles involved and I
wasn't sure if the paths were complete enough, so it's still wIP.

> On a second thought, it is better to just add a proper private-etc to
> mpv.profile. This patch is just a hack until that is fixed. If a private-etc
> definition exists in mpv.profile, this can be reverted.

I think it's fine to merge this patch as is and then try a more encompassing
fix later.

kmk3 (approved):

sudoAlphaX:
> Usually you can run firejail with --trace= and it will say which paths the
program tries to access.

Alright, ill give it a try. Also, is using `--build` okay rather than `--trace`?

kmk3:
> > Usually you can run firejail with --trace= and it will say which paths the
> > program tries to access.
> 
> Alright, ill give it a try. Also, is using `--build` okay rather than
> `--trace`?

`--build=` can be useful as a starter, though many of the profile lines it
generates are hard-coded and IIRC it misses things compared to `--trace=`.  Not
sure about /etc paths.

See src/fbuilder for details.

PR #6789, opened 2025-06-18
build: sort.py: replace items in private-etc with groups

sudoAlphaX:
Modifies sort.py such that when processing private-etc, it removes all
the default automatically included items, and if all items from any
particular group are present, it replaces those private-etc items with
the @group_name.

Uses regex to directly parse etc_groups.h header file

Related: #6783 and 02ad655

### DIFF
--- a/contrib/sort.py
+++ b/contrib/sort.py
@@ -7,6 +7,7 @@
 #  python >= 3.6
 from os import path
 from sys import argv, exit as sys_exit, stderr
+from trim_private_etc import optimize_etc
 
 __doc__ = f"""\
 Strip whitespace and sort the arguments of commands in profiles.
@@ -85,8 +86,10 @@ def check_profile(filename, overwrite):
         fixed_profile = []
         for lineno, original_line in enumerate(lines, 1):
             line = original_line.rstrip()
-            if line[:12] in ("private-bin ", "private-etc ", "private-lib "):
+            if line[:12] in ("private-bin ", "private-lib "):
                 line = f"{line[:12]}{sort_alphabetical(line[12:])}"
+            elif line[:12] == "private-etc ":
+                line = f"{line[:12]}{sort_alphabetical(optimize_etc(line[12:]))}"
             elif line[:13] in ("seccomp.drop ", "seccomp.keep "):
                 line = f"{line[:13]}{sort_alphabetical(line[13:])}"
             elif line[:10] in ("caps.drop ", "caps.keep "):
--- a/contrib/trim_private_etc.py
+++ b/contrib/trim_private_etc.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python3
+
+import re
+import sys
+import os
+
+# C header file containing the char* arrays (relative to the current script)
+etc_groups_file = os.path.join(os.path.dirname(__file__), "../src/include/etc_groups.h")
+default_group = "etc_list"
+
+
+# Regex pattern to capture the array name and its content (in one string).
+default_group_array_pattern = re.compile(
+    r"static\s+char\s*\*\s*"  # Matches "static char *"
+    r"etc_list\s*\[.*?\]\s*=\s*\{"  # Matches default group "etc_list"
+    r"(.*?)"  # Non-greedily captures the entire content (group 1)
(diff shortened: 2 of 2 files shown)


sudoAlphaX:
yea i hate regex

sudoAlphaX:
Example:

Initially: `private-etc @tls-ca,@x11,mailcap,pulse,mime.types,os-release,alternatives,openal,timidity.cfg,alsa,asound.conf,timidity,machine-id,pipewire`

After:
`private-etc @games,@sound,@tls-ca,@x11,mailcap,mime.types,os-release`

* Removed `alternatives` (because automatically included)
* Replaced `openal`, `timidity`, `timidity.cfg` to `@games`
* Replaced `alsa`, `asound.conf`, `machine-id`, `pipewire`, `pulse` with `@sound`
* Sorted (by sort.py)

### Try it out:

```diff
diff --git a/etc/profile-a-l/firefox-common.profile b/etc/profile-a-l/firefox-common.profile
index d55f9da0a..ed2010ab1 100644
--- a/etc/profile-a-l/firefox-common.profile
+++ b/etc/profile-a-l/firefox-common.profile
@@ -86,7 +86,7 @@ disable-mnt
 ?BROWSER_DISABLE_U2F: private-dev
 # Note: The private-etc line below works fine on most distributions but it
 # could cause problems on CentOS.
-private-etc @tls-ca,@x11,mailcap,mime.types,os-release
+private-etc @tls-ca,@x11,mailcap,pulse,mime.types,os-release,alternatives,openal,timidity.cfg,alsa,asound.conf,timidity,machine-id,pipewire
 private-tmp
 
 # Note: `dbus-user none` breaks various desktop integration features like
```

```sh
$ ./sort.py ../etc/profile-a-l/firefox-common.profile;
```

sudoAlphaX:
lmk if I could add a tolerance (kinda) feature too.

Example: in a `private-etc` definition, there is openal and timidity, but no timidity.cfg. But commonly, they all should be together. So we can set tolerance as an integer and if there are at-most n files missing from private-etc which could belong in a group, we just go ahead and add the group.

Example 2:

`@x11` contains `ati`, `dconf`, `drirc`, `gtk-2.0`, `gtk-3.0`, `kde4rc`, `kde5rc`, `machine-id`, `nvidia`, `pango`, `Trolltech.conf`, `vulkan`, `X11`, `xdg`.

`private-etc` of a profile contains: `ati`, `dconf`, `drirc`, `gtk-2.0`, `gtk-3.0`, `kde4rc`, `kde5rc`, `machine-id`, `nvidia`, `pango`, `Trolltech.conf`, `vulkan`.

No `xdg` and no `x11`, but except those two, everything else from `@x11` group belong in the `private-etc`. So, if "tolerance" is 2, we just add `@x11` group because it is likely an x11 app. But if "tolerance" is 1, we dont.

I can integrate it in sort.py or just have it as a standalone feature in the separate python script.

sudoAlphaX:
Ah I see now.

Thank you very much for explaining it to me in detail.

I now understand why it might not be the best idea to always aggregate and group individual private-etc items into groups.

Also, i havent thought about the regex overload (spoiled python developer)

Ill see what i can do about the last point.

Feel free to close this PR. Thanks!

kmk3:
> Ah I see now.
> 
> Thank you very much for explaining it to me in detail.
> 
> I now understand why it might not be the best idea to always aggregate and
> group individual private-etc items into groups.

Glad it helped.

I had an idea about adding a flag to each item in each group to say if it's a
path that is unique to/owned by the group (and thus directly related to the
group) or if it's just a (generic) path that the group happens to need.

Then maybe also run a check to see if all such paths are not duplicated in
other groups.

I might try to add that later and see if it would make sense.

> Also, i havent thought about the regex overload (spoiled python developer)

I'm not sure if it would affect performance or not, but note that sort.py
currently hardcodes some string lengths to try to help with that.

If the difference is negligible, it might make sense to use it in sort.py
itself.

I had a WIP branch that enabled sort.py to format commented lines using regex;
IIRC it added about 20-50ms.

> Ill see what i can do about the last point.

Thanks.

Any improvement to make it generate more accurate/less hardcoded profiles would
be very nice (even if it's unrelated to private-etc).

> Feel free to close this PR. Thanks!

Done.

PR #6812, opened 2025-07-10
profiles: rtv: allow x11 and wayland

ghost:
it blocks media viewers on X

EDIT: and allow wayland

### DIFF
--- a/etc/profile-m-z/rtv.profile
+++ b/etc/profile-m-z/rtv.profile
@@ -6,8 +6,6 @@ include rtv.local
 # Persistent global definitions
 include globals.local
 
-blacklist ${RUNUSER}/wayland-*
-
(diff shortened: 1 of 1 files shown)


kmk3:
This was archived 2 years ago:

* https://github.com/michael-lazar/rtv

And I think that the website API was rate-limited after that.

Does it still work?

Is there a new project URL?

ghost:
Yea it's abandoned.
you can't login any more and edit, but still works as a reader.
It will break for good, whenever reddit updates it's API.

kmk3:
The changes make sense, why close this?

ghost:
it's fundamentay a terminal application, it could stay like that and require the user to open stuff up if they need them.

PR #6818, opened 2025-07-11
build: debian: packaging improvements

### LINKED ISSUE #6811: build: debian packaging minor fixes
add `firejail-profiles` as `provides` and `conflicts`

remove `python` from `suggests`, (`python3` is fine)

ghost:
fixes for the deb packages. Didn't test them, they should work.

Closes #6811.

the provides and conflicts will cause previous installed firejail-profiles to uninstall. So the package just installs
I copied the depends and recommends from the debian package. Maybe they are differences in the compile options.
in suggest, debian has python3, not python.

test that it actually installs.

### DIFF
--- a/config.mk.in
+++ b/config.mk.in
@@ -54,6 +54,7 @@ HAVE_X11=@HAVE_X11@
 
 # Dependency strings for debian control files
 DEPENDS=@DEPENDS@
+RECOMMENDS=@RECOMMENDS@
 
 MANFLAGS = \
 	$(HAVE_APPARMOR) \
--- a/configure
+++ b/configure
@@ -650,6 +650,7 @@ ac_includes_default="\
 ac_header_c_list=
 ac_subst_vars='LTLIBOBJS
 LIBOBJS
+RECOMMENDS
 DEPENDS
 HAVE_ONLY_SYSCFG_PROFILES
 HAVE_FORCE_NONEWPRIVS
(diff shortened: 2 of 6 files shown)


kmk3 (changes requested):
> ```diff
> +Provides: firejail-profiles
> +Conflicts: firejail-profiles
> ```

> the provides and conflicts will cause previous installed firejail-profiles to
> uninstall. So the package just installs

Looks good.

For future reference, in the PPA they are packaged separately (firejail /
firejail-profiles), but here the profiles are part of the same .deb.

> ```diff
> -Suggests: python, python3
> +Suggests: python3
> ```

Looks good.

> ```diff
> -Depends: libc6
> +Depends: libapparmor1, libc6, libselinux1
> ```

These depend on the configuration (and selinux is usually not used on Debian
AFAIK).

In order to properly support this and integrate it with the existing configure
options, the control files should become input files and use the existing
macros from config.mk.in.

> I copied the depends and recommends from the debian package. Maybe they are
> differences in the compile options.

Note that the hardcoding works for the Debian package because it always uses
specific options, while this is a more generic recipe.

> ```diff
> +Recommends: iproute2, iptables, xauth, xdg-dbus-proxy, xpra | xserver-xephyr | xvfb
> ```

Why iproute2?

Is it because of `tc`?

I think these would also be more accurate behind `HAVE_NET`, `HAVE_X11`, etc.

Anyway, for now drop the depends/recommends.

kmk3:
Cc: @reinerh

ghost:
i made depends configurable by configure. I didn't fully tested with a build.
used AI.

ghost:
and recommends

kmk3 (changes requested):
> i made depends configurable by configure.

> I didn't fully tested with a build.

If you're proposing build-related changes, it's your responsibiltiy to test
that they actually work as intended in the relevant combinations.

> used AI.

This code will not work and it's rather obvious why.

It's also poorly written.

Current LLMs are generally helpful for quick explanations and producing small,
self-contained code snippets.

They easily fall apart for anything more complicated than that, needing actual
understanding to fix it.

If you can't properly understand and review the code changes yourself, then you
shouldn't be proposing them.

Code generation aside, the addition of firejail-profiles in provides/conflicts
is pretty good and self-contained (and it looks like it works), so just remove
the AI slop and it should be fine.

ghost:
i think you are burning out with firejail.

Issue #6792, opened 2025-06-26
potential double-free in `procevent_monitor` (SAST Warning)


grey3228:
Hello!

I used the SAST tool **Svace** to analyze Firejail (v0.9.72) and encountered a warning about a potential **double-free** in `src/firemon/procevent.c`. I'd like to verify whether this is a false positive or a legitimate issue.

### Warning Details

In `procevent_monitor`, the tool flags a scenario where `cmd` might be freed twice:

1. **First free** (Line 456):

```c
else {
	sprintf(lineptr, " %s\n", cmd);
	free(cmd);  // First deallocation
}
```

2. **Second free** (Line 471):

```c
// unflag pid for exit events
if (remove_pid) {
	if (pids[pid].option.event.user)
		free(pids[pid].option.event.user);
	if (pids[pid].option.event.cmd)
		free(pids[pid].option.event.cmd); // Potential second deallocation
	memset(&pids[pid], 0, sizeof(Process));
}
```

### Concern

The double-free could occur if:

* pids[pid].option.event.cmd != NULL (i.e., a command was previously cached);
* remove_pid != 0 (i.e., the process is exiting);
* Control flow passes through both code blocks (e.g., for a non-Firejail process exit event).

### Proposed Fix

To prevent this, we could add a check before freeing cmd:

```c
else {
	sprintf(lineptr, " %s\n", cmd);
	if (cmd != pids[pid].option.event.cmd) {  // Only free if not cached
		free(cmd);
	}
}
```

### Questions

1. Is this a false positive, or does a valid execution path trigger this issue?
2. If it's a real issue, would the proposed fix be acceptable?

Thank you for your time and expertise!

kmk3:
Thanks for reporting this.

This is a preliminary response as I haven't finished looking into this, but
after a brief look at the current code, your analysis of the problem seems to
be correct.

> v0.9.72

Note that we do not maintain that version of firejail:

* <https://github.com/netblue30/firejail/blob/master/SECURITY.md>

Versions other than the latest usually have outdated profiles and may contain
bugs and security vulnerabilities that were fixed in later versions.

See also:

* <https://github.com/netblue30/firejail#installing>

Is there a specific reason that you're using 0.9.72?

grey3228:
@kmk3 , thank you for your response! 

[...]

PR #6846, opened 2025-07-31
fix: avoid cmd double-free in procevent_monitor

grey3228:
There is a possible execution path in procevent_monitor function, where allocated memory for cmd may be deallocated twice:

```
char *cmd = pids[pid].option.event.cmd;
if (!cmd) {
	cmd = pid_proc_cmdline(pid);
}
if (add_new) {
    ...
}
else if (proc_ev->what == PROC_EVENT_EXIT && pids[pid].level == 1) {
    ...				
}
else {
	if (!cmd) {
		cmd = pid_proc_cmdline(pid);
	}
	if (cmd == NULL || nodisplay)
		sprintf(lineptr, "\n");
	else {
		sprintf(lineptr, " %s\n", cmd);
		free(cmd);  // <-- First deallocation
	}
	lineptr += strlen(lineptr);
}
	...
// unflag pid for exit events
if (remove_pid) {
	if (pids[pid].option.event.user)
		free(pids[pid].option.event.user);
	if (pids[pid].option.event.cmd)
		free(pids[pid].option.event.cmd);  // <-- Second deallocation
	memset(&pids[pid], 0, sizeof(Process));
}
```

The double-free could occur if:

* pids[pid].option.event.cmd != NULL (i.e., a command was previously cached);
* remove_pid != 0 (i.e., the process is exiting);
* Control flow passes through both code blocks (e.g., for a non-Firejail process exit event).

In PR we add check before deallocating memory.

Fixes: #6792

### DIFF
--- a/src/firemon/procevent.c
+++ b/src/firemon/procevent.c
@@ -496,7 +496,9 @@ static void __attribute__((noreturn)) procevent_monitor(const int sock, pid_t my
 					sprintf(lineptr, "\n");
 				else {
 					sprintf(lineptr, " %s\n", cmd);
-					free(cmd);
+					if (cmd != pids[pid].option.event.cmd) {
+						free(cmd);
+					}
 				}
 				lineptr += strlen(lineptr);
 			}


kmk3 (commented):
Sorry for the delay.

I found the alloc/dealloc flow in `procevent_monitor` rather hard to understand
compared to most functions, so I had decided to leave it for after 0.9.76.

To make procevent.c a bit easier to read, I opened the following PR to refactor
the debugging-related code:

* #6856

Feel free to review it if you want.

Though it breaks CI due to false positives in cppcheck, so I'll try to fix CI
before that PR.

As for this PR, there is another `cmd = pid_proc_cmdline(pid);` line above,
which looks like it could leak memory, so how about doing the following?

<details>
<summary>diff</summary>
<p>

```diff
diff --git a/src/firemon/procevent.c b/src/firemon/procevent.c
--- a/src/firemon/procevent.c
+++ b/src/firemon/procevent.c
@@ -471,8 +471,10 @@ static void __attribute__((noreturn)) procevent_monitor(const int sock, pid_t my
 
 
 			int sandbox_closed = 0; // exit sandbox flag
+			int cmd_dup = 0;
 			char *cmd = pids[pid].option.event.cmd;
 			if (!cmd) {
+				cmd_dup = 1;
 				cmd = pid_proc_cmdline(pid);
 			}
 			if (add_new) {
@@ -490,18 +492,20 @@ static void __attribute__((noreturn)) procevent_monitor(const int sock, pid_t my
 			}
 			else {
 				if (!cmd) {
+					cmd_dup = 1;
 					cmd = pid_proc_cmdline(pid);
 				}
-				if (cmd == NULL || nodisplay)
+				if (!cmd || nodisplay)
 					sprintf(lineptr, "\n");
-				else {
+				else
 					sprintf(lineptr, " %s\n", cmd);
-					free(cmd);
-				}
 				lineptr += strlen(lineptr);
 			}
 			(void) lineptr;
 
+			if (cmd && cmd_dup) {
+				free(cmd);
+				cmd = NULL;
+			}
+
 			// print the event
 			printf("%s", line);
 			fflush(0);
```

</p>
</details>

It's one extra var, but I think it makes the situation clearer and less brittle.

netblue30:
All set, thanks for the bug!

grey3228:
@kmk3 forgive me for my absence, I was able to see your review only now. Regarding your suggestion to add a separate variable "cmd_dup" -- I like it! Perhaps it would be great to refactor this code in the way you suggested in a separate request.

I also took a look at PR https://github.com/netblue30/firejail/pull/6856 and it looks good to me.

Thanks!

kmk3:
> @kmk3 forgive me for my absence, I was able to see your review only now.
> Regarding your suggestion to add a separate variable "cmd_dup" -- I like it!
> Perhaps it would be great to refactor this code in the way you suggested in a
> separate request.
> 
> I also took a look at PR #6856 and it looks good to me.
> 
> Thanks!

All good, thanks for the reviews.

Opened a PR for the suggestion:

* #6872

Issue #5611, opened 2023-01-22
Running a Firejail sandbox on librewolf inside the sway WM has issues with cursor hiding


FarisFiroz:
## Description
On a firejail sandboxed instance of librewolf on swayWM, if the cursor is changed inside the browser, it has trouble changing back. This is most notable when the cursor is hidden such as when viewing a video. Once this happens, the cursor is forever hidden inside that window. Moving the cursor outside the window fixes it, but moving the cursor back inside the window breaks it.

## Steps to reproduce
1. Use the default librewolf firejail profile and launch librewolf inside of firejail on swayWM.
2. Open a video of some sort (Youtube).
3. Put mouse over video or fullscreen and wait for mouse to hide.
4. Once mouse hides, move the mouse.

***Expected Behavior:*** Mouse unhides after moving it.
***Actual Behavior:*** Mouse continues to stay hidden after moving it.

## Extra testing
1. This issue is caused by firejail, running librewolf without firejail has no such issues.
2. This issue does not happen on KDE-PLASMA running wayland
3. This issue does not occur with chromium

## Supplemental Information
*Firejail version:* 0.9.70
*SwayWM version:* 1.7
*Distrobution:* NixOS 22.11

rusty-snake:
Can you test with `--noprofile`.

FarisFiroz:
Yes, --noprofile allows for it to work as expected. I did just notice something though. I have a qt mouse when running with --noprofile or when not sand-boxed at all, while on the librewolf.profile and firefox.profile side, I do not. I believe this may be relevant.

FarisFiroz:
I checked the console and now see this:

```
Gdk-Message: 10:01:17.733: Unable to load left_ptr from the cursor theme
```

but only when sandboxed with firejail. Thats a gtk error so honestly i'm a bit lost, maybe I was wrong in my last message about it being a qt cursor.

glitsj16:
> Gdk-Message: 10:01:17.733: Unable to load left_ptr from the cursor theme

See https://wiki.archlinux.org/title/Cursor_themes#Create_links_to_missing_cursors. In fact, that whole wiki page might be helpful to check. AFAIK mozilla browsers use GTK theming, so if you're using Qt cursor theme(s) this would make sense.

FarisFiroz:
> See https://wiki.archlinux.org/title/Cursor_themes#Create_links_to_missing_cursors. In fact, that whole wiki page might be helpful to check. 

[...]

PR #6858, opened 2025-08-13
docs: github: clarify how to attach logs

kmk3:
Link to the GitHub docs for attaching a file[1].

This should be more straightforward in most cases and would avoid
polluting the user profile with gists unnecessarily (which might get in
the way of using/managing other gists), especially over time when
dealing with many projects/issues/comments.

Keep the gist as a fallback option just in case the file attachment
feature randomly gets broken for an extended period of time, as the
GitHub web UI overall keeps getting slower and jankier over time.

Note: It seems that in both cases (attachment/gist), the file size limit
in the web UI is 25MB[2].

Relates to #5398.

Misc: This was noticed on #5611.

[1] https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files
[2] https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#file-size-limits

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -88,8 +88,14 @@ output goes here
 <summary>Output of <code>LC_ALL=C firejail --debug /path/to/program</code></summary>
 <p>
 
-<!-- If the output is too long to embed it into the comment,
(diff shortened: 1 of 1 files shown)


kmk3:
Cc: @rusty-snake

PR #6876, opened 2025-08-20
profiles: firefox-common: add a comment about mpris

amano-kenji:


etc/profile-a-l/firefox-common.profile
```diff
@@ -109,6 +109,8 @@ dbus-user.talk org.freedesktop.portal.Documents
 # Add the next line to firefox-common.local to allow screensharing under
 # Wayland.
 #dbus-user.talk org.freedesktop.portal.Desktop
+# Add the next line to firefox-common.local to allow media playback control through mpris.
+#dbus-user.own org.mpris.MediaPlayer2.firefox.*
```
kmk3, line 113: Move it above the other mpris comment:

```
# Add the next line to firefox-common.local to allow media playback control through mpris.
#dbus-user.own org.mpris.MediaPlayer2.firefox.*
# Add the next lines to firefox-common.local for plasma browser integration.
#dbus-user.own org.mpris.MediaPlayer2.plasma-browser-integration
```

kmk3 (changes requested):
I see now that this command is already enabled in firefox.profile, so why add
it in a comment here?

etc/profile-a-l/firefox-common.profile
amano-kenji: Done.

amano-kenji:
librewolf and other firefox-based browsers depend on firefox-common.profile. This is actually for librewolf.

kmk3 (approved):

PR #6879, opened 2025-08-24
Add comments indicating how to get qute-pass working in qutebrowser

JustinWayland:
Suffered from this shortfall over the past day, so I felt like contributing this back so no one else has to go through my pain.

Not really sure how to lay this out, both `noblacklist` and `whitelist` are required in qutebrowser.local to make this functionality work. Eliminating either causes qute-pass to error out.

Feedback welcome


kmk3 (approved):
> Feedback welcome

Optional sections usually go below `include globals.local`.

Also, entries are usually sorted.

I moved it and formatted it.

Looks good now; thanks for the PR.

PR #6885, opened 2025-09-01
profiles: organize blacklist sections as per profile.template

kmk3:
On each profile, ensure that the `blacklist` section is right above the
`include disable` section.

See etc/templates/profile.template.

Misc: This appears to affect about a third of the profiles that contain
`blacklist` entries:

    $ git grep -El '^#?blacklist ' -- etc/profile* | wc -l
    158
    $ git diff --name-only f1381b342 | wc -l
    49

Kind of relates to commit 04efbb276 ("profiles: replace x11 socket
blacklist with disable-X11.inc", 2024-03-22) / PR #6286.

### DIFF
--- a/etc/profile-a-l/agetpkg.profile
+++ b/etc/profile-a-l/agetpkg.profile
@@ -7,11 +7,11 @@ include agetpkg.local
 # Persistent global definitions
 include globals.local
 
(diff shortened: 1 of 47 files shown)


kmk3:
Cc: @rusty-snake

PR #6920, opened 2025-10-01
profiles: blink-common-hardened: disable noroot to fix saving files

lucasmz-dev:


kmk3 (commented):
Any idea if the saving files issue is related to the following?

* #6906

Does the following work without disabling `noroot`?

* https://github.com/netblue30/firejail/issues/6906#issuecomment-3316349578

lucasmz-dev:
I'm not sure but those did not work. I am using KDE, notably. Let me see what I can find in a VM (of GNOME)

lucasmz-dev:
Note: It doesn't crash or nothing, literally nothing happens! Nothing new in logs for the app either

lucasmz-dev:
Also does not work in GNOME

lucasmz-dev:
`ignore noroot` works to solve the issue in GNOME as well

kmk3:
Thanks for all the details.

The code looks good, but can you clarify what "breaks saving files" mean?

> Note: It doesn't crash or nothing, literally nothing happens! Nothing new in
> logs for the app either

So the file dialog simply does not appear when it should (such as when using
Ctrl+S)?

Is there anything in `dmesg` (such as seccomp violations)?

> I'm not sure but those did not work. I am using KDE, notably. Let me see what
> I can find in a VM (of GNOME)

AFAICT, the glycin/bwrap issue affects all programs that use gdk-pixbuf2,
even when using other DEs.

rusty-snake:
noroot breaks portals. Maybe not all implementation in all versions but at least some. (#6048)

lucasmz-dev:
It does seem like it is using Portals. The file picker and stuff shows stuff that the sandbox would not be able to access.

Issue #3843, opened 2020-12-21
use sandbox name as xephyr window title


pszxzsd:
Currently Xephyr windows are always named "firejail x11 sandbox" even if a sandbox name is supplied. It seems the name already is used for Xpra, it would be great if the same was applied to Xephyr.

Additionally I would suggest to prepend the sandbox name before any fixed string because desktop taskbars usually cut off long windows titles at the end.

Issue #1983, opened 2018-06-08
An indication to know if a program is Sandboxed


Logmytech:
Almost all Windows based sandboxing programs show an Indication that the program is Sandboxed.

# in Window title #
Window border colors like Red, Yellow, Green, Purple.

Considering the fact that, if Firefox is already running, then a new Firefox opened using Firejail doesn't get sandboxed, I think it is crucial that we have some indication to know that we are protected.

smitsohu:
for context see also #281

chiraag-nataraj:
@netblue30 We should be able to set the window title though, right? I think we could go with that as a stop-gap measure...

rusty-snake:
As a workaround you could use `env` to set an alternative/own GTK+2, GTK+3, GTK+4, QT4, QT5 theme.

PR #6952, opened 2025-10-30
feature: add --xephyr-extra-params= command

yureitzk:
This lets user specify Xephyr options for each sandbox without modifying the `firejai.config` file. 

It can also provide visual indication for sandboxed programs — for example, setting `-name` changes the window's `WM_CLASS` property, which window managers and desktop environments can use in rules to alter its appearance.

Related issues:
- https://github.com/netblue30/firejail/issues/3843
- https://github.com/netblue30/firejail/issues/1983


netblue30:
Thanks for the patch! I'll modify the order as suggested by @kmk3

PR #6958, opened 2025-11-06
bugfix: correct directory traversal checking logic

### LINKED ISSUE #6915: swww: Error: "/run/user/1000/wayland-1-swww-daemon..sock" is an invalid filename
### Description

Whitelisting `/run/user/1000/wayland-1-swww-daemon..sock` is impossible.

### Steps to Reproduce

Execute

```
firejail --include=/etc/firejail/whitelist-runuser-common.inc --whitelist=/run/user/1000/wayland-1-swww-daemon..sock --noprofile bash
```

### Expected behavior

No error

### Actual behavior

```
$ firejail --include=/etc/firejail/whitelist-runuser-common.inc --whitelist=/run/user/1000/wayland-1-swww-daemon..sock --noprofile bash
Reading profile /etc/firejail/whitelist-runuser-common.inc
Error: "/run/user/1000/wayland-1-swww-daemon..sock" is an invalid filename
```

### Environment

- Name/version/arch of the Linux kernel (`uname -srm`): Linux 6.6.101-gentoo x86_64
- Name/version of the Linux distribution (e.g. "Ubuntu 20.04" or "Arch Linux"): Gentoo Linux
- Version of Firejail (`firejail --version`): 
```
firejail version 0.9.74

Compile time support:
        - AppImage support is enabled
        - chroot support is enabled
        - IDS support is disabled
        - Landlock support is enabled
        - output logging is enabled
        - private-lib support is disabled
        - sandbox check is enabled
        - SELinux support is disabled
```

JohnyTheCarrot:
Fixes #6915

The logic checking for the presence of directory traversal 'tokens' yields false positives.
It is perhaps rare though entirely valid for a filename to contain a double dot.
The linked issue is how it manifested in my usage of Firejail.

I aim to fix this by checking any of the following instead:
1. the string is a full match of ".."
2. the string contains "../"
3. the string ends with "/.."

if I am missing any scenarios in which a '..' would ordinarily be valid, please let me know.
PS: should I add `str_ends_with` to the common.h header? I figure it couldn't hurt, no? Let me know.

### DIFF
--- a/src/firecfg/main.c
+++ b/src/firecfg/main.c
@@ -224,7 +224,7 @@ static void parse_config_file(const char *cfgfile, int do_symlink) {
 			continue;
 
 		// do not accept .. and/or / in file name
-		if (strstr(buf, "..") || strchr(buf, '/')) {
+		if (contains_directory_traversal(buf) || strchr(buf, '/')) {
 			fprintf(stderr, "Error: invalid line %d in %s\n", lineno, cfgfile);
 			exit(1);
 		}
--- a/src/firejail/fs_bin.c
+++ b/src/firejail/fs_bin.c
@@ -86,7 +86,7 @@ static int valid_full_path_file(const char *name) {
 
 	if (*name != '/')
 		return 0;
-	if (strstr(name, ".."))
+	if (contains_directory_traversal(name))
 		return 0;
 
 	// do we have a file?
@@ -136,7 +136,7 @@ static void duplicate(char *fname) {
 	EUID_ASSERT();
 	assert(fname);
 
-	if (*fname == '~' || strstr(fname, "..")) {
+	if (*fname == '~' || contains_directory_traversal(fname)) {
 		fprintf(stderr, "Error: \"%s\" is an invalid filename\n", fname);
 		exit(1);
 	}
--- a/src/firejail/fs_etc.c
+++ b/src/firejail/fs_etc.c
@@ -301,7 +301,7 @@ static void duplicate(const char *fname, const char *private_dir, const char *pr
 static void duplicate_globbing(const char *fname, const char *private_dir, const char *private_run_dir) {
 	assert(fname);
 
-	if (*fname == '~' || *fname == '/' || strstr(fname, "..")) {
+	if (*fname == '~' || *fname == '/' || contains_directory_traversal(fname)) {
 		fprintf(stderr, "Error: \"%s\" is an invalid filename\n", fname);
 		exit(1);
 	}
--- a/src/firejail/fs_home.c
+++ b/src/firejail/fs_home.c
@@ -474,7 +474,7 @@ void fs_check_private_cwd(const char *dir) {
 	cfg.cwd = expand_macros(dir);
 
 	// realpath/is_dir not used because path may not exist outside of jail
-	if (strstr(cfg.cwd, "..") || *cfg.cwd != '/')
+	if (contains_directory_traversal(cfg.cwd) || *cfg.cwd != '/')
 		goto errout;
 
 	return;
--- a/src/firejail/fs_lib.c
+++ b/src/firejail/fs_lib.c
@@ -103,7 +103,7 @@ static const char *masked_lib_dirs[] = {
 
 // return 1 if the file is in masked_lib_dirs[]
 static int valid_full_path(const char *full_path) {
-	if (strstr(full_path, ".."))
+	if (contains_directory_traversal(full_path))
 		return 0;
 
 	int i = 0;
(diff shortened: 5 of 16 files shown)


kmk3 (changes requested):
Quoting myself from [here](https://github.com/netblue30/firejail/issues/6915#issuecomment-3502467664):

> As can be seen in [#6958](https://github.com/netblue30/firejail/pull/6958),
> the checks are meant to catch attempts at relative directory traversal (see
> CWE-23).
> 
> I think it would make more sense to change the relevant program to remove the
> extra dot than to increase the risk of a malicious traversal happening.

I think it might be too risky to allow this, especially for something as low
level and ubiquitous as filesystem paths.

The function does seem to catch the cases that I immediately came up with, but
I suspect that the bypassing may be possible with more complicated patterns,
such as paths starting with `//` or `~` or paths using unicode.

In all cases, `..` is the common denominator needed to perform the traversal.

I can't think of a good reason to intentionally use `..` in a filename instead
of something else.

And it seems that the issue of #6915 is going to be fixed directly in the
affected program, so I'd just leave the current code as is.

amano-kenji:
Let's at least wait for confirmation from the upstream.

amano-kenji:
Awww maintainer said he's going to eliminate two consecutive dots in the socket file, but I think firejail should do better than checking whether a file name "contains" two consecutive dots.

kmk3:
> Awww maintainer said he's going to eliminate two consecutive dots in the
> socket file, but I think firejail should do better than checking whether a
> file name "contains" two consecutive dots.

I don't think it's worth the increase in risk and complexity, especially since
seemingly only a single program is affected and only temporarily.

amano-kenji:
There isn't much complexity in detecting `../` or `/..` or `..`. The current behavior is incorrect and can lead to another issue in the future.

I was patient enough to even report this issue. Other people will just stop using firejail because they don't have time and patience to deal with this. If firejail was a physical printer, they would just throw the printer into a trash bin instead of trying to diagnose the issue. Not using firejail isn't going to lead to a better outcome.

kmk3:
> I was patient enough to even report this issue.

And we were patient enough to reply.

> Other people will just stop using firejail because they don't have time and
> patience to deal with this.  If firejail was a physical printer, they would
> just throw the printer into a trash bin instead of trying to diagnose the
> issue. Not using firejail isn't going to lead to a better outcome.

If you want to stop using firejail for this one program (which does not even
have an upstream profile and seemingly never worked with firejail), that sounds
fine.

If you want to do something productive, you can try to help `awww` with
changing the file path or just patch it yourself for the time being.

amano-kenji:
If you are worried about a new bug introduced by this pull request, why don't you ask for a unit test?

To me, this pull request makes sense. The changes are surgically precise and correct. It focuses on one thing only, and I can easily understand the change without knowing much about C or the internal details of firejail. I never wrote C and never read firejail source code before. If this pull request contained a lot of different changes that confuse a mediocre hobbyist programmer like me, then I would agree with your sentiment.

A unit test for `contains_directory_traversal` can give you peace of mind. I think any perceived risk can be managed with a unit test.

I also think that if you accept high-quality pull requests like this, you will get more contribution down the line because people think their efforts won't be wasted. It is a good idea to accept high-quality free labor.

JohnyTheCarrot:
I'll leave closure or merging up to @kmk3 if they think it appropriate.
This PR didn't take me long at all, I'm not too fussed.
I'm no expert in Firejail so I won't claim to know what's best for it.
I will add though that a unit test works well enough for as long as you list any cases you can think of.
The moment someone finds a case you haven't thought of it becomes an issue.
It is increased surface area for attack. The question of worth is one I leave to the maintainer(s).

Edit: wrt being patient enough to report issues, while that's generally appreciated from FOSS' standpoint, you have to remember that the people maintaining your favorite FOSS projects do so for free. I myself am grateful for the time and energy they do spend.

kmk3:
> Edit: wrt being patient enough to report issues, while that's generally
> appreciated from FOSS' standpoint, you have to remember that the people
> maintaining your favorite FOSS projects do so for free. I myself am grateful
> for the time and energy they do spend.

Well said.

> I will add though that a unit test works well enough for as long as you list
> any cases you can think of.  The moment someone finds a case you haven't
> thought of it becomes an issue.  It is increased surface area for attack. The
> question of worth is one I leave to the maintainer(s).

The problem is that path traversal interacts with the lower layers of the stack
(which are not managed by firejail).

That is, `..` can effectively be interpreted as a command by the kernel/libc
(and potentially other layers, such as FUSE or NFS), and potentially in
multiple places.

kmk3:
@JohnyTheCarrot

Hello, are there any further considerations or can this be closed?

JohnyTheCarrot:
Nope. Sorry, I was used to PRs being closed by maintainers. (not a subtle dig)
Thanks for your work and time!

PR #6960, opened 2025-11-10
build: add script to generate syscall headers

pierretom:
This script fetches current system calls from kernel sources then extracts and installs them in the src/include directory.
Syscalls can be updated by regenerating them, ideally once before each release. 
contrib/syntax/lists/syscalls.list is synchronized too.
It generates also etc/templates/new_syscalls.txt, this makes it easier to update groups and to inform users about new syscalls added.
The script must reside in the src/tools directory and requires the cURL CLI program.


netblue30:
Merged, thanks!

Issue #6966, opened 2025-11-12
Error: Firejail configuration file /etc/firejail/firejail.config not found (private-etc)


Antiz96:
### Description

Since d1aeeb4fa1d3c1a2c626267c76db70cd81bdb925, programs (or specific actions in said program) running from a `private-etc` environment fail because `firejail.config` is not accessible from said environment.

I'm guessing this is because `checkcfg(0)` is now called early in the process in [main.c](https://github.com/netblue30/firejail/commit/d1aeeb4fa1d3c1a2c626267c76db70cd81bdb925#diff-6698244f9a67e5c8ae5c03806df74f6d9f1ae1b31ad6176eb09e136f07f3dad9R1078-R1080).

### Steps to Reproduce

_Steps to reproduce the behavior_

From a firejail build including d1aeeb4fa1d3c1a2c626267c76db70cd81bdb925 (thunderbird is just an example).

1. Run in bash `LC_ALL=C firejail /usr/local/bin/thunderbird` (where `/usr/local/bin/thunderbird` is the symlink create by `firecfg`)
2. See error:

```
[...]
Error: Firejail configuration file /etc/firejail/firejail.config not found

Parent is shutting down, bye...
```

_Different scenario_

1. LC_ALL=C firejail /usr/bin/thunderbird
2. Program starts but...
3. Do some actions that should triggers something outside of the sandboxed environment (e.g. clicking a link from a mail to open in your web browser)
4. Nothing happens (link is not being opened) and see error:

```
[...]
Error: Firejail configuration file /etc/firejail/firejail.config not found
```

### Expected behavior

Program to starts (or actions within this program to work reliably)

### Actual behavior

Program fails to start (or specific actions triggering something outside of the sandboxed environment fail).

### Behavior without a profile

Both `LC_ALL=C firejail --noprofile /usr/local/bin/thunderbird` and `LC_ALL=C firejail --noprofile /usr/bin/thunderbird` work as expected (where `/usr/local/bin/thunderbird` is the symlink create by `firecfg`).

### Additional context

Adding `private-etc firejail/firejail.config` to e.g. `~/.config/firejail/thunderbird.local` fixes (or "works around") the issue.

I reported this issue a few days in the related MR at https://github.com/netblue30/firejail/pull/6878#discussion_r2490670980

### Environment

- Name/version/arch of the Linux kernel (`uname -srm`): Linux 6.17.7-arch1-1 x86_64
[...]

PR #6970, opened 2025-11-21
bugfix: make checkcfg init/cfg vars global

kmk3:
Instead of them being local to the translation unit.

This should ensure that firejail.config is not read more than once (as
is intended by the code), especially after options like `private-etc`
are applied (which could render the file inaccessible when calling
`checkcfg()`).

Related commits:

* 8bff773d6 ("add symlink resolution for home directories", 2019-07-09)
  / issue #2877
* d32509945 ("rewrite/partial revert of
  8bff773d6a7bf70c97b3d5b751df9ec0dd6c8b5d", 2019-08-09)
  / issue #2877
* d1aeeb4fa ("feature: add arg-max-count and arg-max-len options to
  firejail.config (#6878)", 2025-10-30)

Relates to #2877 #6966.

### DIFF
--- a/src/firejail/checkcfg.c
+++ b/src/firejail/checkcfg.c
@@ -26,8 +26,8 @@
 
 #define MAX_READ 8192				  // line buffer for profile files
 
-static int firejail_config_init = 0;
-static int cfg_val[CFG_MAX];
+int firejail_config_init = 0;
+int cfg_val[CFG_MAX];
 char *xephyr_screen = "800x600";
 char *xephyr_extra_params = "";
 char *xpra_extra_params = "";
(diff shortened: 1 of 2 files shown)


kmk3:
@Antiz96

Does this change anything with regards to #6966?

Antiz96:
> @Antiz96
> 
> Does this change anything with regards to https://github.com/netblue30/firejail/issues/6966?

@kmk3 It doesn't unfortunately :confused:

kmk3:
I think that this PR by itself is unlikely to fix #6966 and I'd rather debug it
more before seeing if it's worth merging.

Closing for now.

PR #6972, opened 2025-11-23
modif: check for --version during early init

kmk3:
Move it from `run_cmd_and_exit()` to right after the --quiet/--debug
checks.

This simplifies the sandbox check code by removing its own --version
check.

See also commit 5cd597e5d ("fix --version", 2016-06-28).

This is a follow-up to #6969.

Relates to:

* #7137


netblue30:
Thanks, let's go with it. I'll run some tests here.

PR #6985, opened 2025-12-07
profiles: electrum: add noblacklist /usr/share/fonts

ipaqmaster:


ipaqmaster:
Okay.. not sure what I need to change to reproduce this. I can no longer reproduce this problem after commenting out `#noblacklist /usr/share/fonts` in my `~/.config/firejail/electrum.profile`

It seems after running once with access to fonts, it can now consistently use the correct system font. As if it cached it somewhere. Not in ~/.cache/fontconfig it seems.

netblue30:
Merged in, thanks!

ipaqmaster:
This seems to impact a number of other programs when the fix for https://github.com/netblue30/firejail/issues/6906 is active. I think this is due to the whitelist of glycin-loaders creating a whitelist for some programs when there is normally not one.

I'm tweaking prismlauncher.profile at the moment and the same above change fixes its font access as well but it doesn't happen at all if that linked fix isn't applied. This is likely because `whitelist` isn't used in its profile but is being used due to that fix. I guess it just cannot be globally applied then.

This makes me wonder if there's a way we could only apply a whitelist item when a whitelist is already present as an override-only kind of flag.

PR #6990, opened 2025-12-13
build: update gen-syscalls.sh

pierretom:
- Download PA-RISC syscalls table only once
- Add syscalls number and entry points in etc/templates/new_syscalls.txt

Relates to #6960.


netblue30:
Merged, thanks!

PR #6996, opened 2025-12-19
new profile: srb2

HybridDog:
Sonic Robo Blast 2 is implemented in C, occasionally crashes with memory access errors and has a multiplayer mode with insecure network traffic, so I think it's a good candidate for sandboxing.

This is my first time offering a new profile and there are some things which I don't understand yet:
* I've omitted whitelists such as `include whitelist-common.inc`, which are mentioned in the template. Is this a bad choice and if so, which whitelists should we add?
* Do we need to add the landlock commands?
* Do we need to add `noexec ${HOME}/.srb2` and/or `read-write ${HOME}/.srb2` to make nothing in ~/.srb2 executable?
* ~Do we need a `blacklist PATH` or is the current working directory automatically blacklisted?~
* Do we need `private-etc @tls-ca,@games,@x11` if `private-etc @tls-ca` also works?
* What does "Disable input devices." with `noinput` mean? Are the mouse and keyboard "input devices" and if so, why do they work despite `noinput`?

### DIFF
--- a/etc/profile-m-z/srb2.profile
+++ b/etc/profile-m-z/srb2.profile
@@ -22,7 +22,6 @@ include disable-xdg.inc
 mkdir ${HOME}/.srb2
 whitelist ${HOME}/.srb2
 include whitelist-common.inc
-include whitelist-run-common.inc
 include whitelist-runuser-common.inc
 include whitelist-var-common.inc
 


rusty-snake:
> I've omitted whitelists such as include whitelist-common.inc, which are mentioned in the template. Is this a bad choice and if so, which whitelists should we add?

Add all that you can get to work.


> Do we need to add the landlock commands?

No

> Do we need to add noexec ${HOME}/.srb2 and/or read-write ${HOME}/.srb2 to make nothing in ~/.srb2 executable?

Noexec is handled by `include disable-exec.inc`.
read-write is the default

> Do we need a blacklist PATH or is the current working directory automatically blacklisted?

IDK what you mean. CWD is kept if it is avaiable, otherwise $HOME is used as fallback.

> What does "Disable input devices." with noinput mean? Are the mouse and keyboard "input devices" and if so, why do they work despite noinput?

It disables the raw access to input devices (e.g. joysticks). Programs to not interact with mouse/keyboard directly, they get this input from a display server (Wayland/X11).

HybridDog:
Thanks for the quick response.

> IDK what you mean. CWD is kept if it is avaiable, otherwise $HOME is used as fallback.

Sorry, I confused PATH with CWD for some reason today. 

> It disables the raw access to input devices (e.g. joysticks). Programs to not interact with mouse/keyboard directly, they get this input from a display server (Wayland/X11).

The game supports game pads as input, probably via sdl. I assume that the program also gets this input indirectly.

netblue30:
Merged, thanks!

PR #7001, opened 2025-12-27
profiles: freetube: allow mpv integration

pranaovs:
Clicking the launch in mpv button in freetube results in an EACESS
error. Include mpv.profile with some ignore fixes to allow freetube
profile to launch mpv profile

I ran sort.py but didnt see any reasonable auto arranges, so I rearranged the profile to how it made sense to me.

### DIFF
--- a/etc/profile-a-l/freetube.profile
+++ b/etc/profile-a-l/freetube.profile
@@ -6,7 +6,18 @@ include freetube.local
 # Persistent global definitions
 include globals.local
 
+# mpv integration
+ignore include disable-programs.inc
+ignore include disable-shell.inc
+
+private-etc mpv
+
 ignore dbus-user none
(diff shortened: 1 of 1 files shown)


etc/profile-a-l/freetube.profile
```diff
@@ -6,7 +6,18 @@ include freetube.local
 # Persistent global definitions
 include globals.local
 
+# mpv integration
+ignore include disable-programs.inc
```
rusty-snake: No

kmk3 (changes requested):
https://github.com/netblue30/firejail/pull/7001#discussion_r2649113975

pranaovs:
Oops, I was a bit busy. Ill correct this and reopen.

PR #7004, opened 2025-12-28
build: add sort-profiles target for sort.py

kmk3:
And use it in CI.

Note: The target currently always runs sort.py for all profiles.

Before:

    $ ./ci/check/profiles/sort.py etc/inc/*.inc etc/profile*/*.profile

After:

    $ make sort-profiles


kmk3:
Cc: @rusty-snake

kmk3:
Misc: After this PR, I plan to add it to CONTRIBUTING.md and to the PR
template.

PR #7014, opened 2026-01-02
new profile: quakespasm

StellariusOrion592:
QuakeSpasm is a modern multi-platform Quake source port designed as an
improved successor to GLQuake and FitzQuake. It aims to preserve the
classic gameplay and graphics while enhancing compatibility and
modernization.

https://sourceforge.net/projects/quakespasm


kmk3 (approved):

kmk3:
Good to see the details in the commit message.

Though the PR description was worded a bit better (and included the URL), so I
used that as the commit message.

On another note, we are near the release, so please only submit bugfixes until
after the release:

* #6841

PR #7015, opened 2026-01-02
fix zulip profile

mio-19:


kmk3 (changes requested):
> fix zulip profile

What is the fix?

What is not working?

Also, zulip.profile contains many hardening options that are not present in
electron-common.profile and related profiles.

So which lines actually need to be removed?

mio-19:
> What is the fix?

before: zulip complaints about no SUID flag on chrome sandbox. unable to launch zulip
after: zulip works

> Also, zulip.profile contains many hardening options that are not present in electron-common.profile and related profiles.
> 
> So which lines actually need to be removed?

I am sorry I didn't have time to figure out

kmk3:
> > Also, zulip.profile contains many hardening options that are not present in
> > electron-common.profile and related profiles.  So which lines actually need
> > to be removed?
> 
> I am sorry I didn't have time to figure out

Alright, if you find out let us know.

For debugging you can comment/uncomment lines until it works/breaks.

PR #7000, opened 2025-12-27
modif: include new generated syscall headers

pierretom:
Also fix typo in comments.

Other PRs will come after approval.

Relates to:

* #6961
* #6991

PR #7021, opened 2026-01-07
syscall.c: fix indentation of syscall groups

pierretom:
- Fix the indentation

Relates to:

* #7000


pierretom:
The proposed fix was to keep the same indentation than other `#ifdef` but you are right its must be aligned with the first element.

Closing this PR, the indentation is already correct.

PR #7024, opened 2026-01-10
modif: reorganize & update syscall groups (syscall.c) - part 1

pierretom:
- Remove `ni_syscall`, it's an invalid system call
- Remove `execveat` and `prctl` from `@process`, as it's present in `@default-keep` and always whitelisted
- Move `@default-keep` group before `@default-nodebuggers`, to keep the alphabetical order
- Move `execve` before `execveat`, to keep the alphabetical order
- Move `arch_prctl` from `@process` to `@default-keep`, it breaks glibc
- Move `mmap` and `mmap2` from `@file-system` to `@default-keep`, they break loading of shared libraries
- Move `mprotect` from `@system-service` to `@default-keep`, it breaks loading of shared libraries
- Move `epoll_ctl_old` and `epoll_wait_old` from `@io-event` to `@obsolete`
- Move `getrusage` from `@process` to `@resources`
- Allow to define `mincore` in `@resources`

Relates to:

* #7000

### DIFF
--- a/src/lib/syscall.c
+++ b/src/lib/syscall.c
@@ -319,9 +319,6 @@ static const SyscallGroupList sysgroups[] = {
 #ifdef SYS_ioprio_set
 	  "ioprio_set,"
 #endif
-#ifdef SYS_ni_syscall
-	  "ni_syscall,"
-#endif
 #ifdef SYS_syslog
 	  "syslog,"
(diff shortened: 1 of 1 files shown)


netblue30:
All in, thanks!

Issue #6999, opened 2025-12-23
yt-dlp: No supported JavaScript runtime could be found (deno)


hyder365:
### Description

Since recently, yt-dlp pulls in a javascript engine to download Youtube videos. By default it uses deno for this. The profile for yt-dlp needs access to deno if installed for javascript.

### Steps to Reproduce

firejail yt-dlp URL-HERE

### Expected behavior

No warnings

### Actual behavior

> WARNING: [youtube] No supported JavaScript runtime could be found. Only deno is enabled by default; to use another runtime add  --js-runtimes RUNTIME[:PATH]  to your command/config. YouTube extraction without a JS runtime has been deprecated, and some formats may be missing. See  https://github.com/yt-dlp/yt-dlp/wiki/EJS  for details on installing one

### Environment

Arch Linux, latest everything

### Checklist

- [x] I am using firejail [0.9.76 or later](https://github.com/netblue30/firejail/tree/master/SECURITY.md)
- [x] I am using the full program path (e.g. `firejail /usr/bin/vlc` instead of `firejail vlc`; see `https://github.com/netblue30/firejail/issues/2877`)
- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [x] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

dogknowsnx:
Not sure that deno/java is the problem here.

I need to add the following in order to get past python-related errors ever since `yt-dlp` v2025.12.08:

`yt-dlp.local`:

```
noblacklist /usr/lib/python*/*
whitelist /usr/lib/python*/*
```

But I still haven't figured out:

```
$ yt-dlp help
Error: Cannot start application: No such file or directory
```

`firejail version 0.9.79` on Void Linux

netblue30:
[...]

PR #7032, opened 2026-01-14
profiles: add deno paths

kmk3:
Add paths in the same places as nodejs/npm paths.

Deno is a javascript runtime and development tool similar to nodejs.

The following paths seem to be intended for downloading and caching
dependencies (and apparently also artifacts from .ts to .js compilation)
globally during development (as can be done with ~/.npm):

* ~/.cache/deno
* ~/.deno

Note that this commit makes these paths read-only (as npm dependencies
are usually executable code), which may potentially affect users of the
runtime (like yt-dlp).

Related commits:

* f2de86464 ("tentative fix for yt-dlp/javaScript deno profile (#6999)",
  2026-01-13)

### DIFF
--- a/etc/inc/allow-common-devel.inc
+++ b/etc/inc/allow-common-devel.inc
@@ -6,6 +6,10 @@ include allow-common-devel.local
 noblacklist ${HOME}/.arduino15
 noblacklist ${HOME}/Arduino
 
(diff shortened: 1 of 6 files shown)


netblue30:
all in, thanks!

PR #7034, opened 2026-01-18
modif: update syscall groups (syscall.c) - part 3

pierretom:
- Add new syscalls
  - Add `execv`, `exit` and `futex` in `@default-keep`
  - Add two new groups: `@memfd` and `@sandbox`
  - Add `@memfd` and `@sandbox` in the `@system-service` group
  - Move `memfd_create` from `@ipc` to `@memfd`

Thanks to @rusty-snake for information and suggestions.

Relates to:

* #7027

PR #7039, opened 2026-01-22
docs: update syscall groups (syscalls.txt) - part 4

pierretom:
- Remove extra empty lines
- Definition of groups:
  - Add the two new groups `@memfd` and `@sandbox`
  - Add new syscalls
- Inheritance of groups:
  - Redraw it in a clearer form of groups and subgroups
  - Add the two new groups
  - Sort `@mount` and `@obsolete` groups by alphabetical order

This is the last part.

Relates to:

* #7034


netblue30:
Let's grab it as is. We'll probably have to update it later. Thanks!

PR #7044, opened 2026-01-28
modif: place some syscalls with access to a clock into `@default-keep`

pierretom:
Also move `stime` into `@obsolete`, see `man 2 stime`.


netblue30:
all merged, thanks!

PR #7045, opened 2026-01-28
modif: move other syscalls considered deprecated into `@obsolete`

pierretom:
`futimesat`, `gettimeofday`, `remap_file_pages`, `settimeofday`, `tkill` and `utime`.


netblue30:
Try to check them in directly, whatever it is. I've set you up to be able to do it. Let me know if it doesn't work. Thanks!

pierretom:
> Try to check them in directly, whatever it is. I've set you up to be able to do it. Let me know if it doesn't work. Thanks!

Ok.

PR #7049, opened 2026-01-31
feature: add a new option `--debug-syscall-groups` - part 1

pierretom:
This option allows to list all seccomp groups and their syscalls defined for the architecture in use.

Usage examples:
`firejail --debug-syscall-groups`
`firejail --debug-syscall-groups=@chown,@mount,@swap`
`firejail --debug-syscall-groups=@all`

This is the first part.

### DIFF
--- a/src/firejail/main.c
+++ b/src/firejail/main.c
@@ -604,6 +604,22 @@ static void run_cmd_and_exit(int i, int argc, char **argv) {
 		else
 			exit_err_feature("seccomp");
 	}
+	else if (strcmp(argv[i], "--debug-syscall-groups") == 0) {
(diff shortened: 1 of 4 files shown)


rusty-snake:
What is with other groups like @games or @x11? Maybe rename it to drbug-seccomp-groups.

pierretom:
> What is with other groups like @games or @X11? Maybe rename it to drbug-seccomp-groups.

You mean rename it to `--debug-seccomp-groups` ?
Yes, I forgot those groups, seems a good idea to avoid the confusion.

pierretom:
If someone can explain why the PR breaks the test, thanks.
Edit: it's fixed now.

netblue30:
merged, thanks!

PR #7055, opened 2026-02-07
feature: add a new option `--debug-syscall-groups` - part 2

pierretom:
- src/firejail/usage.c
- src/zsh_completion/_firejail.in
  - Add entries for `--debug-syscall-groups`
- src/lib/syscall.c
  - Complete string literals
- src/man/firejail.1.in
  - Add an entry for `--debug-syscall-groups`
  - Add `@memfd` and `@sandbox` syscall groups for the `--seccomp` option

This is the last part.


netblue30:
all set, thanks!

Issue #7063, opened 2026-02-16
Syscall `uname` known as `newuname` (`olduname` also exists, iiuc)


cobratbq:
### Description

I was experimenting with listing `seccomp.keep` syscalls directly. I discovered that (without noticeable warning) `newuname` is unknown so did not have an effect.

Your code (current `master`) contains: `uname`: `63`
Other syscall references list: `newuname`: `63`

I think there's also `olduname`, but I haven't needed it in whitelisting yet. (Just a heads-up.)

Is there any chance you can make an alias for this, or otherwise make `newuname` work? This would help w.r.t. whiteliting syscalls, given that applications already start in "highly precariously restricted" environments. 😃 

### Steps to Reproduce

Add `newuname` to `syscall.keep`-list.

### Expected behavior

Add `newuname` to `syscall.keep`-list and see it show up in seccomp-BPF listing.

### Actual behavior

Discover that it isn't in the produced seccomp-BPF listing. Thus `newuname` call isn't whitelisted. (I didn't know about `uname` until I checked the code.)

### Behavior without a profile

Almost clean profile because I'm screwing around with whitelisting syscalls. So consider it empty, except for a whole swath of syscalls. :-P

### Additional context

Nope.

### Environment

- `firejail` 0.74
- AMD64 arch.

### Checklist

I'm sorry for being obnoxious but I don't think the checklist is necessary. Scold me if you think I'm wrong. 😓 

### Log

Not applicable.

pierretom:
I don't think it can be invoked directly by applications, it's used internally by the kernel.
Maybe `firejail --seccomp.keep=uname` might just be enough in your case ?
Or try with the syscall number on x86_64 `firejail '--seccomp.keep=$63'`

cobratbq:
Wait .. you can do *numbers too*! Arghh .. okay that's useful.

When restricting Firefox it eventually kills on 'newuname', which indeed seems to be 'uname' (63) but seems listed "everywhere" as newuname. So, it seems to arrive there eventually. *Maybe* I got confused by interleaved apparmor messaging or smth, but I don't think I did.

pierretom:
Yes use `uname` with Firefox.

> * `firejail` 0.74

You can try to update Firejail too, there are 2 new releases since 0.9.74.

kmk3:
> Environment

What distribution name/version?

> I was experimenting with listing `seccomp.keep` syscalls directly. I
> discovered that (without noticeable warning) `newuname` is unknown so did not
[...]

PR #7065, opened 2026-02-17
docs: man: clarify --debug-syscalls for seccomp

kmk3:
Some syscalls have multiple versions, which may lead to confusion as to
which is the correct syscall name.

For example, `uname` has two older obsolete versions (see `uname(2)`):

* `__NR_oldolduname`
* `__NR_olduname`
* `__NR_uname`

But only the most recent one is exposed by the syscall map for the
`x86_64` architecture:

    $ firejail --debug-syscalls | grep uname
    63      - uname

Mention `--debug-syscalls` / `--debug-syscalls32` as a way to find the
proper syscall names to be used with `--seccomp` / `--seccomp.32`.

Relates to #7063 #7064.

Reported-by: @cobratbq

### DIFF
--- a/src/man/firejail.1.in
+++ b/src/man/firejail.1.in
@@ -784,17 +784,44 @@ $ firejail \-\-debug-syscall-groups=@chown,@mount,@swap
 $ firejail \-\-debug-syscall-groups=@all
 .TP
 \fB\-\-debug-syscalls
-Print all recognized system calls in the current Firejail software build and exit.
+Print all recognized system calls in the current Firejail software build and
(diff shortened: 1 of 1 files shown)


cobratbq:
Awesome. It would've been convenient to know. (I hope the `$42` syscall-by-number notation is also documented?)

kmk3:
> Awesome. It would've been convenient to know. (I hope the `$42`
> syscall-by-number notation is also documented?)

It is documented; see `--seccomp=` and the following PR:

* #7064

PR #7066, opened 2026-02-17
bugfix: error on invalid syscall name in seccomp

### LINKED ISSUE #2931: --seccomp= unknown syscall names/numbers are silently ignored
If a bogus (unknown) syscall name/number/group-name is supplied to --seccomp=, then it is silently ignored. No warning is printed even with --debug.

This is bad, because a misspelled syscall will not be disabled in the jail, potentially creating a vulnerability. It would be better, if firejail would reject a wrong name/number and refuse to start.

On a related note: why '$' is needed to denote a numeric syscall? If syscall "name" is all numeric, then it is definitely the syscall number, so '$' seems to be redundant.

firejail version: 0.9.56.2-LTS
OS: Ubuntu 16.04.6 LTS

kmk3:
Currently, an error is produced when an invalid syscall group is used:

    $ firejail --quiet --seccomp.drop=@clock /bin/true
    $ firejail --quiet --seccomp.drop=@foo /bin/true
    Error fseccomp: unknown syscall group @foo

But not when an invalid syscall name is used (resulting in it being
silently ignored by firejail).

Make sure to also fail when that happens.

Before:

    $ firejail --quiet --seccomp.drop=uname /bin/true
    $ firejail --quiet --seccomp.drop=foo /bin/true

After:

    $ firejail --quiet --seccomp.drop=uname /bin/true
    $ firejail --quiet --seccomp.drop=foo /bin/true
    Error fseccomp: invalid syscall list entry foo

Misc: From a cursory glance, this appears to have been an issue since
the `syscall_process_name()` function was introduced on commit 322ce2cdc
("seccomp rework", 2016-11-06).

Fixes #2931.

Reported-by: @aoand
Reported-by: @cobratbq

### DIFF
--- a/src/lib/syscall.c
+++ b/src/lib/syscall.c
@@ -2567,6 +2567,10 @@ static void syscall_process_name(const char *name, int *syscall_nr, int *error_n
 				*syscall_nr = SYSCALL_ERROR;
 		}
 	}
+	if (*syscall_nr == SYSCALL_ERROR) {
+		free(str);
+		goto error;
+	}
 
 	free(str);
 	return;


rusty-snake:
I did not looked into detail. Nevermind did you took the feature aspect of silent error into account. I.e. backporting of profiles (which we do not support/recommend) and syscalls that are only defined for specific architectures.

Profiles shouldn't be architecture depended and `--seccomp.drop=foo_x86,foo_arm` should work without changes on every system.

kmk3:
> I did not looked into detail. Nevermind did you took the feature aspect of
> silent error into account. I.e. backporting of profiles (which we do not
> support/recommend) and syscalls that are only defined for specific
> architectures.
> 
> Profiles shouldn't be architecture depended and
> `--seccomp.drop=foo_x86,foo_arm` should work without changes on every system.

I see, that makes sense.

How about at least warning on unknown syscall names?

I don't think any upstream profile uses arch-specific syscalls, so it shouldn't
significantly increase verbosity in the vast majority of cases.

Also, I'm not too familiar with this, but I'd assume that needing to use an
extra or a different syscall name (rather than just a different syscall number)
on a different architecture would be a rather rare occurrence these days.

At least comparing `syscall_x86_64.h` and `syscall_aarch64.h`, the only
difference seems to be that the latter is missing many syscalls (presumably
mostly ones that are obsolete).

If verbosity becomes an issue, maybe a command could be added to control this,
similarly to `seccomp-error-action`:

```
seccomp-error-action=[kill | ERRNO | log]
```

Such as:

```
seccomp-unknown-syscall-action=[exit | warn | none]
```

Thoughts?

pierretom:
Should fix also #2931.

kmk3:
> Should fix also #2931.

Very nice catch, thanks.

rusty-snake:
> presumably
mostly ones that are obsolete

Or architecture specific features.

> How about at least warning on unknown syscall names?

This or (I did not looked into the code) have a list of "known but not implemented for this arch".

Issue #1743, opened 2018-01-20
Can't combine --private with --whitelist


eevee:
With `--private=DIR`, I can replace my entire home directory with a persistent substitute.  With `--whitelist=DIR`, I can preserve _part_ of my home directory while hiding everything else.  But there doesn't seem to be any way to combine these behaviors — e.g., if I want to run a game (which needs all its data files to exist) and still let it write out a save file to some undocumented place elsewhere in my home directory.

This is also troublesome for writing a flexible general-purpose profile.  If I put `private` in the profile, then I can't use `--whitelist` on the command line at all; if I _don't_ put `private` in the profile, then if I don't use `--whitelist` on the command line my entire home directory remains exposed.

In general it would be nice if `--private`, `--whitelist`, and `--overlay` had clearer interactions with each other.

netblue30:
I started adding support for it. So far I have: private-dev, private-tmp and whitelist support for --overlay and --chroot sandboxes. You can try it out.

Support for --private=dir and --private-home is coming soon.

qutetemp:
>whitelist support for --overlay and --chroot sandboxes. You can try it out.
Do I need the latest master because it doesn't seem to work.

netblue30:
Yes, it is only in master. Also, there are some more general fixes related to overlay there. It will go into the next release.

Ninlives:
What's the status for whitelist support for --private now?

H8to:
Also hoping for this feature!
I use firejail as a Firefox profile template engine. For each of my customers the customer-template jail from a fully configured Firefox is copied to "customer-n", which then should be used for customer stuff (--private=customer-n). However I'd like to access e.g. downloads with other applications and this leads to problems as I cannot whitelist paths outside of the jail. Also copying from within the jailed Firefox is not possible as long as I can't whitelist my home/x/Desktop path.

Vincent43:
Perhaps it's not complete solution for you but it's possible to transfer files from/to `--private` sandbox with `--get`/`--put` options. See `FILE TRANSFER` section in manpage.

H8to:
Thank you for your suggestion, I read through it!

I know that's not really the point of a private "jail", but I'd like to be able to download something in the jailed Firefox and then move it to a folder that's shared with the jail via drag and drop. This is the most comforty option. Put and get are demanding to much effort as I could also then just open Nemo and dragNdrop the files or folders from the jail home/downloads folder.

chiraag-nataraj:
[...]

PR #7067, opened 2026-02-17
whitelist: allow placing an overlay into the sandbox

mardy:
The use case for this option might not be that common, but I though of sharing it anyway. If you think that this feature is not of general use, feel free to just close this.

This is similar to what a few comments requested in https://github.com/netblue30/firejail/issues/1743 and can be though of a generalisation of the `--hosts-file=<file>` option. Our use case is that we unpack the application in a special directory, and then when launching it we map some files and directories from the application package into the sandbox. For example:

     whitelist /usr/share/application-data=/path/to/application/data

This operation might deserve its own command name, but since its functionality is 95% the same as the one of the "whitelist" command, I implemented it by just expanding the syntax:

    whitelist <path>[=<overlay>]

Maybe `overlay`, `map` or `replace` would be better names for this.

Note: if "`<path>`" does not exist, it will be created as a new file or directory (depending on whether "`<overlay>`" is a file or a directory).

### DIFF
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -146,6 +146,7 @@ typedef struct profile_entry_t {
 		char *file;		// resolved file path
 		char *link;		// link path
 		TopDir *top;	// top level directory
+		int overlay_fd; // fd of the overlay (-1 if not overlaying)
 	} *wparam;
 
 } ProfileEntry;
--- a/src/firejail/fs_whitelist.c
+++ b/src/firejail/fs_whitelist.c
@@ -129,7 +129,7 @@ static int whitelist_mkpath(const char *parentdir, const char *relpath, mode_t m
 	return fd;
 }
 
-static void whitelist_file(const TopDir * const top, const char *path) {
+static void whitelist_file_with_overlay(const TopDir * const top, const char *path, int overlay_fd) {
 	EUID_ASSERT();
 	assert(top && path);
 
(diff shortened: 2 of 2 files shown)


rusty-snake:
How would you whitelist a path that contains a equal sign?
--> This is a breaking change and unlikely to be merged in this form.

mardy:
> How would you whitelist a path that contains a equal sign? --> This is a breaking change and unlikely to be merged in this form.

We could do like this: if the argument given to `whitelist` is an existing file or directory, do not even attempt to split it on an equal sign.

kmk3 (changes requested):
> How would you whitelist a path that contains a equal sign? --> This is a
> breaking change and unlikely to be merged in this form.

Indeed.

Also, from the perspective of a user, making `path1=path2` seems like a
different operation compared to just allowing an existing path.

It would probably be better to have its own command for it.

> Maybe `overlay`, `map` or `replace` would be better names for this.

`overlay` sounds like it's related to overlayfs, so I would avoid using that
name unless that's the case.

> The use case for this option might not be that common, but I though of
> sharing it anyway. If you think that this feature is not of general use, feel
> free to just close this.

I think having something generic like this could be useful, such as for
overriding certain paths in `$HOME` inside of the sandbox.

But if system paths are allowed (as is the case with `whitelist`), it would
likely have security implications, such as allowing an unprivileged user to
bind-mount paths into /etc/sudoers and /etc/passwd.

Maybe it could work if both paths (or at least the target path) were limited to
being inside of `${HOME}` or `${RUNUSER}`.

Or the user having read/write access to both paths (would also need to be
careful about TOCTOU races).

mardy:
> But if system paths are allowed (as is the case with `whitelist`), it would likely have security implications, such as allowing an unprivileged user to bind-mount paths into /etc/sudoers and /etc/passwd.

Mmm... can you elaborate on the risk? The user is anyway going to run the program under his own privileges, so I don't think that overriding /etc/sudoers or /etc/password would do him much good. Or does firejail allow running setuid binaries?

> Or the user having read/write access to both paths (would also need to be careful about TOCTOU races).

If I bind mount file `/etc/passwd` over `$HOME/passwd`, the owner and permissions of the file now mounted at `$HOME/passwd` are those of `$HOME/passwd`, so, unless I'm mistaken, I don't get any new privileges here.

PR #7071, opened 2026-02-19
new profile: zed-editor, zed

cobratbq:
Profile for the [Zed editor](<https://zed.dev>).

- `zed.profile` for `zed-linux/bin/zed` shell-script
- `zed-editor.profile` for zed-linux/libexec/zed-editor` program binary

According to `lsns`, executed within the firejail-container, Zed-editor does not yet use namespaces. However, I don't think restricting makes sense, due to high likelihood it will get added in the future and the plethora of development tools that can be hosted within Zed-editor or its container.

Best squash this profile when merging.

### DIFF
--- a/etc/profile-m-z/zed-editor.profile
+++ b/etc/profile-m-z/zed-editor.profile
@@ -15,7 +15,7 @@ include disable-exec.inc
 #include disable-interpreters.inc
 include disable-programs.inc
 include disable-xdg.inc
-# TODO: probably want to exchange this for explicits and `disable-...` entries.
+# TODO: probably want to exchange this for explicit and `disable-...` entries.
 #include default.profile
 
 novideo
--- a/etc/profile-m-z/zed-editor.profile
+++ b/etc/profile-m-z/zed-editor.profile
@@ -54,7 +54,9 @@ private-tmp
 
 # TODO: likely not complete.
 dbus-user filter
-dbus-user.talk org.freedesktop.secrets
+dbus-user.talk org.freedesktop.portal.Desktop
+# Add below line to `~/.config/firejail/zed-editor.local` to enable keyring-access.
+#dbus-user.talk org.freedesktop.secrets
 dbus-user.talk org.a11y.Bus
 dbus-system none
 
--- a/etc/profile-m-z/zed-editor.profile
+++ b/etc/profile-m-z/zed-editor.profile
@@ -1,0 +1,33 @@
+# Profile for ./libexec/zed-editor
+include zed-editor.local
+
+noexec /tmp
+noblacklist ${HOME}
+include allow-common-devel.inc
+# NOTE: add `noblacklist` entries here, such as `noblacklist ${HOME}/.cargo`, to
+# support more programming languages and development tools.
+
+# TODO: probably want to exchange this for explicits and `disable-...` entries.
+include default.profile
+
+novideo
+nosound
+nodvd
+notv
+nou2f
+
+private-dev
+private-tmp
+
+# TODO: not tested, seen working.
+dbus-user filter
+dbus-system filter
+dbus-system.talk org.freedesktop.secrets
+
+seccomp
+restrict-namespaces
+caps.drop all
+nonewprivs
+deterministic-exit-code
+deterministic-shutdown
+
--- a/etc/profile-m-z/zed-editor.profile
+++ b/etc/profile-m-z/zed-editor.profile
@@ -1,14 +1,22 @@
 # Profile for ./libexec/zed-editor
 include zed-editor.local
+# Persistent global definitions
+include globals.local
 
-noexec /tmp
 noblacklist ${HOME}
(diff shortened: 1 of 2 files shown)


kmk3 (changes requested):
The profiles do not follow the usual conventions.

See the templates in etc/templates and the profiles of other text editors.

Also related:

* #6002

cobratbq:
Thanks, will do, thanks for early feedback. I marked it draft, because I'm still fine-tuning. I checked the template, but it seemed to follow a different order than most profiles I looked at, so had postponed looking into reordering as I was still fine-tuning.

kmk3:
> Thanks, will do, thanks for early feedback. I marked it draft, because I'm
> still fine-tuning.

No problem.

> I checked the template, but it seemed to follow a different order than most
> profiles I looked at, so had postponed looking into reordering as I was still
> fine-tuning.

Which parts seem to be out of order?

cobratbq:
> Which parts seem to be out of order?

What _seemed_ out of order, were the various restrictions, with apparmor and noXXX etc. all showing up in the same huge list. However, I think I was just left with the wrong impression because profiles are quite distributed over files.

cobratbq:
Questions:
- Any recommendations concerning Landlock? It is likely not easy to apply, because of large variety of requirements due to support for many languages and tools.
- I tried using the _-devel_ includes that seem to be available for this exact purpose, but I am not perfectly familiar with firejail or its available infrastructure.
- Additional feedback is always welcome.
- I welcome testers, because I myself am not using Zed very comprehensively and I might have been too restrictive for that reason.

etc/profile-m-z/zed-editor.profile
kmk3, line 1: 1.

> ```
> noblacklist ${HOME}
> ```

Not blacklisted.

2.

```
# NOTE: add common development-related modifications, such as blacklisting, whitelisting, noexec, and other grants and permissions, to `~/.config/firejail/allow-common-devel.local`.
include allow-common-devel.inc
```

Remove this comment and add the comment from the template.

3. `~/.config/firejail/zed-editor.local` -> zed-editor.local (for consistency
   with the other profiles)

4. Remove landlock section (it's experimental and no profile uses it).

5.

> ```
> # For hardening: `seccomp.block-secondary`
> ```

Why not enable it?

etc/profile-m-z/zed.profile
kmk3, line 1: 6.

> ```
> ignore deterministic-shutdown
> ignore deterministic-exit-code
> ```

Why?

7. The main profile should be for the main program and the redirect profile for
   the shell script.

8. Add `# Redirect` comment.

See the redirect template in etc/template.

kmk3 (changes requested):

etc/profile-m-z/zed-editor.profile
cobratbq: 1. Right, left-over from when I started out including default profile.
2. Comment copied from template.
3. Done.
4. Works for me; it would've been a whole separate investigation.
5. Because I cannot predict if there are, for whatever reason, 32-bit programming tools and from my understanding this would block anything not exactly matching the running architecture.
6. `/opt/zed/bin/zed` is a binary that then executes `/opt/zed/libexec/zed-editor`. None are shell-scripts. One seems to only redirect and possibly assign a socket for some IPC. I'll drop the ignores. It seems that recent versions of Zed have changed behavior for /opt/zed/bin/zed binary.
7. It seems like it is not purely a redirect as a shell-script typically accomplishes, but it seems to do little more.
8. Done.

etc/profile-m-z/zed.profile
cobratbq: See https://github.com/netblue30/firejail/pull/7071#discussion_r3148922554

etc/profile-m-z/zed-editor.profile
kmk3: > 3. Done.

Remove the backticks around filenames (it's fine for commands) for consistency
with the other profiles.

> 5. Because I cannot predict if there are, for whatever reason, 32-bit
>    programming tools and from my understanding this would block anything not
>    exactly matching the running architecture.

In that case it makes sense to be explicit:

```
# Add the following line to zed-editor.local if you don't need to run 32-bit
# programming tools:
#seccomp.block-secondary
```

> 6. `/opt/zed/bin/zed` is a binary that then executes
>    `/opt/zed/libexec/zed-editor`. None are shell-scripts. One seems to only
>    redirect and possibly assign a socket for some IPC.

> 7. It seems like it is not purely a redirect as a shell-script typically
>    accomplishes, but it seems to do little more.

I mean that the main profile should be for the main program (usually in
/usr/bin, /usr/lib or /opt) and that the profile for the wrapper (usually a
shell script in /usr/bin) should be the one that redirects to it (and contains
less entries).

In this case it seems to already be correct then, so nevermind.

> 8. Done.

Remove " (`bin/zed`)".

Still missing things in the header.

Again, see the redirect template.

Issue #6675, opened 2025-03-04
Firefox warning message about user namespace after upgrade to AppArmor 4


vinc17fr:
### Description

Since the upgrade to AppArmor 4 on Debian/unstable, Firefox displays the following warning message: "Some of Firefox's security features may offer less protection on your current operating system."

And there is a link to https://support.mozilla.org/en-US/kb/install-firefox-linux#w_security-features-warning

In particular:
> The sandbox in Firefox makes use of unprivileged user namespaces when creating new processes for enforcing more security. This can be considered a security risk, therefore some Linux distributions have started to restrict its usage and only allow it to work where there is an AppArmor profile.

So the Firefox AppArmor profile has `userns` to allow user namespaces.

The issue is that with Firejail, the AppArmor profile is `firejail-default`, not the Firefox one, thus it does not have `userns`.

### Steps to Reproduce

Run `firejail /usr/bin/firefox` and continue until a Firefox browser window is open.

### Expected behavior

The Firefox window should not show a warning message.

### Actual behavior

The following warning message is shown above the main area: "Some of Firefox's security features may offer less protection on your current operating system."

### Behavior without a profile

_What changed calling `LC_ALL=C firejail --noprofile /path/to/program` in a
terminal?_

There is no warning message. But I suppose that this is because the `firefox` profile uses AppArmor (which is probably wanted), according to the contents of the `firefox-common.profile` file:
```
apparmor
# Fixme!
apparmor-replace
```
while this is not the case without a profile.

### Additional context

My Debian bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1098869

This may be similar to #6368 for Chromium.

### Environment

- Name/version/arch of the Linux kernel (`uname -srm`): `Linux 6.11.10-amd64 x86_64`
- Name/version of the Linux distribution (e.g. "Ubuntu 20.04" or "Arch Linux"): Debian/unstable
- Name/version of the relevant program(s)/package(s) (e.g. "firefox 134.0-1,
  mesa 1:24.3.3-2"): Firefox 135.0.1
- Version of Firejail (`firejail --version`): 0.9.72

### Checklist

- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
[...]

PR #7080, opened 2026-02-25
modif: allow `userns` in firejail-default apparmor profile

### LINKED ISSUE #7078: AppArmor profile does not grant `userns` permissions
### Description

Is AppArmor not granted `userns` privileges when running within AppArmor?

Newer AppArmor versions have a `userns` permission that allows access to user-namespaces when these are otherwise restricted. The profile `firejail-default` does not set this permission.

I suspect this inhibits the capabilities of `firejail` itself. The issue, but specifically for firefox, is also discussed in #6675.

### Steps to Reproduce

1.  Run `/usr/bin/firejail --debug /usr/bin/firefox --no-remote` (without making any changes)
2.  Notice an error message `[56] Sandbox: CanCreateUserNamespace() clone() failure: EACCESS`.

### Expected behavior

No error-message.

### Actual behavior

Application starts, but is restricted from anything concerning user-namespaces even if not restricted by firejail options/profiles, because firejail itself is restricted.

Not tested. Adding `userns` to `/etc/apparmor.d/firejail-default` already solves the problem, i.e. makes the difference.

### Additional context

From reading other information, i.e. #6675, I gather this is a recent change/feature in AppArmor 4, which I guess is in Debian.

This effectively weakens Firefox's inherent sandboxing capabilities because they are restricted from using user-namespaces. It also seems that user-namespaces provide value with Firefox's sandboxing capacity.

Also, given that subsequently capabilities are dropped, there should be reduced risk even if user-namespaces are in use.(?)

### Environment

- Debian trixie
- `6.12.73+deb13-amd64`
- AppArmor enforced for `firejail`

### Checklist

- [X] I am using firejail [0.9.78 or later](https://github.com/netblue30/firejail/tree/master/SECURITY.md)
- [X] I am using the full program path (e.g. `firejail /usr/bin/vlc` instead of `firejail vlc`; see `https://github.com/netblue30/firejail/issues/2877`)
- [X] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [X] I can reproduce the issue without custom modifications (e.g. globals.local).
- [X] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [X] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [X] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [X] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [X] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

N/A

cobratbq:
Add `userns` to the AppArmor profile for firejail, such that with AppArmor enforcing restrictions, firejail is granted sufficient permissions to exert full control over the capabilities and permissions it is managing.

Fixes: #7078 

- question: Is the child-process of firejail already properly handled w.r.t. the AppArmor profile? Already before this patch, the spawned `firejail` child-process would likely be subjected to some influence from apparmor, or not?

The open question should not matter more/less specifically for this patch.

### DIFF
--- a/etc/apparmor/firejail-default
+++ b/etc/apparmor/firejail-default
@@ -15,6 +15,7 @@
 @{PID}={[1-9],[1-9][0-9],[1-9][0-9][0-9],[1-9][0-9][0-9][0-9],[1-9][0-9][0-9][0-9][0-9],[1-9][0-9][0-9][0-9][0-9][0-9],[1-4][0-9][0-9][0-9][0-9][0-9][0-9]}
 
 profile firejail-default flags=(attach_disconnected,mediate_deleted) {
+userns,
 
 ##########
 # Allow D-Bus access. It may negatively affect security. Comment those lines or
--- a/etc/apparmor/firejail-default
+++ b/etc/apparmor/firejail-default
@@ -15,6 +15,12 @@
 @{PID}={[1-9],[1-9][0-9],[1-9][0-9][0-9],[1-9][0-9][0-9][0-9],[1-9][0-9][0-9][0-9][0-9],[1-9][0-9][0-9][0-9][0-9][0-9],[1-4][0-9][0-9][0-9][0-9][0-9][0-9]}
 
 profile firejail-default flags=(attach_disconnected,mediate_deleted) {
+
+##########
+# Allow unprivileged user-namespaces. It may negatively affect security.
+# These privileges are granted to firejail in case they are needed and are
+# expected to be dropped by firejail as required.
+##########
 userns,
 
 ##########


kmk3 (changes requested):
@cobratbq [on Feb 25](https://github.com/netblue30/firejail/issues/7078#issuecomment-3961768082):

> > > IIUC, adding `userns` to the AppArmor-profile for firejail is effectively the
> > > same as previous AppArmor without this addition, because it only recent got
> > > to being managed. It effectively allows user-namespaces thus deferring to
> > > `firejail` to determine whether to continue to allow it.
> > 
> > Nice, since this makes the profile work the same as in AppArmor 3, sounds
> > good to me.
> 
> Assuming that this `abi/4.0` indicates a "language" version, thus available
> keywords, then AppArmor 3 would likely reject the profile, but would do so
> already. (I'm guessing as to its meaning, so I'm not 100% sure.)

Judging by #6675, AppArmor 4 seems to be relatively recent, so I'd expect many
systems to still be using AppArmor 3.x.

If the current profiles work in AppArmor 3.x but fail with this change, then it
might be better to avoid it for now.

Can you test with AppArmor 3.x to confirm what happens?

kmk3:
> Is the child-process of firejail already properly handled w.r.t. the AppArmor
> profile? Already before this patch, the spawned `firejail` child-process
> would likely be subjected to some influence from apparmor, or not?

AIUI firejail itself uses a very permissive apparmor profile
("firejail-base"?), then after setting up the sandbox, it applies a more
restrictive one ("firejail-default" I believe) right before starting the
program.

cobratbq:
> Judging by #6675, AppArmor 4 seems to be relatively recent, so I'd expect many systems to still be using AppArmor 3.x.
> 
> If the current profiles work in AppArmor 3.x but fail with this change, then it might be better to avoid it for now.
> 
> Can you test with AppArmor 3.x to confirm what happens?

I'll see if I can check a couple of things.

We may be able to define profiles with a `include <abi/3.0>` or `include <abi/4.0>` as distinguisher. It depends on whether AppArmor understands to silently skip the incompatible profile. It seems `firejail-default` does not currently include any definition of an ABI, thus making it compatible with older versions.

Converted to draft to reflect the open issue.

kmk3:
> > Judging by #6675, AppArmor 4 seems to be relatively recent, so I'd expect
> > many systems to still be using AppArmor 3.x.  If the current profiles work
> > in AppArmor 3.x but fail with this change, then it might be better to avoid
> > it for now.  Can you test with AppArmor 3.x to confirm what happens?
> 
> I'll see if I can check a couple of things.
> 
> We may be able to define profiles with a `include <abi/3.0>` or `include
> <abi/4.0>` as distinguisher. It depends on whether AppArmor understands to
> silently skip the incompatible profile. It seems `firejail-default` does not
> currently include any definition of an ABI, thus making it compatible with
> older versions.

Good catch.

Maybe AppArmor 4 assumes `abi/4.0` by default (such as by implicitly including
`<abi/4.0>` if no other `<abi/x>` is included).

So if we just add `include <abi/3.0>` and target 3.x, maybe it would allow user
namespaces by default, just like with AppArmor 3.x.

Edit: That is, just for the sake of fixing the user namespace support (and the
profile in general, as other things could be broken as well).

Then we could add a specific profile for 4.x, etc.

cobratbq:
Some AppArmor notes:

- [AppArmor 3.0 release notes](<https://gitlab.com/apparmor/apparmor/-/wikis/Release_Notes_3.0>)
- [`abi` feature](<https://gitlab.com/apparmor/apparmor/-/wikis/AppArmorpolicyfeaturesabi>) introduced in AppArmor 3.0.
- [AppArmor 4.0 release notes](<https://gitlab.com/apparmor/apparmor/-/wikis/Release_Notes_4.0.1>)
- Distrowatch: Debian
  - 13: `4.1`
  - 12: `3`
  - 11: `2.13.6`
  - 10: `2.13.2`
- Distrowatch: Ubuntu
  - newer: `3.0` or higher
  - 20.04 LTS: `2.13.3`

netblue30:
Thanks @cobratbq. It seems to be working fine on Debian stable, but let's wait until next week. We have a new release coming over the weekend, we'll merge it in after that.

cobratbq:
> Thanks @cobratbq. It seems to be working fine on Debian stable, but let's wait until next week. We have a new release coming over the weekend, we'll merge it in after that.

So, I'm not sure what the desired solution is for the profiles? I don't think you can write one profile version, so you would have to conditionally package one conditional on AppArmor version. That's been bugging me a bit.

__On another note__:

Well, if that's in place. Any chance you can check out running Firefox without `apparmor-replace`? I got asked, so I was curious and checked just now. I think it just works:
- Firefox apparmor-profile enables `userns` in `unconfined` mode;
- Firejail in enforcing mode for me;
- running `firejail` with `--apparmor` and `--ignore=apparmor-replace`.

Apparmor ends up in a mixed-mode setup of Firejail `enforcing` and Firefox `unconfined`. (I think)

cobratbq:
@kmk3 @netblue30 did you check https://github.com/netblue30/firejail/pull/7080#issuecomment-3969141566 ? It might be worth evaluating whether `apparmor-replace` still has value.

PR #7084, opened 2026-02-28
profiles: mumble: include whitelist-runuser-common

sofoxe1:


sofoxe1:
it breaks alsa backend for microphone but since pipewire and pulseaudio both work i would say yes

kmk3:
> it breaks alsa backend for microphone but since pipewire and pulseaudio both
> work i would say yes

Interesting, I'm not aware of anything in /run that would affect alsa.

If this is a common path, it might also affect other profiles.

Is this X11 or wayland?

What desktop environment?

Could you try whitelisting paths in /run in ~/.config/firejail/mumble.local
until it works?

Example:

```
whitelist /run/foo1
whitelist /run/foo2
```

kmk3:
Merged just the include of whitelist-runuser-common.

The include of whitelist-run-common can be left as a potential improvement.

If you ever check what /run paths are needed, feel free to post your findings
in here or in a new PR.

Issue #7069, opened 2026-02-19
Add --keep-hostname command


kmk3:
Issue created to track the addition of the `--keep-hostname` command and
related changes.

Related commits:

* cc8b019b5 ("--keep-hostname part 1 (#7048)", 2026-02-03)
* 6f164f415 ("--keep-hostname part 2 (#7048)", 2026-02-03)
* ef7f8ee83 ("--keep-hostname part 3 (#7048)", 2026-02-08)
* 026332a81 ("merges", 2026-02-08)
* fbc94070e ("adding keep-hostname to libreoffice.profile", 2026-02-11)
* e31d872a5 ("profiles: add keep-hostname to profile.template", 2026-02-11)
* df75e45cf ("profiles: add keep-hostname to noprofile.profile", 2026-02-19)

Relates to:

* #7048
* #7062

kmk3:
Removed in:

* #7095

Issue #7048, opened 2026-01-29
No fallback to ipv4 from ipv6


liloman:
### Description

I was digging to resolve a problem with a sandbox till I discovered the root issue, the sandbox wasn't trying both ip stacks ,as a fallback, inside the sandbox . 

### Steps to Reproduce

firejail --noprofile telnet localhost 80
firejail version 0.9.78

Parent pid 46456, child pid 46457
Base filesystem installed in 0.04 ms
Child process initialized in 17.78 ms
Trying ::1...
telnet: connect to address ::1: Connection refused

Parent is shutting down, bye...

### Expected behavior

firejail telnet localhost 80
Trying ::1...
telnet: connect to address ::1: Connection refused
Trying 127.0.0.1...
telnet: connect to address 127.0.0.1: Connection refused

### Additional context

I've been trying to understand what's the difference but without a conclusion. Even using --ignore=private-etc keeps failing.
The real issue was android studio trying to connect to a local port for debugging purposes and failing to do it so, cause not trying on ipv4 after ipv6 failed ...

Unfortunately jail_prober.py wasn't a help here cause it affects even noprofile  :(

### Environment

- Version of Firejail firejail --version
firejail version 0.9.78

### Checklist

- [x ] I am using firejail [0.9.78 or later](https://github.com/netblue30/firejail/tree/master/SECURITY.md)
- [ x] I am using the full program path (e.g. `firejail /usr/bin/vlc` instead of `firejail vlc`; see `https://github.com/netblue30/firejail/issues/2877`)
- [ x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [ x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [ x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [ x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
- [ x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [x ] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ ] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

<details>
<summary>Output of <code>LC_ALL=C firejail /path/to/program</code></summary>
<p>

```
output goes here
```

</p>
</details>

[...]

PR #7095, opened 2026-03-07
modif: replace --keep-hostname with new --hostname-randomize

### LINKED ISSUE #7062: xorg: Authorization required, but no authorization protocol specified
Hey,

I'm using firejail-git version 0.9.78.r114.g88652cdb3-1 on Arch Linux (just built it a few seconds ago). I cannot start  Thunderbird. I have the same problem with firefox:

```
kmille@spring:~# /usr/local/bin/thunderbird
Reading profile /etc/firejail/thunderbird.profile
Reading profile /etc/firejail/firefox-common.profile
Reading profile /etc/firejail/disable-proc.inc
firejail version 0.9.79

Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
Warning: /usr/bin/xdg-dbus-proxy was not found, downgrading dbus-user policy to allow.
To enable DBus filtering, install the xdg-dbus-proxy program.
Ignoring "dbus-user.own org.mozilla.thunderbird.*" and 4 other dbus-user filter rules.
Parent pid 436820, child pid 436821
Private /etc installed in 164.47 ms
Warning: not remounting /run/user/1001/gvfs
Base filesystem installed in 405.62 ms
Seccomp list in: !chroot, check list: @default-keep, prelist: unknown,
Warning: Replacing profile instead of stacking it. It is a legacy behavior that can result in relaxation of the protection. It is here as a temporary measure to unbreak the software that has been broken by switching to the stacking behavior.
Child process initialized in 741.52 ms
[41] Sandbox: CanCreateUserNamespace() clone() failure: EPERM
Authorization required, but no authorization protocol specified

Error: cannot open display: :0

```

Regarding the AppArmor thing (I'm running `6.17.13-hardened1-3-hardened`):

```
kmille@spring:~# sudo aa-enforce firejail-default                   
[sudo] password for kmille: 
Setting /etc/apparmor.d/firejail-default to enforce mode.

ERROR: Cache read/write disabled: interface file missing. (Kernel needs AppArmor 2.4 compatibility patch.)
Warning: unable to find a suitable fs in /proc/mounts, is it mounted?
Use --subdomainfs to override.
```

kmk3:
Changes:

* Keep hostname by default (same as using `--keep-hostname`)
* Add  `--hostname-randomize` command to randomize the hostname
* Ignore `--keep-hostname` command and print a warning if it is used

Setting a different hostname inside of the sandbox may prevent X11
programs from authenticating to the X server and displaying windows at
all (see #7062).

To avoid breakage, keep the hostname as is by default and only set it to
a random value if a new `hostname-randomize` command is used.

This also avoids potentially surprising behavior, as the user might not
expect the hostname to be changed inside of the sandbox, considering
that usually the protections that are applied firejail involve
restricting access to resources (like file paths), rather than modifying
their values inside of the sandbox.

Related commits:

* cc8b019b5 ("--keep-hostname part 1 (#7048)", 2026-02-03)
* 6f164f415 ("--keep-hostname part 2 (#7048)", 2026-02-03)
* ef7f8ee83 ("--keep-hostname part 3 (#7048)", 2026-02-08)
* 026332a81 ("merges", 2026-02-08)
* fbc94070e ("adding keep-hostname to libreoffice.profile", 2026-02-11)
* e31d872a5 ("profiles: add keep-hostname to profile.template", 2026-02-11)
* df75e45cf ("profiles: add keep-hostname to noprofile.profile", 2026-02-19)

Fixes #7062

Relates to #7048 #7069.


netblue30:
merged!

PR #7098, opened 2026-03-11
bugfix: lib: fix memory leaks in `syscall_in_list()`

pierretom:
`asprintf()` overwrites the value of `ptr->xxx` with the new pointer.
Result: the older allocation is never freed.

Should also help to pass a test for the PR #7066.


kmk3:
> Should also help to pass a test for the PR #7066.

Very nice, thanks.

PR #7099, opened 2026-03-11
tests: fix strace color probing in allow-debuggers & seccomp-ptrace

TheJJ:
if strace runs in a terminal, it probes the background color to select its color palette. this probing expects a reply, but due to expect intercepting the io, the answer isn't sent back to strace, so it never starts printing the expected output.

this happens in the ubuntu automatic testing: https://bugs.launchpad.net/ubuntu/+source/firejail/+bug/2143830

### DIFF
--- a/test/environment/allow-debuggers.exp
+++ b/test/environment/allow-debuggers.exp
@@ -5,6 +5,10 @@
 
 set timeout 10
 cd /home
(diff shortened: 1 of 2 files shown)


netblue30:
Merged, thanks!

Issue #7060, opened 2026-02-10
Update Ubuntu PPA to 0.9.76


kmk3:
Links:

* Ubuntu PPA: <https://launchpad.net/~deki/+archive/ubuntu/firejail>

Relates to:

* #6702
* #6842
* #6691

Issue #6842, opened 2025-07-30
Update Debian package to 0.9.76


kmk3:
Links:

* Debian Package Tracker: <https://tracker.debian.org/pkg/firejail>
* Debian Package (Sid): <https://packages.debian.org/sid/firejail>
* Debian CI: <https://salsa.debian.org/reiner/firejail>

Relates to:

* #6702
* #7060
* #6691

kmk3:
Current status on packages.debian.org:

* Debian 11 (Bullseye): firejail 0.9.64.4-2+deb11u1 - <https://packages.debian.org/bullseye/firejail>
* Debian 12 (Bookworm): firejail 0.9.72 - <https://packages.debian.org/bookworm/firejail>
* Debian 13 (Trixie): firejail 0.9.74 - <https://packages.debian.org/trixie/firejail>
* Debian 14 (Forky): firejail 0.9.76 - <https://packages.debian.org/forky/firejail>
* Debian Testing: firejail 0.9.76 - <https://packages.debian.org/testing/firejail>
* Debian Unstable (Sid): firejail 0.9.78 - <https://packages.debian.org/sid/firejail>

Following the link from README.md:

* <https://packages.debian.org/bullseye-backports/firejail>

It says:

> Error: Package not available in this suite.

@reinerh

Does this mean that 0.9.76 cannot be easily installed on Debian 11 anymore?

What is the recommended way of installing firejail on Debian 11-13?

Can users still just install a binary package or should they compile from
source?

If a binary install is still supported, what are the commands to do it?

I'm trying to clarify the current status so that we can update README.md if
needed to help guide users on how to install a supported firejail version on
Debian.

Also, I'm assuming the firejail version cannot be simply upgraded directly
(such as 0.9.74 -> 0.9.76) on supported Debian versions.  If the upgrade can
actually be done, please clarify.

Related comment:

* <https://github.com/netblue30/firejail/issues/7057#issuecomment-3879613393>

PR #7110, opened 2026-03-17
docs: recommend .deb or building from source on debian/ubuntu

kmk3:
Most recent releases:

* firejail 0.9.72: 2023-01-16
* firejail 0.9.74: 2025-03-24
* firejail 0.9.76: 2025-07-30
* firejail 0.9.78: 2026-01-03
* firejail 0.9.80: 2026-03-14

firejail 0.9.76 was released over 6 months ago, but the packages from
both Debian stable (13 / Trixie) and the Ubuntu PPA appear to still be
on firejail 0.9.74, which is over 1 year old[1] [2].

As for installing firejail through Debian backports, it is unclear to me
if that is currently working and if so, which firejail version would be
installed on each Debian version.

Lastly, the packages on Ubuntu seem to still be on firejail 0.9.72,
which is over 3 years old, even on the latest Ubuntu 25.10 and on the
upcoming Ubuntu 26.04[3].

So to avoid bugs and bug reports caused by old firejail versions,
recommend either installing the release .deb file from GitHub or
building from source on Debian/Ubuntu.

Relates to #6842 #7060.

[1] https://tracker.debian.org/pkg/firejail
[2] https://launchpad.net/~deki/+archive/ubuntu/firejail
[3] https://launchpad.net/ubuntu/+source/firejail

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,10 +91,34 @@ For the supported versions, see [SECURITY.md](SECURITY.md).
 
 ### Debian
 
+Note: The versions from Debian stable and backports are likely to be outdated,
+so currently we recommend either downloading and installing the .deb package
+from the latest release:
+
+* <https://github.com/netblue30/firejail/releases/latest>
+
+Or [building from source](#building).
+
+<details>
+<summary>Old instructions</summary>
+
 Debian stable (bullseye): We recommend to use the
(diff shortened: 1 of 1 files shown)


kmk3:
Cc: @reinerh

PR #7112, opened 2026-03-20
profiles: disable-exec: add mount points

pierretom:
Example case: you want to access the photos and have scripts or binaries on the same USB flash drive.
Let's set mount points not executable in disable-exec.inc.

Relates to:

* #7111

### DIFF
--- a/etc/inc/disable-exec.inc
+++ b/etc/inc/disable-exec.inc
@@ -6,6 +6,10 @@ noexec ${HOME}
 noexec ${RUNUSER}
 noexec /dev/mqueue
 noexec /dev/shm
+noexec /media
+noexec /mnt
+noexec /run/media
+noexec /run/mount
 noexec /run/shm
 noexec /tmp
 # /var is noexec by default for unprivileged users


kmk3:
(Offtopic)

> disable-exec: add mount points

This might be pedantic, but I think that a mount point refers to the exact path
where something is mounted, which is usually inside of paths like /run/media
(such as on /run/media/foo), not on /run/media itself (the same would apply to
/mnt and the other paths).

See also `mount(8)`.

pierretom:
> Suggestion:
> 
> ```diff
> diff --git a/etc/inc/disable-exec.inc b/etc/inc/disable-exec.inc
> index d7dcef7e7..e6b076107 100644
> --- a/etc/inc/disable-exec.inc
> +++ b/etc/inc/disable-exec.inc
> @@ -2,6 +2,12 @@
>  # Persistent customizations should go in a .local file.
>  include disable-exec.local
>  
> +# Note: Keep this in sync with disable-write-mnt.inc.
> +noexec /media
> +noexec /mnt
> +noexec /run/media
> +noexec /run/mount
> +
>  noexec ${HOME}
>  noexec ${RUNUSER}
>  noexec /dev/mqueue
> diff --git a/etc/inc/disable-write-mnt.inc b/etc/inc/disable-write-mnt.inc
> index 01f57cb0f..0d6988d10 100644
> --- a/etc/inc/disable-write-mnt.inc
> +++ b/etc/inc/disable-write-mnt.inc
> @@ -2,6 +2,7 @@
>  # Persistent customizations should go in a .local file.
>  include disable-write-mnt.local
>  
> +# Note: Keep this in sync with disable-exec.inc.
>  read-only /media
>  read-only /mnt
>  read-only /run/media
> ```

Thanks for the suggestion, but I don’t see why it would be linked to this file.
We block the execution, this has nothing to do with the write access.

> (Offtopic)
> 
> > disable-exec: add mount points
> 
> This might be pedantic, but I think that a mount point refers to the exact path where something is mounted, which is usually inside of paths like /run/media (such as on /run/media/foo), not on /run/media itself (the same would apply to /mnt and the other paths).
> 
> See also `mount(8)`.

Mount points is correct, /media, /mnt, ... are default mount points, the term is used in manpages, kernel docs and the FHS.
Mount paths is not a standard term.

kmk3:
> Thanks for the suggestion, but I don’t see why it would be linked to this
> file. We block the execution, this has nothing to do with the write access.

The point is not the command used, but that they refer to the same list of
directories.

If a new mount directory is added to one include, it should also be added to
the other.

For example, if `noexec /some/new/mnt` is added to disable-exec.inc, then
`read-only /some/new/mnt` should be added to disable-write-mnt.inc and
vice-versa.

It's easy to forget to do that when there is no indication that the two lists
are related, such as in reviews.

Also, if a user wants to add their own custom mount directory to a .local (such
as ~/mnt), the comments indicate that it should probably be added to the .local
profile of both includes.  Or more simply, to remind the user to use both
`noexec` + `read-only` (rather than just one of them) for the new path.

> > (Offtopic)
> > 
> > > disable-exec: add mount points
> > 
> > 
> > This might be pedantic, but I think that a mount point refers to the exact
> > path where something is mounted, which is usually inside of paths like
> > /run/media (such as on /run/media/foo), not on /run/media itself (the same
> > would apply to /mnt and the other paths).  See also `mount(8)`.
> 
> Mount points is correct, /media, /mnt, ... are default mount points, the term
> is used in manpages, kernel docs and the FHS.

/mnt itself is sometimes used as mount point (such as when installing Arch in
the CLI), but paths like /run/media are usually managed by a system service or
desktop environment and the mount points are usually on paths like
/run/media/foo and /run/media/bar, not directly on /run/media or /media.

That is, "mountpoint" is a specific term with a specific meaning and if
`mountpoint /run/media` returns something other than 0 in effectively all cases
(unless you go out of your way to change that), then I don't think it makes
sense to call /run/media a mountpoint.

> Mount paths is not a standard term.

I don't think there is a standard term to refer to paths like /media and
/run/media.

"mount paths" just refers to "paths related to mounting".

That is, it's not meant to be a term/thing in itself like "mountpoint".

pierretom:
Endless discussion, I'm merging, feel free to make your own changes.

PR #7114, opened 2026-03-25
profiles: steam: allow ~/.cache/nvidia to improve game performance

acuifex:
Improves launch performance for Team Fortress 2 on legacy OpenGL launch option and possibly other games.
Went from 100 seconds to launch the game and ~8% cpu usage during launch (~1 core on my machine?)
To 60 second and ~16% cpu usage (~2 cores)

It might be worth putting that whitelist into other profiles that in any way use OpenGL (e.g. allow-opengl-game.inc and lutris.profile)

### DIFF
--- a/etc/profile-m-z/steam.profile
+++ b/etc/profile-m-z/steam.profile
@@ -103,6 +103,7 @@ mkdir ${HOME}/.steam
 mkdir ${HOME}/Zomboid
 mkfile ${HOME}/.steampath
 mkfile ${HOME}/.steampid
+whitelist ${HOME}/.cache/nvidia
 whitelist ${HOME}/.config/Epic
 whitelist ${HOME}/.config/Loop_Hero
 whitelist ${HOME}/.config/MangoHud


kmk3 (commented):
> Improves launch performance for Team Fortress 2 on legacy OpenGL launch
> option and possibly other games.  Went from 100 seconds to launch the game
> and ~8% cpu usage during launch (~1 core on my machine?) To 60 second and
> ~16% cpu usage (~2 cores)

Nice finding.

Though isn't the default cache path for the proprietary nvidia driver at
`~/.nv/GLCache`?

Did you manually configure it to use `~/.cache/nvidia`?

On what distribution is this?

> It might be worth putting that whitelist into other profiles that in any way
> use OpenGL

> (e.g. allow-opengl-game.inc

The cache path is not blacklisted and there is no whitelisting in `${HOME}` in
that include, so not sure what commands would make sense to add there.

Also, note that this profile seems to be mostly intended to just allow running
shell script wrappers (see also #4071).

> and lutris.profile

Yes, that would be good (along with wine.profile).

Not sure about any other game-related profiles.

acuifex:
> On what distribution is this?

Arch linux. Nothing too fancy, mostly the default configurations. Using these packages if that somehow makes a difference: `nvidia-dkms xlibre-xserver linux-lts`

> Though isn't the default cache path for the proprietary nvidia driver at `~/.nv/GLCache`?
> 
> Did you manually configure it to use `~/.cache/nvidia`?

Seems to be in `~/.cache/nvidia/GLCache` for my machine. `~/.nv/` seems to have yet another cache `~/.nv/ComputeCache`, which sounds less related to games but might still be worth adding for stuff like blender under steam?

i dont think i've configured much beyond what the arch wiki says.

> > (e.g. allow-opengl-game.inc

Yep, fair enough, i for some reason thought it was doing a lot more. I guess i meant it more as a search guide to find opengl apps/games

> Yes, that would be good (along with wine.profile).

Is that a request for me to add that change to these profiles? The change would be very much speculative if i do it, since i dont know of a reliable way to verify any gains or if it even have any effect there.
I'm still very much figuring out the profile building/checking stuff :p
(edit: while it's still on my mind, i want to mention minecraft and all of its launchers)

Might also worth noting of a possible security/privacy compromise from giving sandboxed apps free reign over changing the cache files (code execution bugs and whatnot) which will be shared by host or other sandboxes

kmk3:
> > Though isn't the default cache path for the proprietary nvidia driver at
> > `~/.nv/GLCache`?  Did you manually configure it to use `~/.cache/nvidia`?
> 
> Seems to be in `~/.cache/nvidia/GLCache` for my machine. `~/.nv/` seems to
> have yet another cache `~/.nv/ComputeCache`, which sounds less related to
> games but might still be worth adding for stuff like blender under steam?
> 
> i dont think i've configured much beyond what the arch wiki says.

I see, so maybe the newer driver versions started to use a more XDG-compliant
path by default.

Is there anything else in `~/.cache/nvidia`?

> > > (e.g. allow-opengl-game.inc
> 
> Yep, fair enough, i for some reason thought it was doing a lot more. I guess
> i meant it more as a search guide to find opengl apps/games

Yeah, that file has a rather specific use case, but indeed it might be helpful
to have a common include for gaming or maybe just graphics acceleration.

Though it depends on what commands would be useful.

> I'm still very much figuring out the profile building/checking stuff :p
> (edit: while it's still on my mind, i'll add minecraft and all of its
> launchers to that list)

I'd say it's mostly done using similar profiles and
etc/templates/profile.template + trial and error.

Now that you mention it, there are a few minecraft-related profiles:

* prismlauncher.profile
* minecraft-launcher.profile

Also, I see now that `~/.nv` is blacklisted in disable.programs.inc.

> Might also worth noting of a possible security/privacy compromise from giving
> sandboxed apps free reign over changing the cache files (code execution bugs
> and whatnot) which will be shared by host or other sandboxes

Yes, it's unfortunate that the same cache is shared by all apps.

This is also the case for the mesa cache path at `~/.cache/mesa_shader_cache`.

Though there is at least `private-cache` for when an empty cache does not
significantly impact performance.

You could also try setting a custom cache path for each profile, though a
comprehensive profile would likely need many commands.

Example:

```
ignore noblacklist ~/.cache/mesa_shader_cache
ignore noblacklist ~/.cache/nvidia
ignore noblacklist ~/.nv
ignore whitelist ~/.cache/mesa_shader_cache
ignore whitelist ~/.cache/nvidia
ignore whitelist ~/.nv

noblacklist /some/mesa/path/foo
noblacklist /some/nvidia/path/foo

mkdir /some/mesa/path/foo
mkdir /some/nvidia/path/foo
whitelist /some/mesa/path/foo
whitelist /some/nvidia/path/foo

env MESA_SHADER_CACHE_DIR /some/mesa/path/foo
env __GL_SHADER_DISK_CACHE_PATH /some/nvidia/path/foo
```

kmk3 (commented):
> > Yes, that would be good (along with wine.profile).
> 
> Is that a request for me to add that change to these profiles? The change
> would be very much speculative if i do it, since i dont know of a reliable
> way to verify any gains or if it even have any effect there.

To keep things simple, I think would be fine to just add `~/.cache/nvidia` to
steam.profile for now.

Then we can try to refactor the nvidia cache paths overall.

Though I'd remove the `mkdir`, as this path is not needed on non-nvidia systems
and on nvidia systems it will likely already have been created by something
else that uses opengl/vulkan.

acuifex:
> Is there anything else in `~/.cache/nvidia`?

Only `~/.cache/nvidia/GLCache` for me

> few minecraft-related profiles

multimc.profile!

> Though I'd remove the mkdir, as this path is not needed on non-nvidia systems
and on nvidia systems it will likely already have been created by something
else that uses opengl/vulkan.

Alright. I was following how whitelists were done in steam.profile so i thought there was a reason behind it. Will force push now!

acuifex:
> I see, so maybe the newer driver versions started to use a more XDG-compliant
path by default.

I Looked up nvidia docs and apparently this was changed in 460.27.04:
https://download.nvidia.com/XFree86/Linux-x86_64/455.45.01/README/openglenvvariables.html
https://download.nvidia.com/XFree86/Linux-x86_64/460.27.04/README/openglenvvariables.html

kmk3 (commented):
> > Though I'd remove the mkdir, as this path is not needed on non-nvidia
> > systems and on nvidia systems it will likely already have been created by
> > something else that uses opengl/vulkan.
> 
> Alright. I was following how whitelists were done in steam.profile so i
> thought there was a reason behind it. Will force push now!

Looks good.

To clarify the reason, a path needs to exist in order to be whitelisted (that
is, bind-mounted into the sandbox `${HOME}` in this case).  If it does not
exist, `whitelist` simply ignores it.  If files are created directly in the
sandbox tmpfs, they only last until the program exits.

So it's important to create the path with `mkdir` for user data to avoid data
loss, but cache can be easily recreated (and will likely already exist in this
case).

> > I see, so maybe the newer driver versions started to use a more
> > XDG-compliant path by default.
> 
> I Looked up nvidia docs and apparently this was changed in 460.27.04:
> https://download.nvidia.com/XFree86/Linux-x86_64/455.45.01/README/openglenvvariables.html
> https://download.nvidia.com/XFree86/Linux-x86_64/460.27.04/README/openglenvvariables.html

Good to know, thanks.

kmk3 (approved):

PR #7126, opened 2026-04-02
docs: add install instructions for Arch

madnight:
## Summary
- add an Arch Linux install section under Installing
- point users to the official Arch package in Extra
- remove the outdated yay/AUR firejail-git instructions from Building

## Testing
- not run (README-only change)

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,21 @@ See the following discussions for details:
 
 </details>
 
+### Arch Linux
+
+Arch Linux provides an official
+[firejail](https://archlinux.org/packages/extra/x86_64/firejail/) package in
+the `Extra` repository.
+
+Install it with:
+
+```sh
+sudo pacman -Syu firejail
+```
+
+After installing, verify the version with `firejail --version`. For optional
+desktop integration, see [Desktop integration](#desktop-integration).
+
 ### Other
 
 Firejail is available in multiple Linux distributions:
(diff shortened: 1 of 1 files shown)


kmk3 (changes requested):
> remove the outdated yay/AUR firejail-git instructions from Building

Using firejail-git is the recommended way of building from source on Arch.

There is nothing "outdated" about it.

> add an Arch Linux install section under Installing
> 
> point users to the official Arch package in Extra

Why?

If the firejail package in a given distribution is up-to-date (that is, it uses
the latest released version from GitHub), the user can just install it normally
from there.

This applies to all distributions; there is nothing special about Arch here.

The sections for Debian and Ubuntu only exist because it is common for their
firejail packages to stay on unsupported versions for a very long time (which
is not the case on Arch).

See also:

* <https://github.com/netblue30/firejail/blob/master/SECURITY.md>

kmk3:
@madnight

Are you still interested in this PR or can it be closed?

madnight:
Can be merged

kmk3:
> Can be merged

This reply and the PR changes make no sense and there does not seem to be any
interest in addressing the review, so closing as low quality.

This is probably AI-generated if I had to guess, but not 100% sure.

Cc: @rusty-snake

madnight:
What should I do? I see no actionable items in the Review.

kmk3:
> What should I do? I see no actionable items in the Review.

The changes from the PR do not seem to add anything useful compared to the
current version, as explained in the review.

Unless you can explain why they would, there is indeed nothing actionable and
we can just leave it as is.

If you are interested in improving the project, I'd suggest for example looking
at open issues (such as for programs not starting) and trying to debug/fix
them.

PR #7127, opened 2026-04-04
build: merge fnettrace headers into `fnettrace_common.h`

pierretom:
It's redundant, plus `fnettrace_icmp.h` and `fnettrace_sni.h` use the same guard macro.


kmk3 (commented):
These headers don't really do anything other than having includes, so you can
just remove them:

* `src/fnettrace_dns/fnettrace_dns.h`
* `src/fnettrace_icmp/fnettrace_icmp.h`
* `src/fnettrace_sni/fnettrace_sni.h`

<details>
<summary>diff</summary>
<p>

```diff
diff --git a/src/fnettrace-dns/fnettrace_dns.h b/src/fnettrace-dns/fnettrace_dns.h
deleted file mode 100644
index f7501269d..000000000
--- a/src/fnettrace-dns/fnettrace_dns.h
+++ /dev/null
@@ -1,34 +0,0 @@
-/*
- * Copyright (C) 2014-2026 Firejail Authors
- *
- * This file is part of firejail project
- *
- * This program is free software; you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation; either version 2 of the License, or
- * (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License along
- * with this program; if not, write to the Free Software Foundation, Inc.,
- * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
-*/
-#ifndef FNETTRACE_DNS_H
-#define FNETTRACE_DNS_H
-
-#include "../include/common.h"
-#include <unistd.h>
-#include <sys/stat.h>
-#include <sys/types.h>
-#include <sys/socket.h>
-#include <netinet/in.h>
-#include <time.h>
-#include <stdarg.h>
-#include <fcntl.h>
-#include <sys/mman.h>
-
-#endif
diff --git a/src/fnettrace-dns/main.c b/src/fnettrace-dns/main.c
index d4cba4bb2..d5739336b 100644
--- a/src/fnettrace-dns/main.c
+++ b/src/fnettrace-dns/main.c
@@ -17,12 +17,13 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
-#include "fnettrace_dns.h"
+#include "../include/common.h"
 #include "../include/gcov_wrapper.h"
 #include <sys/ioctl.h>
 #include <time.h>
 #include <linux/filter.h>
 #include <linux/if_ether.h>
+#include <netinet/in.h>
 #include <sys/prctl.h>
 #include <signal.h>
 #define MAX_BUF_SIZE (64 * 1024)
diff --git a/src/fnettrace-icmp/fnettrace_icmp.h b/src/fnettrace-icmp/fnettrace_icmp.h
deleted file mode 100644
index f6e1daf1f..000000000
--- a/src/fnettrace-icmp/fnettrace_icmp.h
+++ /dev/null
@@ -1,34 +0,0 @@
-/*
- * Copyright (C) 2014-2026 Firejail Authors
- *
- * This file is part of firejail project
- *
- * This program is free software; you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation; either version 2 of the License, or
- * (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License along
- * with this program; if not, write to the Free Software Foundation, Inc.,
- * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
-*/
-#ifndef FNETTRACE_SNI_H
-#define FNETTRACE_SNI_H
-
-#include "../include/common.h"
-#include <unistd.h>
-#include <sys/stat.h>
-#include <sys/types.h>
-#include <sys/socket.h>
-#include <netinet/in.h>
-#include <time.h>
-#include <stdarg.h>
-#include <fcntl.h>
-#include <sys/mman.h>
-
-#endif
diff --git a/src/fnettrace-icmp/main.c b/src/fnettrace-icmp/main.c
index 4c39a7544..d91242bcd 100644
--- a/src/fnettrace-icmp/main.c
+++ b/src/fnettrace-icmp/main.c
@@ -17,12 +17,13 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
-#include "fnettrace_icmp.h"
+#include "../include/common.h"
 #include "../include/gcov_wrapper.h"
 #include <sys/ioctl.h>
 #include <time.h>
 #include <linux/filter.h>
 #include <linux/if_ether.h>
+#include <netinet/in.h>
 #include <sys/prctl.h>
 #include <signal.h>
 #define MAX_BUF_SIZE (64 * 1024)
diff --git a/src/fnettrace-sni/fnettrace_sni.h b/src/fnettrace-sni/fnettrace_sni.h
deleted file mode 100644
index f6e1daf1f..000000000
--- a/src/fnettrace-sni/fnettrace_sni.h
+++ /dev/null
@@ -1,34 +0,0 @@
-/*
- * Copyright (C) 2014-2026 Firejail Authors
- *
- * This file is part of firejail project
- *
- * This program is free software; you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation; either version 2 of the License, or
- * (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License along
- * with this program; if not, write to the Free Software Foundation, Inc.,
- * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
-*/
-#ifndef FNETTRACE_SNI_H
-#define FNETTRACE_SNI_H
-
-#include "../include/common.h"
-#include <unistd.h>
-#include <sys/stat.h>
-#include <sys/types.h>
-#include <sys/socket.h>
-#include <netinet/in.h>
-#include <time.h>
-#include <stdarg.h>
-#include <fcntl.h>
-#include <sys/mman.h>
-
-#endif
diff --git a/src/fnettrace-sni/main.c b/src/fnettrace-sni/main.c
index 0ad15cc03..f5c10a1c5 100644
--- a/src/fnettrace-sni/main.c
+++ b/src/fnettrace-sni/main.c
@@ -17,12 +17,13 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
-#include "fnettrace_sni.h"
+#include "../include/common.h"
 #include "../include/gcov_wrapper.h"
 #include <sys/ioctl.h>
 #include <time.h>
 #include <linux/filter.h>
 #include <linux/if_ether.h>
+#include <netinet/in.h>
 #include <sys/prctl.h>
 #include <signal.h>
 #define MAX_BUF_SIZE (64 * 1024)
```

</p>
</details>

pierretom:
> diff

Yes, we make it even less redundant.

src/include/fnettrace_common.h
```diff
@@ -17,18 +17,19 @@
...
-#include <sys/mman.h>
+#include <linux/filter.h>
+#include <linux/if_ether.h>
+#include <netinet/in.h>
+#include <sys/prctl.h>
+#include <signal.h>
+
+#define MAX_BUF_SIZE (64 * 1024)
 
 #endif
```
kmk3, line 36: Usually a header only includes the files that it directly depends on and the
rest is left to the includers.

That is, speaking of redundancy, the current `#include` lines being similar in
the 3 .c files may seem redundant at first glance, but if a new .c file is
added and it uses a different set of headers, then `fnettrace_common.h` would
be left with redundant headers, needing to be changed again to avoid
unnecessary includes.

So a header including only what it actually needs helps avoid redundancy /
extraneous maintenance as the code evolves.

Other than that, the header has:

```
#define MAX_BUF_SIZE (64 * 1024)
```

I don't think a single simple define is worth its own header, especially since
such limits are usually file-specific anyway in the current codebase.

The wrong guard macro and redundant headers are already good findings by
themselves, so just removing the headers would already be a good improvement.

pierretom: > That is, speaking of redundancy, the current `#include` lines being similar in the 3 .c files may seem redundant at first glance, but if a new .c file is added and it uses a different set of headers, then `fnettrace_common.h` would be left with redundant headers, needing to be changed again to avoid unnecessary includes.

Then do in new.c:
```
#include "../include/fnettrace_common.h"
#include "newheaher.h"
```
The PR is correct for me, the code is more maintainable and avoid repetitions.

kmk3: > > That is, speaking of redundancy, the current `#include` lines being similar
> > in the 3 .c files may seem redundant at first glance, but if a new .c file
> > is added and it uses a different set of headers, then `fnettrace_common.h`
> > would be left with redundant headers, needing to be changed again to avoid
> > unnecessary includes.
> 
> Then do in new.c:
> 
> ```
> #include "../include/fnettrace_common.h"
> #include "newheaher.h"
> ```
> 
> The PR is correct for me, the code is more maintainable and avoid
> repetitions.

That would apply if new.c uses more headers, but my point was that it could
also need less headers, in which case including `fnettrace_common.h` would lead
to new.c including headers that it does not actually need (and then to changing
`fnettrace_common.h` to fix that).

In any case, I looked some more at the 3 main.c files and at least they are
very similar (and only 200-250 lines long).

I'm still skeptical but sure, let's go with the header.

PR #7129, opened 2026-04-10
modif: stop following symlinks to /dev/null on disable

### LINKED ISSUE #5803: ssh Couldn't open /dev/null: Permission denied
I had the same problem as with issue #1725 but with version 0.9.72

I tried what one of the users confirming it there tried, deleting ~/.adobe and ~/.macromedia and it fixed it. Looking at backup it looks to me that the cause was that these files were symlinks to /dev/null and the issue can be reproduced by recreating those links.

IIRC this used to be a way of preventing flash cookies. These are old files hanging around in privacy freaks (e.g. me!) home directories.

The relevant lines from running with firejail --debug seem to be:

    Disable /dev/null (requested /home/graeme/.adobe)

and

    Couldn't open /dev/null: Permission denied
    
I do not think .ssh needs to open these files (unless I try to sftp them or something) so I am not clear why there is an attempt to open them. Is there a way to ignore such things rather than stop the command running? They are blacklisted in disable-common.inc

kmk3:
When trying to prevent a file or directory in the user home from being
written to, it is not uncommon to replace it with a symlink to
/dev/null.

If this path is also blacklisted (such as by disable-common.inc), the
symlink will be followed, resulting in /dev/null itself being
blacklisted, which can cause issues with (unrelated) programs that have
their output redirected to /dev/null (for example).

To avoid disabling /dev/null, when applying commands from
`disable_file()` (such as `blacklist` and `read-only`), if a file is a
symlink to /dev/null, avoid following the symlink and perform the
operation on the link itself instead.

Using these commands with "/dev/null" directly as the argument (that is,
without going through a symlink) should still work the same way.

It has been confirmed to work on Linux 3.8[1], so it should work on at
least 3.8 and later.

Closes #5803.

[1] https://github.com/netblue30/firejail/pull/7129#issuecomment-4233141574

Reported-by: @fgpietersz
Suggested-by: @Changaco
Tested-by: @Changaco
Tested-by: @Zopolis4

### DIFF
--- a/src/firejail/fs.c
+++ b/src/firejail/fs.c
@@ -60,11 +60,11 @@ static int disable_file(OPERATION op, const char *filename) {
 	EUID_ASSERT();
 
 	// Resolve all symlinks
-	char* fname = realpath(filename, NULL);
-	if (fname == NULL && errno != EACCES) {
+	char* rpath = realpath(filename, NULL);
+	if (rpath == NULL && errno != EACCES) {
 		return 1;
 	}
-	if (fname == NULL && errno == EACCES) {
+	if (rpath == NULL && errno == EACCES) {
 		// realpath and stat functions will fail on FUSE filesystems
 		// they don't seem to like a uid of 0
 		// force mounting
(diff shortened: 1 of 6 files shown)


Zopolis4:
Builds fine and relevant test passes against kernel `3.8`

kmk3:
> Builds fine and relevant test passes against kernel `3.8`

Thanks for testing.

Added this detail to the commit message.

PR #7135, opened 2026-04-13
new profile: opencode

aminvakil:
https://github.com/anomalyco/opencode

### DIFF
--- a/etc/profile-m-z/opencode.profile
+++ b/etc/profile-m-z/opencode.profile
@@ -35,7 +35,6 @@ whitelist ${HOME}/.config/opencode
 whitelist ${HOME}/.gitconfig
 whitelist ${HOME}/.local/share/opencode
 whitelist ${HOME}/.local/state/opencode
-
 include whitelist-common.inc
 
 apparmor
--- a/etc/profile-m-z/opencode.profile
+++ b/etc/profile-m-z/opencode.profile
@@ -41,6 +41,10 @@ include disable-xdg.inc
 # whitelist ${HOME}/.local/share/opencode
 # whitelist ${HOME}/.local/state/opencode
 # include whitelist-common.inc
+# include whitelist-run-common.inc
+# include whitelist-runuser-common.inc
+# include whitelist-usr-share-common.inc
+# include whitelist-var-common.inc
 
 apparmor
 caps.drop all
--- a/etc/profile-m-z/opencode.profile
+++ b/etc/profile-m-z/opencode.profile
@@ -1,0 +1,71 @@
+# Firejail profile for opencode
+# Description: The open source coding agent
+# This file is overwritten after every install/update
+quiet
+# Persistent local customizations
+include opencode.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.config/opencode
+noblacklist ${HOME}/.cache/opencode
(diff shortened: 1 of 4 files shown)


etc/profile-m-z/opencode.profile
```diff
@@ -0,0 +1,72 @@
...
+quiet
+# Persistent local customizations
+include opencode.local
+# Persistent global definitions
+include globals.local
+
+noblacklist ${HOME}/.cache/opencode
+noblacklist ${HOME}/.config/opencode
+noblacklist ${HOME}/.local/share/opencode
+noblacklist ${HOME}/.local/state/opencode
```
kmk3, line 13: Add new paths to disable-programs.inc.

Same for:

* #7136

etc/profile-m-z/opencode.profile
```diff
@@ -0,0 +1,72 @@
...
+include disable-programs.inc
+include disable-x11.inc
+include disable-xdg.inc
+
+whitelist ${HOME}/.cache/opencode
+whitelist ${HOME}/.config/git
+whitelist ${HOME}/.config/opencode
+whitelist ${HOME}/.gitconfig
+whitelist ${HOME}/.local/share/opencode
+whitelist ${HOME}/.local/state/opencode
```
kmk3, line 41: Why `mkdir` new paths in #7136 but not in here?

etc/profile-m-z/opencode.profile
```diff
@@ -0,0 +1,72 @@
...
+
```
kmk3:
```suggestion
```

kmk3 (changes requested):

etc/profile-m-z/opencode.profile
aminvakil: Done in https://github.com/netblue30/firejail/pull/7135/changes/6cb502002a4acf01f749719826e8286760147be0.

etc/profile-m-z/opencode.profile
aminvakil: Because I forgot :) Fixed in https://github.com/netblue30/firejail/pull/7135/changes/b4bd2f2d2f1d95b76ad1627d3dbf2c0bd2cbad64.

aminvakil:
So I've tested these past days with both `opencode` and `pi` using these profiles and wanted to give feedback on my own PRs.

I had created a `pi.local` and `opencode.local` like this:
```
 $ cat ~/.config/firejail/pi.local 
whitelist ${HOME}/my_workspace
```

So I can work in `my_workspace` directory without problems and it was all fine.

The only problem arises when it created a `test.py` or `test.sh` file and tried to execute it, but got permission denied, so model retried with `python test.py` or `bash test.sh`, using a system-prompt like this fixed that issue as well.

```
You are operating inside a strict Firejail sandbox.
1. Avoid executing repository-local scripts directly when an interpreter or runtime can be used instead (for example, use `python main.py` instead of `./main.py`).
2. Assume you do not have sudo or root privileges. Do not attempt system package installation or system-wide configuration changes.
3. Be cautious with long-running daemons or network-listening services; prefer code editing, static analysis, tests, and standard non-privileged CLI builds.
```

kmk3 (approved):

PR #7136, opened 2026-04-13
new profile: pi

aminvakil:
https://github.com/badlogic/pi-mono/tree/main/packages/coding-agent


aminvakil:
@kmk3 Your comment in https://github.com/netblue30/firejail/pull/7135#discussion_r3080506940 has been fixed in https://github.com/netblue30/firejail/pull/7136/commits/4d379cda4638b4a400af76522b45fcd8b7f7e5c3.

aminvakil:
So I've tested these past days with both `opencode` and `pi` using these profiles and wanted to give feedback on my own PRs.

I had created a `pi.local` and `opencode.local` like this:
```
 $ cat ~/.config/firejail/pi.local 
whitelist ${HOME}/my_workspace
```

So I can work in `my_workspace` directory without problems and it was all fine.

The only problem arises when it created a `test.py` or `test.sh` file and tried to execute it, but got permission denied, so model retried with `python test.py` or `bash test.sh`, using a system-prompt like this fixed that issue as well.

```
You are operating inside a strict Firejail sandbox.
1. Avoid executing repository-local scripts directly when an interpreter or runtime can be used instead (for example, use `python main.py` instead of `./main.py`).
2. Assume you do not have sudo or root privileges. Do not attempt system package installation or system-wide configuration changes.
3. Be cautious with long-running daemons or network-listening services; prefer code editing, static analysis, tests, and standard non-privileged CLI builds.
```

kmk3 (approved):

PR #7142, opened 2026-04-18
feature: firecfg: add and use firejail-symlink wrapper

### LINKED ISSUE #7140: firejail intercepts arguments
### Discussed in https://github.com/netblue30/firejail/discussions/7137

<div type='discussions-op-text'>

<sup>Originally posted by **gcb** April 15, 2026</sup>
Sorry if this is something obvious, but seems very strange and i still haven't had time to investigate, so asking for a sanity check.

Here's qpdf used without firejail:

```console
$ /bin/qpdf --version
qpdf version 12.3.2
Run qpdf --copyright to see copyright and license information.
```

And here's firejail stealing the show:

```console
$ qpdf --version
firejail version 0.9.80

Compile time support:
        - AppArmor support is enabled
        - AppImage support is enabled
        - chroot support is enabled
        - Landlock support is enabled
...
```

I don't have anything special for qpdf in firejail, it's all "upstream". archlinux package  0.9.80-1 (well, the version is up there too :)

What is going on?

Only for this qpdf command, firejail gets in the way and mess up invocation to the point I cannot use it. Everything else works fine.
</div>

---

Relates to:

* #6969
* #6972

kmk3:
Add a new `firejail-symlink` wrapper script that sets
`FIREJAIL_SYMLINK=1` before calling firejail and use that script as the
symlink in /usr/local/bin instead of `firejail`.

Check both paths in firecfg for backwards compatibility (especially for
`firecfg --clean`).

Place the script in `$(libexecdir)` (which is not currently used), as
that is a more appropriate location than `$(libdir)` and having a stable
target path for the symlinks makes maintenance and support easier (as
opposed to potentially moving the script path later).

Background: The early firejail init code is brittle due to certain
interdependencies between different parts (such as when checking
environment variables, flags and firejail.config options).

This should make the code more robust by not having to guess when
firejail is called as a symlink (which affects argument parsing).

This fixes a bug added on commit 907916d04 ("modif: check for --version
during early init", 2025-11-23) / PR #6972, in which running
`foo --version` (and with `foo` being a symlink to firejail) results in
firejail parsing `--version`, printing its version and exiting instead
of just forwarding the argument to the program:

    $ qpdf --version
    firejail version 0.9.80
    [...]

Misc: PR #6972 was still marked as a draft when it was merged.

Fixes #7140.

Relates to #6969 #6972.

Reported-by: @gcb
Bisected-by: @rusty-snake

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -225,6 +225,9 @@ endif
 	$(INSTALL) -m 0755 -t $(DESTDIR)$(libdir)/firejail src/profstats/profstats
 	$(INSTALL) -m 0755 -t $(DESTDIR)$(libdir)/firejail src/etc-cleanup/etc-cleanup
 	$(INSTALL) -m 0755 -t $(DESTDIR)$(libdir)/firejail src/fbwrap/fbwrap
+	# other scripts and executables
+	$(INSTALL) -m 0755 -d $(DESTDIR)$(libexecdir)/firejail
+	$(INSTALL) -m 0755 src/firecfg/firejail-symlink.sh $(DESTDIR)$(libexecdir)/firejail/firejail-symlink
 	# plugins w/o read permission (non-dumpable)
 	$(INSTALL) -m 0711 -t $(DESTDIR)$(libdir)/firejail $(SBOX_APPS_NON_DUMPABLE)
 	$(INSTALL) -m 0711 -t $(DESTDIR)$(libdir)/firejail src/fshaper/fshaper.sh
--- a/config.mk.in
+++ b/config.mk.in
@@ -17,6 +17,7 @@ prefix=@prefix@
 exec_prefix=@exec_prefix@
 bindir=@bindir@
 libdir=@libdir@
+libexecdir=@libexecdir@
 datarootdir=@datarootdir@
 docdir=@docdir@
 mandir=@mandir@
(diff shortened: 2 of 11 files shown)


kmk3:
@rusty-snake

Thoughts?

Anything else that would need to be updated (such as maybe firecfg.py)?

Also, it could maybe work to check `argv[0]`, though I'm not sure how it would
work when using firejail as a login shell (or any other corner cases that I'm
not aware of).

Using a separate script a bit more complex, but it should solve the early
symlink detection once and for all (and maybe allow moving more checks
earlier).

In any case, feel free to suggest alternative approaches.

rusty-snake:
How do existing systems behave? Do we need to migrate them or is this change backward compatible?

kmk3:
> How do existing systems behave?

I didn't change/remove the existing code related to symlink detection, so it
should work the same way (other than having an extra way of detecting that
firejail is running from a symlink).

> Do we need to migrate them or is this change backward compatible?

Nothing changes until running `sudo firecfg`, which will make use of
`firejail-symlink`.

`sudo firecfg` already removes any existing symlinks from /usr/local/bin (see
`clean()`) before installing new ones, so it would remove `firejail` symlinks
and install `firejail-symlink` symlinks.

The one compatibility issue that I can think of would be when downgrading
firejail, in which case the user would have to run `sudo firecfg --clean`
before uninstalling, as the `firejail-symlink` file would not be present on
older firejail versions (and so the symlinks would have no effect and `firecfg`
would not remove them / would not replace them with symlinks to `firejail`).

Though note that doing this is already recommended in:

* <https://github.com/netblue30/firejail#uninstalling>

netblue30:
Thanks! I'll do some tests here also.

PR #7143, opened 2026-04-21
profiles: torbrowser-launcher: add no3d

cobratbq:
`no3d` added to `torbrowser-launcher`-profile.

Without GPU acceleration, even video rendering is still possible. However, with GPU acceleration and 3D rendering capabilities, and especially with whole libraries hosted in a Javascript-engine, it is easy to gain and take advantage of direct access. Considering the nature and intentions of tor-browser, it is probably best to default to disabling 3D-acceleration, unless explicitly granted.

### DIFF
--- a/etc/profile-m-z/torbrowser-launcher.profile
+++ b/etc/profile-m-z/torbrowser-launcher.profile
@@ -45,6 +45,7 @@ include whitelist-var-common.inc
 #apparmor
 caps.drop all
 netfilter
+no3d
 nodvd
 nogroups
 noinput


cobratbq:
As an aside: maybe it's better to remove `#restrict-namespaces` from the profile to avoid giving false hints. IIUC, user-namespaces are significant to browser-sandboxing and with `caps.drop all` many risks are already mitigated due to administrative capabilities being dropped prior to running the hosted child-process.

rusty-snake:
> IIUC, user-namespaces are significant to browser-sandboxing

Yes. Though seccomp + broker architecture is more.

>  caps.drop all many risks are already mitigated due to administrative capabilities being dropped prior to running the hosted child-process.

Unrelated to `restrict-namespaces`. Even with an empty bounding set in your namespace, entering a user namespaces grants you all capabilities in this namespace.

cobratbq:
> > IIUC, user-namespaces are significant to browser-sandboxing
> 
> Yes. Though seccomp + broker architecture is more.

Okay, IIUC with unprivileged user-namespaces: within the newly created user-namespace one acquires full administrative privileges. This allows setting other protective measures. However, with all capabilities accessible/available/permissable, there is also a risk that other capabilities cross the namespace-boundary (hence posing risks) and by dropping capabilities and preventing acquiring new privileges, you mitigate a significant amount of the additional surface-area for attacks.

> > caps.drop all many risks are already mitigated due to administrative capabilities being dropped prior to running the hosted child-process.
> 
> Unrelated to `restrict-namespaces`. Even with an empty bounding set in your namespace, entering a user namespaces grants you all capabilities in this namespace.

No, but that's the point. Within the namespace one has by matter of permission access to administrative privileges, but do the capabilities restrictions (e.g. `caps.drop all`) not function as a mask to prevent actually accessing/executing corresponding capabilities, even if administrative permissions are there?

IIUC `capabilities` as a kernel security module. That is, the process acquires ability to create any type of namespace, but is prevented from accessing capabilities that are not deemed relevant by firejail (profile).

Note: I understand that seccomp and bpf allow installing small "interception programs" to catch and interfere/interrupt/process specific syscalls, etc.

> Even with an empty bounding set in your namespace, ...

I think this contradicts my statement. Is there no way to make capability restrictions work witnin the namespace? I mean, that's essentially what one is aiming for to achieve, right?

rusty-snake:
user-namespaces play two roles when it comes to unprivileged sandbox creation.

First they grant you full namespaced capabilities within this namespace no matter what capabilities you had in permitted/effective or bounding set before. Capabilities like namespaced-`CAP_SYS_ADMIN` together with unshared mount namespace allow you to use mount-syscalls to create a sandboxed filesystem view. You can bind mount, mount apifs (`tmpfs`, `procfs`, ...) but not blockfs (`ext4`, ...) and change mount attributes (`ro`, `noexec`, `nosuid`, ...). Also some other stuff like a few sysctls are namespaced.

Then they are required to unshare any other namespace (network, pid, ...). Which isolates access to resources of that kind to the one available in this namespace. It's a low-level thing and what is covered by which namespaces is not that easy.

The second part is a good thing. It's not a sandbox, but can be part of it. The first part is required to effectively limit filesystem access w/o seccomp+broker, Landlock it's fully capable yet. But it also means that every process which can unshare namespaces gets access to a lot more codepaths in the kernel = attack surface.
If you only need the second part but not any namespaced capabilities, there are ways hooks for LSMs to create user namespaces with zero capabilities.

cobratbq:
I get the point you're making. I don't have anything to add atm, with my current understanding.

Actually, maybe, to add: in my understanding, it isn't that `firejail`'s ability to spawn a process in a new user-namespace is the problem. It can set desired capabilities. It is that hosted child-process can create new user-namespaces and `firejail` cannot control if and how the capabilities within those user-namespaces are bounded. (That is, if I ignore all related matters and concerns, and drill down to the core issues for this particular case.)

etc/profile-m-z/torbrowser-launcher.profile
```diff
@@ -45,6 +45,8 @@ include whitelist-var-common.inc
 #apparmor
 caps.drop all
 netfilter
+# To re-enable GPU acceleration, add `ignore no3d` to a local profile.
```
kmk3:
```suggestion
```

This is already explained in `man firejail`.

cobratbq: That is true, but this _does_ emphasize this rather invasive change from previous expectations. Anyone looking at the profile to understand what's going on, would be alerted to this fact. Is that not a desired property?

It could be of assistence to less informed/knowledgeable users.

kmk3: > That is true, but this _does_ emphasize this rather invasive change from
> previous expectations. Anyone looking at the profile to understand what's
> going on, would be alerted to this fact. Is that not a desired property?

Invasive how?

Expectations of what?

Also, the changes between releases are already listed in the RELNOTES / in the
GitHub Release.

If the user finds an issue with a profile after upgrading firejail, that would
be a good place to look at.

> It could be of assistence to less informed/knowledgeable users.

If the user is capable of finding and reading the profile, they can probably
figure out that `no3d` blocks 3D acceleration.

Especially since this is a fairly common option.

cobratbq: I'll remove it. 'Invasive' is probably wrong choice of words. It is unexpected that browser had previously enjoyed GPU acceleration, but now suddenly it does not. It seemed like anyone reading the profile would notice the addition of `no3d` more easily because the comment emphasizes it and explains how to undo.

cobratbq:
@kmk3 please check the other fix. I'm happy to separate it into a new PR if needed.

cobratbq:
On a side-note, I wonder if current firejail-profile is interfering with proper detection of GPU hardware. I get these reports, though I'm not completely sure that I am not causing this with local changes. It doesn't seem to be.

```
Crash Annotation GraphicsCriticalError: |[0][GFX1-]: glxtest: ManageChildProcess failed
 (t=0.639868) [GFX1-]: glxtest: ManageChildProcess failed

Crash Annotation GraphicsCriticalError: |[0][GFX1-]: glxtest: ManageChildProcess failed
 (t=0.639868) |[1][GFX1-]: No GPUs detected via PCI
 (t=0.639868) [GFX1-]: No GPUs detected via PCI
```

It might be connected to these `dmesg` reports:

```
[10309.722074] audit: type=1400 audit(1777313836.069:355): apparmor="DENIED" operation="file_receive" class="file" info="Failed name lookup - disconnected path" error=-13 profile="torbrowser_firefox" name="dev/dri/card0" pid=10894 comm="glxtest" requested_mask="wr" denied_mask="wr" fsuid=1001 ouid=0
[10309.724432] audit: type=1400 audit(1777313836.073:356): apparmor="DENIED" operation="file_receive" class="file" info="Failed name lookup - disconnected path" error=-13 profile="torbrowser_firefox" name="dev/dri/card0" pid=10894 comm="glxtest" requested_mask="wr" denied_mask="wr" fsuid=1001 ouid=0
```

Any chance that `/etc/apparmor.d/firejail-default` or `/etc/apparmor.d/local/firejail-default` is blocking this?

kmk3:
> @kmk3 please check the other fix. I'm happy to separate it into a new PR if
> needed.

I think it would be better as a separate PR, as `no3d` is hardening (and could
make a noticeable difference as you mentioned) and the fix allows more things.

kmk3:
> On a side-note, I wonder if current firejail-profile is interfering with
> proper detection of GPU hardware. I get these reports, though I'm not
> completely sure that I am not causing this with local changes. It doesn't
> seem to be.
> 
> ```
> Crash Annotation GraphicsCriticalError: |[0][GFX1-]: glxtest: ManageChildProcess failed
>  (t=0.639868) [GFX1-]: glxtest: ManageChildProcess failed
> 
> Crash Annotation GraphicsCriticalError: |[0][GFX1-]: glxtest: ManageChildProcess failed
>  (t=0.639868) |[1][GFX1-]: No GPUs detected via PCI
>  (t=0.639868) [GFX1-]: No GPUs detected via PCI
> ```
> 
> It might be connected to these `dmesg` reports:
> 
> ```
> [10309.722074] audit: type=1400 audit(1777313836.069:355): apparmor="DENIED" operation="file_receive" class="file" info="Failed name lookup - disconnected path" error=-13 profile="torbrowser_firefox" name="dev/dri/card0" pid=10894 comm="glxtest" requested_mask="wr" denied_mask="wr" fsuid=1001 ouid=0
> [10309.724432] audit: type=1400 audit(1777313836.073:356): apparmor="DENIED" operation="file_receive" class="file" info="Failed name lookup - disconnected path" error=-13 profile="torbrowser_firefox" name="dev/dri/card0" pid=10894 comm="glxtest" requested_mask="wr" denied_mask="wr" fsuid=1001 ouid=0
> ```
> 
> Any chance that `/etc/apparmor.d/firejail-default` or
> `/etc/apparmor.d/local/firejail-default` is blocking this?

Potentially, yes.

It could also be the upstream apparmor profile blocking it (as it says
`"profile="torbrowser_firefox"`).

Is there a change with `ignore apparmor`?

etc/profile-m-z/torbrowser-launcher.profile
kmk3: > I'll remove it. 'Invasive' is probably wrong choice of words. It is
> unexpected that browser had previously enjoyed GPU acceleration, but now
> suddenly it does not.

It seems unlikely that the tor browser would be used for graphics-intensive
browser applications, so I don't think adding `no3d` makes a functional
difference in the vast majority of cases (other than having less smooth GUI
animations).

If anything I'm surprised that this profile didn't already have `no3d`.

> It seemed like anyone reading the profile would notice the addition of `no3d`
> more easily because the comment emphasizes it and explains how to undo.

There is a balance between information that is helpful and noise.

If the point of the comment is to be a temporary notice, then it follows that
at some point it stops being useful and should be removed, which is a
maintenance burden (which is best saved for other things).

cobratbq:
Created #7152 for fixes. This PR will now only add `no3d`. It is ready.

kmk3 (approved):

Issue #3464, opened 2020-06-11
Atom 1.48.0 breaks with Firejail 0.9.58.2


kmotoko:
**Bug and expected behavior**
- Describe the bug.
Upgrading from `atom` `1.46.0` to `1.48.0` breaks starting `atom` under `firejail`.
- What did you expect to happen?
`atom` to start normally.

**No profile or disabling firejail**
- What changed calling `firejail --noprofile PROGRAM` in a shell?
`atom` starts-up normally.
- What changed calling the program *by path*=without firejail (check `whereis PROGRAM`, `firejail --list`, `stat $programpath`)?
`atom` starts-up normally.

**Reproduce**
1. Run in bash `firejail atom`
2. See error:
```
Reading profile /etc/firejail/atom.profile
Reading profile /etc/firejail/atom.local
Reading profile /etc/firejail/globals.local
Reading profile /etc/firejail/disable-common.inc
Reading profile /etc/firejail/disable-passwdmgr.inc
Reading profile /etc/firejail/disable-programs.inc
Warning: networking feature is disabled in Firejail configuration file
Parent pid 13545, child pid 13546
Child process initialized in 47.07 ms
/usr/bin/atom: line 190:    14 Trace/breakpoint trap   nohup "$ATOM_PATH" --executed-from="$(pwd)" --pid=$$ "$@" > "$ATOM_HOME/nohup.out" 2>&1
[14:0611/105406.114486:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /usr/share/atom/chrome-sandbox is owned by root and has mode 4755.

Parent is shutting down, bye...

```

**Environment**
 - Linux distribution and version (ie output of `lsb_release -a`)
```
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 10 (buster)
Release:	10
Codename:	buster
```
 - Firejail version (output of `firejail --version`) exclusive or used git commit (`git rev-parse HEAD`)
```
firejail version 0.9.58.2

Compile time support:
	- AppArmor support is enabled
	- AppImage support is enabled
	- chroot support is enabled
	- file and directory whitelisting support is enabled
	- file transfer support is enabled
	- networking support is enabled
	- overlayfs support is enabled
	- private-home support is enabled
	- seccomp-bpf support is enabled
	- user namespace support is enabled
[...]

PR #7148, opened 2026-04-25
build: remove etc-fixes directory

kmk3:
This directory contains profiles from older firejail versions with
backported bugfixes.

All of the firejail versions in question are many years old (and
unsupported by us), with the most recent version (0.9.60) being from
2019 and the last backport-related change in etc-fixes appearing to be
from 2020:

* 55906959a ("More fixes for #3464", 2020-06-13)

Assuming that they are not being updated/maintained (/used) anyway,
these profiles can add noise when using `git grep` (depending on the
search pattern used) and when trying to update many profiles at once in
general.

Future backports can be done using git branches (such as using one
branch per release), which may be less convenient for updating many
versions at once but would avoid polluting the main tree (and would be a
more common and arguably more reliable method).

So just remove this directory.

Note: This also removes a README file and a binary file:

* etc-fixes/seccomp-join-bug/README
* etc-fixes/seccomp-join-bug/eecf35c-backports.zip


netblue30:
Absolutely, we need to get rid of them!

PR #7151, opened 2026-04-27
feature: allow subpaths in xdg macros

### LINKED ISSUE #2359: Support subpaths in macros (like `${PICTURES}/Screenshots`)

kmk3:
Currently it is not possible to use XDG-related macros (such as
`${DOCUMENTS}`) with subpaths (such as `${DOCUMENTS}/foo`) and so
profiles just use `${HOME}` with a hardcoded path using the English
directory name and the subpath (such as `${HOME}/Documents/foo`).

Allow using subpaths after XDG macros, so that they automatically use
the auto-detected XDG path, just as when currently using the XDG macros
without subpaths.

Before:

    ${HOME}/Documents/foo

After:

    ${DOCUMENTS}/foo

This is a follow-up to #7147.

Closes #2359.

Relates to #4229.

### DIFF
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -542,10 +542,10 @@ uint32_t arp_assign(const char *dev, Bridge *br);
 
 // macros.c
 char *expand_macros(const char *path);
-char *resolve_macro(const char *name);
+char *resolve_macro(const char *path);
 void invalid_filename(const char *fname, int globbing);
-int is_macro(const char *name);
-int macro_id(const char *name);
+int is_macro(const char *path);
+int macro_id(const char *path);
 
 
 // util.c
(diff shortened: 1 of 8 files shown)


kmk3:
@rusty-snake

Any thoughts?

I plan to update the profiles in a new PR.

Issue #7132, opened 2026-04-10
can't disable blacklist of ~/.config/firejail


carlsampurna:
### Description

firejail calls disable_config() if not running as root, which blacklists ~/.config/firejail.  There is no way to tell firejail not to do this, which means when I sandbox a program to back up my home directory, my firejail configs don't get backed up.

### Steps to Reproduce

run from your home directory:

$ firejail --noprofile ls -ld .config/firejail
firejail version 0.9.80

Parent pid 2765731, child pid 2765732
Base filesystem installed in 0.05 ms
Child process initialized in 18.53 ms
dr-------- 2 root root 40 Mar 30 09:24 .config/firejail

Parent is shutting down, bye...

### Expected behavior

I did not expect the firejail source to contain hardcoded blacklists that cannot be disabled, though from the many calls to disable_file() in fs.c this is only one of many, and as far as I can tell this is undocumented behavior.  

### Environment

Arch
Linux 6.19.10-zen1-1-zen x86_64
firejail 0.9.80-1

### Checklist

- [x] I am using firejail [0.9.80 or later](https://github.com/netblue30/firejail/tree/master/SECURITY.md)
- [x] I am using the full program path (e.g. `firejail /usr/bin/vlc` instead of `firejail vlc`; see `https://github.com/netblue30/firejail/issues/2877`)
- [x] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [x] I have performed a short search for similar issues (to avoid opening a duplicate).

### Log

<details>
<summary>Output of <code>LC_ALL=C firejail --debug /path/to/program</code></summary>
<p>

```
firejail version 0.9.80

DISPLAY=:0 parsed as 0
Parent pid 2782310, child pid 2782311
Looking for kernel processes
Found kthreadd process, we are not running in a sandbox
pid=2782310: locking /run/firejail/firejail-run.lock ...
pid=2782310: locked /run/firejail/firejail-run.lock
pid=2782310: unlocking /run/firejail/firejail-run.lock ...
pid=2782310: unlocked /run/firejail/firejail-run.lock
Building quoted command line: 'ls' '-ld' '.config/firejail' 
Command name #ls#
pid=2782310: locking /run/firejail/firejail-run.lock ...
pid=2782310: locked /run/firejail/firejail-run.lock
[...]

PR #7154, opened 2026-04-29
security: create `~/.config/firejail` to prevent sandbox escape

Changaco:
A quick-and-dirty fix for https://github.com/netblue30/firejail/issues/7132#issuecomment-4239282977


kmk3 (approved):
Good idea.

Issue #5116, opened 2022-04-23
Filter netlink families with seccomp


rusty-snake:
### Is your feature request related to a problem? Please describe.

`AF_NETLINK` exposes a (huge) attack surface for the kernel. See https://github.com/netblue30/firejail/issues/4013#issuecomment-786780504 and #4020 for previous discussions.

cc @kris7t @glitsj16 

### Describe the solution you'd like

The socket syscall looks like `int socket(int domain, int type, int protocol);`. We already filter the first argument (`domain`) with `protocol unix,inet,inet6,netlink`. To minimize the attack surface of `AF_NETLINK` we should filter the third/last argument (`protocol`) if the first is `AF_NETLINK` (`netlink_socket = socket(AF_NETLINK, socket_type, netlink_family);`).

```
protocol unix,inet,inet6,netlink
protocol.netlink netfilter,audit,generic,route,...
```

### Describe alternatives you've considered

N/A

### Additional context

N/A

glitsj16:
Would make a nice enhancement. FWIW I always try to run apps without netlink via local overrides. Only in very rare cases it is actually needed. So, in other words, besides netlink filtering we could drop netlink from protocol for known 'good' applications too IMO. More of a note for future reference when doing PR's.

PR #7165, opened 2026-05-13
profile: blink-common-hardened: drop netlink & add mdwe comment

cobratbq:
I would propose the following change, considering that this is the _hardened_-profile.

1.  Remove `netlink` being an unnecessary protocol which, in my understanding, provides access into kernel mechanisms.
2.  Provide a comment hinting at a possibility of `memory-deny-write-execute` if Chromium is run with certain command-line argument.

### DIFF
--- a/etc/profile-a-l/blink-common-hardened.inc.profile
+++ b/etc/profile-a-l/blink-common-hardened.inc.profile
@@ -5,7 +5,7 @@ include blink-common-hardened.inc.local
 caps.drop all
 nonewprivs
 #noroot breaks saving files
-protocol unix,inet,inet6,netlink
+protocol unix,inet,inet6
 seccomp !chroot
 
 #restrict-namespaces
--- a/etc/profile-a-l/chromium-common-hardened.inc.profile
+++ b/etc/profile-a-l/chromium-common-hardened.inc.profile
@@ -6,5 +6,8 @@ include chromium-common-hardened.inc.local
 # added by caller profile
 #include globals.local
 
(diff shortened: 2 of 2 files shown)


rusty-snake:
1. Yes, I already opened #5116 to filter them. It's basically only `NETLINK_GENERIC` and `NETLINK_ROUTE` that are of use for user applications. `socket` (and therefore `io_uring_*`) are probably the biggest vectors into the kernel, see CopyFail for example.<br>
  However we should give users a hint why WebRTC and other things do not work.
2. Not sure. There could be more cases.

cobratbq:
> 1. Yes, I already opened [Filter netlink families with seccomp #5116](https://github.com/netblue30/firejail/issues/5116) to filter them. It's basically only `NETLINK_GENERIC` and `NETLINK_ROUTE` that are of use for user applications. `socket` (and therefore `io_uring_*`) are probably the biggest vectors into the kernel, see CopyFail for example.

True. Yeah, CopyFail is indeed one of the reasons I checked. Your solution is better. In part, I noticed torbrowser-launcher does this, so I figured it might be worth it for hardened version.

> However we should give users a hint why WebRTC and other things do not work.

Yeah, that's probably not so easy. Maybe allow for a terminal report/comment (with appropriate distinction/emphasis), right before executing guest application. Very relevant and I have no idea how to best solve that.

> 2. Not sure. There could be more cases.

My intention is to make super-users aware that the possibility exists to enforce W^X under the right circumstances. Given the somewhat notorious nature of web-browsers, it might be a plus.

It is probably better to close this and go with your approach.

PR #7166, opened 2026-05-14
new profile: tldr

haplo:
This is a profile for tldr official Python CLI:

* https://github.com/tldr-pages/tldr
* https://github.com/tldr-pages/tldr-python-client

It's working on my machine (Arch Linux) with stable Firejail (0.9.80). I tried to make it as restrictive as possible, as it's only downloading files and putting them in `~/.cache/tldr`.

### DIFF
--- a/etc/inc/disable-programs.inc
+++ b/etc/inc/disable-programs.inc
@@ -239,6 +239,7 @@ blacklist ${HOME}/.cache/systemsettings
 blacklist ${HOME}/.cache/telepathy
 blacklist ${HOME}/.cache/thunderbird
 blacklist ${HOME}/.cache/tiny-rdm
+blacklist ${HOME}/.cache/tldr
 blacklist ${HOME}/.cache/torbrowser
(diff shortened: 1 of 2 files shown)


kmk3 (changes requested):
> ```
> include allow-python3.inc
> ```

Add comment from template.

> ```
> # Commands that reduce access to resources.
> ```

Remove this.

> ```
> private-etc alternatives,ca-certificates,crypto-policies,ld.so.cache,ld.so.conf,ld.so.conf.d,ld.so.preload,locale,locale.alias,locale.conf,localtime,mime.types,protocols,resolv.conf,ssl,xdg
> ```

Use `private-etc` groups (see `etc_groups.h` and other profiles).

haplo:
Thank you for the review @kmk3, I have implemented the changes and tested that the `private-etc` groups work on my machine.

kmk3 (approved):
Looks good, thanks.

PR #7172, opened 2026-05-22
profiles: templates: improve formatting & clarify rules

kmk3:
Use `## ` for comments that should be ignored / not copied to profiles
and `# ` for comments that should.

This should clarify which lines are which when creating new profiles.


kmk3:
Cc: @rusty-snake

netblue30:
Merged, thanks!

PR #7184, opened 2026-06-14
bugfix: use `O_NOFOLLOW` in sbox helper exec

hyder365:
When executing helper programs with root privileges via `sbox_run()` or `sbox_run_root()`, the `sbox.c` helper opens the executable path with `open(arg[0], O_PATH | O_CLOEXEC)` without `O_NOFOLLOW`. While ownership (uid=0, gid=0) and writability are verified via `fstat()` on the opened fd, a TOCTOU window exists between `open()` and `fstat()`. If `arg[0]` is a symlink that an attacker can swap, the `open()` could follow the symlink to a different file before `fstat()` runs.

Impact: The check after `fstat()` mitigates most risk (only root-owned, non-world-writable files can execute), but an attacker with root-owned file creation capabilities could potentially use this to redirect execution. Adding `O_NOFOLLOW` closes the window and follows the principle of least privilege.

Fix: Added `O_NOFOLLOW` to the `open()` flags so symlinks are never followed when opening helper executables.

### DIFF
--- a/src/firejail/sbox.c
+++ b/src/firejail/sbox.c
@@ -241,7 +241,7 @@ static int __attribute__((noreturn)) sbox_do_exec_v(unsigned filtermask, char * 
 	else assert(0);
 
 	if (arg[0]) { // get rid of scan-build warning
-		int fd = open(arg[0], O_PATH | O_CLOEXEC);
+		int fd = open(arg[0], O_PATH | O_NOFOLLOW | O_CLOEXEC);
 		if (fd == -1) {
 			if (errno == ENOENT) {
 				fprintf(stderr, "Error: %s does not exist\n", arg[0]);


kmk3 (changes requested):
> While ownership (uid=0, gid=0) and writability are verified via `fstat()` on
> the opened fd, a TOCTOU window exists between `open()` and `fstat()`. If
> `arg[0]` is a symlink that an attacker can swap, the `open()` could follow
> the symlink to a different file before `fstat()` runs.

Wrong, `open()` follows symlinks and returns a file descriptor to the real
file.

The same file descriptor is used in `fstat()` and later calls, so there is no
TOCTOU issue in that regard.

Also, somehow this seems to break related tests:

From <https://github.com/netblue30/firejail/actions/runs/27515182936/job/82171395876?pr=7184>:

```
TESTING: network IP (net_ip.exp)
spawn /bin/bash
firejail --net=br0  cat /proc/1/net/dev
runner@runnervmqtt2i:~/work/firejail/firejail/test/network$ 
<st/network$ firejail --net=br0  cat /proc/1/net/dev
Reading profile /etc/firejail/landlock-common.inc

firejail version 0.9.81

Parent pid 4306, child pid 4307
Error: /sbin/iptables-restore is world writable, refusing to execute
Error: failed to run /sbin/iptables-restore: exit status 1, exiting...
Error: proc 4306 cannot sync with peer: unexpected EOF
Peer 4307 unexpectedly exited with status 1
runner@runnervmqtt2i:~/work/firejail/firejail/test/network$ TESTING ERROR 0
```

PR #7204, opened 2026-07-15
profiles: firefox-common: add seccomp.block-secondary

StellariusOrion592:
Add seccomp.block-secondary to firefox-common profile.

### DIFF
--- a/etc/profile-a-l/firefox-common.profile
+++ b/etc/profile-a-l/firefox-common.profile
@@ -82,6 +82,7 @@ protocol unix,inet,inet6,netlink
 # Note: The seccomp line below still permits the chroot syscall; see
 # https://github.com/netblue30/firejail/issues/2506 for possible workarounds.
 seccomp !chroot
+seccomp.block-secondary
 
 disable-mnt
 ?BROWSER_DISABLE_U2F: private-dev


etc/profile-a-l/firefox-common.profile
```diff
@@ -62,8 +62,6 @@ include whitelist-usr-share-common.inc
 include whitelist-var-common.inc
 
 apparmor
-# Fixme!
-apparmor-replace
```
rusty-snake, line 66: This is not a fix

StellariusOrion592: The apparmor-replace is an old deprecated command that can result in less security.
I noticed Firefox based browsers and software derived from Firefox such as Thunderbird would display the following warning:

“Warning: Replacing profile instead of stacking it. It is a legacy behavior that can result in relaxation of the protection. It is here as a temporary measure to unbreak the software that has been broken by switching to the stacking behavior.”

When I removed apparmor-replace the warning was no longer displayed and
Thunderbird, Firefox, and other Firefox based browsers such as Waterfox and Zen-browser still continued to work just fine.

Could please explain further?

rusty-snake: > When I removed apparmor-replace the warning was no longer displayed

`2>&1 >/dev/null` would do the same

> Thunderbird, Firefox, and other Firefox based browsers such as Waterfox and Zen-browser still continued to work just fine.

Tested on which system?

> Could please explain further?

900db66

StellariusOrion592: Thank you for explaining apparmor-replace. I was simply concerned about security because the warning said that it will result in the “relaxation of the protection”. I understand it was used to unbreak certain software but that was a long time ago. 

When I ran Firefox from a terminal like for example: firejail /usr/bin/firefox that is when I noticed the warning about apparmor-replace. I removed it from the firefox-common profile and tested Firefox to see if it still works. Everything worked just fine without any problems and that resolved the warning issue. That’s why I removed it to further improve security and fix that old problem.

On Arch-Linux I tested the following:
Firefox, LibreWolf, Waterfox, Zen Browser, Tor Browser, and Thunderbird.

And I also tested on Debian 13 (stable):
Firefox, LibreWolf, Tor Browser, Thunderbird, and Pale Moon.

All were tested on the latest stable version of Firejail 0.9.80 with great success.

kmk3 (changes requested):
From <https://github.com/netblue30/firejail/pull/7204/changes/887de324a9b63d44ebf9ca776793d4ba452bed75>:

> fix: remove deprecated apparmor-replace directive
> 
> - remove apparmor-replace command that can relax AppArmor protection
> - resolve firejail warning about legacy profile replacement behavior
> - align with stacking behavior for improved security posture

The explanation makes no sense in the context of the change and looks like AI
slop.

Removing `apparmor-replace` makes it less hardened, not more.

A proper fix would be to replace it with `apparmor` and make sure that it works
in the case that required changing it to `apparmor-replace`.

Adding `seccomp.block-secondary` looks good, so please undo the apparmor change
for this PR.

You can create a separate PR for changing `apparmor-replace` into `apparmor` if
you want to discuss it further.

See also:

* #5462
* #5475
* #7078

StellariusOrion592:
Thank you kmk3 very much for your feedback.
I made the requested changes.
The warning message in Firejail and AppArmor’s profile stacking behavior is
what got me confused. I thought apparmor-replace was deprecated
since I couldn’t find it in the Firejail manual.
But you are correct apparmor-replace is important and it does harden security.
Please feel free to request any more changes and
I will gladly do them.

kmk3 (approved):

PR #7158, opened 2026-05-05
profiles: add llm-agent-common.inc

aminvakil:
Regarding @cameronj86 comment in https://github.com/netblue30/firejail/issues/1139#issuecomment-3721450281 and @kmk3 comment in https://github.com/netblue30/firejail/pull/6936#issuecomment-4359189937.

Also I cannot remember why I hadn't added `env NO_BROWSER=true` in `pi.profile`.

Most probably because I have misremembered :) Please tell me if I should add it to `llm-agent-common.inc` as well, or it should be in another PR.

PR #7212, opened 2026-07-22
profiles: llm-agent-common: disable private-cache

aminvakil:
This is a follow-up on my previous multiple PRs for adding profiles of coding agents.

Working with them in the past couple of months, I have allowed some `.cache` access in my `.local` file like this:
```
whitelist ${HOME}/.cache/go-build
whitelist ${HOME}/.cache/huggingface
```

I think it's better for coding agents to access `.cache` by default, therefore this PR.

Also note to myself in future, keep in mind that `disable-programs.inc` already disables accessing many sensitive `.cache` directories like browsers' caches, so with removing `private-cache`, firejail still protects browsers' caches.

Relates to #7158.

### DIFF
--- a/etc/profile-a-l/llm-agent-common.profile
+++ b/etc/profile-a-l/llm-agent-common.profile
@@ -49,7 +49,6 @@ seccomp
 seccomp.block-secondary
 
 disable-mnt
-private-cache
 private-dev
 private-etc @network,@tls-ca
 private-tmp


kmk3 (approved):

aminvakil:
@kmk3 Thanks for fixing up the PR title and message, I didn't remember the name convention.

Sorry for your trouble and thanks again.

PR #7224, opened 2026-08-09
bugfix: fbwrap: wait for target and propagate exit code

### LINKED ISSUE #7081: fbwrap sleeps instead of waiting for child to exit, does not preserve return code
### Description

fbwrap sleeps instead of waiting for the child to exit. This breaks programs that expect to wait for the sandboxed child. All xfce4 applications appear to be affected as is geeqie.

### Steps to Reproduce

Launch any xfce4 application using firejail. There will be an extra two second delay before the program starts.
```
time firejail --noprofile /usr/bin/xfce4-terminal --disable-server -e /bin/true
time firejail --noprofile --allow-bwrap /usr/bin/xfce4-terminal --disable-server -e /bin/true
```

### Expected behavior

Both commands should take close to the same amount of time to complete.

### Actual behavior

The version using firejail's fbwrap takes 4 seconds longer than the one using native bwrap.

With or without a profile makes no difference for any affected application that has one.

### Additional context

Some programs wait for the wrapped program to exit and may check the exit code. xfce4 applications do this on startup causing a delay before their window opens.

From the `bwrap` man page:
```
EXIT STATUS
       The bwrap command returns the exit status of the initial application
       process (pid 2 in the sandbox).
```

The problem is in the following code in `src/fbwrap/main.c`:
```
	// wait child to finish
	//int status;
	//waitpid(child, &status, 0);

	// don't bother waiting
	sleep(2);

	return 0;
```
It should be changed to:
```
	// wait for child to finish
	int status;
	waitpid(child, &status, 0);

	return status;
```

### Environment

Linux 6.18.12+deb14-amd64 x86_64
Debian forky/sid
xfce4-terminal 1.1.5 (Xfce 4.20) (For demonstration. All Xfce 4.20 programs are affected.)
firejail version 0.9.78

### Checklist

- [x] I am using firejail [0.9.78 or later](https://github.com/netblue30/firejail/tree/master/SECURITY.md)
- [x] I am using the full program path (e.g. `firejail /usr/bin/vlc` instead of `firejail vlc`; see `https://github.com/netblue30/firejail/issues/2877`)
- [ ] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
  - [ ] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ ] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

<details>
<summary>Output of <code>time firejail --noprofile /usr/bin/xfce4-terminal --disable-server -e /bin/true</code></summary>
<p>

```
firejail version 0.9.78

Parent pid 183678, child pid 183679
Base filesystem installed in 0.04 ms
Child process initialized in 6.62 ms

real	0m4.117s
user	0m0.074s
sys	0m0.044s
```

</p>
</details>

<details>
<summary>Output of <code>time firejail --noprofile --allow-bwrap /usr/bin/xfce4-terminal --disable-server -e /bin/true</code></summary>
<p>

```
firejail version 0.9.78

Parent pid 183707, child pid 183708
Base filesystem installed in 0.05 ms
Child process initialized in 6.59 ms

real	0m0.140s
user	0m0.080s
sys	0m0.041s
```

</p>
</details>

Relates to:

* #7009

ousamabenyounes:
## Problem

`fbwrap` (the bwrap replacement used inside the sandbox with `--allow-bwrap`)
forked the target program and then, instead of waiting for it:

```c
// don't bother waiting
sleep(2);
return 0;
```

Consequences:
- a fixed **2-second delay** on every invocation (the extra delay reported for
  xfce4 apps and geeqie);
- the child's **exit status is discarded** — `fbwrap` always returned `0`;
- any program still running after 2s is **killed** via `PR_SET_PDEATHSIG` when
  `fbwrap` returns.

## Fix

Wait for the child with `waitpid()` and return its exit status (`128+signo`
when it is killed by a signal), matching the real `bwrap` it stands in for.
Also `_exit(127)` when `execvp()` fails instead of returning `0`.

## Test verification (RED → GREEN)

New test `test/utils/fbwrap.exp` (wired into `test/utils/utils.sh`) asserts the
exit code is propagated, for a clean exit, a non-zero exit, and signal death.

RED (unmodified `master`, which does `sleep(2); return 0`):

```
$ fbwrap /usr/bin/false ; echo fbwrap-rc-$?
fbwrap-rc-0            # wrong: should be 1
# timing: ~2.003s      # fixed sleep
# fbwrap.exp -> "TESTING ERROR 1"
```

GREEN (with this patch):

```
$ fbwrap /usr/bin/false ; echo fbwrap-rc-$?
fbwrap-rc-1                              # correct, no fixed delay (~0.003s)
$ fbwrap /usr/bin/true ; echo fbwrap-rc-$?
fbwrap-rc-0
$ fbwrap /bin/sh -c 'kill -TERM $$' ; echo fbwrap-rc-$?
fbwrap-rc-143                            # 128 + SIGTERM(15)
# fbwrap.exp -> "all done"
```

Full build is clean (`-Wall -Wextra`, no new warnings).

Fixes #7081

### DIFF
--- a/src/fbwrap/main.c
+++ b/src/fbwrap/main.c
@@ -154,9 +154,8 @@ int main(int argc, char **argv) {
 		_exit(127); // 127: standard "command not found/not executable" status
 	}
 
-	// wait for the target and propagate its exit status: callers (and the real
-	// bwrap that fbwrap stands in for) expect fbwrap to block until the child
-	// exits and to return the child's return code, not a fixed delay
+	// Note: Callers expect bwrap to block until the child exits and to
+	// return the child's exit status (see #7081).
 	int status;
 	while (waitpid(child, &status, 0) == -1) {
 		if (errno == EINTR)
--- a/src/fbwrap/main.c
+++ b/src/fbwrap/main.c
@@ -148,8 +148,9 @@ int main(int argc, char **argv) {
 		// kill the target if the parent dies
 		prctl(PR_SET_PDEATHSIG, SIGKILL, 0, 0, 0);
 		execvp(arglist[0], arglist);
+
 		// execvp only returns on failure
-		fprintf(stderr, "Error: fbwrap cannot execute %s\n", arglist[0]);
+		fprintf(stderr, "Error: fbwrap cannot execute %s: %s\n", arglist[0], strerror(errno));
 		_exit(127); // 127: standard "command not found/not executable" status
 	}
 
@@ -157,12 +158,11 @@ int main(int argc, char **argv) {
 	// bwrap that fbwrap stands in for) expect fbwrap to block until the child
 	// exits and to return the child's return code, not a fixed delay
 	int status;
-	pid_t rv;
-	do {
-		rv = waitpid(child, &status, 0);
-	} while (rv == -1 && errno == EINTR);
-	if (rv == -1) {
-		fprintf(stderr, "Error: fbwrap cannot wait for the target program\n");
+	while (waitpid(child, &status, 0) == -1) {
+		if (errno == EINTR)
+			continue;
+
+		fprintf(stderr, "Error: fbwrap cannot wait for %s: %s\n", arglist[0], strerror(errno));
 		exit(1);
 	}
 	if (WIFEXITED(status))
--- a/src/fbwrap/main.c
+++ b/src/fbwrap/main.c
@@ -18,6 +18,7 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 #define _GNU_SOURCE
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
@@ -156,7 +157,11 @@ int main(int argc, char **argv) {
 	// bwrap that fbwrap stands in for) expect fbwrap to block until the child
 	// exits and to return the child's return code, not a fixed delay
 	int status;
-	if (waitpid(child, &status, 0) == -1) {
+	pid_t rv;
+	do {
+		rv = waitpid(child, &status, 0);
+	} while (rv == -1 && errno == EINTR);
+	if (rv == -1) {
 		fprintf(stderr, "Error: fbwrap cannot wait for the target program\n");
 		exit(1);
 	}
--- a/src/fbwrap/main.c
+++ b/src/fbwrap/main.c
@@ -25,6 +25,7 @@
 #include <sys/stat.h>
 //#include <linux/prctl.h>
 #include <sys/prctl.h>
+#include <sys/wait.h>
 #include <signal.h>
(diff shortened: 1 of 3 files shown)


monerostar:
Ubuntu 24.04 on a local 5800X box. Built old fbwrap (pre-PR sleep path) and this PR tip, ran both outside the full firejail install.

| case | old rc / time | this PR rc / time |
|---|---|---|
| `/usr/bin/false` | 0 / ~2.02s | 1 / ~0.004s |
| `/usr/bin/true` | 0 / ~2.02s | 0 / ~0.001s |
| `/bin/sh -c 'kill -TERM $$'` | 0 / ~2.02s | 143 / ~0.004s |

Matches the expect cases in the PR. Fix looks right: waitpid + exit status, and `_exit(127)` on execvp failure is better than returning 0.

One small nit, not a blocker: `waitpid(child, &status, 0)` has no EINTR loop. If waitpid returns -1 for EINTR, fbwrap exits 1 while `PR_SET_PDEATHSIG, SIGKILL` can still kill a live child. A tight retry on EINTR would make that path safer:

```c
while (waitpid(child, &status, 0) == -1) {
    if (errno != EINTR) {
        fprintf(stderr, "Error: fbwrap cannot wait for the target program\n");
        exit(1);
    }
}
```

Happy to re-run on another tip if useful.

ousamabenyounes:
Addressed the interrupted waitpid() case in 001a30d: fbwrap now retries on EINTR, with a regression test covering the failure path. Validation: reproduced exit 1 before the fix, then exit 0 after it; the fatal-warning build and changed-line coverage gate pass.

kmk3 (changes requested):
Remove changes from latest commit (other than in main.c):

https://github.com/netblue30/firejail/commit/001a30dcdf168c32be1504d6077831dfc8a90f65

ousamabenyounes:
Done in `2335889` — stripped everything but `src/fbwrap/main.c` from the latest commit, as requested. It now carries only the EINTR retry loop:

```c
pid_t rv;
do {
    rv = waitpid(child, &status, 0);
} while (rv == -1 && errno == EINTR);
if (rv == -1) { ... }
```

Removed the LD_PRELOAD regression helper (`test/utils/fbwrap-eintr.c`) and its `test/utils/fbwrap.exp` / `test/utils/utils.sh` wiring. Force-pushed (`001a30d` → `2335889`).

Rebuilt `fbwrap` clean with `-Wall -Wextra`; exit-code propagation still holds: `true` → 0, `false` → 1, `sh -c 'kill -TERM $$'` → 143.

kmk3 (approved):

PR #7235, opened 2026-08-21
profiles: keepassxc: add ~/.local/state/keepassxc

0xIO32:
Keepassxc now uses XDG_STATE_HOME instead of XDG_CACHE_HOME for its application state.
See https://github.com/keepassxreboot/keepassxc/commit/509e218676cebc30c72f15234d900528bc5e06fe


kmk3 (commented):
Also add to disable-programs.inc.

kmk3 (approved):

PR #7237, opened 2026-08-27
profiles: mpv: allow dbus mpris

### LINKED ISSUE #7134: mpv: program does not open
### Description

When using the firejail profile with mpv, mpv will crash 

### Steps to Reproduce

with mpv installed, run sudo firecfg

### Expected behavior
mpv should open or open the video clicked on

### Actual behavior

mpv will just not open, it outright crashes.
Often it still is active when running firejail --list

works as expected without a profile

### Additional context

N/A

### Environment

kernel - 6.19.11-1-cachyos
Cachyos
MPV 1:0.41.0-3
firejail 0.9.80

### Checklist

- [x] I am using firejail [0.9.80 or later](https://github.com/netblue30/firejail/tree/master/SECURITY.md)
- [x ] I am using the full program path (e.g. `firejail /usr/bin/vlc` instead of `firejail vlc`; see `https://github.com/netblue30/firejail/issues/2877`)
- [x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [x] The profile (and redirect profile if exists) hasn't already been fixed [upstream](https://github.com/netblue30/firejail/tree/master/etc).
  - [x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [x] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

### Log

Cannot create an output.
Runing

firejail --profile=/etc/firejail/mpv.profile /usr/bin/mpv --player-operation-mode=pseudo-gui -- %U
it just crashes outright with no output

tomhense:
When using mpv-mpris without this fix, mpv will fail to open (unless --no-config is specified).

Fixes #7134.

### DIFF
--- a/etc/profile-m-z/mpv.profile
+++ b/etc/profile-m-z/mpv.profile
@@ -92,12 +92,8 @@ private-bin env,mpv,python*,waf,youtube-dl,yt-dlp
 #private-cache
 private-dev
 
-# If you use mpv-mpris you should add this to your mpv.local file
-#ignore dbus-user none
-#dbus-user filter
-#dbus-user.own org.mpris.MediaPlayer2.mpv
-
-dbus-user none
+dbus-user filter
+dbus-user.own org.mpris.MediaPlayer2.mpv
 dbus-system none
 
 restrict-namespaces
--- a/etc/profile-m-z/mpv.profile
+++ b/etc/profile-m-z/mpv.profile
@@ -92,6 +92,11 @@ private-bin env,mpv,python*,waf,youtube-dl,yt-dlp
 #private-cache
 private-dev
 
+# If you use mpv-mpris you should add this to your mpv.local file
+#ignore dbus-user none
+#dbus-user filter
+#dbus-user.own org.mpris.MediaPlayer2.mpv
+
 dbus-user none
 dbus-system none
 


etc/profile-m-z/mpv.profile
```diff
@@ -92,6 +92,11 @@ private-bin env,mpv,python*,waf,youtube-dl,yt-dlp
 #private-cache
 private-dev
 
+# If you use mpv-mpris you should add this to your mpv.local file
+#ignore dbus-user none
+#dbus-user filter
+#dbus-user.own org.mpris.MediaPlayer2.mpv
+
 dbus-user none
```
kmk3:
```suggestion
dbus-user filter
dbus-user.own org.mpris.MediaPlayer2.mpv
```

mpris seems to be common enough.

kmk3 (approved):

tomhense:
The mpv-mpris update last week seems to have broken things again. Version [1.3](https://github.com/hoyon/mpv-mpris/releases/tag/1.3) now includes a unique dbus identifier name
For example: `org.mpris.MediaPlayer2.mpv.instance-DdfbdgCJ`

mpv also shows a warning message during startup like this:
```
(process:64): GLib-GIO-WARNING **: 22:52:35.676: Error releasing name org.mpris.MediaPlayer2.mpv.instance-DdfbdgCJ: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: org.freedesktop.DBus.Error.ServiceUnknown
```

Unlike the situation before mpv now works fine but refuses to quit (unless one sends a SIGKILL). Also the mpris functionality does not work.

Is it possible to fix the dbus filter rule? I tried but I couldn't figure out how.

If it is not possible to fix the filter rule it would unfortunately probably be a good idea to undo the merged filter change.
The latest version now works when the user dbus isn't available (they fixed the previous [issue](https://github.com/hoyon/mpv-mpris/issues/144)).

PR #7238, opened 2026-08-30
build: test: sync `DISTFILES_TEST` with `TESTS`

kmk3:
Standardize and deduplicate the test items by using the same test paths
in `make dist` as the test targets in `$(TESTS)`.

This should fix the errors caused by missing test files when running
`make test` from the tarball.

Added tests:

* apparmor
* appimage
* chroot
* firecfg
* private-etc
* seccompextra

Removed tests:

* private-lib

This is a follow-up to #7223.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -327,21 +327,6 @@ DISTFILES = \
 	src \
 	testcompile.sh
 
-DISTFILES_TEST = \
-	test/Makefile \
-	test/apps \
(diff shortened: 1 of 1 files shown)


kmk3:
@reinerh

I remember the errors on Debian CI being related to tests and missing files.

Any idea if this helps?

PR #7242, opened 2026-09-05
bugfix: sandbox: search the default PATH when the environment has none

### LINKED ISSUE #5939: inconsistent behavior when PATH in not set
### Description

When PATH is not set, the behavior depends on whether the profile contains a ${PATH} variable
_Describe the bug_

### Steps to Reproduce

```
# create an empty profile
$ rm -f firejail.profile && touch firejail.profile 
# unset PATH
$ unset PATH
$ /usr/bin/firejail --profile=./firejail.profile ls
# shows: Error: no suitable ls executable found

# Now, writes a directive that references PATH in the profile
$ echo 'blacklist ${PATH}/systemctl' > firejail.profile
$ /usr/bin/firejail --profile=./firejail.profile ls
# now, the ls command is executed properly
```

### Expected behavior
I expect the behavior to be coherent. Either a default PATH is always set, either it is never set.

### Checklist

- [x ] The issues is caused by firejail (i.e. running the program by path (e.g. `/usr/bin/vlc`) "fixes" it).
- [ x] I can reproduce the issue without custom modifications (e.g. globals.local).
- [ x] The program has a profile. (If not, request one in `https://github.com/netblue30/firejail/issues/1139`)
- [ x] I have performed a short search for similar issues (to avoid opening a duplicate).
  - [ x] I'm aware of `browser-allow-drm yes`/`browser-disable-u2f no` in `firejail.config` to allow DRM/U2F in browsers.
- [ x] I used `--profile=PROFILENAME` to set the right profile. (Only relevant for AppImages)

ousamabenyounes:
`ok_to_run()` looks the program up in `env_get("PATH")`, which is firejail's
copy of the caller's environment. When the caller has no PATH that lookup finds
nothing and firejail gives up, even though it goes on to execute the program
with the default PATH it sets for itself in `env_apply_whitelist()`.

The lookup works anyway whenever something else in the run reaches
`init_paths()` first, because that stores the default PATH in the environment
list as a side effect. Expanding a `${PATH}` macro is the usual trigger and
`disable-common.inc` has 145 of them, so the stock profile hides the problem
while `--noprofile` (or any profile that never mentions `${PATH}`) shows it:

```console
$ env -u PATH firejail --quiet --noprofile ls
Error: no suitable ls executable found
$ env -u PATH firejail --quiet ls
<lists the directory>
```

That is the "depends on the content of the profile" behaviour reported in the
issue.

This searches the same default list in `ok_to_run()`, so the lookup matches
what the program is actually executed with. The default was already spelled out
twice (`env_apply_whitelist()` and `init_paths()`), so the first commit hoists
it into a single `DEFAULT_PATH` macro before the fix adds a third use.

Two things worth flagging:

* The `if (path1)` guard in `ok_to_run()` is always true now, so the block is
  de-indented. `git diff -w` reduces the change to +7/-2.
* With `--rmenv=PATH` a bare program name now runs instead of erroring out,
  because `execvp()` falls back to `confstr(_CS_PATH)`. A program that only
  lives in `/usr/local/bin` or an sbin directory still fails, with
  `Cannot start application` instead of `no suitable ... executable found`.

### Testing

`test/environment/path-unset.exp` run on unmodified `master` with only that
file added (RED):

```console
$ cd test/environment && ./path-unset.exp
spawn /bin/bash
stty -echo
[expect echoes the spawned shell prompt]
Warning: /usr/bin/bwrap was not disabled
Error: no suitable printf executable found
TESTING ERROR 2
```

Same test on this branch (GREEN):

```console
$ cd test/environment && ./path-unset.exp
spawn /bin/bash
stty -echo
[expect echoes the spawned shell prompt]
Warning: /usr/bin/bwrap was not disabled
pathunset

all done
```

Full local suite, run on `master` and on this branch with identical results:
the default build and the `gcc-12 --enable-analyzer` build, the clang build,
`make install` + `make installcheck`, `make test-environment`,
`scan-build-18 make`, `make cppcheck`, `make codespell`, and the profile
checks. `make test-environment` reports no `TESTING ERROR` on either side.

Fixes #5939

### DIFF
--- a/src/firejail/env.c
+++ b/src/firejail/env.c
@@ -313,7 +313,7 @@ void env_apply_whitelist(void) {
 	env_apply_list(env_whitelist, ARRAY_SIZE(env_whitelist));
 
 	// hardcoding PATH
-	if (setenv("PATH", "/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin", 1) < 0)
+	if (setenv("PATH", DEFAULT_PATH, 1) < 0)
 		errExit("setenv");
 }
 
--- a/src/firejail/firejail.h
+++ b/src/firejail/firejail.h
@@ -31,6 +31,9 @@
 
 
 
+// PATH used when the calling environment doesn't set one
+#define DEFAULT_PATH "/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin"
+
 // profiles
 #define DEFAULT_USER_PROFILE	"default"
 #define DEFAULT_ROOT_PROFILE	"server"
--- a/src/firejail/paths.c
+++ b/src/firejail/paths.c
@@ -29,7 +29,7 @@ static void init_paths(void) {
 	const char *env_path = env_get("PATH");
 	char *p;
 	if (!env_path) {
-		env_path = "/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/sbin";
+		env_path = DEFAULT_PATH;
 		env_store_name_val("PATH", env_path, SETENV);
 	}
 	char *path = strdup(env_path);
--- a/src/firejail/sandbox.c
+++ b/src/firejail/sandbox.c
@@ -428,44 +428,49 @@ static int ok_to_run(const char *program) {
 			return 1;
 	}
 	else { // search $PATH
+		// the program runs with DEFAULT_PATH when the calling environment has
+		// none (env_apply_whitelist()), so search that list here as well;
+		// otherwise the lookup depends on whether something else in this run
+		// called init_paths() first, e.g. a ${PATH} macro in the profile
 		const char *path1 = env_get("PATH");
-		if (path1) {
+		if (!path1)
+			path1 = DEFAULT_PATH;
+
+		if (arg_debug)
+			printf("Searching $PATH for %s\n", program);
+		char *path2 = strdup(path1);
+		if (!path2)
+			errExit("strdup");
+
+		// use path2 to count the entries
+		char *ptr = strtok(path2, ":");
+		while (ptr) {
+			char *fname;
+
+			if (asprintf(&fname, "%s/%s", ptr, program) == -1)
+				errExit("asprintf");
 			if (arg_debug)
-				printf("Searching $PATH for %s\n", program);
-			char *path2 = strdup(path1);
-			if (!path2)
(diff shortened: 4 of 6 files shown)


ousamabenyounes:
Hi @kmk3, could you please clarify why this PR was closed and labeled `ai-slop`?

If there is a specific technical problem, a project guideline I missed, or something that should be simplified, I would genuinely appreciate the feedback and would be happy to correct it.

kmk3:
> Hi @kmk3, could you please clarify why this PR was closed and labeled
> `ai-slop`?

This PR is low quality and clearly AI-generated.

> If there is a specific technical problem, a project guideline I missed, or
> something that should be simplified, I would genuinely appreciate the
> feedback and would be happy to correct it.

Besides the above, it's not clear whether this fully fixes the issue (and if it
doesn't, it might just mask the problem), as there are other parts of the code
related to this.

I'm not interested in investigating that at the moment and I don't trust AI to
do it properly, especially whichever one was used to generate this PR.

As for general feedback, you could try reading and writing the code and text
yourself.

Human-generated PRs are usually easier and more worthwhile to review.

ousamabenyounes:
Thanks for the feedback. I retested the exact #5939 reproduction with real Firejail, comparing the unmodified base (`541dd07`) and this PR (`e8159c0`):

```text
PATH unset, empty profile: base → "no suitable ls executable found"; PR → ls runs (exit 0)
PATH unset, profile with `blacklist ${PATH}/systemctl`: ls runs on both
path-unset.exp: base → TESTING ERROR 2; PR → pathunset / all done
```

The patch was AI-assisted, but these are real executions, not mocked results. They show that this PR fixes the inconsistency reported in #5939. If you see another case within that issue that it misses, could you share a reproducer? I'll gladly test it and extend the fix.

---

**Update:** you were right that other code is involved.

This patch is at the wrong layer and it breaks `--rmenv=PATH`: `env_get()` returns NULL for an RMENV entry by design (`env.c:251`), so a default inside `ok_to_run()` would resolve a bare command name even when the user removed PATH explicitly. Not something to do in a setuid binary.

What I'd missed: `init_paths()` already stores the default (`paths.c:33`), but it's only reached via `build_paths()`, called conditionally from `fs.c` and `landlock.c` — that's why #5939 is profile-dependent. `run_symlink.c:44` and `fs_lib.c:60` have the same gap.

Keep it closed. Analysis on #5939 if anyone picks it up.

